Issue #12, opened 2022-09-22
Test more conditions in GitHub actions


joshlf:
Use GitHub actions to test the following axes:
- Currently stable, current beta, current nightly, MSRV stable (1.51.0)
- All of the different targets that affect us (currently, this is only relevant for the `simd` and `simd-nightly` features, which emit impls for various architecture-specific types)
- All of the Cargo features we support
- `cargo check`, `cargo test`, and `cargo miri test`

Note that, depending on how we implement this, some combinations of the above may not be possible to run. E.g., if we run on an x86_64 machine, it may be possible to run `cargo miri test` while targeting powerpc (I'm not sure about this), but it definitely won't be possible to run `cargo test` while targeting powerpc.

As a possible stretch goal, also include a test to make sure that `README.md` is kept up to date (#18).

AntoniosBarotsis:
Hello! I was wondering if this still up for taking and if so whether you could explain the issue in a bit more detail (I'm also new to rust in general so I probably missed some stuff).

From what I understand you want to:

- Check, build and test your crate + the 3 features, namely `alloc`, `simd` and `simd-nightly`
- Run all these against some predetermined targets

I am not sure if running these in both stable and nightly (when applicable) changes anything.

joshlf:
Hi @AntoniosBarotsis , thanks for offering! Yes, it's still open; feel free to assign it to yourself if you feel like taking it on!

There are a few considerations here; I'll break them down by the "axes" I mentioned in the original comment:

### Version

We want to make sure to test with different versions of the toolchain for a few reasons:
- The nightly toolchain is required in order for us to test using unstable Rust features, which are required for the `simd-nightly` Cargo feature that we expose (we could also add other nightly-only Cargo features in the future)
- The beta toolchain is just us being a good citizen - the Rust project always needs more people to test code using the current beta
- The MSRV stable ensures that we successfully compile on our published MSRV
- The current stable is technically redundant given that we're also going to be testing on MSRV (if tests succeed on MSRV, they should succeed on current stable as well), but I'm a bit paranoid that if we ever removed tests for MSRV in the future and we didn't also have a test for current stable, we might not notice and would end up in a situation where we weren't testing on _any_ stable, which would be bad. It also has the advantage that, if a PR breaks MSRV stable but not current stable, it means that it's relying on a feature that was introduced between MSRV stable and current stable, and suggests that we might want to update our MSRV in the next breaking release.

### Targets

[...]

Issue #18, opened 2022-10-01
Test that `cargo readme` output matches `README.md`


joshlf:
We use the following command to auto-generate the contents of `README.md` from the crate-level doc comment in `src/lib.rs`:

```shell
cargo readme | sed 's/\[\(`[^`]*`\)]/\1/g'
```

The `sed` command removes code links like:

```rust
/// Here is a link to [`Vec<u8>`].
```

We should add a test that the output of that command matches the current contents of `README.md`, which will ensure PRs which edit that doc comment but do not update `README.md` can't be spuriously merged.

joshlf:
Fixed in fa7e3e83af64a219c7a3d86925875cb65c71e95d.

PR #19, opened 2022-10-02
Adding workflow

AntoniosBarotsis:
Related to #12.

## Roadmap

- [X] Stable and nightly channels
- [X] Multiple architectures (implemented but not specified)
- [X] Features
- [X] Miri test

## Links

- [Workflow file](https://github.com/AntoniosBarotsis/zerocopy/blob/githubActions/.github/workflows/ci.yml)
- [Workflow runs](https://github.com/AntoniosBarotsis/zerocopy/actions)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
     # If target is x86 or x86_64, 
     # or we are using simd-nightly while not on the nightly channel
     # do not continue on error
-    continue-on-error: ${{ !(contains(matrix.target, 'x86_64') || !contains(matrix.target, 'i686')) && ((contains(matrix.features, 'simd-nightly')) && (matrix.channel != 'nightly')) }}
+    continue-on-error: ${{ !(contains(matrix.target, 'x86_64') || contains(matrix.target, 'i686')) || ((contains(matrix.features, 'simd-nightly')) && (matrix.channel != 'nightly')) }}
 
     name: Build & Test (${{ matrix.channel }} for ${{ matrix.target }}, features set to ${{ matrix.features != '' }})
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,45 @@
+name: Build & Tests
+
+on:
+  push:
+  pull_request:
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  build_test:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        channel: [ "stable", "nightly" ]
+        target: [ "aarch64-unknown-linux-gnu" ]
+
+    continue-on-error: ${{ matrix.channel == 'nightly' }}
+    name: Build & Test - Channel ${{ matrix.channel }}
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Setup Rust ${{ matrix.channel }} ${{ matrix.target }}
+      uses: ATiltedTree/setup-rust@v1
+      with:
+        rust-version: ${{ matrix.channel }}
+        targets: ${{ matrix.target }}
+
+    - name: Check ${{ matrix.channel }}
+      run: cargo +${{ matrix.channel }} check --target ${{ matrix.target }} --verbose
+
+    - name: Build ${{ matrix.channel }}
+      run: cargo +${{ matrix.channel }} build --target ${{ matrix.target }} --verbose
+
+    - name: Run ${{ matrix.channel }} tests
+      run: cargo +${{ matrix.channel }} test --target ${{ matrix.target }} --verbose
+    
+    - name: Install cargo-miri
+      run: cargo install miri
+    
+    - name: Run miri test
+      run: cargo miri test -y
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
     # Prevents a workflow run from failing when a job fails
     # If target is x86 or x86_64, 
     # or we are using simd-nightly while not on the nightly channel
-    # do not continue on error
+    # do not continue on error (so crash on error)
     continue-on-error: ${{ !(contains(matrix.target, 'x86_64') || contains(matrix.target, 'i686')) || ((contains(matrix.features, 'simd-nightly')) && (matrix.channel != 'nightly')) }}
 
     name: Build & Test (${{ matrix.channel }} for ${{ matrix.target }}, features set to ${{ matrix.features != '' }})
@@ -45,11 +45,22 @@ jobs:
     - name: Build ${{ matrix.channel }}
       run: cargo +${{ matrix.channel }} build --target ${{ matrix.target }} ${{ matrix.features }} --verbose
 
-    # Only run tests in the architecture that the runners are using
+    # Only run tests in the x86 architecture
     - name: Run ${{ matrix.channel }} tests
       run: cargo +${{ matrix.channel }} test --target ${{ matrix.target }} ${{ matrix.features }} --verbose
       if: ${{ contains(matrix.target, 'x86_64') || contains(matrix.target, 'i686') }}
 
+    # - name: Check README consistency
+    #   run: |
+    #     cargo install cargo-readme
+    #     cargo readme
+    #     DIFF=$(diff <(cat README.md) <(cargo readme))
+
+    #     if [ "$DIFF" != "" ]; then
+    #       exit 1
+    #     else
+    #       exit 0
+    #     fi
     
     - name: Install cargo-miri
       run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
     - name: Install dependencies
       run: |
         sudo apt-get update && sudo apt-get upgrade
-        sudo apt-get install libffi-devel libxml2-devel zlib-devel libxkbcommon-dev
+        sudo apt-get install libxkbcommon-dev
 
     - name: Check ${{ matrix.channel }}
       run: cargo +${{ matrix.channel }} check --target ${{ matrix.target }} --verbose
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,8 @@ jobs:
     
     - name: Rust Cache
       uses: Swatinem/rust-cache@v2.0.0
+      with:
+        key: ${{ matrix.channel }}-${{ matrix.target }}-${{ matrix.features }}-${{ hashFiles('**/Cargo.lock') }}
 
     - name: Check ${{ matrix.channel }}
       run: cargo +${{ matrix.channel }} check --target ${{ matrix.target }} ${{ matrix.features }} --verbose
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,17 +57,17 @@ jobs:
       run: cargo +${{ matrix.channel }} test --target ${{ matrix.target }} ${{ matrix.features }} --verbose
       if: ${{ contains(matrix.target, 'x86_64') || contains(matrix.target, 'i686') }}
 
-    - name: Check README consistency
-      run: |
-        cargo install cargo-readme
-        cargo readme
-        DIFF=$(diff <(cat README.md) <(cargo readme))
+    # - name: Check README consistency
+    #   run: |
+    #     cargo install cargo-readme
+    #     cargo readme
+    #     DIFF=$(diff <(cat README.md) <(cargo readme))
 
-        if [ "$DIFF" != "" ]; then
-          exit 1
-        else
-          exit 0
-        fi
+    #     if [ "$DIFF" != "" ]; then
+    #       exit 1
+    #     else
+    #       exit 0
+    #     fi
     
     - name: Install cargo-miri
       run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,7 +61,7 @@ jobs:
     #       exit 0
     #     fi
     
-    - name: Install cargo-miri
+    - name: Install Miri
       run: |
         rustup toolchain install nightly-${{ env.MIRI_TOOLCHAIN_VERSION }}
         rustup component add miri --toolchain nightly-${{ env.MIRI_TOOLCHAIN_VERSION }}-${{ matrix.target }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,14 +37,14 @@ jobs:
         targets: ${{ matrix.target }}
 
     - name: Check ${{ matrix.channel }}
-      run: cargo +${{ matrix.channel }} check --target ${{ matrix.target }} --verbose
+      run: cargo +${{ matrix.channel }} check --target ${{ matrix.target }} ${{ matrix.features }} --verbose
 
     - name: Build ${{ matrix.channel }}
-      run: cargo +${{ matrix.channel }} build --target ${{ matrix.target }} --verbose
+      run: cargo +${{ matrix.channel }} build --target ${{ matrix.target }} ${{ matrix.features }} --verbose
 
     # Only run tests in the architecture that the runners are using
     - name: Run ${{ matrix.channel }} tests
-      run: cargo +${{ matrix.channel }} test --target ${{ matrix.target }} --verbose
+      run: cargo +${{ matrix.channel }} test --target ${{ matrix.target }} ${{ matrix.features }} --verbose
       if: ${{ matrix.target == 'x86_64-unknown-linux-gnu' }}
     
     - name: Install cargo-miri
@@ -53,4 +53,4 @@ jobs:
         rustup component add miri --toolchain nightly-${{ env.TOOLCHAIN_VERSION }}-${{ matrix.target }}
     
     - name: Run miri test
-      run: cargo +nightly-${{ env.TOOLCHAIN_VERSION }}-${{ matrix.target }} miri test
+      run: cargo +nightly-${{ env.TOOLCHAIN_VERSION }}-${{ matrix.target }} miri test ${{ matrix.features }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,7 @@ jobs:
     - name: Rust Cache
       uses: Swatinem/rust-cache@v2.0.0
       with:
-        key: ${{ matrix.channel }}-${{ matrix.target }}-${{ matrix.features }}-${{ hashFiles('**/Cargo.lock') }}
+        key: ${{ matrix.channel }}-${{ matrix.target }}-${{ matrix.features == '' }}-${{ hashFiles('**/Cargo.lock') }}
 
     - name: Check ${{ matrix.channel }}
       run: cargo +${{ matrix.channel }} check --target ${{ matrix.target }} ${{ matrix.features }} --verbose
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,10 +61,10 @@ jobs:
     #       exit 0
     #     fi
     
-    - name: Install Miri
-      run: |
-        rustup toolchain install nightly-${{ env.MIRI_TOOLCHAIN_VERSION }}
-        rustup component add miri --toolchain nightly-${{ env.MIRI_TOOLCHAIN_VERSION }}-${{ matrix.target }}
+    # - name: Install Miri
+    #   run: |
+    #     rustup toolchain install nightly-${{ env.MIRI_TOOLCHAIN_VERSION }}
+    #     rustup component add miri --toolchain nightly-${{ env.MIRI_TOOLCHAIN_VERSION }}-${{ matrix.target }}
     
-    - name: Run miri test
-      run: cargo +nightly-${{ env.MIRI_TOOLCHAIN_VERSION }}-${{ matrix.target }} miri test ${{ matrix.features }} -- --skip ui
+    # - name: Run miri test
+    #   run: cargo +nightly-${{ env.MIRI_TOOLCHAIN_VERSION }}-${{ matrix.target }} miri test ${{ matrix.features }} -- --skip ui
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,13 +9,17 @@ env:
 
 jobs:
   build_test:
+    # Used for installing miri
+    # https://rust-lang.github.io/rustup-components-history/
+    env:
+      TOOLCHAIN_VERSION: '2022-09-26'
 
     runs-on: ubuntu-latest
 
     strategy:
       matrix:
         channel: [ "stable", "nightly" ]
-        target: [ "aarch64-unknown-linux-gnu" ]
+        target: [ "x86_64-unknown-linux-gnu" ]
 
     # Prevents a workflow run from failing when a job fails
     continue-on-error: ${{ matrix.channel == 'nightly' }}
@@ -42,8 +46,8 @@ jobs:
     
     - name: Install cargo-miri
       run: |
-        rustup toolchain install nightly-2022-09-26
-        rustup component add miri --toolchain nightly-2022-09-26-x86_64-unknown-linux-gnu
+        rustup toolchain install nightly-${{ TOOLCHAIN_VERSION }}
+        rustup component add miri --toolchain nightly-${{ TOOLCHAIN_VERSION }}-${{ matrix.target }}
     
     - name: Run miri test
-      run: cargo +nightly-2022-09-26-x86_64-unknown-linux-gnu miri test
+      run: cargo +nightly-${{ TOOLCHAIN_VERSION }}-${{ matrix.target }} miri test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,14 +31,14 @@ jobs:
         rust-version: ${{ matrix.channel }}
         targets: ${{ matrix.target }}
 
+    - name: Install dependencies
+      run: sudo apt-get install libffi-devel libxml2-devel zlib-devel
+
     - name: Check ${{ matrix.channel }}
       run: cargo +${{ matrix.channel }} check --target ${{ matrix.target }} --verbose
 
     - name: Build ${{ matrix.channel }}
       run: cargo +${{ matrix.channel }} build --target ${{ matrix.target }} --verbose
-
-    - name: Clean
-      run: cargo clean
 
     - name: Run ${{ matrix.channel }} tests
       run: cargo +${{ matrix.channel }} test --target ${{ matrix.target }} --verbose
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
     # If target is x86 or x86_64, 
     # or we are using simd-nightly while not on the nightly channel
     # do not continue on error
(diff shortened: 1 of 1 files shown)


AntoniosBarotsis:
I'm currently having issues running the tests for a particular target, `build` and `check` seem to be working fine.

For the tests in stable `aarch64-unknown-linux-gnu` I am currently getting:

```sh
error: linking with `cc` failed: exit status: 1
```

See [logs](https://github.com/AntoniosBarotsis/zerocopy/actions/runs/3169566963/jobs/5161566154).

I tried a few things but I didn't find too much online. I am assuming some dependency is missing, do you happen to have any idea what might be wrong?

AntoniosBarotsis:
I am currently trying to add `miri test` and it seems that it is failing somewhere, check [these logs](https://github.com/AntoniosBarotsis/zerocopy/actions/runs/3169696208/jobs/5161785065) for example. Is this expected or should it be passing?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,56 @@
...
+on:
+  push:
+  pull_request:
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  build_test:
+    # Used for installing miri
```
joshlf, line 12: Uber nit: it's spelled Miri (capital M)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,56 @@
...
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        channel: [ "stable", "beta", "nightly" ]
+        target: [ "x86_64-unknown-linux-gnu" ]
+        features: [ "" , "--features alloc,simd,simd-nightly"]
+
+    # Prevents a workflow run from failing when a job fails
+    continue-on-error: ${{ matrix.channel != 'stable' }}
```
joshlf, line 26: I think this is actually too permissive. AFAICT, the only times we want to continue on error are:
- We're compiling for a non-x86 or x86_64 target and running `cargo test`
- We're compiling for non-nightly and using the `simd-nightly` feature

Any other failure - even if on non-stable - should "count".

.github/workflows/ci.yml
```diff
@@ -0,0 +1,56 @@
...
+    - name: Check ${{ matrix.channel }}
+      run: cargo +${{ matrix.channel }} check --target ${{ matrix.target }} ${{ matrix.features }} --verbose
+
+    - name: Build ${{ matrix.channel }}
+      run: cargo +${{ matrix.channel }} build --target ${{ matrix.target }} ${{ matrix.features }} --verbose
+
+    # Only run tests in the architecture that the runners are using
+    - name: Run ${{ matrix.channel }} tests
+      run: cargo +${{ matrix.channel }} test --target ${{ matrix.target }} ${{ matrix.features }} --verbose
+      if: ${{ matrix.target == 'x86_64-unknown-linux-gnu' }}
```
joshlf, line 48: Once you add other targets to these tests, I _think_ you should be able to run x86 (ie, 32-bit) as well even on an x86_64 machine.

.github/workflows/ci.yml
joshlf: Ah, I see your conditional logic below for running `cargo test` on the wrong target, so I guess it's just the `simd-nightly` issue.

AntoniosBarotsis: Should I make a case where I only use `simd-nightly` for this or is using all features fine?

joshlf:
This looks awesome; thanks so much! Do you know how we can get GitHub to actually run these for us before we merge this PR so we can test them out?

Also, how would you feel about addressing the following in this PR?
- Compile for other targets (even if we don't list every possible target, at least a few would be good to make sure the machinery works)
- Test on MSRV stable
- https://github.com/google/zerocopy/issues/18

Finally, to answer your question from the thread on #12 :

> Could you outline the exact targets that you want because I'm not sure what the differences between some of these are.

Yeah sorry, the list of architectures and list of targets seems to not perfectly overlap, which is confusing (e.g., "x86" seems to refer to a family of targets which includes i586 and i686). Here's a list of architectures (the thing in `#[cfg(target_arch = "xxx")]`) and corresponding targets you should be able to use (sourced from the [Platform Support page](https://doc.rust-lang.org/nightly/rustc/platform-support.html
) of the rustc book):


- Tier 1 (precompiled "std" with strong stability guarantees - these should work out of the box)
  - x86 - i686-unknown-linux-gnu
  - x86_64 - x86_64-unknown-linux-gnu
  - aarch64 - aarch64-unknown-linux-gnu
- Tier 2 (precompiled "std" with weak stability guarantees - these should work, but they may break sporadically in the future)
  - arm - arm-unknown-linux-gnueabi
  - powerpc - powerpc-unknown-linux-gnu
  - powerpc64 - powerpc64-unknown-linux-gnu
  - wasm32 - wasm32-wasi

AntoniosBarotsis:
I *think* I resolved everything you mentioned, could you take a look

.github/workflows/ci.yml
```diff
@@ -45,7 +48,7 @@ jobs:
     # Only run tests in the architecture that the runners are using
     - name: Run ${{ matrix.channel }} tests
       run: cargo +${{ matrix.channel }} test --target ${{ matrix.target }} ${{ matrix.features }} --verbose
-      if: ${{ matrix.target == 'x86_64-unknown-linux-gnu' }}
+      if: ${{ contains(matrix.target, 'x86') }}
```
joshlf, line 51: I think this'll probably need to be either an exact comparison or at least a check for the substrings `x86_64` and `i686`.

AntoniosBarotsis: Yes I just saw your response, will update

.github/workflows/ci.yml
```diff
@@ -23,7 +23,10 @@ jobs:
         features: [ "" , "--features alloc,simd,simd-nightly"]
 
     # Prevents a workflow run from failing when a job fails
-    continue-on-error: ${{ matrix.channel != 'stable' }}
+    # If target is x86 or x86_64, 
+    # or we are using simd-nightly while not on the nightly channel
+    # do not continue on error
+    continue-on-error: ${{ !contains(matrix.target, 'x86') || (contains(matrix.features, 'simd-nightly') && matrix.channel != 'nightly') }}
```
joshlf, line 29: I think the "x86" check will need to be either an exact comparison against the various targets or at least a check for the substrings `x86_64` and `i686`.

AntoniosBarotsis:
We replied at the same time!

> Do you know how we can get GitHub to actually run these for us before we merge this PR so we can test them out?

I *believe* this is not possible because the CI is in my fork but once it is merged it should spin up

> Compile for other targets

Do you mean produce executables and put them somewhere or just build? I am adding the targets you mentioned so if you just mean build then this should be fine

> Test on MSRV stable

I could add another matrix parameter for the rust version although not sure if I can use something like `latest` so we'd have to manually change that once a new one comes out (of course MSRV should be constant).

> #18

I could add that right after `cargo test` yes

.github/workflows/ci.yml
```diff
@@ -0,0 +1,59 @@
...
+    # https://rust-lang.github.io/rustup-components-history/
+    env:
+      TOOLCHAIN_VERSION: '2022-09-26'
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        channel: [ "stable", "beta", "nightly" ]
+        target: [ "i686-unknown-linux-gnu", "x86_64-unknown-linux-gnu", "aarch64-unknown-linux-gnu", "arm-unknown-linux-gnueabi", "powerpc-unknown-linux-gnu", "powerpc64-unknown-linux-gnu", "wasm32-wasi" ]
```
joshlf, line 22: Uber nit: Could you put `arm-unknown-linux-gnueabi` before `aarch64-unknown-linux-gnu` since arm 32-bit and aarch64 is 64-bit (and that's the pattern you've used for the other targets)?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,59 @@
...
+    - name: Check ${{ matrix.channel }}
+      run: cargo +${{ matrix.channel }} check --target ${{ matrix.target }} ${{ matrix.features }} --verbose
+
+    - name: Build ${{ matrix.channel }}
+      run: cargo +${{ matrix.channel }} build --target ${{ matrix.target }} ${{ matrix.features }} --verbose
+
+    # Only run tests in the architecture that the runners are using
+    - name: Run ${{ matrix.channel }} tests
+      run: cargo +${{ matrix.channel }} test --target ${{ matrix.target }} ${{ matrix.features }} --verbose
+      if: ${{ !contains(matrix.target, 'x86_64') || !contains(matrix.target, 'i686') }}
```
joshlf, line 51: If I'm reading this correctly, won't this run only if we're _not_ on i686 or x86_64?

AntoniosBarotsis: Oops yes you are right, nice catch

joshlf: Thanks 😃

joshlf:
> > Compile for other targets
> 
> Do you mean produce executables and put them somewhere or just build? I am adding the targets you mentioned so if you just mean build then this should be fine

Yeah I just meant check/build/Miri - you already covered this

> > Test on MSRV stable
> 
> I could add another matrix parameter for the rust version although not sure if I can use something like latest so we'd have to manually change that once a new one comes out (of course MSRV should be constant).

You can just add another channel for `1.51.0`; the channel syntax (according to `rustup toolchain`) is:

```text
stable|beta|nightly|<major.minor>|<major.minor.patch>
```

> > https://github.com/google/zerocopy/issues/18
> 
> I could add that right after cargo test yes

Perfect, thanks! Maybe add it at the very bottom (after Miri) though?

AntoniosBarotsis:
> You can just add another channel

Oh cool, didn't know

> Maybe add it at the very bottom (after Miri) though?

The idea was to always get feedback on this since Miri is failing even currently on the repo and any subsequent tasks won't execute but it's up to you.

About the README stuff, I wrote the following:

```sh
DIFF=$(diff <(cat README.md) <(cargo readme))

if [ "$DIFF" == "" ]; then
  exit 0
else
  exit 1
fi
```

But the 2 are indeed different right now, is this expected?

<details>
  <summary>Diff output</summary>

	$ diff <(cat README.md) <(cargo readme)
	
	1,44c1,40
	< # zerocopy
	<
	< Utilities for safe zero-copy parsing and serialization.
	<
	< This crate provides utilities which make it easy to perform zero-copy
	< parsing and serialization by allowing zero-copy conversion to/from byte
	< slices.
	<
	< This is enabled by three core marker traits, each of which can be derived
	< (e.g., `#[derive(FromBytes)]`):
	< - `FromBytes` indicates that a type may safely be converted from an
	<   arbitrary byte sequence
	< - `AsBytes` indicates that a type may safely be converted *to* a byte
	<   sequence
	< - `Unaligned` indicates that a type's alignment requirement is 1
	<
	< Types which implement a subset of these traits can then be converted to/from
	< byte sequences with little to no runtime overhead.
	<
	< Note that these traits are ignorant of byte order. For byte order-aware
	< types, see the `byteorder` module.
	<
	< ## Features
	<
	< `alloc`: By default, `zerocopy` is `no_std`. When the `alloc` feature is
	< enabled, the `alloc` crate is added as a dependency, and some
	< allocation-related functionality is added.
	<
	< `simd`: When the `simd` feature is enabled, `FromBytes` and `AsBytes` impls
	< are emitted for all stable SIMD types which exist on the target platform.
	< Note that the layout of SIMD types is not yet stabilized, so these impls may
	< be removed in the future if layout changes make them invalid. For more
	< information, see the Unsafe Code Guidelines Reference page on the [Layout of
	< packed SIMD vectors][simd-layout].
	<
	< `simd-nightly`: Enables the `simd` feature and adds support for SIMD types
	< which are only available on nightly. Since these types are unstable, support
	< for any type may be removed at any point in the future.
	<
	< [simd-layout]: https://rust-lang.github.io/unsafe-code-guidelines/layout/packed-simd-vectors.html
	<
	< ## Dislcaimer
	<
	< Disclaimer: Zerocopy is not an officially supported Google product.
	\ No newline at end of file
	---
	> # zerocopy
	>
	> Utilities for safe zero-copy parsing and serialization.
	>
	> This crate provides utilities which make it easy to perform zero-copy
	> parsing and serialization by allowing zero-copy conversion to/from byte
	> slices.
	>
	> This is enabled by three core marker traits, each of which can be derived
	> (e.g., `#[derive(FromBytes)]`):
	> - [`FromBytes`] indicates that a type may safely be converted from an
	>   arbitrary byte sequence
	> - [`AsBytes`] indicates that a type may safely be converted *to* a byte
	>   sequence
	> - [`Unaligned`] indicates that a type's alignment requirement is 1
	>
	> Types which implement a subset of these traits can then be converted to/from
	> byte sequences with little to no runtime overhead.
	>
	> Note that these traits are ignorant of byte order. For byte order-aware
	> types, see the [`byteorder`] module.
	>
	> ## Features
	>
	> `alloc`: By default, `zerocopy` is `no_std`. When the `alloc` feature is
	> enabled, the `alloc` crate is added as a dependency, and some
	> allocation-related functionality is added.
	>
	> `simd`: When the `simd` feature is enabled, `FromBytes` and `AsBytes` impls
	> are emitted for all stable SIMD types which exist on the target platform.
	> Note that the layout of SIMD types is not yet stabilized, so these impls may
	> be removed in the future if layout changes make them invalid. For more
	> information, see the Unsafe Code Guidelines Reference page on the [Layout of
	> packed SIMD vectors][simd-layout].
	>
	> `simd-nightly`: Enables the `simd` feature and adds support for SIMD types
	> which are only available on nightly. Since these types are unstable, support
	> for any type may be removed at any point in the future.
	>
	> [simd-layout]: https://rust-lang.github.io/unsafe-code-guidelines/layout/packed-simd-vectors.html
</details>

joshlf:
>> Maybe add it at the very bottom (after Miri) though?
>
> The idea was to always get feedback on this since Miri is failing even currently on the repo and any subsequent tasks won't execute but it's up to you.

Oh gotcha - can you point me to those failures?

It's late here, so I've got to get to bed, but I'll follow up with the rest of this tomorrow morning. Thanks again for all your awesome work on this!

AntoniosBarotsis:
The `contains` if checks seem to not be working properly, they seem to always return true, I'll debug later and let you know what that's resolved.

AntoniosBarotsis:
We could add dependency caching with [this](https://github.com/Swatinem/rust-cache) workflow which could save around 45 seconds each run, although the vast majority of the workflow is `miri test` at 4 minutes.

AntoniosBarotsis:
I'm looking at this a bit on and off but I just noticed that [this](https://github.com/AntoniosBarotsis/zerocopy/actions/runs/3173112141/jobs/5168358756#logs) job failed with 1.51 on x86_64 without the features (it also failed with the features but that is set to continue on error unlike this one).

I am not sure if I misread your earlier message;

> I think this is actually too permissive. AFAICT, the only times we want to continue on error are:
> -  We're compiling for a non-x86 or x86_64 target and running cargo test 
> -  We're compiling for non-nightly and using the simd-nightly feature

or if I got something wrong in the logic, there's so many logical negations I am unsure at this point.

Did you mean that both of these should be true or either one?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,71 @@
...
+      matrix:
+        channel: [ "1.51.0", "stable", "beta", "nightly" ]
+        target: [ "i686-unknown-linux-gnu", "x86_64-unknown-linux-gnu", "arm-unknown-linux-gnueabi", "aarch64-unknown-linux-gnu", "powerpc-unknown-linux-gnu", "powerpc64-unknown-linux-gnu", "wasm32-wasi" ]
+        features: [ "" , "--features alloc,simd,simd-nightly"]
+
+    # Prevents a workflow run from failing when a job fails
+    # If target is x86 or x86_64, 
+    # or we are using simd-nightly while not on the nightly channel
+    # do not continue on error (so crash on error)
+    continue-on-error: ${{ !(contains(matrix.target, 'x86_64') || contains(matrix.target, 'i686')) || ((contains(matrix.features, 'simd-nightly')) && (matrix.channel != 'nightly')) }}
```
joshlf, line 29: Until I get an answer on https://github.com/rust-lang/rust/issues/48556#issuecomment-1264564803, let's assume that we don't need to worry about `simd-nightly`, so this can just be:

```text
continue-on-error: ${{ !(contains(matrix.target, 'x86_64') || contains(matrix.target, 'i686')) }}
```

joshlf:
> For an older Miri failure, you can check [here](https://github.com/AntoniosBarotsis/zerocopy/actions/runs/3169696208/jobs/5161785065).

Ah, I see what's happening. Miri is trying to run our `ui` test, which uses the `trybuild` crate to actually invoke the compiler. Part of doing that requires making syscalls, which Miri doesn't support. You should be able to add `-- --skip ui` to the end of your `miri test` invocation to skip that test.

> About the README stuff, I wrote the following:
> 
> ```shell
> DIFF=$(diff <(cat README.md) <(cargo readme))
> 
> if [ "$DIFF" == "" ]; then
>   exit 0
> else
>   exit 1
> fi
> ```
> 
> But the 2 are indeed different right now, is this expected?

Ah I'll have to fix that. Let's just leave this out for the time being - I'll follow up with another PR once this one is merged, and I'll both fix that and add this test.

> We could add dependency caching with [this](https://github.com/Swatinem/rust-cache) workflow which could save around 45 seconds each run, although the vast majority of the workflow is miri test at 4 minutes.

Seems worth it to me - 16% speedup is pretty good!

> I'm looking at this a bit on and off but I just noticed that [this](https://github.com/AntoniosBarotsis/zerocopy/actions/runs/3173112141/jobs/5168358756#logs) job failed with 1.51 on x86_64 without the features (it also failed with the features but that is set to continue on error unlike this one).

Ah, that's my mistake. I determined the MSRV using `cargo check` as the command to verify whether a particular Rust version was supported. It looks like `once_cell` (which is a transitive dependency of one of our `dev-dependencies` that only get built for `cargo test`, but not for `cargo check`) is incompatible with 1.51.0. I re-ran `cargo msrv` using `cargo test` as the verification command, and found that, when we take `cargo test` into account, our MSRV is 1.56.1. If you update to use that as the MSRV, it should work.

> Let me know if you figure something out yourself and want me to change it otherwise I think it would be easier for you to make the last few edits considering you actually know what should compile under what circumstances.

Either works for me! Just want to make sure you get credit for this - if I take your code and upload it as a new PR, I don't think there's a way for me to make it appear that you were the author. I _think_ that with the changes I mentioned in this comment (omitting the `README.md` check, using `rust-cache`, and updating the MSRV to 1.56.1) and in a comment on the code itself (removing the `continue-on-error` logic for `simd-nightly`), everything should work now. Maybe just make those changes, and if everything works, I can merge as-is, and if not, I can take over from there?

AntoniosBarotsis:
I'll read through these tomorrow as it is getting late here and my brain isn't really working :)

Thanks for the reply! Should be able to make all the changes tomorrow!

AntoniosBarotsis:
I believe everything is fine now. Sorry for the messy commit history, always ends like this whenever I mess with workflows haha.

I am not entirely sure how the key is computed for the cache action, specifically not sure if it works with the matrix stuff we are using so I just added my own which is a combination of all the parameters and the lock file which should be fine.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,78 @@
...
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  build_test:
+    # Used for installing Miri
+    # https://rust-lang.github.io/rustup-components-history/
+    env:
+      TOOLCHAIN_VERSION: '2022-09-26'
```
joshlf, line 15: Maybe call this `MIRI_TOOLCHAIN_VERSION`?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,78 @@
...
+      matrix:
+        channel: [ "1.56.1", "stable", "beta", "nightly" ]
+        target: [ "i686-unknown-linux-gnu", "x86_64-unknown-linux-gnu", "arm-unknown-linux-gnueabi", "aarch64-unknown-linux-gnu", "powerpc-unknown-linux-gnu", "powerpc64-unknown-linux-gnu", "wasm32-wasi" ]
+        features: [ "" , "--features alloc,simd,simd-nightly"]
+
+    # Prevents a workflow run from failing when a job fails
+    # If target is x86 or x86_64, 
+    # or we are using simd-nightly while not on the nightly channel
+    # do not continue on error (so crash on error)
+    continue-on-error: ${{ !(contains(matrix.target, 'x86_64') || contains(matrix.target, 'i686')) }}
```
joshlf, line 29: Unfortunately I just realized that this isn't sufficient. If `cargo check`, `cargo build`, or `cargo miri test` fail, we should report those failures regardless of what target we're building for. It's only `cargo test` that should be allowed to fail on non-x86_64/non-i686. Is there a way to test for which command we're running?

AntoniosBarotsis: Not that I'm aware of, you can still inspect the tasks by browsing through the workflow runs though

.github/workflows/ci.yml
```diff
@@ -0,0 +1,78 @@
...
+
+    - name: Build ${{ matrix.channel }}
+      run: cargo +${{ matrix.channel }} build --target ${{ matrix.target }} ${{ matrix.features }} --verbose
+
+    # Only run tests in the x86 architecture
+    - name: Run ${{ matrix.channel }} tests
+      run: cargo +${{ matrix.channel }} test --target ${{ matrix.target }} ${{ matrix.features }} --verbose
+      if: ${{ contains(matrix.target, 'x86_64') || contains(matrix.target, 'i686') }}
+
+    - name: Check README consistency
```
joshlf, line 60: Doesn't this currently fail? Should we comment it out until it succeeds?

AntoniosBarotsis: You mean the test or the README stuff?

joshlf: Yeah

.github/workflows/ci.yml
joshlf: Oh wait it looks like we don't need this at all! You're already only running the `Run ${{ matrix.channel }} tests` job conditionally:

```text
if: ${{ contains(matrix.target, 'x86_64') || contains(matrix.target, 'i686') }}
```

joshlf: So we should be able to just remove this entirely. I can add it back later if we add more nightly features or something.

AntoniosBarotsis: Ah alright!

AntoniosBarotsis: Just to be sure, you mean remove the `if` check you just pasted right?

joshlf: No sorry, I mean remove this `continue-on-error` line (including the comment on it and the commented-out `continue-on-error` line below it). IIUC, the `if` in `Run ${{ matrix.channel }} tests` should be all that we need.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,70 @@
...
+    - name: Setup Rust ${{ matrix.channel }} ${{ matrix.target }}
+      uses: ATiltedTree/setup-rust@v1
+      with:
+        rust-version: ${{ matrix.channel }}
+        targets: ${{ matrix.target }}
+    
+    - name: Rust Cache
+      uses: Swatinem/rust-cache@v2.0.0
+      with:
+        key: ${{ matrix.channel }}-${{ matrix.target }}-${{ matrix.features == '' }}-${{ hashFiles('**/Cargo.lock') }}
```
joshlf, line 39: Won't `${{ matrix.features == '' }}` have the effect of treating all sets of features as identical? Would it make more sense to do `${{ matrix.features }}` so that the exact feature set is part of the key? If that causes syntax issues, you could update the `features` variable defined above to:

```text
features: [ "", "alloc,simd,simd-nightly" ]
```

...and then use it in the various cargo commands below as `--features ${{ matrix.features }}` (although you might need to put that in quotes in case it's the empty string - `"${{ matrix.features }}"` - I'm not sure whether the `run` command passes its input to the shell as a single string or as separate arguments).

AntoniosBarotsis: `${{ matrix.features == '' }}` will return a boolean and since the features vector is either empty or `--features alloc,simd,simd-nightly` it should be `[true, false]`

joshlf: Agreed, but we might want to update the `features` vector in the future (for example, to have a separate feature set for stable features: `[ "", "alloc,simd", "alloc,simd,simd-nightly" ]`).

AntoniosBarotsis: The issue with this is that we can't have commas in the key because that's what I tried initially.

joshlf: Ah understood. I guess let's merge as-is and I can try to change this after it's merged. I'll submit an issue for this once this is merged.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,70 @@
...
+    #       exit 0
+    #     fi
+    
+    - name: Install cargo-miri
+      run: |
+        rustup toolchain install nightly-${{ env.MIRI_TOOLCHAIN_VERSION }}
+        rustup component add miri --toolchain nightly-${{ env.MIRI_TOOLCHAIN_VERSION }}-${{ matrix.target }}
+    
+    - name: Run miri test -- --skip ui
+      run: cargo +nightly-${{ env.MIRI_TOOLCHAIN_VERSION }}-${{ matrix.target }} miri test ${{ matrix.features }}
```
joshlf, line 70: Looks like this isn't actually doing `-- --skip ui`? Is this not failing for you?

AntoniosBarotsis: Yes that's because I put it in the name instead of the command lmao my bad

.github/workflows/ci.yml
```diff
@@ -0,0 +1,70 @@
...
+    #     cargo readme
+    #     DIFF=$(diff <(cat README.md) <(cargo readme))
+
+    #     if [ "$DIFF" != "" ]; then
+    #       exit 1
+    #     else
+    #       exit 0
+    #     fi
+    
+    - name: Install cargo-miri
```
joshlf, line 64: Can we rename this to `Install Miri`? `cargo-miri` implies that it's a crate that's installed via `cargo install cargo-miri` rather than a rustup component.

joshlf:
OK awesome, this looks good! Can you kick off a run with the current version of the file? Once that passes successfully, I'll merge!

AntoniosBarotsis:
You can see the run [here](https://github.com/AntoniosBarotsis/zerocopy/actions/runs/3184463768), have been getting some errors for a while.

joshlf:
Hey, it looks like that stdsimd thing is breaking us after all! Why don't you just remove `simd-nightly` from the list of features and we can try running the build again. I can follow up later to get that working.

AntoniosBarotsis:
[Running](https://github.com/AntoniosBarotsis/zerocopy/actions/runs/31845398080)

AntoniosBarotsis:
Most of my commits were trying to stop these 2 from failing haha.

[Running again](https://github.com/AntoniosBarotsis/zerocopy/actions/runs/3184571865).

joshlf:
Hmmm, still failing. How would you feel about me taking this over so that I can iterate more quickly and kick off tests on my own and then we can co-own the PR and commit that results?

AntoniosBarotsis:
These linkers errors are probably happening because the target is not the same as the runner's but yes feel free to pick it up from here, pretty sure this was considered "valid" for Hacktober when you added the label anyway.

Good luck!

joshlf:
Merged manually in fa7e3e83af64a219c7a3d86925875cb65c71e95d.

PR #34, opened 2022-10-08
Fix as_bytes() test for big endian targets

### LINKED ISSUE #23: `test_as_bytes_methods` fails on powerpc
From [this test job](https://github.com/google/zerocopy/actions/runs/3185927469/jobs/5196030735):

```text
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `[0, 0, 0, 1, 0, 0, 0, 2, 0, 0, 0, 0]`,
 right: `[1, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0]`', src/lib.rs:2885:9
```

Fix the bug and re-enable the test in `.github/workflows/ci.yml`.

frazar:
The tests for the as_bytes() method assumed that bytes are stored in little endian order, which is not the case for PowerPC. Fix the alignment tests so that they work on both little and big endian targets.

Fixes #23

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2872,13 +2872,21 @@ mod tests {
         }
 
         let mut foo = Foo { a: 1, b: Wrapping(2), c: None };
+
         // Test that we can access the underlying bytes, and that we get the
         // right bytes and the right number of bytes.
-        assert_eq!(foo.as_bytes(), [1, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0]);
+        let expected: Vec<u8> = if cfg!(target_endian = "little") {
+            vec![1, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0]
+        } else {
+            vec![0, 0, 0, 1, 0, 0, 0, 2, 0, 0, 0, 0]
+        };
+        assert_eq!(foo.as_bytes(), expected.as_bytes());
+
         // Test that changes to the underlying byte slices are reflected in the
         // original object.
         foo.as_bytes_mut()[0] = 3;
-        assert_eq!(foo, Foo { a: 3, b: Wrapping(2), c: None });
+        let expected_a = if cfg!(target_endian = "little") { 0x00_00_00_03 } else { 0x03_00_00_01 };
+        assert_eq!(foo, Foo { a: expected_a, b: Wrapping(2), c: None });
 
         // Do the same tests for a slice, which ensures that this logic works
         // for unsized types as well.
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -2872,33 +2872,50 @@ mod tests {
         }
 
         let mut foo = Foo { a: 1, b: Wrapping(2), c: None };
+
         // Test that we can access the underlying bytes, and that we get the
         // right bytes and the right number of bytes.
-        assert_eq!(foo.as_bytes(), [1, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0]);
+
+        #[cfg(target_endian = "little")]
```
joshlf, line 2879: Can you use `cfg!` instead to make this a bit cleaner?

```rust
let expected: Vec<u8> = if cfg!(target_endian = "little") {
    vec![1, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0]
} else {
    vec![0, 0, 0, 1, 0, 0, 0, 2, 0, 0, 0, 0]
};
```

Same below.

src/lib.rs
```diff
@@ -2872,33 +2872,50 @@ mod tests {
...
+        #[cfg(target_endian = "big")]
+        let expected: Vec<u8> = vec![0, 0, 0, 1, 0, 0, 0, 2, 0, 0, 0, 0];
+        assert_eq!(foo.as_bytes(), expected.as_bytes());
+
         // Test that changes to the underlying byte slices are reflected in the
         // original object.
         foo.as_bytes_mut()[0] = 3;
-        assert_eq!(foo, Foo { a: 3, b: Wrapping(2), c: None });
+        #[cfg(target_endian = "little")]
+        let expected = Foo { a: 0x00_00_00_03, b: Wrapping(2), c: None };
```
joshlf, line 2889: Can you pull `a` out separately? Something like:

```rust
let a = if cfg!(target_endian = "little") { 0x00_00_00_03 } else { 0x03_00_00_01 };
let expected = Foo { a, b: Wrapping(2), c: None };
```

Same below.

joshlf:
Awesome, thanks @frazar! Left a few comments.

frazar:
Great! Should be ready now.

joshlf (approved):

joshlf:
Thanks again!

joshlf:
> May I ask you to add the "Hacktoberfest-accepted" label to this PR? Thanks!

Done!

frazar:
Great, thanks!

PR #35, opened 2022-10-09
Add include_bytes! macro

joshlf:
The `include_bytes!` macro is like the macro from `core` of the same name, except that it transmutes the file's contents to an arbitrary type (which implements `FromBytes`).

TODO:
- [x] ~~Change the example usage in the doc comment to use something more complex like the fixed-length header of a file format?~~
  - As discussed [here](https://github.com/google/zerocopy/pull/505#discussion_r1361326169), better to have a focused example that doesn't distract with other details.
- [x] ~~Are there other variants we should introduce? Perhaps one for slice types? Can we design it in such a way that it would work with other (non-slice) DSTs in the future if we add support for those?~~
  - Per https://github.com/google/zerocopy/discussions/524#discussioncomment-7362622, we're not trying to plan this far out anymore.


src/lib.rs
```diff
@@ -1924,6 +1924,51 @@ macro_rules! transmute_ref {
...
+///     let as_u32: u32 = include_bytes!("data");
+///     assert_eq!(as_u32, U32::<NativeEndian>::from([b'a', b'b', b'c', b'd']).get());
+///     let as_i32: i32 = include_bytes!("data");
+///     assert_eq!(as_i32, I32::<NativeEndian>::from([b'a', b'b', b'c', b'd']).get());
+/// }
+/// ```
+#[macro_export]
+macro_rules! include_bytes {
+    ($file:expr $(,)?) => {
+        // TODO: Is this hygienic?
```
joshlf, line 1967: TODO

@jswrenn You may know more about this than I do.

src/lib.rs
```diff
@@ -1924,6 +1924,51 @@ macro_rules! transmute_ref {
...
+///     assert_eq!(as_u32, u32::from_ne_bytes([b'a', b'b', b'c', b'd']));
+///     let as_i32: i32 = include_bytes!("data");
+///     assert_eq!(as_i32, i32::from_ne_bytes([b'a', b'b', b'c', b'd']));
+/// }
+/// ```
+#[macro_export]
+macro_rules! include_bytes {
+    ($file:expr $(,)?) => {
+        // TODO: Is this hygienic?
+        $crate::transmute!(*core::include_bytes!($file))
```
jswrenn, line 1968: Following the advice of [this guide](https://gist.github.com/Kestrer/8c05ebd4e0e9347eb05f265dfb7252e1):
```suggestion
        $crate::transmute!(*::core::include_bytes!($file))
```

jswrenn (commented):
Nice! When I [blogged about this a few years ago](https://jack.wrenn.fyi/blog/include-transmute/), I opted to match the standard library's behavior of returning a reference. I.e., this:
```rust
#[macro_export]
macro_rules! include_bytes {
    ($file:expr $(,)?) => {
        &$crate::transmute!(*::core::include_bytes!($file))
    };
}
```

I think it's fine to deviate from the standard library, but we *probably* should have some principled idea of why we're doing so.

src/lib.rs

joshlf:
> Nice! When I [blogged about this a few years ago](https://jack.wrenn.fyi/blog/include-transmute/), I opted to match the standard library's behavior of returning a reference. I.e., this:
> 
> ```rust
> #[macro_export]
> macro_rules! include_bytes {
>     ($file:expr $(,)?) => {
>         &$crate::transmute!(*::core::include_bytes!($file))
>     };
> }
> ```
> 
> I think it's fine to deviate from the standard library, but we _probably_ should have some principled idea of why we're doing so.

tbh I think the stdlib is wrong here, but I agree that we should mirror it unless we have a good reason to diverge. Done.

src/lib.rs
```diff
@@ -2069,6 +2069,51 @@ macro_rules! transmute_ref {
...
+///     let as_u32: &u32 = include_bytes!("data");
+///     assert_eq!(as_u32, &u32::from_ne_bytes([b'a', b'b', b'c', b'd']));
+///     let as_i32: &i32 = include_bytes!("data");
+///     assert_eq!(as_i32, &i32::from_ne_bytes([b'a', b'b', b'c', b'd']));
+/// }
+/// ```
+#[macro_export]
+macro_rules! include_bytes {
+    ($file:expr $(,)?) => {
+        &$crate::transmute!(*::core::include_bytes!($file))
```
joshlf, line 2113: Is the `$crate::transmute!` hygienic? The [blog post](https://gist.github.com/Kestrer/8c05ebd4e0e9347eb05f265dfb7252e1) you linked to doesn't specifically address `$crate` or naming macros by path.

jswrenn: I believe so. That blog post says:
> Declarative macros provide a very useful feature, $crate, which is a special path root that unambiguously refers to the crate in which the declarative macro was written. 

The rust documentation supports this: https://doc.rust-lang.org/beta/reference/macros-by-example.html#hygiene

jswrenn:
Agreed — I think that's a good reason for us to *not* return a reference.

src/lib.rs
joshlf: I'm specifically worried about use in older editions. IIRC, it used to be the case that `foo::bar!()` wasn't valid - you just needed to do `#[macro_use] extern crate foo;` and then `bar!()` separately?

joshlf: Discussed offline; this is good as is.

jswrenn (dismissed):

jswrenn (approved):

PR #40, opened 2022-10-10
Make more fns/macros const, update MSRV to 1.61.0

joshlf:
Make more functions and the `transmute!` macro const. This requires updating our MSRV to 1.61.0.

While we're doing this, document our MSRV in the crate root and our `README.md`. Make `.github/workflows/ci.yml` a bit more bug-proof with respect to pinned toolchain versions.


joshlf:
Closed in favor of #60.

PR #41, opened 2022-10-10
Refactor zeroing for loops with the [u8]::fill method

### LINKED ISSUE #38: Replace for loop with `[u8]::fill` method
In `map_zeroed` and `map_prefix_tuple_zeroed`, we zero a byte slice by doing:

```rust
for b in lv.0.iter_mut() {
    *b = 0;
}
```

Instead, we could use [`[u8]::fill`](https://doc.rust-lang.org/std/primitive.slice.html#method.fill):

```rust
lv.0.fill(0);
```

MSalah73:
Refactored map_zeroed & map_prefix_tuple_zeroed's for-loops with the [u8]::fill method.

Resolves #38

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1041,9 +1041,7 @@ fn map_zeroed<B: ByteSliceMut, T: ?Sized>(
 ) -> Option<LayoutVerified<B, T>> {
     match opt {
         Some(mut lv) => {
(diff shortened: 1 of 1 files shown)


joshlf (approved):
Awesome, thanks for this, @MSalah73!

PR #28, opened 2022-10-06
Fix alignment tests for targets with pointer width < 8 bytes

frazar:
The alignment tests relied on the fact that the pointer width is 8 bytes, which is not the case for i686.
Fix the alignment tests so that they work on targets with pointer width of 4 or even 2 bytes.

Fixes #21

Issue #21, opened 2022-10-04
`test_new_error` fails on i686


joshlf:
From [this test job](https://github.com/google/zerocopy/actions/runs/3185233602/jobs/5194573658):

```text
error: test failed, to rerun pass '--lib'
thread 'tests::test_new_error' panicked at 'assertion failed: LayoutVerified::<_, u64>::new(&buf.buf[4..]).is_none()', src/lib.rs:2784:9
```

Fix the bug and re-enable the test in `.github/workflows/ci.yml`.

PR #42, opened 2022-10-10
Re-enable test_new_error() for i686 target

frazar:
The `test_new_error()` test was disabled for i686 due to Issue #21. This issue has since been fixed in #28, but the test was not re-enabled.


joshlf (approved):
Oh good catch, thanks!

PR #44, opened 2022-10-11
Little typo in `README`

yvan-sraka:


joshlf (changes requested):
Hey, thanks for the fix! You'll actually need to update this in `generate-readme.sh` - `README.md` is automatically generated and shouldn't be edited directly. (You'll also need to run `./generate-readme.sh > README.md` to actually do that update.)

joshlf:
Superseded by #74.

PR #47, opened 2022-10-11
added the comment

Varad0124:
added the comment in generate-readme.sh file

### DIFF
--- a/generate-readme.sh
+++ b/generate-readme.sh
@@ -3,6 +3,8 @@
 # Copyright 2018 The Fuchsia Authors. All rights reserved.
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
+# README.md is generated automatically
+# Edits should be made in the doc comment on src/lib.rs or in generate-readme.sh
 
 set -e
 


joshlf (changes requested):
Can you put this comment in the code that is emitted into `README.md` instead of here? Basically in here: https://github.com/google/zerocopy/blob/7b1f41410ab4028e9067c2821b91342ef8020fb1/generate-readme.sh#L15-L20

joshlf (changes requested):
Hmm, I tried this change locally, and it didn't seem to have any effect. Can you update the `cat` command like this instead?

```shell
cat > $COPYRIGHT_HEADER <<EOF
<!-- Copyright 2022 The Fuchsia Authors. All rights reserved.
Use of this source code is governed by a BSD-style license that can be
found in the LICENSE file.

WARNING: DO NOT EDIT THIS FILE. It is generated automatically. Edits should be
made in the doc comment on `src/lib.rs` or in `generate-readme.sh`.
-->

EOF
```

joshlf:
Superseded by 77e95f0e929226a36e3e2f2898720cce5687fb2a.

PR #51, opened 2022-10-11
[ci.yml] Add clippy_check job

### LINKED ISSUE #49: Test Clippy in CI
Exactly what it says on the tin.

@frazar is working on this in #51

frazar:
Add a CI job to run the clippy linting software.

Fixes #49

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
           #
           # [1]
           # https://github.com/actions/runner/issues/409#issuecomment-752775072
-          components: ${{ contains(matrix.channel, 'nightly') && 'miri' || '' }}
+          components: clippy ${{ contains(matrix.channel, 'nightly') && ', miri' || '' }}
 
     # The features string contains commas which cannot be part of the cache
     # key for the Rust Cache action. Instead, we hash the features
@@ -108,6 +108,9 @@ jobs:
       #
       # TODO(#22): Re-enable testing on wasm32-wasi once it works.
       if: ${{ contains(matrix.channel, 'nightly') && matrix.target != 'wasm32-wasi' }}
+
+    - name: Clippy check
+      run: cargo +${{ matrix.channel }} clippy --target ${{ matrix.target }} --features "${{ matrix.features }}" --verbose
 
   check_fmt:
     runs-on: ubuntu-latest


frazar:
To test how this works when a clippy lint is triggered, I tried creating a dummy PR [here](https://github.com/frazar/zerocopy/actions/runs/3230148879/) that should trigger a clippy warning.

joshlf (changes requested):
Thanks for this! Can you move the check to be one of the steps under the existing "Build & Test" matrix? Specifically, I'd suggest adding it just after the existing "Check" step - it should be basically a carbon copy of that step. It'd be good to make sure we get coverage on all toolchain/target combinations.

frazar:
Sure! I didn't realize that clippy warning could differ based on toolchain/target.

joshlf:
Yeah, any code that is conditionally compiled, and possibly even code that isn't but whose behavior differs by platform (e.g., some clippy lints (like [this one](https://rust-lang.github.io/rust-clippy/master/#trivially_copy_pass_by_ref)) make use of the word size on the platform).

joshlf:
I think you can fix the current CI failures by invoking Clippy directly from the command line. It comes default with Rust installations now, so you don't need a separate action for it. All I did here was copy-paste the "Check" action and edit it to invoke `clippy` instead:

```toml
    - name: Check
      run: cargo +${{ matrix.channel }} clippy --target ${{ matrix.target }} --features "${{ matrix.features }}" --verbose
```

frazar:
Unfortunately, I didn't have much time to experiment with this. Surely your solution would be easier. I thought the GitHub Action could be more convenient since it can also show inline error messages like this:

Screenshot_20221015-083247~2

Would that be desirable for you?

I think I'll have more time to work on this next week. If instead it's a priority for you, feel free to close this PR!

joshlf:
Ah that looks really nice! Unfortunately, it [looks like it won't work](https://github.com/actions-rs/clippy-check#limitations) for people submitting PRs from forked repositories (which is basically everyone other than me). Also, since we're enforcing this (rather than just using Clippy lints as suggestions), hopefully most PRs won't have many if any lints, and so a nice UI isn't that important. I opened #69 to track making it easier to run all of our tests (including Clippy) locally.

Given those considerations, I think https://github.com/google/zerocopy/pull/51#issuecomment-1279193054 makes the most sense.

frazar:
Oh, I see! I didn't understand this limitation existed.. Trying now the changes you suggested.

frazar:
So, it seems like some clippy errors are triggered when the "alloc" features is enabled!

joshlf:
Ooops! I think I fixed the rest in #71 .

joshlf:
Looks like you'll need to push again (and maybe even rebase on `main`?) to get the tests to run against #71. I am able to trigger the tests to re-run from the UI, but it looks like it's not including #71.

frazar:
Rebased and pushed. Looks good now!

joshlf (approved):
Woo, thanks @frazar !

PR #55, opened 2022-10-12
Varadtote

varadtote:
added the comment


generate-readme.sh
```diff
@@ -3,6 +3,8 @@
 # Copyright 2018 The Fuchsia Authors. All rights reserved.
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
+# README.md is generated automatically
+# Edits should be made in the doc comment on src/lib.rs or in generate-readme.sh
```
joshlf, line 7: Can you remove these two lines of comment here? You've added it below in the `cat` command, which is where it should go.

joshlf (changes requested):

generate-readme.sh
joshlf: Nvm, I fixed it!

joshlf (approved):
Thanks for this, @varadtote!

Issue #43, opened 2022-10-10
Rust Cache warnings for tests with more than one feature enabled


frazar:
Looking at [this CI log](https://github.com/google/zerocopy/actions/runs/3222541405/jobs/5271716415#step:4:30)
```
Warning:  ValidationError: Key Validation Error: v0-rust-1.56.1-i686-unknown-linux-gnu-alloc,simd--build_test-a732c1bed0bf174cd2e9b37132d29ed1815eb7cd-df69c652838aa8f1d3f759765b37bdf011baf599 cannot contain commas.
    at checkKey (/home/runner/work/_actions/Swatinem/rust-cache/v2.0.0/dist/restore/index.js:58:15)
    at Object.<anonymous> (/home/runner/work/_actions/Swatinem/rust-cache/v2.0.0/dist/restore/index.js:90:13)
    at Generator.next (<anonymous>)
    at /home/runner/work/_actions/Swatinem/rust-cache/v2.0.0/dist/restore/index.js:15:71
    at new Promise (<anonymous>)
    at __webpack_modules__.7799.__awaiter (/home/runner/work/_actions/Swatinem/rust-cache/v2.0.0/dist/restore/index.js:11:12)
    at Object.restoreCache (/home/runner/work/_actions/Swatinem/rust-cache/v2.0.0/dist/restore/index.js:80:12)
    at run (/home/runner/work/_actions/Swatinem/rust-cache/v2.0.0/dist/restore/index.js:62066:40)
```
it seems like Rust Cache does not like commas in keys.

This occurs because keys [defined as](https://github.com/google/zerocopy/blob/6aea74e78a85f7e8227d485f7e11ae4957329120/.github/workflows/ci.yml#L49)
```
key: "${{ matrix.channel }}-${{ matrix.target }}-${{ matrix.features }}-${{ hashFiles('**/Cargo.lock') }}"
```

and they end up containing commas when `${{ matrix.features }}` expands to two or more features, like "alloc,simd,simd-nightly".

Maybe the `key` definition should be changed to a comma-less verison. Also, a way to turn RustCache warnings into hard errors should be considered.

joshlf:
Thanks for catching this! Definitely should fix, although I'm a bit on the fence about making RustCache warnings hard errors - the tests still test what we want them to, they just run somewhat more slowly. I wonder if there's a middle ground where we can make sure we're notified of failures but they don't block the rest of the tests from running, and they don't block PRs from being merged?

joshlf:
[...]

PR #59, opened 2022-10-14
Fix Rust Cache invalid key warning

AntoniosBarotsis:
Addresses #43 by using the length of the feature string instead of the features themselves. This avoids the commas which were causing the issues and the key should be unique so long as the feature strings are of different lengths (which for the time being is the case).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,6 @@
 name: Build & Tests
 
 on:
-  push:
   pull_request:
 
 env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,9 +45,7 @@ jobs:
 
     - name: Set feature string for cache key
       run: |
-        features=${{matrix.features}}
-        features_key=${#features}
-        echo "FEATURES=$(echo $features_key)" >> $GITHUB_ENV
+        echo "FEATURES_HASH=$(echo ${{ matrix.features }} | sha256sum | cut -d ' ' -f 1)" >> $GITHUB_ENV
 
     - name: Rust Cache
       uses: Swatinem/rust-cache@v2.0.0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,13 +44,16 @@ jobs:
           # https://github.com/actions/runner/issues/409#issuecomment-752775072
           components: ${{ contains(matrix.channel, 'nightly') && 'miri' || '' }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -43,10 +43,16 @@ jobs:
           # https://github.com/actions/runner/issues/409#issuecomment-752775072
           components: ${{ contains(matrix.channel, 'nightly') && 'miri' || '' }}
 
+    - name: Set feature string for cache key
+      run: |
+        features=${{matrix.features}}
+        features_key=${#features}
+        echo "FEATURES=$(echo $features_key)" >> $GITHUB_ENV
```
joshlf, line 50: This is really clever! So long as you're executing arbitrary shell commands, maybe you could just hash the feature set instead so this will keep working even if we have two feature sets of the same character length?

```yaml
    - name: Set feature string for cache key
      run: |
        echo "FEATURES_HASH=$(echo ${{ matrix.features }} | sha256sum | cut -d ' ' -f 1)" >> $GITHUB_ENV
```

joshlf (changes requested):

.github/workflows/ci.yml
```diff
@@ -43,10 +44,14 @@ jobs:
           # https://github.com/actions/runner/issues/409#issuecomment-752775072
           components: ${{ contains(matrix.channel, 'nightly') && 'miri' || '' }}
 
+    - name: Set feature string for cache key
+      run: |
+        echo "FEATURES_HASH=$(echo ${{ matrix.features }} | sha256sum | cut -d ' ' -f 1)" >> $GITHUB_ENV
```
joshlf, line 49: Can you add a comment here explaining why we need to do this?

joshlf (changes requested):

.github/workflows/ci.yml
AntoniosBarotsis: That's actually a great idea, just added that!

.github/workflows/ci.yml
AntoniosBarotsis: I wrote the following, let me know if you think I should mention something else

```md
    # The features string contains commas which cannot be part of the cache
    # key for the Rust Cache action. Instead, we hash the features
    # to get a string of legal characters.
```

joshlf: Perfect!

joshlf (approved):
Awesome, thanks for this!!

PR #60, opened 2022-10-14
Make Unalign::into_inner const

joshlf:


src/lib.rs
```diff
@@ -817,9 +817,24 @@ impl<T> Unalign<T> {
     }
 
     /// Consumes `self`, returning the inner `T`.
-    pub fn into_inner(self) -> T {
-        let Unalign(val) = self;
-        val
+    pub const fn into_inner(self) -> T {
+        // Use this instead of `mem::transmute` since the latter can't tell
+        // that `Unalign<T>` and `T` have the same size.
+        union Transmute<T> {
```
djkoloski, line 823: This union should be `repr(C)` to guarantee that all members are aligned to the beginning of the union's storage.

joshlf: Oh crap good catch! Fixed.

djkoloski (approved):

PR #79, opened 2022-10-19
[DO NOT MERGE] Validate that #70 catches MSRV mismatch

joshlf:
This PR builds on #70 and adds a commit which intentionally introduces an MSRV mismatch to make sure that the CI test in #70 catches it.


joshlf:
Closing this now that #70 has merged.

PR #96, opened 2022-10-22
[CI] Use `--all-targets` with `cargo check`

### LINKED ISSUE #95: [CI] Use cargo check --all-targets flag

jswrenn:
Ensures that examples, benchmarks, etc. are also checked.

Fixes #95.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,7 +122,7 @@ jobs:
         key: "${{ env.ZC_TOOLCHAIN }}-${{ matrix.target }}-${{ env.FEATURES_HASH }}-${{ hashFiles('**/Cargo.lock') }}"
 
     - name: Check
-      run: cargo +${{ env.ZC_TOOLCHAIN }} check --manifest-path ${{ matrix.manifest-path }} --target ${{ matrix.target }} --features "${{ matrix.features }}" --verbose
+      run: cargo +${{ env.ZC_TOOLCHAIN }} check --manifest-path ${{ matrix.manifest-path }} --target ${{ matrix.target }} --features "${{ matrix.features }}" --all-targets --verbose
 
     - name: Build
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -122,7 +122,7 @@ jobs:
         key: "${{ env.ZC_TOOLCHAIN }}-${{ matrix.target }}-${{ env.FEATURES_HASH }}-${{ hashFiles('**/Cargo.lock') }}"
 
     - name: Check
-      run: cargo +${{ env.ZC_TOOLCHAIN }} check --manifest-path ${{ matrix.manifest-path }} --target ${{ matrix.target }} --features "${{ matrix.features }}" --verbose
+      run: cargo +${{ env.ZC_TOOLCHAIN }} check --manifest-path ${{ matrix.manifest-path }} --target ${{ matrix.target }} --features "${{ matrix.features }}" --verbose --all-targets
```
joshlf, line 125: Uber nit: Can you swap the flag order so it's `--all-targets --verbose` at the end?

joshlf (changes requested):
Thanks! One uber nit, but otherwise looks good.

.github/workflows/ci.yml
joshlf: Actually, another thought: Can you also add `--all-targets` to the `cargo build` step?

jswrenn: Might I suggest scrapping the `cargo check` step entirely in favor of `cargo build --all-targets`?

joshlf: Good question; I'll look into that and we can follow up in a separate PR if it turns out that's the way to go.

joshlf:
Looks like the current errors are caused by not passing the right linker as described in https://github.com/rust-lang/rust/issues/28924. It may be worth just punting on this for `cargo build` and only doing it for `cargo check` unless there's an easy way to provide the linker flag that doesn't significantly increase the complexity of our CI tests or their configs.

joshlf:
This is captured by our build matrix.

PR #99, opened 2022-10-23
[derive] document Rust changes to priv-in-pub lint

jswrenn:
This commit corrects outdated code comments in zerocopy-derive. As of Rust 1.59.0, it is no longer an error to name private types in the `where` clauses of public trait impls for public types.

### DIFF
--- a/zerocopy-derive/src/lib.rs
+++ b/zerocopy-derive/src/lib.rs
@@ -426,28 +426,7 @@ fn impl_block<D: DataExt>(
     // While there's no getting around this requirement for us, it does have
     // some pretty serious downsides that are worth calling out:
     //
-    // 1. You lose the ability to have fields of generic type with reduced visibility.
-    //
-    //     #[derive(Unaligned)]
-    //     #[repr(C)]
-    //     pub struct Public<T>(Private<T>);
-    //
-    //     #[derive(Unaligned)]
-    //     #[repr(C)]
-    //     struct Private<T>(T);
-    //
-    //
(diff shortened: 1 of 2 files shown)


zerocopy-derive/tests/priv_in_pub.rs
```diff
@@ -0,0 +1,20 @@
+// Copyright 2019 The Fuchsia Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+use zerocopy::{AsBytes, FromBytes, Unaligned};
+
+// These derives do not result in E0446 as of Rust 1.59.0, because of
+// https://github.com/rust-lang/rust/pull/90586
```
joshlf, line 8: Uber nit: Trailing period? (See the section on comments in `CONTRIBUTING.md`)

zerocopy-derive/src/lib.rs
```diff
@@ -426,28 +426,7 @@ fn impl_block<D: DataExt>(
     // While there's no getting around this requirement for us, it does have
```
joshlf, line 426: Now that there's only one item in this list, can you make it not a list? Something like:

```rust
    // While there's no getting around this requirement for us, it does have the
    // pretty serious downside that, when lifetimes are involved, the trait
    // solver ties itself in knots:
```

joshlf (changes requested):
This is awesome! I had no idea this had been resolved. Two tiny nits, but otherwise looks good.

joshlf (approved):

PR #104, opened 2022-10-26
Cargo.toml: use `license` instead of `license-file`

gburgessiv:
When using a standard SPDX license, Cargo.toml should only have the license field. `license-file` is for use when the license being used is nonstandard, per:
https://doc.rust-lang.org/cargo/reference/manifest.html


joshlf (dismissed):
Looks good, thanks @gburgessiv!

joshlf:
Looks like you'll need to add `--no-license` to the `cargo readme` invocation in `generate-readme.sh` to fix [this build failure](https://github.com/google/zerocopy/actions/runs/3333266895/jobs/5515284071).

gburgessiv:
Thanks for the quick review! Updated as requested :)

joshlf (approved):

PR #106, opened 2022-10-27
[CI] Optimize (and fix) caching

### LINKED ISSUE #85: Optimize caching in CI
In CI, we use the [Swatinem/rust-cache](https://github.com/Swatinem/rust-cache) GitHub Action to cache certain build artifacts across runs. However, it seems from the README that we might be using it in a suboptimal way. A few observations:
- Zerocopy has only one dependency, and `rust-cache` only caches dependencies, not crate artifacts (note that zerocopy-derive has more dependencies - and dependencies that are expensive to compile - so presumably benefits much more than zerocopy)
- We generate a custom cache key which includes:
  - A hash of the Cargo features, but this presumably shouldn't have any effect on dependencies (at least in our case, since none of our Cargo features depend on dependencies' Cargo features)
  - A hash of `**/Cargo.lock`, but this is documented to be the default behavior anyway

We should:
- Figure out if there's a way to squeeze more benefit out for zerocopy (can we get `rust-cache` to cache any more artifacts?)
- Don't include Cargo features in the cache
- Don't include a hash of `**/Cargo.lock` in the cache
- Go back over our use of `rust-cache` and its README with a fine toothed comb and see what else we're doing suboptimally

cc @Swatinem in case there's anything obvious that jumps out at you!

glokta1:
- Remove redundant flags when using rust-cache
- Add 'push to main' as trigger to reuse cache across PRs. See https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows\#restrictions-for-accessing-a-cache

Resolves #85


joshlf (approved):
Looks great, thanks @glokta1!

joshlf:
I'm getting an error that might be related to the contents of this PR. I've had issues on mobile before, so it also might just be that. I'll try again later through the web UI and see if that works.

image

glokta1:
@joshlf Hey, can I ask you to please add the 'Hacktoberfest-Accepted' label? Thanks!

joshlf:
Done!

Issue #109, opened 2022-10-28
Simplify zerocopy-derive-generated code


joshlf:
We ought to be able to simplify the code generated by zerocopy-derive's `AsBytes` impl by adding the following module to zerocopy:

```rust
/// Utilities used by `zerocopy-derive`.
///
/// These are defined in `zerocopy` rather than in code generated by
/// `zerocopy-derive` so that they can be compiled once rather than recompiled
/// for every pair of type and trait (in other words, if they were defined in
/// generated code, then deriving `AsBytes` and `FromBytes` on three different
/// types would result in the code in question being emitted and compiled six
/// different times).
#[doc(hidden)]
#[allow(missing_debug_implementations)]
pub mod derive_util {
    /// Implemented for `Bool<true>`.
    pub trait True {}

    /// A boolean constant value which can be used in type bounds.
    pub struct Bool<const TERM: bool>();

    impl True for Bool<true> {}

    /// Does the struct type `$t` have padding?
    ///
    /// `$ts` is the list of the type of every field in `$t`. `$t` must be a
    /// struct type, or else `has_padding!`'s result may be meaningless.
    ///
    /// Note that `has_padding!`'s results are independent of `repr` since they
    /// only consider the size of the type and the sizes of the fields. Whatever
    /// the repr, the size of the type already takes into account any padding
    /// that the compiler has decided to add. Note that while this is *probably*
    /// also true of `repr(rust)`, the author is not confident of that fact, and
    /// it should not be relied upon for soundness.
    #[doc(hidden)] // `#[macro_export]` bypasses this module's `#[doc(hidden)]`.
    #[macro_export]
    macro_rules! has_padding {
        ($t:ty, $($ts:ty),*) => {
            core::mem::size_of::<$t>() > 0 $(+ core::mem::size_of::<$ts>())*
        };
    }
}
```

Then, for a type like:

```rust
#[derive(AsBytes)]
#[repr(C)]
struct Foo(u8, u16);
```

...we'd emit an impl like:

```rust
unsafe impl AsBytes for Foo where Bool<{!zerocopy::has_padding!(Foo, u8, u16)}>: True {
    fn only_derive_is_allowed_to_implement_this_trait()
    where
        Self: Sized,
    {}
}
```

Notably, this ought to make it a very simple change to support type parameters by emitting code like:

```rust
unsafe impl AsBytes for Bar<T> where Bool<{!zerocopy::has_padding!(Bar<T>, T)}>: True {
[...]

PR #113, opened 2022-11-01
Move padding checks into `derive_util` macros

djkoloski:
This moves the padding checks into macros in a new `derive_util` module and emits implementations with an additional where clause. Some of the proposed names from #109 has been slightly changed to make the emitted error messages more understandable:

```
error[E0277]: the trait bound `HasPadding<Foo, true>: ShouldBe<false>` is not satisfied
 --> tests/ui-stable/asbytes-illegal.rs:5:10
  |
5 | #[derive(zerocopy::AsBytes)]
  |          ^^^^^^^^^^^^^^^^^ the trait `ShouldBe<false>` is not implemented for `HasPadding<Foo, true>`
  |
  = help: the trait `ShouldBe<VALUE>` is implemented for `HasPadding<T, VALUE>`
```


tests/ui/asbytes-illegal.rs
```diff
@@ -0,0 +1,17 @@
+// Copyright 2022 The Fuchsia Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#[derive(zerocopy::AsBytes)]
```
joshlf, line 5: Can you do `use zerocopy::AsBytes` at the top and then `#[derive(AsBytes)]` here (just to be consistent with other UI tests)? Same below.

zerocopy-derive/src/lib.rs
```diff
@@ -552,39 +552,38 @@ fn impl_block<D: DataExt>(
...
-                impl FieldsAreSameSize<FIELDS_ARE_SAME_SIZE> for #type_ident {}
-                let _ = assert_fields_are_same_size::<#type_ident>;
-            };
-        ),
-    };
+    match (field_types.is_empty(), padding_check) {
+        (true, _) | (false, PaddingCheck::None) => (),
+        (false, PaddingCheck::Struct) => {
+            let fields = field_types.iter();
+            // `parse_quote!` doesn't parse macro invocations in const generics
```
joshlf, line 559: Can you add to this comment showing an example of what the below code generates? It's hard to see what it's doing on its own. Same in the last match arm.

joshlf (changes requested):
This looks awesome! Thanks so much for doing this! A few small comments inline, but also: It looks like the new code generates padding errors during an earlier compiler pass than most of the other errors (at least that's my guess as to why many errors are removed in this diff). Could you split up the UI test source files so that we still get test coverage for those errors?

djkoloski:
The `syn` issue mentioned can be fixed by enabling the `full` feature. I timed a few runs of `cargo clean; cargo build -p zerocopy-derive`:

```
"full" off:
    real    0m4.433s
    user    0m8.858s
    sys     0m2.084s

    real    0m4.382s
    user    0m8.990s
    sys     0m1.963s

    real    0m4.447s
    user    0m8.900s
    sys     0m2.032s

    real    0m4.364s
    user    0m8.859s
    sys     0m2.005s

    real    0m4.339s
    user    0m8.678s
    sys     0m2.197s

4.393 +-0.1s

"full" on:
    real    0m5.643s
    user    0m11.430s
    sys     0m2.180s

    real    0m5.840s
    user    0m11.407s
    sys     0m2.192s

    real    0m5.577s
    user    0m11.174s
    sys     0m2.202s

    real    0m5.608s
    user    0m11.192s
    sys     0m2.183s

    real    0m5.720s
    user    0m11.271s
    sys     0m2.243s

5.6776 =- 0.23s
```

So it looks like enabling `full` adds about 1.2s to each build of zerocopy-derive (+29.2% for this package alone). Should we continue with a workaround or enable the `full` feature?

joshlf:
Yeah let's stick to the hack; it's pretty minor, and that's a lot of compilation time to add.

joshlf:
> It looks like the new code generates padding errors during an earlier compiler pass than most of the other errors (at least that's my guess as to why many errors are removed in this diff). Could you split up the UI test source files so that we still get test coverage for those errors?

Looks like this wasn't addressed yet - just a heads up in case you're waiting for a review on the latest changes. If you were planning on addressing this later, then just disregard this comment.

zerocopy-derive/tests/ui/padding_check.rs
```diff
@@ -0,0 +1,27 @@
+// Copyright 2019 The Fuchsia Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#[macro_use]
+extern crate zerocopy;
+
+fn main() {}
+
+// These tests exercise our checks for padding bytes in types.
```
joshlf, line 10: Can you also mention why they're in a separate file from the other tests?

zerocopy-derive/tests/ui/padding_check.rs
```diff
@@ -0,0 +1,27 @@
...
+
+#[repr(C, align(4))]
+struct AlignedU32(u32);
+
+#[derive(AsBytes)]
+#[repr(C)]
+struct AsBytesStruct {
+    foo: u8,
+    bar: AlignedU32,
```
joshlf, line 19: Just so I understand correctly, are you using this type as opposed to `u32` because the latter's alignment isn't 4 on all platforms?

zerocopy-derive/tests/ui/padding_check.stderr
```diff
@@ -0,0 +1,21 @@
+error[E0277]: the trait bound `HasPadding<AsBytesStruct, true>: ShouldBe<false>` is not satisfied
+  --> tests/ui/padding_check.rs:15:10
+   |
+15 | #[derive(AsBytes)]
+   |          ^^^^^^^ the trait `ShouldBe<false>` is not implemented for `HasPadding<AsBytesStruct, true>`
+   |
+   = help: the trait `ShouldBe<VALUE>` is implemented for `HasPadding<T, VALUE>`
+   = help: see issue #48214
+   = help: add `#![feature(trivial_bounds)]` to the crate attributes to enable
```
joshlf, line 9: Uh oh, it looks like this feature actually just breaks the `where` bound approach entirely: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=2ad3a2c4527a3f8b5b3baa33a4c0dfa6

joshlf (changes requested):

zerocopy-derive/tests/ui/padding_check.rs
djkoloski: Yep, the alignment of `u32` could be as low as 1 so I manually set the align to 4 to guarantee padding bytes.

zerocopy-derive/tests/ui/padding_check.stderr
djkoloski: If you actually use it, then you get an error: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=0adc598549ab840bce52c61705142053

This is true of all generic bounds, so if the plan is to eventually make this compatible with generics then I think it's the right approach. Maybe for non-generic types we could emit a constant to force immediate resolution? Something like:

```rs
fn must_be_as_bytes<T: AsBytes>() {}
const _: fn() = must_be_as_bytes::<Foo>;
```

That would force resolution for concrete types.

joshlf: OK yeah that sounds good to me! I didn't realize that this would fail at resolution time. Forcing immediate resolution sounds like the right approach.

zerocopy-derive/tests/ui/padding_check.rs
djkoloski: Originally I thought I needed them in a separate pass. Moved back into the main files.

zerocopy-derive/tests/ui/padding_check.stderr
djkoloski: Added a constant usage for concrete types.

zerocopy-derive/tests/ui/late_compile_pass.rs
```diff
@@ -24,11 +24,13 @@ struct FromBytes1 {
 // AsBytes errors
 //
 
+#[repr(C)]
+struct MyU32(u32);
```
joshlf, line 28: Derive `AsBytes` for this? Looks like the errors you're getting are now because this doesn't implement `AsBytes`, not because of padding in `AsBytes1`.

zerocopy-derive/tests/ui/struct.rs
```diff
@@ -15,6 +15,16 @@ fn main() {}
 #[repr(C)]
 struct AsBytes1<T>(T);
 
+#[repr(C, align(4))]
+struct AlignedU32(u32);
```
joshlf, line 19: Same here - `derive(AsBytes)`.

joshlf (changes requested):

zerocopy-derive/tests/ui/late_compile_pass.rs
djkoloski: That's the intent of this test, I renamed a few things to make it more obvious. Because the check for `AsBytes` happens in a later compile pass, it needs to be tested for in this file.

zerocopy-derive/tests/ui/struct.rs
djkoloski: This is fine to derive because the padding issue will error anyway. Adding this gives the test a higher signal to noise ratio so I think it's a good add.

zerocopy-derive/tests/ui/late_compile_pass.rs
joshlf: Ah I understand; thanks!

joshlf (approved):

PR #118, opened 2022-11-03
Don't assume primitive type alignments

### LINKED ISSUE #116: Don't assume primitive alignments
Currently, a lot of our test code assumes that all primitive numerical types have an alignment equal to their size. This isn't true on all platforms, and so we should replace these uses with types with guaranteed alignment.

Credit to @djkoloski for pointing this out in https://github.com/google/zerocopy/pull/113#discussion_r1011042531.

joshlf:
Previously, we used the `u16` and `u64` types in a number of tests,
relying on them having the alignments 2 and 8 respectively. While those
types have those alignments on many platforms, those are not guaranteed
to be their alignments on all platforms. In this commit, we replace
most of these uses with types of the same size but guaranteed alignment.
In a few places, we have to leave the uses as-is, but this isn't a big
deal since our tests run in CI on platforms where the alignments are
actually the values we're relying upon.

Closes #116


joshlf:
Hey @djkoloski , mind reviewing this one? (Looks like I can't add you as a reviewer, at least not yet - I might be able to once you've commented on this thread)

djkoloski:
Yep, feel free to add me as a reviewer. 🙂

djkoloski (approved):
It would be neat if we could put this helper type in a `mod`, but I like how simple the test glob is in `trybuild.rs`. Maybe if more commonly used helper types are added, we could add some extra logic to handle that.

joshlf:
Do you think it'd be worth using [`include!`](https://doc.rust-lang.org/stable/std/macro.include.html)?

djkoloski:
Eh, `include!` for source is also pretty nasty. Maybe we could add the module, keep the glob, and "test" that the helper module doesn't have any compiler errors?

joshlf:
Good call; done! I was able to use the `#[path = ...]` attribute to make the file usable from both UI and non-UI tests.

djkoloski (approved):

PR #119, opened 2022-11-03
Move field type bounds into where clauses

djkoloski:
This cleans up some of the error messages without introducing any functional changes. This will also make it easier to enable deriving on generic types in the future.


joshlf (approved):
Nice, this is so much simpler!

Issue #69, opened 2022-10-16
Make it easier to run CI tests locally


joshlf:
We currently run a large suite of tests in CI, which includes testing with different tools across different toolchains and different targets. It also involves miscellaneous things like confirming that `README.md` is up to date, that code is properly formatted with `rustfmt`, etc. In other words, it's not as simple as just running `cargo test`.

As a result, it's easy to make changes that appear fine locally but break CI, requiring an edit-commit-push cycle just to see if all of the CI tests will pass. We should make it easier to run most or all of these tests locally. For especially fast tests (like `README.md` and code formatting), we could even provide them as commit hooks.

@djkoloski also reports finding this to be painful (https://github.com/google/zerocopy/issues/69#issuecomment-1304693936):

> I struggled specifically with formatting and forgetting to make stderr match in tests. Adding a way to format test files and an easier way to update all the stderr files for tests might be helpful. 🙂

The challenge here is going to be making these tests available without either a) moving everything out of our `ci.yml` and into a script (which means we don't get nice job output and error messages) or, b) duplicating everything (which is brittle).

AntoniosBarotsis:
You could also look into something like [Toast](https://github.com/stepchowfun/toast).

memark:
Some interesting tools in this space are
- [`act`](https://github.com/nektos/act)
- [`earthly.dev`](https://earthly.dev)

AntoniosBarotsis:
One thing to keep in mind is that Act and I think Toast as well don't support Windows images (or at least didn't last time I tried IIRC).

djkoloski:
I struggled specifically with formatting and forgetting to make stderr match in tests. Adding a way to format test files and an easier way to update all the stderr files for tests might be helpful. 🙂

joshlf:
Awesome, added that to the issue description. This'll definitely be important to solve moving forward.

joshlf:
@djkoloski I took a stab at pulling the scripts out of `ci.yml` and into a shell script [here](https://github.com/google/zerocopy/compare/main...test-scripts#diff-5ebe12a1851cda0190cf2ed6fcf42049b071c8eeaa9c495dc20b0adab7b1f773). There's still work to be done, but I probably won't get it to a state where I can merge for a few days, so I figured I'd send you that in case you want to just keep a copy of it locally until I get it merged.

joshlf:
[...]

PR #120, opened 2022-11-03
Transparent generic structs

### LINKED ISSUE #9: Support generic parameters when deriving `AsBytes` on a `#[repr(transparent)]` type
*Migrated from https://fxbug.dev/84475*

*Originally reported by @Kestrer*

Hi, I'm not quite sure how to open an issue with Zerocopy directly so I hope I put this in the right place.

I'm the maintainer of the `bounded-integer` crate (https://docs.rs/bounded-integer) and I would like to support implementing the Zerocopy traits on the bounded integers. My code looks something like this:

```rust
type Inner = u8;

#[derive(Unaligned, AsBytes)]
#[repr(transparent)]
pub struct Bounded<const MIN: Inner, const MAX: Inner>(Inner);
```

However `zerocopy` complains because it doesn't support generic parameters.

*Comment by @cmyr*

I've also just run into this, and I'd be happy to prepare a patch if that would be welcome (and if I can remember how gerrit works).

I've taken a quick skim through the current impl, and stared into space for a few minutes, and it isn't obvious to me why this constraint exists, currently. I trust there is a reason (given that the limitation does not exist for FromBytes) but it isn't obvious to me what it is, so any useful information here would be welcome. :)

*Comment by @joshlf*

Hi folks, sorry for the delay in responding to this.

Generic parameters are supported for `Unaligned` - try deriving that trait on its own without `AsBytes`.

The reason that `AsBytes` doesn't support generic parameters is that I haven't figured out a way to ensure that the resulting type has no padding, which is a requirement for `AsBytes`. In particular, consider the following type:

```rust
#[repr(C)]
struct Foo<T, U> {
    t: T,
    u: U,
}
```

In order for `Foo<T, U>` to be `AsBytes`, there can't be any padding. That, however, depends on both the sizes and alignments of `T` and `U`. I haven't figured out how to write an impl bound that expresses the right requirements.

Take a look at [the implementation of the `AsBytes` derive](https://cs.opensource.google/fuchsia/fuchsia/+/master:src/lib/zerocopy/zerocopy-derive/src/lib.rs;l=569;drc=d2a5240f1b3f83e8d0e9b44a06e3e3a04a4d7b8b) for how we do this when there are no generic parameters - we use const code to compare the size of the type to the sum of the sizes of its fields. That approach doesn't translate since const generics aren't powerful enough yet to just use the same code in an impl bound. We'd need another approach.

*Comment by @joshlf*

Update: Some of the other developers realized that we can probably support your specific use case more easily since your type is `#[repr(transparent)]`, and thus we don't need to worry about any layout computation - we just need to bound `Inner: AsBytes`. I'm changing this issue to track that more narrow goal.

djkoloski:
Fixes #9. Depends on #119 because the generated generic bounds cannot be used in a constant context.

### DIFF
--- a/zerocopy-derive/src/lib.rs
+++ b/zerocopy-derive/src/lib.rs
@@ -195,16 +195,26 @@ fn derive_from_bytes_union(ast: &DeriveInput, unn: &DataUnion) -> proc_macro2::T
 // - `repr(packed)`
 
 fn derive_as_bytes_struct(ast: &DeriveInput, strct: &DataStruct) -> proc_macro2::TokenStream {
-    // TODO(#10): Support type parameters.
-    if !ast.generics.params.is_empty() {
-        return Error::new(Span::call_site(), "unsupported on types with type parameters")
-            .to_compile_error();
+    let reprs = try_or_print!(STRUCT_UNION_AS_BYTES_CFG.validate_reprs(ast));
+    let is_transparent = reprs.contains(&StructRepr::Transparent);
+
+    // TODO(#10): Support type parameters for non-transparent structs.
+    if !ast.generics.params.is_empty() && !is_transparent {
+        return Error::new(
+            Span::call_site(),
+            "unsupported on generic structs that are not repr(transparent)",
+        )
+        .to_compile_error();
     }
 
-    let reprs = try_or_print!(STRUCT_UNION_AS_BYTES_CFG.validate_reprs(ast));
-    let padding_check =
-        if reprs.contains(&StructRepr::Packed) { PaddingCheck::None } else { PaddingCheck::Struct };
-
+    // We don't need a padding check if the struct is repr(transparent) since
+    // the layout and ABI of the whole struct is the same as its only non-ZST
+    // field, and we require that field to be `AsBytes`.
+    let padding_check = if is_transparent || reprs.contains(&StructRepr::Packed) {
+        PaddingCheck::None
+    } else {
+        PaddingCheck::Struct
+    };
     impl_block(ast, strct, "AsBytes", true, padding_check)
 }
 
--- a/zerocopy-derive/tests/ui/derive_transparent.rs
+++ b/zerocopy-derive/tests/ui/derive_transparent.rs
@@ -1,0 +1,35 @@
+// Copyright 2019 The Fuchsia Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+extern crate zerocopy;
+
+#[path = "../util.rs"]
+#[macro_use]
+mod util;
+
+use self::util::{NotAsBytes, AU16};
+
+fn main() {}
+
+use core::marker::PhantomData;
+use zerocopy::{AsBytes, FromBytes, Unaligned};
+
+// Test generic transparent structs
+
+#[derive(AsBytes, FromBytes, Unaligned)]
+#[repr(transparent)]
+struct TransparentStruct<T> {
+    inner: T,
+    _phantom: PhantomData<()>,
+}
+
+// A type that does not implement `AsBytes`.
+pub struct NotAsBytes;
+
+// It should be legal to derive these traits on a transparent struct, but it
+// must also ensure the traits are only implemented when the inner type
+// implements them.
+assert_is_as_bytes!(TransparentStruct<NotAsBytes>);
+assert_is_from_bytes!(TransparentStruct<char>);
+assert_is_unaligned!(TransparentStruct<AU16>);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3251,4 +3251,22 @@ mod tests {
             assert_eq!(<()>::new_zeroed(), ());
         }
     }
+
+    #[test]
+    fn test_transparent_generic_struct() {
+        #[derive(AsBytes, FromBytes, Unaligned)]
+        #[repr(transparent)]
+        struct Foo<T> {
+            _bar: T,
+            _phantom: PhantomData<()>,
+        }
+
+        fn assert_impls_asbytes<T: AsBytes>() {}
+        fn assert_impls_frombytes<T: FromBytes>() {}
+        fn assert_impls_unaligned<T: Unaligned>() {}
+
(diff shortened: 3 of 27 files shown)


zerocopy-derive/src/lib.rs
```diff
@@ -195,16 +195,20 @@ fn derive_from_bytes_union(ast: &DeriveInput, unn: &DataUnion) -> proc_macro2::T
...
 fn derive_as_bytes_struct(ast: &DeriveInput, strct: &DataStruct) -> proc_macro2::TokenStream {
-    // TODO(#10): Support type parameters.
-    if !ast.generics.params.is_empty() {
+    let reprs = try_or_print!(STRUCT_UNION_AS_BYTES_CFG.validate_reprs(ast));
+    let is_transparent = reprs.contains(&StructRepr::Transparent);
+
+    // TODO(#10): Support type parameters for non-transparent structs.
+    if !ast.generics.params.is_empty() && !is_transparent {
         return Error::new(Span::call_site(), "unsupported on types with type parameters")
             .to_compile_error();
```
joshlf, line 204: Maybe change this to "unsupported on non-repr(transparent) types with type parameters" or something similar?

zerocopy-derive/src/lib.rs
```diff
@@ -195,16 +195,20 @@ fn derive_from_bytes_union(ast: &DeriveInput, unn: &DataUnion) -> proc_macro2::T
...
     }
 
-    let reprs = try_or_print!(STRUCT_UNION_AS_BYTES_CFG.validate_reprs(ast));
-    let padding_check =
-        if reprs.contains(&StructRepr::Packed) { PaddingCheck::None } else { PaddingCheck::Struct };
-
+    let padding_check = if is_transparent || reprs.contains(&StructRepr::Packed) {
```
joshlf, line 207: Can you add a comment explaining why `is_transparent` implies that we don't need a padding check? Off the top of my head, this hangs on two factors (but maybe there's more I'm missing that needs to go into a justification):
- Rust itself will enforce that `repr(transparent)` only compiles if there's exactly one non-ZST field
- `repr(transparent)` means that the layout is identical to that of the non-ZST type, so `AsBytes` validity here is equivalent to `AsBytes` validity for that type, which is representable with a trait bound

joshlf (changes requested):
This is awesome!! Would you mind rebasing on `main` now that #119 has merged so that the diff doesn't include those changes as well?

zerocopy-derive/src/lib.rs
djkoloski: Done

zerocopy-derive/src/lib.rs
djkoloski: Yep, and both of those are true. I left a comment that I think conveys that.

zerocopy-derive/tests/ui/derive_transparent.rs
```diff
@@ -0,0 +1,27 @@
...
+struct TransparentStruct<T> {
+    inner: T,
+    _phantom: PhantomData<()>,
+}
+
+fn assert_is_asbytes<T: AsBytes>() {}
+fn assert_is_frombytes<T: FromBytes>() {}
+fn assert_is_unaligned<T: Unaligned>() {}
+
+const _: fn() = assert_is_asbytes::<TransparentStruct<(u8, u32)>>;
```
joshlf, line 25: Can you use a type that's guaranteed to not be `AsBytes` here? `u32` could have alignment 1 on some platforms, and the layout of tuples might be guaranteed in the future. Maybe just a `struct NotAsBytes` or `MaybeUninit<u8>` or something?

joshlf (changes requested):
One more tiny nit, but otherwise lgtm!

zerocopy-derive/tests/ui/derive_transparent.rs
```diff
@@ -0,0 +1,29 @@
...
+
+fn assert_is_asbytes<T: AsBytes>() {}
+fn assert_is_frombytes<T: FromBytes>() {}
+fn assert_is_unaligned<T: Unaligned>() {}
+
+struct NotAsBytes;
+
+const _: fn() = assert_is_asbytes::<TransparentStruct<NotAsBytes>>;
+const _: fn() = assert_is_frombytes::<TransparentStruct<char>>;
+const _: fn() = assert_is_unaligned::<TransparentStruct<u32>>;
```
joshlf, line 29: Same here - maybe we can merge #118 first and then you can use `AU16`?

joshlf (changes requested):

zerocopy-derive/tests/ui/derive_transparent.rs
djkoloski: Yep, sorry missed the update on that PR. I left an approval and will rebase this once it merges.

joshlf: Merged!

djkoloski:
I wanted to incorporate some additional `util` module cleanup but it seems pretty difficult to have separate PRs that are inter-related on GitHub. I kind of miss gerrit?? I can separate those changes into a separate PR if you'd prefer.

joshlf:
> I wanted to incorporate some additional `util` module cleanup but it seems pretty difficult to have separate PRs that are inter-related on GitHub. I kind of miss gerrit?? I can separate those changes into a separate PR if you'd prefer.

Yeah, I miss gerrit all the time 😞 GitHub's UI is great, but the fact that you can't stack PRs is a travesty. I haven't been doing dev on GH long enough to know the real tricks, but so far my approach has been to just carve off one commit at a time to put through the PR cycle. It's a lot easier for me since I'm self-approving a lot of the more minor changes. I feel like where stacked PRs is a huge deal is when you want to have multiple reviews ongoing at a time.

joshlf:
Oh but to respond to your comment about this PR: I saw, and those changes look great! I'm happy to just keep them in this PR and we can figure out a better workflow later if it keeps being a problem.

djkoloski:
Got some friction with CI, #69 would help improve the contribution experience.

joshlf (approved):

PR #126, opened 2022-11-07
Allow dereferencing an Unalign, add trait impls

joshlf:
Provide methods and trait implementations to generate `&T` and `&mut T` from `&Unalign<T>` and `&mut Unalign<T>` respectively. If `T: Unaligned`, then these are infallible. Otherwise, a runtime check must be performed to ensure that the `Unalign<T>` satisfies `T`'s alignment requirement.

Since dereferencing is now supported, we can also implement various traits using `Deref` and `DerefMut` impls.


joshlf:
@djkoloski can you take a look at this when you get a chance? Please be brutal about the safety comments. I want to make sure we're 100% bulletproof.

src/lib.rs
```diff
@@ -852,20 +853,31 @@ mod simd {
...
 /// [`into_inner`]: Unalign::into_inner
+/// [`try_deref`]: Unalign::try_deref
+/// [`try_deref_mut`]: Unalign::try_deref_mut
+/// [`deref_unchecked`]: Unalign::deref_unchecked
+/// [`deref_mut_unchecked`]: Unalign::deref_mut_unchecked
 #[allow(missing_debug_implementations)]
-#[derive(FromBytes, Unaligned, Copy)]
+#[derive(FromBytes, Unaligned, Default, Copy)]
 #[repr(C, packed)]
 pub struct Unalign<T>(T);
```
djkoloski, line 882: Not actionable, just a note: we don't need to write a custom drop that aligns the inner value before dropping. The compiler generates a shim that does that for us. See https://github.com/rust-lang/reference/issues/488

src/lib.rs
```diff
@@ -970,6 +1062,65 @@ unsafe impl<T: AsBytes> AsBytes for Unalign<T> {
     }
 }
 
+impl<T: Unaligned> Deref for Unalign<T> {
+    type Target = T;
+
+    fn deref(&self) -> &T {
+        // SAFETY: `deref_unchecked`'s safety requirement is that `self` is
+        // aligned to `align_of::<T>()`, which is guaranteed by `T: Unaligned`.
```
djkoloski, line 1070: Suggest: `Unaligned` types are guaranteed to have an alignment of one. All pointers are one-aligned because all addresses are divisible by 1.

src/lib.rs
```diff
@@ -1819,18 +1970,71 @@ where
...
-    // Currently, `as usize` is the only way to accomplish this.
-    #[allow(clippy::as_conversions)]
-    let addr = bytes.as_ptr().cast::<()>() as usize;
-    // `mem::align_of::<T>()` is guaranteed to return a non-zero value, which in
+fn aligned_to<T: AsAddress, U>(t: T) -> bool {
+    // `mem::align_of::<U>()` is guaranteed to return a non-zero value, which in
     // turn guarantees that this mod operation will not panic.
     #[allow(clippy::arithmetic_side_effects)]
-    let remainder = addr % mem::align_of::<T>();
+    let remainder = t.addr() % mem::align_of::<U>();
```
djkoloski, line 2037: NB: This can just be `t.addr() & (mem::align_of::<U>() - 1) == 0`. If you want to micro-optimize (you might not) then that might give slightly better codegen since we know `mem::align_of` always returns a power of two.

djkoloski (approved):
There are a few small changes that could be made (your discretion).

src/lib.rs
joshlf: Ack 👍 

Were you thinking it'd be worth putting that in a comment or something for future readers? I wouldn't be opposed.

djkoloski: Not sure, it could be worth documenting so the knowledge isn't lost though. I struggled to find any documentation on how packed drops work.

joshlf: Yeah, sounds reasonable. Added.

src/lib.rs
joshlf: Good call; done.

joshlf:
Thanks for the review!

PR #128, opened 2022-11-07
[derive] Clean up AsBytes support on repr(packed)

### LINKED ISSUE #127: Support deriving `AsBytes` on generic `repr(packed)` structs
We support deriving `AsBytes` on generic `repr(transparent)` structs because we don't need to explicitly check for padding, and so we don't need to emit code that's incompatible with generic type arguments. We ought to be able to use the same approach to support generic `repr(packed)` structs as well.

EDIT: Looks like we do support them, but there are no tests for that support, and our error messages imply that we don't support them. We should clean that up.

cc @djkoloski

joshlf:
Unify our handling of `repr(packed)` and `repr(transparent)` when deriving `AsBytes`, and bring error messages up to date with the fact that we support both of these reprs on generic structs.

Closes #127

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3253,20 +3253,31 @@ mod tests {
     }
 
     #[test]
-    fn test_transparent_generic_struct() {
+    fn test_transparent_packed_generic_struct() {
         #[derive(AsBytes, FromBytes, Unaligned)]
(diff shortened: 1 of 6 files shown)


joshlf:
@djkoloski can you take a look at this when you get a chance?

PR #131, opened 2022-11-08
[byteorder] Derive AsBytes for byteorder types

joshlf:
This was previously a hand-rolled impl, but now that our `AsBytes` derive supports generic `repr(transparent)` types, we can use the derive instead.


joshlf:
@djkoloski an easy and fun review 🎉

PR #136, opened 2022-11-09
Clean up trait impls, add some safety comments

joshlf:
Previously, impls of the `FromBytes`, `AsBytes`, and `Unaligned` traits were spread across a number of inconsistent and hard-to-follow macro invocations. This commit cleans that up by introducing a single macro and a single section of `lib.rs` to hold macro invocations.

This commit also adds some safety comments, but there are still some left as TODOs. Note that, since this commit does not add or remove any trait impls (see `test_impls` for a confirmation that the exact set of impls are tested), these TODOs are acceptable to add in this commit; this commit moves us strictly forward in terms of safety comment coverage.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -159,134 +159,6 @@ mod zerocopy {
     pub(crate) use crate::*;
 }
 
-// Implements an unsafe trait for a range of container types.
-macro_rules! impl_for_composite_types {
-    ($trait:ident) => {
-        // TODO(#61): Add a "SAFETY" comment and remove this `allow`.
-        #[allow(clippy::undocumented_unsafe_blocks)]
-        unsafe impl<T: ?Sized> $trait for PhantomData<T> {
-            fn only_derive_is_allowed_to_implement_this_trait()
-            where
-                Self: Sized,
-            {
-            }
-        }
-        // SAFETY: `ManuallyDrop` has the same layout as `T`, and accessing the
-        // inner value is safe (meaning that it's unsound to leave the inner
-        // value uninitialized while exposing the `ManuallyDrop` to safe code).
-        // - `FromBytes`: Since it has the same layout as `T`, any valid `T` is
-        //   a valid `ManuallyDrop<T>`. Since `T: FromBytes`, any sequence of
-        //   bytes is a valid `T`, and thus a valid `ManuallyDrop<T>`.
-        // - `AsBytes`: Since it has the same layout as `T`, and since it's
-        //   unsound to let safe code access a `ManuallyDrop` whose inner value
-        //   is uninitialized, safe code can only ever access a `ManuallyDrop`
-        //   whose contents are a valid `T`. Since `T: AsBytes`, this means that
-        //   safe code can only ever access a `ManuallyDrop` with all
-        //   initialized bytes.
-        // - `Unaligned`: `ManuallyDrop` has the same layout (and thus
-        //   alignment) as `T`, and `T: Unaligned` guarantees that that
-        //   alignment is 1.
-        unsafe impl<T: $trait + ?Sized> $trait for ManuallyDrop<T> {
-            fn only_derive_is_allowed_to_implement_this_trait()
-            where
-                Self: Sized,
-            {
-            }
-        }
-        // TODO(#61): Add a "SAFETY" comment and remove this `allow`.
-        #[allow(clippy::undocumented_unsafe_blocks)]
-        unsafe impl<T: $trait> $trait for [T] {
-            fn only_derive_is_allowed_to_implement_this_trait()
-            where
-                Self: Sized,
-            {
-            }
-        }
-        // SAFETY: According to the `Wrapping` docs, "`Wrapping<T>` is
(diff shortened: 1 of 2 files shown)


joshlf:
@djkoloski could I get a review on this? Please be brutal w/ safety comments. Some are left as TODOs (see the commit message for an explanation of why this is okay) - lmk if you think any of those shouldn't be left as TODOs here.

src/lib.rs
```diff
@@ -723,8 +551,281 @@ pub unsafe trait Unaligned {
...
+    ///
+    /// TODO(https://github.com/rust-lang/rust/pull/104081): Cite guaranteed
+    /// size and alignment.
+    unsafe_impl!(T: ?Sized => FromBytes for PhantomData<T>);
+    unsafe_impl!(T: ?Sized => AsBytes for PhantomData<T>);
+    unsafe_impl!(T: ?Sized => Unaligned for PhantomData<T>);
+}
+safety_comment! {
+    /// SAFETY:
+    /// `Wrapping<T>` is guaranteed to have the same layout as `T`.
```
djkoloski, line 767: "... because it is `#[repr(transparent)]` with a single public field"

src/lib.rs
```diff
@@ -723,8 +551,281 @@ pub unsafe trait Unaligned {
...
+    unsafe_impl!(T: Unaligned => Unaligned for Wrapping<T>);
+}
+
+safety_comment! {
+    // `MaybeUninit<T>` is `FromBytes`, but never `AsBytes` since it may contain
+    // uninitialized bytes.
+    //
+    /// SAFETY:
+    /// `MaybeUninit<T>` has no restrictions on its contents.
+    unsafe_impl!(T => FromBytes for MaybeUninit<T>);
```
djkoloski, line 779: This should also implement `Unaligned` for `MaybeUninit<T>` where `T: Unaligned`.

src/lib.rs
```diff
@@ -1056,17 +1159,14 @@ impl<T: Copy> Unalign<T> {
...
-    {
-    }
+safety_comment! {
+    /// SAFETY:
+    /// Since `T: AsBytes`, we know that it's safe to construct a `&[u8]` from
+    /// an aligned `&T`. Since `&[u8]` itself has no alignment requirements, it
+    /// must also be safe to construct a `&[u8]` from a `&T` at any address.
+    /// Since `Unalign<T>` is `#[repr(C, packed)]`, everything about its layout
+    /// except for its alignment is the same as `T`'s layout.
+    unsafe_impl!(T: AsBytes => AsBytes for Unalign<T>);
```
djkoloski, line 1169: Also `FromBytes`?

src/lib.rs
```diff
@@ -722,8 +550,281 @@ pub unsafe trait Unaligned {
...
+    // than 1.
+    ($ty:ty: Unaligned) => {
+        // We only compile this assertion under `cfg(test)` to avoid making this
+        // crate more expensive to compile for our dependents.
+        #[cfg(test)]
+        const _: () = { static_assertions::const_assert_eq!(core::mem::align_of::<$ty>(), 1); };
+        unsafe impl Unaligned for $ty { fn only_derive_is_allowed_to_implement_this_trait() {} }
+    };
+    // Implement `$trait` for `$ty` with no bounds.
+    ($ty:ty: $trait:tt) => {
```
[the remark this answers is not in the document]
joshlf: Works for this and the one below, but not the ones below that:

```text
error: `$r#trait:ty` is followed by `for`, which is not allowed for `ty` fragments
   --> src/lib.rs:601:32
    |
601 |     ($tyvar:ident => $trait:ty for $ty:ty) => {
    |                                ^^^ not allowed after `ty` fragments
    |
    = note: allowed there are: `{`, `[`, `=>`, `,`, `>`, `=`, `:`, `;`, `|`, `as` or `where`
```

We could change the syntax or the word "for", but we also don't currently have any traits that take type arguments, and probably won't for a long time, so I'm not sure it's worth the genericity seeing as we can always add it later if/when we need it.

src/lib.rs
joshlf: Agreed, but I'm gonna follow up on that in a subsequent PR. For the time being, I want this change to just be a refactor and adding safety arguments, not actually adding new impls.

src/lib.rs
joshlf: I don't think it's universally agreed that `#[repr(transparent)]` represents a stability guarantee. I'm actually just citing the documentation here: https://doc.rust-lang.org/nightly/core/num/struct.Wrapping.html#layout-1

joshlf:
Thanks for all the comments! Pushed a new change that addresses many of them.

src/lib.rs
joshlf: Ah, TIL. Updated.

src/lib.rs
joshlf: jk, we already implement it:
https://github.com/google/zerocopy/blob/140b89c7ad5989459b550f3f198c2ef8a7bffcf3/src/lib.rs#L887-L889

PR #140, opened 2022-11-10
Fix bug that made a static assertion a no-op

joshlf:
The `unsafe_impl!` macro has a branch that special-cases `Unaligned` and adds an extra static assertion that the type's alignment is 1. It turns out that this branch is unreachable when the macro is invoked with multiple traits like:

  unsafe_impl!(ty: TraitA, TraitB, Unaligned);

This commit fixes the bug by marking the macro item matching trait names as a `tt` rather than a `ty`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -590,11 +590,11 @@ macro_rules! unsafe_impl {
         unsafe impl Unaligned for $ty { fn only_derive_is_allowed_to_implement_this_trait() {} }
     };
     // Implement `$trait` for `$ty` with no bounds.
-    ($ty:ty: $trait:ty) => {
+    ($ty:ty: $trait:tt) => {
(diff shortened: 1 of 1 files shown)


joshlf:
@djkoloski Now I remember why I had that as `tt` 🤣

joshlf:
I'm not sure; I think it's maybe the opposite - it's changing it _to_ a `tt` that fixed the bug, so I think it might be that the `ty` made it so that the metavariable was treated as opaque and thus not equal to `Unaligned` even if the literal text was the same. That's just a guess though.

joshlf:
Looks like we're going to have to remove this logic [here](https://github.com/google/zerocopy/pull/141/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R612-R627) anyway, so let's not bother with this.

PR #141, opened 2022-11-10
Add new trait impls

joshlf:
Add the following impls:
- `str: Unaligned`
- `NonZeroU8: Unaligned`
- `NonZeroI8: Unaligned`
- `Option<NonZeroU8>: Unaligned`
- `Option<NonZeroI8>: Unaligned`
- `MaybeUninit<T>: Unaligned` where `T: Unaligned`

Previously, `unsafe_impl!` special-cased `Unaligned` and emitted a static assertion that the type in question actually had alignment 1. Since this used `core::mem::align_of`, which only works for `Sized` types, it was incompatible with adding an impl of `Unaligned` for `str` in this commit. In order to avoid this issue, this commit removes that special-casing, and introduces a new `assert_unaligned!` macro. Existing `Unaligned` implementations are modified to also call `assert_unaligned!`. `assert_unaligned!(str)` is not used since `str: !Unaligned`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -576,19 +576,6 @@ macro_rules! safety_comment {
 
 /// Unsafely implements trait(s) for a type.
 macro_rules! unsafe_impl {
-    // Implement `Unaligned` for `$ty` with no bounds.
-    //
-    // For `Unaligned` in particular, it's possible to assert at compile time
-    // that the trait impl is sound. This provides a small speed bump to
-    // accidentally implementing `Unaligned` for a type with alignment greater
-    // than 1.
-    ($ty:ty: Unaligned) => {
-        // We only compile this assertion under `cfg(test)` to avoid making this
-        // crate more expensive to compile for our dependents.
-        #[cfg(test)]
-        const _: () = { static_assertions::const_assert_eq!(core::mem::align_of::<$ty>(), 1); };
-        unsafe impl Unaligned for $ty { fn only_derive_is_allowed_to_implement_this_trait() {} }
-    };
     // Implement `$trait` for `$ty` with no bounds.
     ($ty:ty: $trait:ty) => {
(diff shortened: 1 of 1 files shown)


joshlf:
@djkoloski you may have thoughts here. I don't yet have safety comments for `NonZeroX8: Unaligned` or `Option<NonZeroX8>: Unaligned`, and my plan is to wait for https://github.com/rust-lang/rust/pull/104082 to merge and then cite the updated documentation and then merge this PR at that point. However, if you can think of a different way to justify the soundness of those impls today (we can leave a TODO comment to cite the documentation once that PR lands), that'd be awesome.

djkoloski:
Maybe you could argue that since `Option<NonZeroU8>` and `Option<NonZeroI8>` are both [guaranteed](https://doc.rust-lang.org/stable/std/num/struct.NonZeroI8.html) to be the same size as `u8` and `i8`, that they must be the same size or smaller and so must have size 1. A size of 1 also implies an align of 1 since size must be a multiple of align.

joshlf:
> Maybe you could argue that since `Option<NonZeroU8>` and `Option<NonZeroI8>` are both [guaranteed](https://doc.rust-lang.org/stable/std/num/struct.NonZeroI8.html) to be the same size as `u8` and `i8`, that they must be the same size or smaller and so must have size 1. A size of 1 also implies an align of 1 since size must be a multiple of align.

I was on the fence about taking that approach, but thinking about it more (and reinforced by you suggesting the same), I guess it's basically unthinkable that they would ever change that aspect of the layout of those types. I was thinking that the wording in that documentation is sufficiently weak (it says "is", not "is guaranteed to be" or something similar - it reads as though it could just be saying "in practice"), but I think that's probably not enough of a concern to block this. I'll go with that approach.

joshlf:
OK done! Ready for review whenever you get a chance.

djkoloski (approved):

PR #143, opened 2022-11-11
Introduce FromZeroes trait

### LINKED ISSUE #30: Relax requirements for deriving `FromZeroes` on enums
This issue used to track `FromZeroes`, which has now been merged. However, support for `#[derive(FromZeroes)]` is incomplete - the rules implemented when deriving on an enum are equivalent to the rules for deriving `FromBytes` on an enum. Those rules are unnecessarily restrictive.

Now, this issue tracks:
- Figuring out what rules enums need to follow in order to be `FromZeroes`
- Modifying our derive to implement those rules

## Old text

Add an `unsafe` marker trait called `FromZeroes` (or similar) which indicates that a type can be safely constructed from all 0 bytes. Add a custom derive to `zerocopy-derive` for this trait.

I would expect the API to look something like this:

```rust
unsafe trait FromZeroes {
    #[doc(hidden)]
    fn only_derive_is_allowed_to_implement_this_trait() where Self: Sized;

    fn zero(&mut self);

    // These would all be moved from `FromBytes`.

    fn new_zeroed() -> Self where Self: Sized { ... }
    #[cfg(feature = "alloc")]
    fn new_box_zeroed() -> Box<Self> where Self: Sized { ... }
    #[cfg(feature = "alloc")]
    fn new_box_slice_zeroed() -> Box<Self> where Self: Sized { ... }

    // If we also implement `Zeroed` (#31):
    fn read_from_zeroed<B: ByteSlice>(bytes: Zeroed<B>) -> Option<Self> where Self: Sized { ... }
    fn read_from_zeroed_prefix<B: ByteSlice>(bytes: Zeroed<B>) -> Option<Self> where Self: Sized { ... }
    fn read_from_zeroed_suffix<B: ByteSlice>(bytes: Zeroed<B>) -> Option<Self> where Self: Sized { ... }
}

// `FromBytes` would gain a `FromZeroes` bound.
unsafe trait FromBytes: FromZeroes { ... }
```

joshlf:
Closes #30

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,8 +14,10 @@ This crate provides utilities which make it easy to perform zero-copy
 parsing and serialization by allowing zero-copy conversion to/from byte
 slices.
 
-This is enabled by three core marker traits, each of which can be derived
+This is enabled by four core marker traits, each of which can be derived
 (e.g., `#[derive(FromBytes)]`):
+- `FromZeroes` indicates that a sequence of zero bytes represents a valid
+  instance of a type
 - `FromBytes` indicates that a type may safely be converted from an
   arbitrary byte sequence
 - `AsBytes` indicates that a type may safely be converted *to* a byte
@@ -34,12 +36,12 @@ types, see the `byteorder` module.
 enabled, the `alloc` crate is added as a dependency, and some
 allocation-related functionality is added.
 
-`simd`: When the `simd` feature is enabled, `FromBytes` and `AsBytes` impls
-are emitted for all stable SIMD types which exist on the target platform.
-Note that the layout of SIMD types is not yet stabilized, so these impls may
-be removed in the future if layout changes make them invalid. For more
-information, see the Unsafe Code Guidelines Reference page on the [Layout of
-packed SIMD vectors][simd-layout].
+`simd`: When the `simd` feature is enabled, `FromZeroes`, `FromBytes`, and
+`AsBytes` impls are emitted for all stable SIMD types which exist on the
+target platform. Note that the layout of SIMD types is not yet stabilized,
+so these impls may be removed in the future if layout changes make them
+invalid. For more information, see the Unsafe Code Guidelines Reference page
+on the [Layout of packed SIMD vectors][simd-layout].
 
 `simd-nightly`: Enables the `simd` feature and adds support for SIMD types
 which are only available on nightly. Since these types are unstable, support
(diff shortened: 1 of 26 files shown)


joshlf:
@djkoloski ready for a review when you get a chance!

README.md
```diff
@@ -14,8 +14,10 @@ This crate provides utilities which make it easy to perform zero-copy
 parsing and serialization by allowing zero-copy conversion to/from byte
 slices.
 
-This is enabled by three core marker traits, each of which can be derived
+This is enabled by four core marker traits, each of which can be derived
 (e.g., `#[derive(FromBytes)]`):
+- `FromZeroes` indicates that a sequence of zero bytes represents a valid
```
djkoloski, line 19: This wording is a little awkward, it could be interpreted as a sequence of length 0. Maybe something like "`FromZeroes` indicates that a sequence of bytes all set to zero represents a valid instance of a type.

zerocopy-derive/src/lib.rs
```diff
@@ -107,6 +118,63 @@ const STRUCT_UNION_ALLOWED_REPR_COMBINATIONS: &[&[StructRepr]] = &[
...
 
+// A struct is `FromZeroes` if:
+// - all fields are `FromZeroes`
+
+fn derive_from_zeroes_struct(ast: &DeriveInput, strct: &DataStruct) -> proc_macro2::TokenStream {
+    impl_block(ast, strct, "FromZeroes", true, PaddingCheck::None)
+}
+
+// An enum is `FromZeroes` if:
+// - Every possible bit pattern must be valid, which means that every bit
```
djkoloski, line 129: Does this comment need fixup? Or is this just a stricter guarantee for now until we go back and fix it up?

README.md
joshlf: Good call; I've updated it, but kept the "Types..." prefix to be consistent with other traits. I can change all of them if you'd prefer.

zerocopy-derive/src/lib.rs
joshlf: The latter - right now, this is just a copy-paste from the comment on the `FromBytes` derive.

README.md
djkoloski: Missed change? I don't see this in the latest revision.

joshlf: Here's the diff I'm looking at: https://github.com/google/zerocopy/compare/afcea57c53372fd04d28bca8b2ea90db429a86e6..3fb2a0e342304d60ca354102d8bc2d7d2181d6fc

djkoloski (approved):

PR #145, opened 2022-11-15
Relax requirements for deriving `FromZeroes` on enums

### LINKED ISSUE #30: Relax requirements for deriving `FromZeroes` on enums

djkoloski:
Closes #30. @joshlf the wording of [the reference](https://doc.rust-lang.org/reference/items/enumerations.html#custom-discriminant-values-for-fieldless-enumerations) seems to imply that we can derive `FromZeroes` on enums without an explicit repr. Do you feel comfortable doing that?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -187,8 +187,11 @@ mod zerocopy {
 /// `FromZeroes` for that type:
 /// - If the type is a struct:
 ///   - All of its fields must implement `FromZeroes`
-/// - If the type is an enum, it must follow the same rules as [`FromBytes`]
-///   enums. this will likely change soon.
+/// - If the type is an enum, all of its variants must be fieldless and it must
+///   have a variant with a discriminant of `0` (see [the reference] for a
+///   description of how discriminant values are chosen)
+///
+/// [the reference]: https://doc.rust-lang.org/reference/items/enumerations.html#custom-discriminant-values-for-fieldless-enumerations
 ///
 /// # Rationale
 ///
--- a/zerocopy-derive/src/lib.rs
+++ b/zerocopy-derive/src/lib.rs
@@ -33,7 +33,7 @@ use {
     syn::visit::{self, Visit},
     syn::{
         parse_quote, punctuated::Punctuated, token::Comma, Data, DataEnum, DataStruct, DataUnion,
-        DeriveInput, Error, GenericParam, Ident, Lifetime, Type, TypePath,
+        DeriveInput, Error, Expr, ExprLit, GenericParam, Ident, Lifetime, Lit, Type, TypePath,
     },
 };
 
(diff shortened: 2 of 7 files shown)


joshlf:
> @joshlf the wording of [the reference](https://doc.rust-lang.org/reference/items/enumerations.html#custom-discriminant-values-for-fieldless-enumerations) seems to imply that we can derive `FromZeroes` on enums without an explicit repr. Do you feel comfortable doing that?

Seems reasonable to me! Are we guaranteed that, regardless of representation, the 0 is represented as all 0 bits? That's the only pedantry I can come up with that makes me question the validity of deriving on an enum w/ no explicit repr, but I'm pretty confident that's not a problem.

Also, could you update the doc comment on `FromZeroes` to document the rules? Same with [the comment](https://github.com/google/zerocopy/pull/145/files#diff-977c29e6cf079bd7307f811df24f99633df603962785c583145eb6250b0e12daR128-R140) that precedes `derive_from_zeroes_enum`.

joshlf:
Oh also: Could you add success tests (not UI tests) in `zerocopy-derive/tests/enum_from_zeroes.rs`?

joshlf:
> Closes #143

Also also: I think you mean "Closes #30" 🙂

src/lib.rs
```diff
@@ -187,8 +187,9 @@ mod zerocopy {
 /// `FromZeroes` for that type:
 /// - If the type is a struct:
 ///   - All of its fields must implement `FromZeroes`
-/// - If the type is an enum, it must follow the same rules as [`FromBytes`]
-///   enums. this will likely change soon.
+/// - If the type is an enum, all of its variants must be fieldless and it must
+///   have a variant with a discriminant of `0` (if the first variant does not
+///   have an explicit discriminant, its discriminant is set to 0).
```
joshlf, line 192: How would you feel about removing this parenthetical? IMO, in this position, it feels stylistically like it's describing a requirement of the trait rather than describing a detail of the language. It's also likely the case that a given programmer either already knows about discriminant values and how to set them (and thus the parenthetical isn't telling them something they don't know) or they don't know about how to set discriminant values (in which case they won't know what "explicit discriminant" means). Maybe we could reword this like so?

```rust
/// - If the type is an enum, all of its variants must be fieldless, and it must
///   have a variant with a discriminant of `0` (see [the reference] for
///   a description of how discriminant values are chosen)
...
/// [the reference]: https://doc.rust-lang.org/reference/items/enumerations.html#custom-discriminant-values-for-fieldless-enumerations
```

Also, can you remove the trailing period to be consistent with the other bullet points?

joshlf (changes requested):

zerocopy-derive/src/lib.rs
```diff
@@ -145,22 +145,23 @@ fn derive_from_zeroes_enum(ast: &DeriveInput, enm: &DataEnum) -> proc_macro2::To
             .to_compile_error();
     }
 
-    let reprs = try_or_print!(ENUM_FROM_BYTES_CFG.validate_reprs(ast));
+    let has_explicit_zero_discriminant =
```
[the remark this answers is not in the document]
djkoloski: Whoops, missed that one.

src/lib.rs
djkoloski: Changed wording and removed trailing period

joshlf (approved):

PR #156, opened 2023-02-16
Update MSRV to 1.65.0

joshlf:
Also update the versions of the stable and nightly compiler toolchains that are pinned in CI.


joshlf:
Looks like the tests are failing due to https://github.com/rust-lang/miri/issues/2791.

PR #163, opened 2023-03-11
Disable incremental compilation

AntoniosBarotsis:
> compiling a crate for the first time in incremental mode can be slower than compiling it in non-incremental mode

This should speed up your CI runs by a 1-2 minutes :)


AntoniosBarotsis:
Hm, on further testing it looks like this is rather inconsistent and not gaining as much of a speedup as I'd expect it to. I'll look into this a bit further and if I can't figure out why that is the case I'll close the PR.

PR #166, opened 2023-03-27
Migrate to syn-2

maurer:


joshlf:
@djkoloski Could I get your eyeballs on this? You understand this area of the codebase better than I do.

djkoloski (approved):
LGTM

PR #168, opened 2023-04-26
Clarify Unalign ABI guarantees

### LINKED ISSUE #164: `Unalign` ABI promises are wrong
https://docs.rs/zerocopy/latest/zerocopy/struct.Unalign.html states

> `Unalign<T>` has the same size and ABI as `T`

This is wrong, as the type is `#[repr(C, packed)]` and not `repr(transparent)`.

The [Rust reference](https://doc.rust-lang.org/reference/type-layout.html#the-transparent-representation) about `repr(transparent)`:

> This is different than the C representation because a struct with the C representation will always have the ABI of a C struct while, for example, a struct with the transparent representation with a primitive field will have the ABI of the primitive field.

Here's an example of such an ABI mismatch in practice: https://godbolt.org/z/bhP71dnh4

The docs should be relaxed to not promise full ABI compatibility, but only promise what `repr(C)` promises (which is probably strong enough for most use cases).

joshlf:
Closes #164

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1139,9 +1139,10 @@ mod simd {
 /// A type with no alignment requirement.
 ///
 /// An `Unalign` wraps a `T`, removing any alignment requirement. `Unalign<T>`
-/// has the same size and ABI as `T`, but not necessarily the same alignment.
-/// This is useful if a type with an alignment requirement needs to be read from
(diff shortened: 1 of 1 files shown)


djkoloski (dismissed):
LGTM

PR #171, opened 2023-05-20
Implement FromBytes and AsBytes for raw pointers

### LINKED ISSUE #170: Implement `FromBytes` and `AsBytes` for raw pointers
Historically, we've avoided adding these impls for fear that they could present a footgun, making it easier to accidentally transmute into a type, breaking its safety invariants.

Today, I was working on replacing some `unsafe` code in Fuchsia's Starnix project with zerocopy, and ran into some structs for which I can't derive `FromBytes` or `AsBytes` because they contain raw pointers. Currently, the lack of impls on raw pointers is entirely preventing this `unsafe` code from being replaced by zerocopy, which feels to me like throwing the baby out with the bathwater. IMO, it's more important to unlock this use case than to avoid a minor and theoretical footgun.

Note that, while the [layout of raw pointers is well-defined](https://doc.rust-lang.org/reference/type-layout.html#pointers-and-references-layout), it's less clear that their bit validity is well-defined.

TODO:
 - [ ] [Confirm that these impls would actually be sound](https://github.com/rust-lang/unsafe-code-guidelines/issues/401)
 - [ ] Resurrect [this PR](https://github.com/google/zerocopy/pull/171)

joshlf:
EDIT: This is now blocked on https://github.com/rust-lang/unsafe-code-guidelines/issues/401

Previously, we omitted these impls out of a fear of creating a footgun in which it would be too easy to mistakenly transmute into a type with internal safety invariants. Recently, this omission meant that I was unable to get rid of a number of lines of `unsafe` code in a codebase that used raw pointers, and it made me realize that it is more important to enable this use case than to avoid a theoretical footgun.

Closes #170

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -775,6 +775,30 @@ safety_comment! {
 
 safety_comment! {
     /// SAFETY:
+    /// - Raw pointers to `Sized` types have the same size and alignment as
+    ///   `usize` [1]
+    /// - We know that any bit pattern is valid for a raw pointer to a `Sized`
+    ///   type because it is a safe operation to `as` cast from a `usize`
+    /// - We know that raw pointers to `Sized` types do not contain
+    ///   uninitialized or padding bytes - and more generally that it is always
+    ///   valid to view the bytes of a raw pointer to a `Sized` type as a `[u8]`
+    ///   - because it is a safe operation to `as` cast such a pointer to a
+    ///   `usize`
+    ///
+    /// TODO: Make the bit validity argument in terms of the reference once bit
+    /// validity is described in the reference
+    ///
+    /// [1] https://doc.rust-lang.org/reference/type-layout.html#pointers-and-references-layout
+    unsafe_impl!(T: Sized => FromZeroes for *const T);
+    unsafe_impl!(T: Sized => FromBytes for *const T);
+    unsafe_impl!(T: Sized => AsBytes for *const T);
+    unsafe_impl!(T: Sized => FromZeroes for *mut T);
+    unsafe_impl!(T: Sized => FromBytes for *mut T);
+    unsafe_impl!(T: Sized => AsBytes for *mut T);
+}
+
+safety_comment! {
+    /// SAFETY:
     /// - `FromZeroes`, `FromBytes`: the `{f32,f64}::from_bits` constructors'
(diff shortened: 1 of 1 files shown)


djkoloski:
Per tests, looks like `tests/ui/transmute-illegal.stderr` needs to be updated now that pointers may be transmuted.

joshlf:
Oh damn, that may actually be a clue to a soundness hole. For those following along, here's the error I get:

```rust
error[E0080]: evaluation of constant value failed
  --> tests/ui/transmute-illegal.rs:10:1
   |
10 | const POINTER_VALUE: usize = zerocopy::transmute!(&0usize as *const usize);
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^ unable to turn pointer into raw bytes
   |
   = help: this code performed an operation that depends on the underlying bytes representing a pointer
   = help: the absolute address of a pointer is not known at compile-time, so such operations are not supported
```

Are we guaranteed that rustc's const eval will catch every instance of turning a pointer into raw bytes? If so, this should be okay - it'll just result in a potentially confusing compile-time error for the user. If this is a best effort analysis, though, then introducing those trait impls may be unsound.

djkoloski:
AFAIK the compiler will catch all cases where a pointer is transmuted to bytes at compile time. If you want to guarantee that we can't do that in const expressions, the macro could expand to a non const function that is immediately invoked. I don't think the guarantees around const UB are well defined.

joshlf:
Submitted https://github.com/rust-lang/unsafe-code-guidelines/issues/401 to get a clearer answer; if the answer is that it's guaranteed, I'll follow up with a change to the Rust reference to make it clear that it's guaranteed.

This also brings up another question, which is: None of our safety comments justifying `FromBytes`, `AsBytes`, and `Unaligned` impls mention the const context. If there are other operations which are sound in non-const code but unsound in const code, then those safety arguments may be incomplete. Do we need to come up with a general strategy for reasoning about const soundness?

djkoloski:
I anticipate the answer will be that any time const soundness is more restrictive than runtime soundness, the compiler is guaranteed to emit a compile error. If that's not the case, then it's difficult to make a single coherent strategy

joshlf:
Closing for now; I'll reopen depending on the resolution of https://github.com/rust-lang/unsafe-code-guidelines/issues/401

PR #172, opened 2023-05-20
Move now-stable aarch64 types to "simd" feature

joshlf:


joshlf:
@djkoloski Hey, removed you as reviewer because I realized after adding you that this wasn't worth bothering you about. Let me know if you have thoughts, obviously - I can still revert.

PR #176, opened 2023-05-23
Make zerocopy-derive an optional dependency

### LINKED ISSUE #169: Make zerocopy-derive an optional Cargo dependency
## Overview

Make zerocopy-derive an optional dependency in order to significantly improve compile times for those who don't need the derives.

## Motivation

Compile times are important, and are a major blocker for zerocopy adoption, especially for high-profile ecosystem crates who are very strict with their compile time budgets.

~~[Benchmarking suggests](https://github.com/google/zerocopy/issues/169#issuecomment-1555342914) that 85-90% of zerocopy's build time is due to zerocopy-derive.~~ The [most recent numbers](https://github.com/google/zerocopy/pull/176#issuecomment-1559860045) from the current implementation imply that it's closer to 95%.

If we can make zerocopy-derive an optional dependency, then the `--no-default-features` configuration will have zero dependencies and will take [less than 1/3 of a second](https://github.com/google/zerocopy/pull/176#issuecomment-1559860045) to build on some platforms.

## Design

There are two reasons that zerocopy depends on zerocopy-derive:
- zerocopy re-exports the derives for convenience
- zerocopy uses the derives for its own types

The first of these is a non-issue - users disabling the zerocopy-derive dependency are presumably not making use of the derives anyway.

The second presents a bigger problem. The derives don't just save boilerplate, they also perform important soundness verification that we'd otherwise have to reason through ourselves. It would be a big step backwards in terms of confidence in the correctness of our code to get rid of the derives entirely and revert to manual implementations.

A first stab at this might be to feature-gate our derives - in other words, use them when the `derive` feature is enabled, and use manual impls when it's disabled. This would in theory ensure that, so long as we test with the `derive` feature in CI, the manual impls can never be unsound.

However, that's not quite enough. The soundness of a trait impl isn't _just_ based on its presence or absence. It's also based on the trait bounds placed on that impl. Consider:

```rust
#[repr(transparent)]
struct Wrapper<T>(T);

unsafe impl<T> FromBytes for Wrapper<T> {}
```

In this example, the `FromBytes` impl is unsound because `T` is unbound. The impl is only sound if we add a `T: FromBytes` bound (which is exactly what the custom derive does).

So we not only need to verify that _some_ impl is sound, but that the impls we emit - with their particular trait bounds - are sound. In order to solve this problem, we introduce an `impl_or_verify!` macro:

```rust
macro_rules! impl_or_verify {
    // Implement `$trait` for `$ty` with no bounds.
    ($ty:ty: $trait:ty) => {
        impl_or_verify!(@impl { unsafe_impl!($ty: $trait); });
        impl_or_verify!(@verify $trait, { impl Subtrait for $ty {} });
    };
    // Implement all `$traits` for `$ty` with no bounds.
    ($ty:ty: $($traits:ty),*) => {
        $( foobar!($ty: $traits); )*
    };
    // For all `$tyvar` with no bounds, implement `$trait` for `$ty`.
    ($tyvar:ident => $trait:ident for $ty:ty) => {
        impl_or_verify!(@impl { unsafe_impl!($tyvar => $trait for $ty); });
        impl_or_verify!(@verify $trait, { impl<$tyvar> Subtrait for $ty {} });
    };
    // For all `$tyvar: ?Sized` with no bounds, implement `$trait` for `$ty`.
    ($tyvar:ident: ?Sized => $trait:ident for $ty:ty) => {
        impl_or_verify!(@impl { unsafe_impl!($tyvar: ?Sized => $trait for $ty); });
        impl_or_verify!(@verify $trait, { impl<$tyvar: ?Sized> Subtrait for $ty {} });
    };
    // For all `$tyvar: $bound`, implement `$trait` for `$ty`.
    ($tyvar:ident: $bound:path => $trait:ident for $ty:ty) => {
        impl_or_verify!(@impl { unsafe_impl!($tyvar: $bound => $trait for $ty); });
        impl_or_verify!(@verify $trait, { impl<$tyvar: $bound> Subtrait for $ty {} });
    };
    // For all `$tyvar: $bound + ?Sized`, implement `$trait` for `$ty`.
    ($tyvar:ident: ?Sized + $bound:path => $trait:ident for $ty:ty) => {
        impl_or_verify!(@impl { unsafe_impl!($tyvar: ?Sized + $bound => $trait for $ty); });
        impl_or_verify!(@verify $trait, { impl<$tyvar: ?Sized + $bound> Subtrait for $ty {} });
    };
    // For all `$tyvar: $bound` and for all `const $constvar: $constty`,
    // implement `$trait` for `$ty`.
    ($tyvar:ident: $bound:path, const $constvar:ident: $constty:ty => $trait:ident for $ty:ty) => {
        impl_or_verify!(@impl { unsafe_impl!($tyvar: $bound, const $constvar: $constty => $trait for $ty); });
        impl_or_verify!(@verify $trait, { impl<$tyvar: $bound, const $constvar: $constty> Subtrait for $ty {} });
    };
    (@impl $impl_block:tt) => {
        #[cfg(not(any(feature = "derive", test)))]
        const _: () = { $impl_block };
    };
    (@verify $trait:ident, $impl_block:tt) => {
        #[cfg(any(feature = "derive", test))]
        const _: () = {
            trait Subtrait: $trait {}
            $impl_block
        };
    };
}
```

When the `derive` feature is disabled, `impl_or_verify!` simply emits the impl. When the `derive` feature is enabled, though, it emits code which only compiles if the manual impl's trait bounds are at least as restrictive as those emitted by the derive. This effectively ties our manual impls to the custom derive, ensuring that the manual impls are sound so long as the custom derive is sound.

There's one final problem, however. There's no way to actually guarantee that the `impl_or_verify!` macro is ever used in "derive" mode. Consider:

```rust
#[cfg_attr(feature = "derive", derive(FromBytes))]
#[repr(transparent)]
struct Wrapper<T>(T);

#[cfg(not(feature = "derive"))]
impl_or_verify!(T => FromBytes for Wrapper<T>);
```

In this case, the unsound manual impl would never be detected because the `impl_or_verify!` macro would never be compiled with the `derive` feature enabled. In order to hedge against this (albeit slight) risk, we opt to still require a safety comment for uses of `impl_or_verify!`. That is, when combined with our use of [`deny(clippy::undocumented_unsafe_blocks)`](https://rust-lang.github.io/rust-clippy/master/index.html#/undocumented_unsafe_blocks), a call to `impl_or_verify!` will only compile when combined with `safety_comment!`:

```rust
safety_comment! {
    /// SAFETY:
    /// `Wrapper<T>` has the same layout as `T`, so `T: FromBytes` implies that `Wrapper<T>: FromBytes`.
    impl_or_verify!(T: FromBytes => FromBytes for Wrapper<T>);
}
```

This serves both to sanity-check our impl in case it's only compiled when `not(feature = "derive")` and also serves to document the reasoning behind a set of trait bounds, which is useful on its own independent of the `not(feature = "derive")` issue.

joshlf:
This requires a few changes to accomodate:
- We still want to be able to derive zerocopy traits on our own types so
  that we can rely on it to validate the soundness of those impls for
  us. That means that, when the `derive` feature is enabled, we still
  use custom derives, but when it's disabled, we implement traits
  manually. However, we also need to ensure that our manual trait impls
  have the correct bounds required in order for those impls to be sound.
  In order to ensure this, we introduce an `impl_or_verify!` macro
  which, when building with `derive` enabled, verifies that the manual
  impl matches the one emitted by our custom derive.
- Since some doc examples use custom derives, we can no longer run doc
  tests unconditionally in CI - when the `derive` feature is disabled,
  doc tests fail. Instead, we disable doc tests by default (in order to
  run them, `cargo test` must be run with the `--doc` flag). We
  introduce a separate "Run doc tests" step in CI - this step only runs
  when the `derive` feature is enabled.

Closes #169

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
         # versions.
         toolchain: [ "msrv", "stable", "nightly" ]
         target: [ "i686-unknown-linux-gnu", "x86_64-unknown-linux-gnu", "arm-unknown-linux-gnueabi", "aarch64-unknown-linux-gnu", "powerpc-unknown-linux-gnu", "powerpc64-unknown-linux-gnu", "wasm32-wasi" ]
-        features: [ "--no-default-features", "" , "--features __internal_use_only_features_that_work_on_stable", "--all-features" ]
+        features: [ "--no-default-features", "", "--features __internal_use_only_features_that_work_on_stable", "--all-features" ]
         crate: [ "zerocopy", "zerocopy-derive" ]
         exclude:
           # Exclude any combination which uses a non-nightly toolchain but
@@ -166,6 +166,20 @@ jobs:
       # TODO(#22): Re-enable testing on wasm32-wasi once it works.
       if: matrix.toolchain == 'nightly' && matrix.target != 'wasm32-wasi'
 
+    - name: Run doc tests
+      # We explicitly pass `--doc` here because doc tests are disabled by
+      # default in zerocopy's `Cargo.toml`. This is because some doc examples
+      # make use of derives, and so would fail without the `derive` feature
+      # enabled. We skip this step for `zerocopy` when the `derive` feature is
+      # omitted for that reason.
+      run: cargo +${{ env.ZC_TOOLCHAIN }} test --doc --package ${{ matrix.crate }} --target ${{ matrix.target }} ${{ matrix.features }} --verbose
+      # Only run tests when targetting x86 (32- or 64-bit) - we're executing on
+      # x86_64, so we can't run tests for any non-x86 target.
+      #
+      # TODO(https://github.com/dtolnay/trybuild/issues/184#issuecomment-1269097742):
+      # Run compile tests when building for other targets.
+      if: ${{ (contains(matrix.target, 'x86_64') || contains(matrix.target, 'i686')) && !(matrix.crate == 'zerocopy' && !contains(matrix.features, 'derive')) }}
+
     - name: Clippy check
       run: cargo +${{ env.ZC_TOOLCHAIN }} clippy --package ${{ matrix.crate }} --target ${{ matrix.target }} ${{ matrix.features }} --tests --verbose
 
@@ -255,8 +269,14 @@ jobs:
           ver_zerocopy=$(version zerocopy)
           ver_zerocopy_derive=$(version zerocopy-derive)
 
+          # The non-dev dependency version (`.kind == null` filters out the dev
+          # dependency).
           zerocopy_derive_dep_ver=$(cargo metadata --format-version 1 \
-            | jq -r ".packages[] | select(.name == \"zerocopy\").dependencies[] | select(.name == \"zerocopy-derive\").req")
+            | jq -r ".packages[] | select(.name == \"zerocopy\").dependencies[] | select((.name == \"zerocopy-derive\") and .kind == null).req")
+          # The dev dependency version (`.kind == \"dev\"` selects only the dev
+          # dependency).
+          zerocopy_derive_dev_dep_ver=$(cargo metadata --format-version 1 \
+            | jq -r ".packages[] | select(.name == \"zerocopy\").dependencies[] | select((.name == \"zerocopy-derive\") and .kind == \"dev\").req")
 
           if [[ "$ver_zerocopy" == "$ver_zerocopy_derive" ]]; then
             echo "Same crate version ($ver_zerocopy) found for zerocopy and zerocopy-derive." | tee -a $GITHUB_STEP_SUMMARY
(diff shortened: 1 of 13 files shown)


joshlf:
Some issues to address before it's ready for review/merge, but the numbers look great! In my cloud VM, compiling without zerocopy-derive is ~20x faster:

```text
$ rm -r target/* && time cargo check
   Compiling proc-macro2 v1.0.56
   Compiling unicode-ident v1.0.4
   Compiling quote v1.0.26
    Checking byteorder v1.4.3
   Compiling syn v2.0.15
   Compiling zerocopy-derive v0.7.0-alpha.4 (.../zerocopy/zerocopy-derive)
    Checking zerocopy v0.7.0-alpha.4 (.../zerocopy)
    Finished dev [unoptimized + debuginfo] target(s) in 5.05s

real    0m5.112s
user    0m8.476s
sys     0m1.562s

$ rm -r target/* && time cargo check --no-default-features
    Checking zerocopy v0.7.0-alpha.4 (.../zerocopy)
    Finished dev [unoptimized + debuginfo] target(s) in 0.23s

real    0m0.289s
user    0m0.246s
sys     0m0.045s
```

cc @djkoloski

joshlf:
@djkoloski You might have ideas here. While all of the types with trait impls use `derive` when `cfg(any(feature = "derive", test))`, all that tells us is that it's valid to derive the relevant trait. However, it doesn't tell us what bounds are used in the impl that the derive emits. As a result, it's still possible that our manual impls could have looser bounds on type parameters than the impl emitted by the derive. That means that, as written, we can't really rely on the "this impl is sound because the derive succeeds" argument. Any thoughts on how we can do this without just resorting to safety arguments from scratch?

djkoloski:
Am I understanding correctly that you want to use `zerocopy-derive` to write your impls when the `derive` feature is disabled, but want to continue using it when the feature is enabled? It seems easier and more straightforward (to me) to only keep manual impls. Maybe there's a way to:

1. Move all of the derived impls into a set of files
2. Mark `zerocopy-derive` as a dev-dependency
3. `cargo expand` (or manually invoke the same derive methods) on that set of files into zerocopy source that gets checked in

That could very easily be overengineering it though.

joshlf:
Yeah specifically I want to make sure that we're able to rely on the derive to do the soundness analysis for us. I'd like to keep the amount of manually verified unsafe code as small as possible.

joshlf:
Okay, the latest version contains a macro - `impl_or_verify!` - which does what I was proposing above. There's a small hole - namely, you could write code like this:

```rust
#[cfg(not(any(feature = "derive", test)))]
impl_or_verify!(T => FromZeroes for Unalign<T>);
```

...which would only compile the impls, but would not compile when the derive was used, and thus wouldn't perform its verification function. I'll try to figure out how to plug this hole, but a) I suspect it may be hard to do ergonomically and, b) I consider it a pretty minor risk.

djkoloski (approved):

Issue #159, opened 2023-02-16
Add `transmute_ref!` and `transmute_mut!` macros


joshlf:
We currently have a `transmute!` macro which safely transmutes an `AsBytes` type into a `FromBytes` type. We should add macros which do the same for mutable and immutable references.

A candidate implementation of `transmute_ref!` is in https://github.com/google/zerocopy/pull/183. We may also want to ditch it in favor of something based on [this approach](https://github.com/google/zerocopy/pull/183#discussion_r1208678157) suggested by @gootorov, but we'd need to first confirm that invalid uses are guaranteed to generate compile-time rather than runtime errors (https://github.com/rust-lang/rust/issues/112090). EDIT: Based on the discussion there, it seems that there is such a guarantee, but we'd need to be careful about how we structure things to make sure that we actually write code that benefits from that guarantee.

Note that https://github.com/rust-lang/rust/issues/112301 will likely make @gootorov's technique a bit annoying for users (although the error messages are _way_ better than the existing technique, and the implementation is _way_ simpler, so IMO it's still worth it on balance).

EDIT: Another issue with this approach is that generic APIs might never fail until they're used by a downstream crate since monomorphization only happens at that point. One potential solution would be to keep the functions `#[doc(hidden)]` and provide macros to invoke them which also call `mem::transmute` in order to ensure that they can only be called from type-concrete contexts. This would artificially hamstring the API, but the footgun is big enough that it might be worth it.

joshlf:
Closed in https://github.com/google/zerocopy/pull/183 and https://github.com/google/zerocopy/pull/550.

PR #183, opened 2023-05-25
Add `transmute_ref!` macro

joshlf:
This macro is like the existing `transmute!`, but it transmutes immutable references rather than values.

Issue #159

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1652,7 +1652,7 @@ macro_rules! transmute {
             // `AsBytes` and that the type of this macro invocation expression
             // is `FromBytes`.
             const fn transmute<T: $crate::AsBytes, U: $crate::FromBytes>(_t: T) -> U {
-                unreachable!()
+                loop {}
             }
             transmute(e)
         } else {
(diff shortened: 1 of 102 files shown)


joshlf:
@djkoloski Pinging you in case you have opinions about this.

src/lib.rs
```diff
@@ -1463,6 +1463,177 @@ macro_rules! transmute {
...
+///
+/// This is saying that `max(align_of::<T>(), align_of::<U>()) !=
+/// align_of::<T>()`, which is equivalent to `align_of::<T>() <
+/// align_of::<U>()`.
+#[macro_export]
+macro_rules! transmute_ref {
+    ($e:expr) => {{
+        // NOTE: This must be a macro (rather than a function with trait bounds)
+        // because there's no way, in a generic context, to enforce that two
+        // types have the same size or alignment.
```
gootorov, line 1548: There's one "trick" with traits and associated constants you can do:
```Rust
use std::mem;
use std::marker::PhantomData;

struct AlignChecker<T, U>(PhantomData<T>, PhantomData<U>);

trait AlignCheck {
    const ALIGN_OF_T: usize;
    const ALIGN_OF_U: usize;
    const ALIGN_EQ_PROOF: () =
        assert!(Self::ALIGN_OF_T == Self::ALIGN_OF_U, "T and U have different alignment");
    
    fn noop();
}

impl<T, U> AlignCheck for AlignChecker<T, U> {
    const ALIGN_OF_T: usize = mem::align_of::<T>();
    const ALIGN_OF_U: usize = mem::align_of::<U>();
    
    #[inline(always)]
    fn noop() {
        // NOTE: This does not compile to anything.
        // However, this `let` must be present to force the compiler to const evaluate our checks.
        let _consteval_proof = Self::ALIGN_EQ_PROOF;
    }
}
```
Then, `noop` as-is is zero-cost*
```Rust
// does not compile to anything*
AlignChecker::<u8, u8>::noop();

// but this produces a compilation error
AlignChecker::<u8, u16>::noop();
```
So, you can modify `noop` to contain your generic code.

* About zero-cost. So, `noop` is compiled to just a single `ret`. Calling `noop` produces a `call` instruction, unless it is inlined. https://godbolt.org/z/YnaMvxYqh

Sorry if this isn't useful/already known. Sharing just in case, because I do not see this technique used often :)

joshlf: This is a fantastic idea, thanks! We'll definitely consider it. Just want to confirm first that the check is guaranteed to catch any unsoundness: https://github.com/rust-lang/unsafe-code-guidelines/issues/409

joshlf: Here's another thread about it: https://github.com/rust-lang/rust/issues/112090

joshlf: > Here's another thread about it: [rust-lang/rust#112090](https://github.com/rust-lang/rust/issues/112090)

Okay, based on the discussion there, it seems that there is such a guarantee, but we'd need to be careful about how we structure things to make sure that we actually write code that benefits from that guarantee.

gootorov: Ah, and maybe in the future it would be possible to make error messages a bit nicer!
```Rust
#![feature(const_type_name)] // (and const formatting)
use std::any::type_name;

const SAFE_TO_TRANSMUTE: () = assert!(
    mem::size_of::<T>() == mem::size_of::<U>() && mem::align_of::<T>() >= mem::align_of::<U>(),
    "It is not safe to transmute {} into {}",
    type_name::<T>(),
    type_name::<U>()
);
```

joshlf: Yeah something like this would be awesome! I'm not gonna get a chance to work on this much for the next few weeks, but feel free to leave more comments here, and I'll get to them when I'm working on this again!

joshlf: Note that https://github.com/rust-lang/rust/issues/112301 will likely make this technique a bit annoying for users (although the error messages are _way_ better than the existing technique, and the implementation is _way_ simpler, so IMO it's still worth it on balance).

jswrenn: I've made a branch experimenting with this approach: https://github.com/google/zerocopy/pull/190

src/lib.rs
```diff
@@ -1674,6 +1674,140 @@ macro_rules! transmute {
...
+/// outlive the lifetime of the output type, `&U`.
+///
+/// # Alignment increase error message
+///
+/// Because of limitations on macros, the error message generated when
+/// `transmute_ref!` is used to transmute from a type of lower alignment to a
+/// type of higher alignment is somewhat confusing. For example, the following
+/// code:
+///
+/// ```rust,compile_fail
```
jswrenn, line 1695:
```suggestion
/// ```compile_fail
```

src/lib.rs
```diff
@@ -1674,6 +1674,140 @@ macro_rules! transmute {
...
+        let e = &*$e;
+
+        #[allow(unused, clippy::diverging_sub_expression)]
+        if false {
+            // This branch, though never taken, ensures that the type of `e` is
+            // `&T` where `T: 't + Sized + AsBytes`, that the type of this macro
+            // expression is `&U` where `U: 'u + Sized + FromBytes`, and that
+            // `'t` outlives `'u`.
+            const fn transmute<'u, 't: 'u, T: 't + Sized + $crate::AsBytes, U: 'u + Sized + $crate::FromBytes>(_t: &'t T) -> &'u U {
+                unreachable!()
```
jswrenn, line 1736:
```suggestion
                loop {}
```
Replacing these `unreachable!()` invocations with `loop {}` will reduce codgen bloat, and keep panics out of `transmute!`, which may be relevant for #202.

src/lib.rs
```diff
@@ -3810,6 +3944,92 @@ mod tests {
...
+        }
+
+        assert_t_align_gteq_u_align!(u8, u8, true);
+        assert_t_align_gteq_u_align!(AU64, AU64, true);
+        assert_t_align_gteq_u_align!(AU64, u8, true);
+        assert_t_align_gteq_u_align!(u8, AU64, false);
+    }
+
+    #[test]
+    fn test_transmute_ref() {
```
jswrenn, line 3999: Can you also add a check here that `transmute_ref`'s argument is only evaluated once?

tests/ui-nightly/transmute-ref-dst-mutable.stderr
jswrenn, line 1: The triple-reporting of the same error is unfortunate. I think we can punt on the issue for now (let's land something that works, then we can hone error messages), but we should revisit this — especially if we manage to re-implement this macro atop const fns.

tests/ui-nightly/transmute-ref-alignment-increase.stderr
```diff
@@ -0,0 +1,9 @@
+error[E0512]: cannot transmute between types of different sizes, or dependently-sized types
+  --> tests/ui-nightly/transmute-ref-alignment-increase.rs:15:35
+   |
+15 | const INCREASE_ALIGNMENT: &AU16 = transmute_ref!(&[0u8; 2]);
+   |                                   ^^^^^^^^^^^^^^^^^^^^^^^^^
+   |
+   = note: source type: `MaxAlignsOf<[u8; 2], AU16>` (16 bits)
+   = note: target type: `AlignOf<[u8; 2]>` (8 bits)
```
jswrenn, line 8: In the implementation of `transmute_ref`, is it possible to reverse the order of the alignment-checking transmute, such that `AlignOf<[u8; 2]` is the 'source type' in this error message?

tests/ui-nightly/transmute-ref-dst-mutable.stderr
joshlf: Agreed.

tests/ui-nightly/transmute-ref-alignment-increase.stderr
joshlf: Good idea! Done.

jswrenn (approved):

jswrenn (approved):

src/lib.rs
gootorov: Exciting news! With `inline_const` being stabilized, the following check will soon be possible on stable compiler:
```Rust
fn check_align<T, U>() {
    const {
        assert!(
            mem::size_of::<T>() == mem::size_of::<U>() 
                && mem::align_of::<T>() >= mem::align_of::<U>()
        );
    }
}
```
Link to nightly playground:
https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=24d5ca56b2834fdfde8ec7681359e729

joshlf: Unfortunately our MSRV is currently 1.56, so we won't be able to use that for a while, but I'm sure we'll make use of it once our MSRV is high enough.

PR #189, opened 2023-06-05
Add more UI tests of `transmute!`

jswrenn:
I found that I was unable to easily review changes made atop #183 due to its omnibus UI test producing very long stderr files. This PR lifts the tests of `transmute!` added in #183, but takes a test-per-file-approach. A test-per-file approach is taken so that reviewers of future PRs that alter the implementation details of `transmute!` can assess each dimension of change discretely and incrementally — as opposed to contending with `.stderr` files hundreds of lines in length.


joshlf:
Implemented in https://github.com/google/zerocopy/pull/456.

PR #190, opened 2023-06-05
Replace `transmute!` macro with `transmute` fn

jswrenn:
As a consequence of this change, the phase in which some invalid transmutes are detected differs. The `transmute!` macro depended internally on `mem::transmute`, which cannot be used in generic contexts but provides pre-monomorphization errors if the source and destination types have different sizes. By contrast, the `transmute` function can be used in generic contexts, but does not report size errors until monomorphization.

This increase in expressivity comes at the cost of some usability: size errors are currently only reported by invocations of `cargo build`; not `cargo check`.


src/lib.rs
```diff
@@ -1413,54 +1413,48 @@ impl<T: Unaligned + Display> Display for Unalign<T> {
     }
 }
 
-// Used in `transmute!` below.
 #[doc(hidden)]
-pub use core::mem::transmute as __real_transmute;
+pub trait TransmuteFrom<Src>: Sized {
```
kupiakos, line 1417: This trait does not need to be (and shouldn't be) `pub`, because it has a blanket implementation on `T: Sized` that `transmute` can use.

PR #194, opened 2023-06-30
Migrate UI tests to `ui-test` crate

jswrenn:
The `ui-test` crate allows for annotating the tests themselves with the expected error message. This will make changes to `zerocopy` or its tests that alter error messages easier to review, as the `.stderr` file can be merely skimmed.

This PR is marked as a draft, pending the resolution of a `ui-test` bug that causes it to fail to resolve `zerocopy` when `zerocopy`'s version is a pre-release.


oli-obk:
I am unable to reproduce the pre-release failure with either the pinned nightly or the pinned stable. I'm guessing there was some update that wasn't tracked (`Cargo.lock` is not committed, so I can't tell).

Cargo.toml
```diff
@@ -49,10 +49,9 @@ optional = true
...
-# Pinned to a specific version so that the version used for local development
-# and the version used in CI are guaranteed to be the same. Future versions
-# sometimes change the output format slightly, so a version mismatch can cause
-# CI test failures.
-trybuild = "=1.0.80"
+test-util = { path = "./test-util" }
+
+[[test]]
+name = "ui"
+harness = false
```
joshlf, line 57: This is problematic because, in CI, we do `cargo test -- --skip ui` in order to skip this test when running under Miri. With a custom test harness, the command line flags aren't parsed (unless we do it manually), and so `--skip ui` does nothing. Is there any way that we could use the normal test harness and retain the same behavior in this PR?

test-util/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+pub use ui_test::*;
+
+#[rustversion::nightly(2023-05-24)]
+pub static TOOLCHAIN: &str = "nightly-2023-05-25";
+
+#[rustversion::stable(1.69.0)]
+pub static TOOLCHAIN: &str = "1.69.0";
+
+#[rustversion::stable(1.65.0)]
+pub static TOOLCHAIN: &str = "1.65.0";
```
joshlf, line 15: Since you wrote runtime code to extract these versions below, maybe we could make everything runtime-dynamic, including this? That would mean that we wouldn't have to update these version numbers in source code when we update the versions we pin in CI - we'd only have to update in `Cargo.toml`.

In particular, I'm thinking that we could do the following at runtime:
- Extract pinned versions like you currently do in `pinned_toolchains`
- Check what version is being used (either to build the code or to run the code - since this is a test environment, they should be the same version anyway)
- Match that against the pinned versions or error out if there's no match

test-util/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+/// Bless tests with `cargo test -- -- --bless`.
+pub fn should_bless() -> bool {
+    std::env::args().any(|arg| arg == "--bless")
+}
+
+/// Run (and optionally bless) UI tests in a given folder, within a given toolchain.
+pub fn ui_test(toolchain: &str, folder: &str, bless: bool) -> color_eyre::eyre::Result<()> {
+    let toolchain = String::from("+") + toolchain;
+
+    // Make sure we can depend on zerocopy in ui tests.
```
joshlf, line 62: Not sure I understand how this line of code accomplishes what's documented in this comment; maybe add a bit more detail?

zerocopy-derive/Cargo.toml
```diff
@@ -25,9 +25,9 @@ syn = { version = "2", features = ["visit"] }
...
-# and the version used in CI are guaranteed to be the same. Future versions
-# sometimes change the output format slightly, so a version mismatch can cause
-# CI test failures.
-trybuild = "=1.0.80"
+test-util = { path = "../test-util" }
 zerocopy = { path = "../" }
+
+[[test]]
+name = "ui"
+harness = false
```
joshlf, line 33: Same comment here about not using the default test harness.

Cargo.toml
oli-obk: we can make the test itself detect `--skip ui` arguments and just return early. Or we use `#[cfg(miri)]` to bail out immediately

PR #204, opened 2023-07-29
Replace generic ByteSlices with concrete slices

joshlf:
In `FromBytes` and `AsBytes`, we have some methods which took a `B: ByteSlice` or `B: ByteSliceMut` bound which was unnecessary. In this commit, we simplify those methods by just taking `&[u8]` or `&mut [u8]` instead. This simplifies the API and also may reduce monomorphization bloat for some users.


djkoloski (approved):
LGTM, I assume this is because all `ByteSlice` can be deref'd to `&[u8]` or `&mut [u8]` right?

Issue #61, opened 2022-10-15
Add SAFETY comments to all unsafe code


joshlf:
Deny the [`undocumented_unsafe_blocks`](https://rust-lang.github.io/rust-clippy/master/#undocumented_unsafe_blocks) Clippy lint and document all currently-undocumented unsafe blocks with a `// SAFETY: ...` comment.

briansmith:
In issue #716, @ijackson asked for a postmortem (my characterization) which I think would be useful as part of this effort.

I think it's also important to look at how the CVE/RUSTSEC advisory and yanking for #716 affected many crates that depend on ahash (and probably other crates that depend on zerocopy), even though almost none of them had a real impact due to #716 since they weren't using the affected API in an affected way (AFAICT), but vulnerability scanning / advisory response software available to us isn't advanced enough to "see" that. The cost of these kinds of false-positive external impacts need to be accounted for in some way. I haven't seen this aspect mentioned elsewhere.

Issue #8, opened 2022-09-19
Document soundness requirements around references


joshlf:
*Migrated from https://fxbug.dev/108820*

Currently, `FromBytes` and `AsBytes` are documented as simply requiring that a type may be converted from an arbitrary sequence of bytes or may be converted to a sequence of bytes (respectively). However, this isn't actually sufficient in practice given the way we use these traits. In particular, we allow converting from `&[u8]` to `&T` (where `T: FromBytes`) and from `&T` to `&[u8]` (where `T: AsBytes`). The `UnsafeCell<T>` type can be converted from an arbitrary sequence of bytes if `T` can. However, if we were to implement `FromBytes` for `UnsafeCell<T>`, it would make these reference conversions unsound, as code with a `&T` could perform interior mutability which code with the `&[u8]` wouldn't know about.

We need to expand the soundness requirements on `FromBytes` and `AsBytes` to explicitly mention this reference-safety requirement.

See also https://github.com/google/zerocopy/issues/251 for an alternate approach.

PR #227, opened 2023-08-05
Document some unsafe blocks

joshlf:
Makes progress on #61

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -568,15 +568,33 @@ pub unsafe trait AsBytes {
     /// `as_bytes` provides access to the bytes of this value as an immutable
     /// byte slice.
     fn as_bytes(&self) -> &[u8] {
-        // TODO(#61): Add a "SAFETY" comment and remove this `allow`.
-        #[allow(clippy::undocumented_unsafe_blocks)]
-        unsafe {
-            // Note that this method does not have a `Self: Sized` bound;
-            // `size_of_val` works for unsized values too.
-            let len = mem::size_of_val(self);
-            let slf: *const Self = self;
-            slice::from_raw_parts(slf.cast::<u8>(), len)
-        }
+        // Note that this method does not have a `Self: Sized` bound;
+        // `size_of_val` works for unsized values too.
+        let len = mem::size_of_val(self);
+        let slf: *const Self = self;
+
+        // SAFETY:
+        // - `slf.cast::<u8>()` is valid for reads for `len *
+        //   mem::size_of::<u8>()` many bytes because...
+        //   - `slf` is the same pointer as `self`, and `self` is a reference
+        //     which points to an object whose size is `len`. Thus...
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -568,15 +568,33 @@ pub unsafe trait AsBytes {
...
-            slice::from_raw_parts(slf.cast::<u8>(), len)
-        }
+        // Note that this method does not have a `Self: Sized` bound;
+        // `size_of_val` works for unsized values too.
+        let len = mem::size_of_val(self);
+        let slf: *const Self = self;
+
+        // SAFETY:
+        // - `slf.cast::<u8>()` is valid for reads for `len *
+        //   mem::size_of::<u8>()` many bytes because...
```
djkoloski, line 578: Nit: `mem::size_of::<u8>()` is guaranteed to be 1, so we can just say `len` many bytes.

src/lib.rs
```diff
@@ -568,15 +568,33 @@ pub unsafe trait AsBytes {
...
+        //     - The entire region of `len` bytes starting at `slf` is contained
+        //       within a single allocation.
+        //     - `slf` is non-null.
+        //   - `slf` is trivially aligned to `align_of::<u8>() == 1`.
+        // - `Self: AsBytes` ensures that all of the bytes of `slf` are
+        //   initialized.
+        // - Since `slf` is derived from `self`, and `self` is an immutable
+        //   reference, the only other references to this memory region that
+        //   could exist are other immutable references, and those don't allow
+        //   mutation.
```
djkoloski, line 590: This argument is technically not sufficient, as a `Cell<usize>` or other type with interior mutability could mutate the bytes after getting the byte slice. This case is covered by the safety requirements of `AsBytes` which state that implementors guarantee that it is sound to treat the type as a byte slice of `size_of::<T>()` len.

I think an appropriate change here would be:
- More explicitly ban interior mutability for `AsBytes` types.
- Call out that `AsBytes` allows us to treat this as a byte slice.

Same thing below for `as_bytes_mut`.

djkoloski (changes requested):

src/lib.rs
joshlf: Yeah, I'm just trying to be pedantic and mirror the way the requirements are laid out in the `from_raw_parts` docs.

src/lib.rs
joshlf: Yeah, we have that tracked: https://github.com/google/zerocopy/issues/8

I was hoping that the following `TODO` would capture that since, tbh, `AsBytes` is basically entirely unsound right now since the safety requirements don't mention interior mutability.

As for `as_bytes_mut`, my understanding is that interior mutability isn't relevant since `&mut self` ensures that there are no other references full stop. Does that sound right to you?

djkoloski: Yep, that sounds right to me. If we have it tracked then LGTM

djkoloski (approved):

PR #234, opened 2023-08-08
Reduce MSRV to 1.61.0

joshlf:
It turned out that there were only a few lines of code that required our MSRV to be as high as it was before this commit (1.65.0), and those lines were easy to modify to be compatible with this new MSRV.


src/lib.rs
```diff
@@ -334,15 +334,22 @@ pub unsafe trait FromZeroes {
...
+            .checked_mul(len)
+            .expect("mem::size_of::<Self>() * len overflows `usize`");
+        let align = mem::align_of::<Self>();
+        // On stable Rust versions <= 1.64.0, `Layout::from_size_align` has a
+        // bug in which sufficiently-large allocations (those which, when
+        // rounded up to the alignment, overflow `isize`) are not rejected,
+        // which can cause undefined behavior. See #64 for details.
+        //
+        // TODO(#67): Once our MSRV is > 1.64.0, remove this assertion.
+        let max_alloc = (usize::MAX / 2).saturating_sub(align);
```
djkoloski, line 347: Nit: I think this would be clearer as `(isize::MAX as usize).saturating_sub(align)`

djkoloski (approved):
LGTM

src/lib.rs
joshlf: Good call; done.

Issue #230, opened 2023-08-07
Install OpenSSF Scorecard and consider adopting its recommendations


joshlf:
This derives from a request by Google's security team, which is reproduced here:

> We ask that you please:
> 1. install the [OpenSSF Scorecard GitHub Action](https://github.com/ossf/scorecard-action/) ([instructions](https://github.com/ossf/scorecard-action/#installation)) in your repository.
> 2. adopt its suggestions to improve your project's security posture.
> 
> A preliminary run of the [OpenSSF Scorecard](https://securityscorecards.dev/) has identified the following improvements that can be made to the project, followed by their risk level and a summary of the remediation steps:
> 
> * [Token-Permissions](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) (High): declare GitHub workflow permissions top level as read only, and grant any write permission on job level. The changes can be seen at https://app.stepsecurity.io/secureworkflow/google/zerocopy/ci.yml/main?enable=permissions.
> * [Branch-Protection](https://github.com/ossf/scorecard/blob/main/docs/checks.md#branch-protection) (High): configure the recommended branch protection settings.
> * [Dependency-Update-Tool](http://dependency-update-tool/) (High): if applicable, use tools to help update the project's dependencies (package github-actions for example).
> * [Code-Review](https://github.com/ossf/scorecard/blob/main/docs/checks.md#code-review) (High): ensure to review code before merging it (Branch-Protection can help to ensure that).
> * [Pinned-Dependencies](https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies) (Medium): declare and hash pin all dependencies. The changes can be seen at https://app.stepsecurity.io/secureworkflow/google/zerocopy/ci.yml/main?enable=pin.
> * [SAST](https://github.com/ossf/scorecard/blob/main/docs/checks.md#sast) (Medium): use static code analysis tools, e.g. CodeQL, SonarCloud.
> * [CII-Best-Practices](https://github.com/ossf/scorecard/blob/main/docs/checks.md#cii-best-practices) (Low): add an OpenSSF (formerly CII) Best Practices Badge.
> * [Fuzzing](https://github.com/ossf/scorecard/blob/main/docs/checks.md#fuzzing) (Medium): use fuzzing tools, e.g. [OSS-Fuzz](https://github.com/google/oss-fuzz).

[...]

PR #235, opened 2023-08-08
Add badge to the README.md

joycebrum:
It helps on easily see what are the current scores.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,8 @@ made in the doc comment on `src/lib.rs` or in `generate-readme.sh`.
 
 # zerocopy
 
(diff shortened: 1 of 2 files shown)


joshlf:
We auto-generate our README from the doc comment at the top of `src/lib.rs`, so I'll need to figure out a way to include this separately. (That's why [this job](https://github.com/google/zerocopy/actions/runs/5801788174/job/15726843503?pr=235) is failing.)

joycebrum:
Fixed @joshlf

joshlf:
This was very useful to spur work on #230 , but I don't think this badge is something our Rust users will care about, so I'm going to close this. Really appreciate all your help on #230 though!

PR #272, opened 2023-08-16
[CI] Test on loongarch64-unknown-linux-gnu target

joshlf:


joshlf:
Turns out this isn't supported by Miri, which means this doesn't buy us much over the existing riscv target, and adds some complexity to CI, so not worth it IMO.

PR #279, opened 2023-08-22
[WIP] Try from bytes

joshlf:
Feel free to review or comment, but DO NOT MERGE; field projection needs to be landed in a separate PR first before this can be merged.

### DIFF
--- a/src/derive_util.rs
+++ b/src/derive_util.rs
@@ -62,3 +62,62 @@ macro_rules! union_has_padding {
         false $(|| core::mem::size_of::<$t>() != core::mem::size_of::<$ts>())*
     };
 }
+
+/// Implements `TryFromBytes` for a struct type by delegating to existing
+/// implementations for each of its fields, and optionally supports a custom
+/// validation method.
+///
+/// ```rust
+/// # use zerocopy::impl_try_from_bytes_for_struct;
+///
+/// #[repr(C)]
+/// struct Foo {
+///     a: u8,
+///     b: u16,
+/// }
+///
+/// impl_try_from_bytes_for_struct!(Foo { a: u8, b: u16 });
+///
+/// #[repr(transparent)]
+/// struct Bar(Foo);
+///
+/// impl Bar {
+///     fn is_valid(&self) -> bool {
+///         u16::from(self.0.a) < self.0.b
+///     }
+/// }
+///
+/// impl_try_from_bytes_for_struct!(Bar { 0: Foo } => is_valid);
+/// ```
+///
+/// # Safety
+///
+/// `$ty` must be a struct type, `$f` must list every field's name, and `$f_ty`
+/// must be the correct types for those fields.
+#[doc(hidden)] // `#[macro_export]` bypasses this module's `#[doc(hidden)]`.
+#[macro_export]
+macro_rules! impl_try_from_bytes_for_struct {
+    ($ty:ty { $($f:tt: $f_ty:ty),* } $(=> $validation_method:ident)?) => {
+        // SAFETY: The caller promises that all fields are listed with their
+        // correct types. We validate that every field is valid, which is the
+        // only requirement for the entire struct to be valid. Thus, we
+        // correctly implement `is_bit_valid` as required by the trait's safety
+        // documentation.
+        #[allow(unused_qualifications)]
+        unsafe impl zerocopy::TryFromBytes for $ty {
+            fn is_bit_valid(bytes: &zerocopy::MaybeValid<Self>) -> bool {
+                true $(&& {
+                    let f: &zerocopy::MaybeValid<$f_ty> = zerocopy::project!(&bytes.$f);
+                    zerocopy::TryFromBytes::is_bit_valid(f)
+                })*
+                $(&& {
+                    // SAFETY: We just validated that all of the struct's fields
+                    // are valid, which means that the struct itself is valid.
+                    // That is the only precondition of `assume_valid_ref`.
+                    let slf = unsafe { bytes.assume_valid_ref() };
+                    slf.$validation_method()
+                })?
+            }
+        }
+    }
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -162,6 +162,12 @@ pub mod byteorder;
 #[doc(hidden)]
 pub mod derive_util;
 
+// TODO(#196): Once we figure out how to make the `project!` macro safe
+// (specifically, by statically rejecting projection through `deref`), remove
+// this `#[doc(hidden)]` and make `project` part of our public API.
+#[doc(hidden)]
+pub mod project;
+
 #[cfg(feature = "byteorder")]
 pub use crate::byteorder::*;
 #[cfg(any(feature = "derive", test))]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -188,6 +188,8 @@ use core::{
     ptr, slice,
 };
 
+use project::Projectable;
+
 #[cfg(feature = "alloc")]
 extern crate alloc;
 #[cfg(feature = "alloc")]
@@ -399,7 +401,8 @@ pub unsafe trait FromZeroes {
         #[allow(clippy::as_conversions)]
         let max_alloc = (isize::MAX as usize).saturating_sub(align);
         assert!(size <= max_alloc);
-        // TODO(https://github.com/rust-lang/rust/issues/55724): Use `Layout::repeat` once it's stabilized.
+        // TODO(https://github.com/rust-lang/rust/issues/55724): Use
+        // `Layout::repeat` once it's stabilized.
         let layout =
             Layout::from_size_align(size, align).expect("total allocation size overflows `isize`");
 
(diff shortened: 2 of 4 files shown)


src/derive_util.rs
```diff
@@ -62,3 +62,68 @@ macro_rules! union_has_padding {
...
+                true $(&& {
+                    let f: &zerocopy::MaybeValid<$f_ty> = zerocopy::project!(&bytes.$f);
+                    zerocopy::TryFromBytes::is_bit_valid(f)
+                })*
+                $(&& {
+                    // SAFETY: We just validated that all of the struct's fields
+                    // are valid, which means that the struct itself is valid.
+                    // That is the only precondition of `assume_valid_ref`.
+                    let slf = unsafe { bytes.assume_valid_ref() };
+                    // TODO: What about interior mutability? One approach would
```
kupiakos, line 118: For now, we just forbid interior mutability for `TryFromBytes`

joshlf: Updated the comment.

While we're on the subject, though, how do we actually ban interior mutability in `MaybeValid`? Currently, we implement `AsMaybeUninit` for all sized types and slices of sized types, which means that impls like the following exist automatically thanks to those blanket impls:

```rust
unsafe impl<T> AsMaybeUninit for UnsafeCell<T> {
    type MaybeUninit = MaybeUninit<UnsafeCell<T>>;
}
```

That, in turn, means that it's valid to write `MaybeValid<UnsafeCell<T>>`.

Maybe we could add safety preconditions to methods like `assume_valid_ref` that `T::MaybeUninit` doesn't contain any `UnsafeCell`s? Then, so long as we only implement `TryFromBytes` for types without `UnsafeCell`s, and we only ever emit calls to `assume_valid_ref` in such derived impls, then we should be okay.

Any other ideas? It'd be nice to not have this be so viral.

src/macros.rs
```diff
@@ -13,51 +13,93 @@
...
+            unsafe_impl!(@method $trait $(; |$candidate: &$repr| $is_bit_valid)?);
         }
     };
+
+    (@method TryFromBytes ; |$candidate:ident: &$repr:ty| $is_bit_valid:expr) => {
+        fn is_bit_valid(candidate: &MaybeValid<Self>) -> bool {
+            // SAFETY: TODO
+            // TODO: Why is this `allow` necessary? Why is it acceptable?
+            #[allow(clippy::as_conversions)]
+            let $candidate = unsafe { &*(candidate as *const MaybeValid<Self> as *const $repr) };
```
joshlf, line 92: @kupiakos since you're the as-casting expert, would you mind taking a look at the as casts in this PR and confirming that they're kosher? I'm pretty sure they are, but Clippy is making me question my sanity...

src/derive_util.rs
kupiakos: > how do we actually ban interior mutability in MaybeValid

There's no reason to. The only place where shared mutability gets involved in `MaybeValid` is through the `TryFromBytes::is_bit_valid` method - that's the one that needs guarantees that there is only a single reference, either through `unsafe` or through banning `UnsafeCell` for `TryFromBytes` entirely.

> That, in turn, means that it's valid to write MaybeValid<UnsafeCell<T>>.

Yes, you can go from `&mut MaybeValid<T>` to `&mut T` on any type via a `.write()`. That's fine though - this doesn't involve shared mutability due to `&mut`.

> Maybe we could add safety preconditions to methods like assume_valid_ref that T::MaybeUninit doesn't contain any UnsafeCells

Again, I don't see why this is in _any_ way necessary.

> See https://github.com/rust-lang/unsafe-code-guidelines/issues/455 for more discussion.

That's essentially everything I told you offline - SB bans it, TB is fine. `MaybeUninit` does not interact with shared mutability, and neither should the `AsMaybeUninit` trait.

src/macros.rs
kupiakos: I'm not able to reproduce the clippy warning [on the playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ce252f166d962b764ee6c716f7761ecb). Are you running a version of clippy with a bug?

joshlf: It's only produced in 1.61 (our MSRV, which is tested in CI).

src/macros.rs
```diff
@@ -13,51 +13,93 @@
...
 
 /// Unsafely implements trait(s) for a type.
+///
+/// # Safety
+///
+/// The trait impl must be sound.
+///
+/// When implementing `TryFromBytes`:
+/// - If no `is_bit_valid` impl is provided, then it must be valid for
+///   `is_bit_valid` to unconditionally return `true`.
```
kupiakos, line 39: Why allow there to be no `is_bit_valid` impl at all? Why would `true` implementations be common enough to justify a default method?

src/project.rs
```diff
@@ -0,0 +1,659 @@
...
+        #[allow(unused_unsafe)]
+        let token = unsafe { $crate::project::unsafe_token::UnsafeToken::new() };
+        use ::core::borrow::$borrow_trait as _;
+        $crate::project::$project_fn(
+            token,
+            // We call `borrow` or `borrow_mut` so that this macro works
+            // regardless of whether `$c` is owned or borrowed.
+            //
+            // TODO(#196): Is there any way to make sure this calls
+            // `Borrow::borrow` or `BorrowMut::borrow_mut` even if the type has
```
kupiakos, line 369: Not without spelling the type of `$c`.

`type alias impl Trait` could make this possible.

src/macros.rs
joshlf: It's used whenever you're implementing `TryFromBytes` for a type which is also `FromBytes` - in this PR, ~30 times.

src/project.rs
joshlf: Yeah, that's what I figured.

Also, could you leave comments for this module on https://github.com/google/zerocopy/pull/292? I hate GitHub...

src/derive_util.rs
joshlf: Okay, I believe I've completely handled the interior mutability issue with [this update](https://github.com/google/zerocopy/compare/1ecc13acb232d7916a29ff64e9fa59951f7d55ab..6a4ff328652e1dc4fc7782bd79571c97295d3df1). Lmk if this looks sound to you?

For a bit more detail on my thinking (I think we're both on the same page now, but just to be sure), I had two concerns:
- Is it sound to have `&MaybeValid<T>` and `&T` at the same time? This is required in order to call custom validators.
  - This is now sound thanks to the requirement that `T` and `T::MaybeUninit` have `UnsafeCell`s at the same offsets.
- Is it sound to have `&MaybeValid<T>` and `&[u8]` at the same time? This is required in order to implement `try_from_ref` and friends since they take a `&[u8]` and convert it to a `&MaybeValid<T>` in order to call `is_bit_valid`.
  - This is sound so long as `T::MaybeUninit` doesn't contain any `UnsafeCell`s. Thanks to the `UnsafeCell`s-at-the-same-locations requirement, this is sound so long as `T` doesn't contain any `UnsafeCell`s. Since we only implement `FromZeroes` and `FromBytes` for `MaybeValid<T>` where `T: TryFromBytes` (and since `T: TryFromBytes` bans `UnsafeCell`s), we're okay. Given those impls, we only use safe code (namely, `Ref` and `FromBytes` methods) to convert `&[u8]` to `&MaybeValid<T>`.

src/project.rs
kupiakos: Oh wait, I just tried something. So long as the output type is non-ambiguous you can write `<_ as $borrow_trait>::$borrow_method($c)`. How does that work?

joshlf: dafuq...

I wonder if the compiler desugars them to the same thing (with the caveat that `$c.$borrow_method()` needs to also perform method resolution) and so they both get the same auto-ref treatment.

src/lib.rs
```diff
@@ -508,6 +517,425 @@ pub unsafe trait FromBytes: FromZeroes {
...
+    /// impls of those traits. Perhaps we need to distinguish between "contains
+    /// an `UnsafeCell`" and "allows interior mutability" (where the latter
+    /// means something like "contains an `UnsafeCell` and allows it to be used
+    /// to perform shared mutation").
+    unsafe_impl!(T => TryFromBytes for MaybeValid<T>);
+    unsafe_impl!(T => TryFromBytes for MaybeValid<[T]>);
+    unsafe_impl!(T => FromZeroes for MaybeValid<T>);
+    unsafe_impl!(T => FromZeroes for MaybeValid<[T]>);
+    unsafe_impl!(T => FromBytes for MaybeValid<T>);
+    unsafe_impl!(T => FromBytes for MaybeValid<[T]>);
```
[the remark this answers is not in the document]
joshlf: Should be fixed with [this update](https://github.com/google/zerocopy/compare/1ecc13acb232d7916a29ff64e9fa59951f7d55ab..6a4ff328652e1dc4fc7782bd79571c97295d3df1).

src/project.rs
kupiakos: In the `<_ as $borrow_trait>::$borrow_method($c)` you are _guaranteeing_ you're not running an inherent method and are running the trait method you intended, except you're leaving the concrete type that implements the trait as inferred.

src/macros.rs
kupiakos: Ah, I would instead have that be covered by the implementing macro instead of exposing that as an extra thing for external users to consider

joshlf: We could do that, it'd just make the usage sites a lot more verbose. Currently, [you can do](https://github.com/google/zerocopy/pull/279/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R1211-R1223):

```rust
unsafe_impl!(u8: TryFromBytes, FromZeroes, FromBytes, AsBytes, Unaligned);
unsafe_impl!(i8: TryFromBytes, FromZeroes, FromBytes, AsBytes, Unaligned);
unsafe_impl!(u16: TryFromBytes, FromZeroes, FromBytes, AsBytes);
unsafe_impl!(i16: TryFromBytes, FromZeroes, FromBytes, AsBytes);
unsafe_impl!(u32: TryFromBytes, FromZeroes, FromBytes, AsBytes);
unsafe_impl!(i32: TryFromBytes, FromZeroes, FromBytes, AsBytes);
unsafe_impl!(u64: TryFromBytes, FromZeroes, FromBytes, AsBytes);
unsafe_impl!(i64: TryFromBytes, FromZeroes, FromBytes, AsBytes);
unsafe_impl!(u128: TryFromBytes, FromZeroes, FromBytes, AsBytes);
unsafe_impl!(i128: TryFromBytes, FromZeroes, FromBytes, AsBytes);
unsafe_impl!(usize: TryFromBytes, FromZeroes, FromBytes, AsBytes);
unsafe_impl!(isize: TryFromBytes, FromZeroes, FromBytes, AsBytes);
```

If you needed to supply a closure, you'd have to do:

```rust
unsafe_impl!(u8: FromZeroes, FromBytes, AsBytes, Unaligned);
unsafe_impl!(i8: FromZeroes, FromBytes, AsBytes, Unaligned);
unsafe_impl!(u16: FromZeroes, FromBytes, AsBytes);
unsafe_impl!(i16: FromZeroes, FromBytes, AsBytes);
unsafe_impl!(u32: FromZeroes, FromBytes, AsBytes);
unsafe_impl!(i32: FromZeroes, FromBytes, AsBytes);
unsafe_impl!(u64: FromZeroes, FromBytes, AsBytes);
unsafe_impl!(i64: FromZeroes, FromBytes, AsBytes);
unsafe_impl!(u128: FromZeroes, FromBytes, AsBytes);
unsafe_impl!(i128: FromZeroes, FromBytes, AsBytes);
unsafe_impl!(usize: FromZeroes, FromBytes, AsBytes);
unsafe_impl!(isize: FromZeroes, FromBytes, AsBytes);

unsafe_impl!(u8: TryFromBytes; |_: &MaybeValid<u8>| true);
unsafe_impl!(i8: TryFromBytes; |_: &MaybeValid<i8>| true);
unsafe_impl!(u16: TryFromBytes; |_: &MaybeValid<u16>| true);
unsafe_impl!(i16: TryFromBytes; |_: &MaybeValid<i16>| true);
unsafe_impl!(u32: TryFromBytes; |_: &MaybeValid<u32>| true);
unsafe_impl!(i32: TryFromBytes; |_: &MaybeValid<i32>| true);
unsafe_impl!(u64: TryFromBytes; |_: &MaybeValid<u64>| true);
unsafe_impl!(i64: TryFromBytes; |_: &MaybeValid<i64>| true);
unsafe_impl!(u128: TryFromBytes; |_: &MaybeValid<u128>| true);
unsafe_impl!(i128: TryFromBytes; |_: &MaybeValid<i128>| true);
unsafe_impl!(usize: TryFromBytes; |_: &MaybeValid<usize>| true);
unsafe_impl!(isize: TryFromBytes; |_: &MaybeValid<isize>| true);
```

Since we require safety comments on all impls, you're forced to write down a justification for why a type is `TryFromBytes`. IMO it'd be pretty hard for both you and the reviewer to fail to notice that you'd implemented `TryFromBytes` unconditionally on a type that isn't actually `TryFromBytes`. Even if you didn't know how the macro works (namely, that omitting the validator means all bit patterns are valid), you presumably would be unsure of the conditions under which the type was `TryFromBytes` and would go read the macro docs to see how it behaves.

src/project.rs
joshlf: Marking this as resolved; we can [continue on the project PR](https://github.com/google/zerocopy/pull/292#discussion_r1309130530).

src/lib.rs
```diff
@@ -566,6 +575,484 @@ pub unsafe trait FromBytes: FromZeroes {
...
+    // - By the same token, their raw pointer types are trivially `as` castable
+    //   and preserve size.
+    // - By the same token, `[MaybeUninit<T>]` contains `UnsafeCell`s at the
+    //   same byte ranges as `MaybeValid<[T]>` does.
+    type MaybeUninit = [MaybeUninit<T>];
+}
+
+impl<T> Default for MaybeValid<T> {
+    fn default() -> MaybeValid<T> {
+        MaybeValid { inner: MaybeUninit::uninit() }
```
joshlf, line 767: This is unsound! It violates the initialization invariant of `MaybeValid`.

PR #283, opened 2023-08-25
Improve trait safety documentation

### LINKED ISSUE #8: Document soundness requirements around references

joshlf:
For each of our unsafe traits, clarify who can skip reading the Safety section.

For each of our unsafe traits (except for `Unaligned`):
- Clarify that it must be sound to construct `&[u8]` and `&T` to the same memory region (this addresses #8)
- Clarify that, in order to implement the trait, the type's fields need to satisfy the same requirements, but don't actually need to implement the trait
- Clarify that, in order to implement the trait, the type must not contain any `UnsafeCell`s

Closes #8

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -199,19 +199,40 @@ mod zerocopy {
 ///
 /// # Safety
 ///
-/// If `T: FromZeroes`, then unsafe code may assume that it is sound to treat
-/// any initialized sequence of zero bytes of length `size_of::<T>()` as a `T`.
+/// *This section describes what is required in order for `T: FromZeroes`, and
+/// what unsafe code may assume of such types. `#[derive(FromZeroes)]` only
+/// permits types which satisfy these requirements. If you don't plan on
+/// implementing `FromZeroes` manually, and you don't plan on writing unsafe
+/// code that operates on `FromZeroes` types, then you don't need to read this
+/// section.*
+///
+/// If `T: FromZeroes`, then unsafe code may assume that:
+/// - It is sound to treat any initialized sequence of zero bytes of length
+///   `size_of::<T>()` as a `T`.
+/// - Given `b: &[u8]` where `b.len() == size_of::<T>()`, `b` is aligned to
+///   `align_of::<T>()`, and `b` contains only zero bytes, it is sound to
+///   construct a `t: &T` at the same address as `b`, and it is sound for both
+///   `b` and `t` to be live at the same time.
+///
 /// If a type is marked as `FromZeroes` which violates this contract, it may
 /// cause undefined behavior.
 ///
-/// If a type has the following properties, then it is safe to implement
+/// If a type has the following properties, then it is sound to implement
 /// `FromZeroes` for that type:
-/// - If the type is a struct, all of its fields must implement `FromZeroes`
+/// - If the type is a struct, all of its fields must satisfy the requirements
+///   to be `FromZeroes` (they do not actually have to be `FromZeroes`).
 /// - If the type is an enum, it must be C-like (meaning that all variants have
-///   no fields) and it must have a variant with a discriminant of `0` (see [the
-///   reference] for a description of how discriminant values are chosen)
+///   no fields) and it must have a variant with a discriminant of `0`. See [the
+///   reference] for a description of how discriminant values are chosen.
+/// - The type must not contain any [`UnsafeCell`]s (this is required in order
+///   for it to be sound to construct a `&[u8]` and a `&T` to the same region of
+///   memory). The type may contain references or pointers to `UnsafeCell`s so
+///   long as those values can themselves be initialized from zeroes
+///   (`FromZeroes` is not currently implemented for, e.g.,
+///   `Option<&UnsafeCell<_>>`, but it could be one day).
 ///
 /// [the reference]: https://doc.rust-lang.org/reference/items/enumerations.html#custom-discriminant-values-for-fieldless-enumerations
+/// [`UnsafeCell`]: core::cell::UnsafeCell
 ///
 /// # Rationale
 ///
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -546,27 +581,43 @@ pub unsafe trait FromBytes: FromZeroes {
...
 ///     or `repr(packed)`).
-///   - All of its fields must be `AsBytes`
+///   - All of its fields must satisfy the requirements to be `AsBytes` (they do
+///     not actually have to be `AsBytes`).
 ///   - Its layout must have no padding. This is always true for
 ///     `repr(transparent)` and `repr(packed)`. For `repr(C)`, see the layout
 ///     algorithm described in the [Rust Reference].
 /// - If the type is an enum:
-///   - It must be a C-like enum (meaning that all variants have no fields)
+///   - It must be a C-like enum (meaning that all variants have no fields).
```
kupiakos, line 611: This is also more strict than necessary. It needs to have no padding, which can be guaranteed with some enums with fields, such as:

```rust
#[repr(u32)]
enum Foo {
  A(i32),
  B([u8; 4]),
}
```

If the type is an enum with fields, it must either:
  - Have an explicit representation `#[repr(Int)]`. Each field must be the same size. No field data can have alignment greater than `Int`. All field data must be `AsBytes`.
  - Have a single variant with data that is `AsBytes`. All variants without data are represented as an illegal representation (a niche) of the data type, such that the size of the enum is the size of the data in the single variant. This covers `Option<NonZeroUsize>`. (this could be phrased more cleanly)

I believe both of these could be `derive`d.

src/lib.rs
```diff
@@ -546,27 +581,43 @@ pub unsafe trait FromBytes: FromZeroes {
...
 ///   - It must have a defined representation (`repr`s `C`, `u8`, `u16`, `u32`,
 ///     `u64`, `usize`, `i8`, `i16`, `i32`, `i64`, or `isize`).
+/// - The type must not contain any [`UnsafeCell`]s (this is required in order
+///   for it to be sound to construct a `&[u8]` and a `&T` to the same region of
+///   memory).
```
kupiakos, line 616: The type must not contain any `MaybeUninit` fields either.

src/lib.rs
```diff
@@ -546,27 +581,43 @@ pub unsafe trait FromBytes: FromZeroes {
...
+///   size_of_val(t)` at the same address as `t`, and it is sound for both `b`
+///   and `t` to be live at the same time.
+///
+/// If a type is marked as `AsBytes` which violates this contract, it may cause
 /// undefined behavior.
 ///
-/// If a type has the following properties, then it is safe to implement
-/// `AsBytes` for that type
+/// If a type has the following properties, then it is sound to implement
+/// `AsBytes` for that type:
```
kupiakos, line 601: So, there are multiple kinds of safety requirements listed here.

First it states what operations that unsafe code are allowed to perform. Then, it states that certain types are sound to implement for.

However, there's a missing step here. It should first say what _core properties_ a type must uphold in order for those operations to be sound. Those core safety requirements are:

- No bytes that can possibly be uninitialized:
  - No padding
  - No `MaybeUninit`
- No `UnsafeCell`

These are the requirements for every data type, struct, enum, union, etc.

Then, it can expand upon what types naturally fit these requirements.
The bits about "if the type is a struct" are essentially _requirements for the derive_ to ensure that the above safety requirements are met, and are stricter than the actual safety requirements. Some of them are also requirements to guarantee _stability_ of sound implementation, such as requiring `#[repr(C)]` on structs.

This is less applicable for `FromBytes` since the core safety requirements are essentially identical to the guarantees unsafe code can assume. Might be worth throwing in a "it cannot have any invalid representations"?

src/lib.rs
```diff
@@ -199,19 +199,36 @@ mod zerocopy {
...
-/// - If the type is a struct, all of its fields must implement `FromZeroes`
+/// - If the type is a struct, all of its fields must satisfy the requirements
+///   to be `FromZeroes` (they do not actually have to be `FromZeroes`).
 /// - If the type is an enum, it must be C-like (meaning that all variants have
 ///   no fields) and it must have a variant with a discriminant of `0` (see [the
-///   reference] for a description of how discriminant values are chosen)
+///   reference] for a description of how discriminant values are chosen).
+/// - The type must not contain any [`UnsafeCell`]s (this is required in order
+///   for it to be sound to construct a `&[u8]` and a `&T` to the same region of
+///   memory).
```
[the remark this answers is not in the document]
joshlf: I wasn't sure how to phrase this since we don't implement `FromZeroes` for `Option<&...>` today. How does this sound?

> The type must not contain any `UnsafeCell`s (this is required in order for it to be sound to construct a `&[u8]` and a `&T` to the same region of memory). The type may contain references or pointers to `UnsafeCell`s so long as those values can themselves be initialized from zeroes (`FromZeroes` is not currently implemented for, e.g., `Option<&UnsafeCell<_>>`, but it could be one day).

src/lib.rs
joshlf: True, but that's covered by the other parts of this definition (no padding).

kupiakos: No, it is not. An explicit `MaybeUninit` field is disallowed.

src/lib.rs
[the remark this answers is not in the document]
kupiakos: That sounds good, other than the generally applicable technical writing advice to avoid parentheticals longer than a word or two.

src/lib.rs
```diff
@@ -199,19 +199,36 @@ mod zerocopy {
 ///
 /// # Safety
 ///
-/// If `T: FromZeroes`, then unsafe code may assume that it is sound to treat
-/// any initialized sequence of zero bytes of length `size_of::<T>()` as a `T`.
+/// *This section describes what is required in order for `T: FromZeroes`, and
+/// what unsafe code may assume of such types. `#[derive(FromZeroes)]` only
+/// permits types which satisfy these requirements. If you don't plan on
+/// implementing `FromZeroes` manually, and you don't plan on writing unsafe
```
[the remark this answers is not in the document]
joshlf: I still want to strongly discourage users from implementing these traits themselves, so IMO it'd be good to still have a speed bump. Maybe we could rename it to something like `i_promise_i_need_to_implement_this_manually_and_cant_use_the_custom_derive`?

src/lib.rs
joshlf: IIUC, `MaybeUninit` doesn't satisfy the `AsBytes` requirement because its `uninit` field has padding:

```rust
pub union MaybeUninit<T> {
    uninit: (),
    value: ManuallyDrop<T>,
}
```

src/lib.rs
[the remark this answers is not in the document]
kupiakos: `i_promise_i_need_to_implement_this_manually_and_cant_use_the_custom_derive` is good

src/lib.rs
kupiakos: That's an implementation detail of `MaybeUninit` that users should not be expected to know.

src/lib.rs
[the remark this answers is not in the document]
joshlf: Opened https://github.com/google/zerocopy/issues/287 to track this.

src/lib.rs
```diff
@@ -199,19 +199,36 @@ mod zerocopy {
...
-/// If a type has the following properties, then it is safe to implement
+/// If a type has the following properties, then it is sound to implement
 /// `FromZeroes` for that type:
-/// - If the type is a struct, all of its fields must implement `FromZeroes`
+/// - If the type is a struct, all of its fields must satisfy the requirements
+///   to be `FromZeroes` (they do not actually have to be `FromZeroes`).
 /// - If the type is an enum, it must be C-like (meaning that all variants have
 ///   no fields) and it must have a variant with a discriminant of `0` (see [the
-///   reference] for a description of how discriminant values are chosen)
+///   reference] for a description of how discriminant values are chosen).
```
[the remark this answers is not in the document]
joshlf: Filed https://github.com/google/zerocopy/issues/288 to track this.

src/lib.rs
joshlf: Filed https://github.com/google/zerocopy/issues/288 to track this.

src/lib.rs
joshlf: Filed https://github.com/google/zerocopy/issues/288 to track this.

src/lib.rs
joshlf: Filed https://github.com/google/zerocopy/issues/288 to track this.

src/lib.rs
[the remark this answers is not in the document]
joshlf: Gonna merge this for now; feel free to follow up with more discussion here and I'd be happy to update the docs in a follow-up PR.

PR #289, opened 2023-08-25
Cargo check before get the metadata

### LINKED ISSUE #225: CI step "Set toolchain version" can fail without stopping CI job
In [this CI job](https://github.com/google/zerocopy/actions/runs/5754097055/job/15598706930?pr=222), it appears that a network error causes issues with the "Set toolchain version" step, but the GitHub runner does not fail at that point. Instead, later steps are executed with bad data (namely, the `ZC_TOOLCHAIN` environment variable is set to an empty string), causing confusing errors.

I suspect that what's happening is that the shell script executed for that step has a bug that results in errors inside of `$(...)` to not respect the `set -e` at the top of the script.

zoo868e:
Fixes #225
The reason `set -e` failed to interrupt the workflow is due to the output redirection to `jq`. Since `jq` returns a value of 0, indicating success, the script continues to execute. You can try running these scripts locally to gain insight into the issue.

To illustrate:

Script without interruption:
```sh
set -e

function pkg-meta {
  cargo metadata --format-version 1 | jq -r ".packages[] | select(.name\
  == \"zerocopy\").$1"
}

ZC_TOOLCHAIN=\"$(pkg-meta 'metadata.ci.\"pinned-nightly\"')\"

echo \"Discovered that the nightly toolchain is $ZC_TOOLCHAIN\"
echo \"ZC_TOOLCHAIN=$ZC_TOOLCHAIN\"
```

Script with interruption:
```sh
set -e

function pkg-meta {
  cargo metadata --format-version 1"
}

ZC_TOOLCHAIN=\"$(pkg-meta 'metadata.ci.\"pinned-nightly\"')\"

echo \"Discovered that the nightly toolchain is $ZC_TOOLCHAIN\"
echo \"ZC_TOOLCHAIN=$ZC_TOOLCHAIN\"
```

Additionally, you can display the return value after the command execution. This output will confirm the successful execution of the command.

```sh
set -e

function pkg-meta {
  cargo metadata --format-version 1 | jq -r \".packages[] | select(\
  .name == \"zerocopy\").$1\"
  echo
  echo \$?
}

ZC_TOOLCHAIN=\"$(pkg-meta 'metadata.ci.\"pinned-nightly\"')\"

echo \"Discovered that the nightly toolchain is $ZC_TOOLCHAIN\"
echo \"ZC_TOOLCHAIN=$ZC_TOOLCHAIN\"
```

Therefore, I've incorporated `cargo check` to validate the package retrieval. Subsequent to this check, there's no necessity to fetch the package again using `cargo metadata`. This modification should rectify the point of failure in the Git action.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,7 +111,7 @@ jobs:
 
     - name: Configure environment variables
       run: |
-        set -e
+        set -eo pipefail
         if [[ '${{ matrix.toolchain }}' == 'nightly' ]]; then
           RUSTFLAGS="$RUSTFLAGS $ZC_NIGHTLY_RUSTFLAGS"
           MIRIFLAGS="$MIRIFLAGS $ZC_NIGHTLY_MIRIFLAGS"
@@ -214,7 +214,7 @@ jobs:
       - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
       - name: Check Rust formatting
         run: |
-          set -e
+          set -eo pipefail
           cargo fmt --check -p zerocopy
           cargo fmt --check -p zerocopy-derive
           shopt -s globstar
@@ -231,7 +231,7 @@ jobs:
         uses: Swatinem/rust-cache@e207df5d269b42b69c8bc5101da26f7d31feddb4 # v2.6.2
       - name: Check README.md
         run: |
-          set -e
+          set -eo pipefail
           cargo install cargo-readme --version 3.2.0
           diff <(./generate-readme.sh) README.md
           exit $?
@@ -245,7 +245,7 @@ jobs:
       # files are the same.
       - name: Check MSRVs match
         run: |
-          set -e
+          set -eo pipefail
 
           # Usage: msrv <crate-name>
           function msrv {
@@ -274,7 +274,7 @@ jobs:
       # `INTERNAL.md` for an explanation of why we do this.
       - name: Check crate versions match
         run: |
-          set -e
+          set -eo pipefail
 
           # Usage: version <crate-name>
           function version {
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -86,6 +86,7 @@ jobs:
       run: |
         set -e
 
+        cargo check
```
joshlf, line 89: Can you change this to `cargo metadata --format-version 1 > /dev/null`? `cargo check` may fail for other reasons (ie, there's an issue with the code uploaded in a PR).

joshlf (commented):
Thanks so much for this, @zoo868e!

joshlf:
[This build failure](https://github.com/google/zerocopy/actions/runs/5980678106/job/16238736921?pr=289) illustrates that your technique works perfectly!

.github/workflows/ci.yml
zoo868e: Hello @joshlf,

I've assessed that both `cargo check` and `cargo metadata` aren't appropriate solutions for addressing this issue. They don't align with the workflow of this job. The root problem lies in the job's failure to pause when a command encounters an error, and then pass that outcome to the subsequent command in the pipeline.

Given this, I am proposing a solution that revolves around adjusting the `set` command. To be more precise, by introducing the `-o pipefail` option, we can make the command proficient at recognizing failures. This, in turn, empowers the `-e` flag to conclude the process. Consequently, the refined `set` command would appear as follows: `set -eo pipefail`. For further elaboration on the `set` command, additional details can be found using `set --help`.

Additionally, I've noticed that other jobs also use `set -e` and command piping, which might have similar issues. Would it be acceptable to apply these changes to all those jobs?

Your guidance on this matter would be greatly appreciated.

joshlf: Yeah, setting `set -eo pipefail` everywhere would be great!

joshlf (approved):

PR #291, opened 2023-08-28
Support the associated proc macro pattern

jswrenn:
The "associated proc macro pattern" is a trick for ensuring that `zerocopy` and `zerocopy-derive` have equal verions, even when the optional `derive` feature of `zerocopy` isn't used.

The `derive` feature of `zerocopy` provides a convenient way for end-users to depend on `zerocopy-derive` that ensures that `zerocopy` and `zerocopy-derive` have exactly equal versions. Unfortunately, using this feature creates a dependency between these two crates that prevents them from being built in parallel. A compile-time-conscious user of `zerocopy` can avoid this by depending on both of these crates explicitly, but this places the onus on them to ensure that their versions are exactly equal.

The "associated proc macro pattern" is a trick for ensuring that `zerocopy` and `zerocopy-derive` have equal verions, even when the optional `derive` feature of `zerocopy` isn't used. With it, an end-user of zerocopy can write:

    [dependencies]
    zerocopy = "MAJOR.MINOR"
    zerocopy-derive = "MAJOR.MINOR"

...and cargo will ensure that the actual versions that are used are exactly equal.

For more information, see: https://github.com/matklad/macro-dep-test


joshlf (commented):
Looks great! Can we leave a suggestion in the crate doc comment to suggest that this is an option? We'll also want to maybe open an issue to track making this a more strongly-worded suggestion once this technique has baked in the ecosystem and folks agree it's the way to go.

README.md
```diff
@@ -30,35 +30,51 @@ byte sequences with little to no runtime overhead.
...
+  order conversions. This can be useful in handling file formats, network
+  packet layouts, etc which don't provide alignment guarantees and which may
+  use a byte order different from that of the execution platform.
+
+- **`derive`**
+  Provides derives for the core marker traits via the `zerocopy-derive`
+  crate. These derives are re-exported from `zerocopy`, so it is not
+  necessary to depend on `zerocopy-derive` directly.
+
+  However, you may experience better compile times if you explicitly depend
```
joshlf, line 54: Maybe "...if you don't use the `derive` feature and instead explicitly depend..." to make it clear that this is instead of the feature, not in addition to it?

Also, maybe s/explicitly/directly/?

joshlf (changes requested):
Does the approach of putting a newline after the feature name play nicely with markdown (for the readme) and `cargo doc` (for lib.rs)?

joshlf:
Oh and one other thing: could you also bump the version numbers to 0.7.1 in this PR? I'll release that once this merges so folks can take advantage of it.

jswrenn:
> Does the approach of putting a newline after the feature name play nicely with markdown (for the readme) and cargo doc (for lib.rs)?

Mostly. Unfortunately, `cargo readme` appears to be stripping out the trailing triple-space from those lines, so Github's rendering doesn't have a newline after each feature heading, but the rustdoc does. I don't think this is an issue. Github is also a little more generous with spacing between list items, so the extra newline isn't as important for visually delineating cargo features. Both renderings look good to my eye.

> Oh and one other thing: could you also bump the version numbers to 0.7.1 in this PR? I'll release that once this merges so folks can take advantage of it.

Done.

README.md
jswrenn: Good point. I opted for slightly more concise wording, but the effect should be the same.

joshlf: Doesn't look like these changes actually got pushed?

jswrenn: Sigh. I totally typed them in README.md instead of lib.rs, and then gleefully overwrote README.md with generate-readme.sh.

jswrenn: Pushed!

joshlf (approved):

Issue #196, opened 2023-07-25
Support field projection inside of `Unaligned` (and other wrappers?)


joshlf:
TODO: Fill in details here.

SkiFire13:
> ```rust
>         unsafe {
>             $crate::$project_fn(
>                 $c.$borrow_method(),
>                 |outer| outer as $ptr_ty,
>                 |inner| ::core::ptr::$addr_of!((*inner) $($f)*),
>                 |field| field as $ptr_ty,
>             )
>         }
> ```

Note that this allowes `c` and `$($f)*` to do `unsafe` operations without actually using the `unsafe` keyword (because it's hidden in your macro)

joshlf:
Good catch, thanks for that, @SkiFire13 ! Took a quick stab at fixing it and ran into inference issues (specifically when assigning `|inner| ...` to a variable before the `unsafe` block), but I've added this to the design doc; we'll obviously need to fix that before the design is finalized.

SkiFire13:
> Took a quick stab at fixing it and ran into inference issues (specifically when assigning `|inner| ...` to a variable before the `unsafe` block)

A couple of ideas for this:

- you could use those expressions somewhere else, where they are never executed, but outside of an `unsafe` block so the compiler still checks them. For example inside an `if false { ...; unreachable!() }` block.

- you could try making an identity function that just specifies the bounds on the closure, helping type inference. Something like:

  ```
  fn help_outer_to_inner_inference<P, OuterToInner>(f: OuterToInner) -> OuterToInner
  where
      P: Projectable + ?Sized,
      OuterToInner: FnOnce(*const P) -> *const P::Inner
  {
      f
  }
  ```

  and likewise for the other closures.

- another approach could be moving the required use of `unsafe` somewhere else, for example by making the function safe, but having it take something like an `UnsafeToken`, which can only be created by using `unsafe`. This way it's as if the function is `unsafe`, but you're not required to wrap the whole call expression, including the parameters, in it.

joshlf:
Found unsoundness in the way we're using `addr_of!`: https://users.rust-lang.org/t/how-to-use-addr-of-to-create-an-unaligned-pointer/98524

joshlf:
> * another approach could be moving the required use of `unsafe` somewhere else, for example by making the function safe, but having it take something like an `UnsafeToken`, which can only be created by using `unsafe`. This way it's as if the function is `unsafe`, but you're not required to wrap the whole call expression, including the parameters, in it.

[...]

PR #292, opened 2023-08-29
[project] Initial commit

joshlf:
Makes progress on #196


joshlf:
This is basically ready, except there's one outstanding [safety doc comment](https://github.com/google/zerocopy/pull/292/files#diff-f36a5c348f86854b0177520d601d75ed1dc2ac736aebe2e6b928389485b75021R162-R168) that I can't figure out how to write.

I know the safety reasoning here is pretty convoluted, so I'd love folks to double-check my work and/or suggest ways to simplify or strengthen the arguments. Referring to #196 may help in understanding the design.

cc @kupiakos @jswrenn @djkoloski

src/project.rs
```diff
@@ -0,0 +1,659 @@
...
+///     type Inner = T;
+/// }
+/// ```
+///
+/// # Safety
+///
+/// `Self` must be a `repr(transparent)`, `repr(packed)`, or `repr(C)` struct
+/// with a single non-zero-sized field of type `Self::Inner`. `W` must be a
+/// `repr(transparent)`, `repr(packed)`, or `repr(C)` struct with a single
+/// non-zero field of type `F`.
```
joshlf, line 112: Oops, this is just wrong. E.g., it doesn't apply to `MaybeValid<T: AsMaybeUninit>` from #279, whose inner field is not `T`, but `T::MaybeUninit`.

src/project.rs
```diff
@@ -0,0 +1,659 @@
...
+        let token = unsafe { $crate::project::unsafe_token::UnsafeToken::new() };
+        use ::core::borrow::$borrow_trait as _;
+        $crate::project::$project_fn(
+            token,
+            // We call `borrow` or `borrow_mut` so that this macro works
+            // regardless of whether `$c` is owned or borrowed.
+            //
+            // TODO(#196): Is there any way to make sure this calls
+            // `Borrow::borrow` or `BorrowMut::borrow_mut` even if the type has
+            // an inherent method of the same name?
```
joshlf, line 370: @kupiakos [suggested](https://github.com/google/zerocopy/pull/279#discussion_r1309126974) that the following works:

```
<_ as $borrow_trait>::$borrow_method($c)
```

joshlf: @kupiakos I wasn't actually able to get this to work on this PR. Did you test with this PR, or just a minimized version?

src/project.rs
joshlf: Okay, I think this should be [fixed now](https://github.com/google/zerocopy/compare/dcc4affa21f6a30a3fa98e3724ea1c016a6209ce..35847cbd750d79f4401c11bd039a58a7d939d022)?

PR #294, opened 2023-08-29
Implement FromZeroes for thin raw pointers

joshlf:
Makes progress on #170


src/lib.rs
```diff
@@ -945,6 +945,40 @@ safety_comment! {
     unsafe_impl!(T: ?Sized + Unaligned => Unaligned for ManuallyDrop<T>);
     assert_unaligned!(ManuallyDrop<()>, ManuallyDrop<u8>);
 }
+safety_comment! {
+    /// SAFETY:
+    /// The all-zeroes const and mut raw pointers are valid, and it is sound to
+    /// materialize them from nothing. The existence of `ptr::null` [1] and
```
jswrenn, line 951: Is it clearly documented anywhere that `0` is a valid value for the null pointer? Infamously, the C standard does not require that null pointers have an all-zero bit-pattern, only that they compare as equal to `0`.

joshlf:
Current plan is to wait until [this change](https://github.com/rust-lang/reference/pull/1392#issuecomment-1696768191) (or something similar) lands so the soundness of this PR is clearly backed by the Rust reference.

PR #301, opened 2023-08-31
Retrieve the toolchain version from Cargo.toml

### LINKED ISSUE #295: CI step "Set toolchain version" is flaky due to network timeouts
Update: #296 doesn't seem to have done the trick. A few more possible approaches:
- Retry the entire job in order to give more time for the network issue to resolve itself
- Maybe part of the issue is that the connection is bad, and the retry is reusing the bad connection? Maybe we could implement our own manual retry logic?
- Maybe we could somehow cache the needed crates so the jobs usually don't have to re-download?

## Old text

Over the past few weeks, most full CI runs (currently 149 jobs) have had at least one failure due to Cargo failing to download a package even after 3 retries (the default value for the `net.retry` configuration option). Hopefully increasing the number of retries will make this less common.

Note that this was previously encountered in #225. #289 made the errors more obvious, but didn't make them any less common.

zoo868e:
Fixes #295

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,18 +97,19 @@ jobs:
         set -eo pipefail
 
         function pkg-meta {
-          cargo metadata --format-version 1 | jq -r ".packages[] | select(.name == \"zerocopy\").$1"
+          grep $1 Cargo.toml | sed 's/.*= "\(.*\)"/\1/g'
+
         }
 
         case ${{ matrix.toolchain }} in
           msrv)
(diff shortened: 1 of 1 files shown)


joshlf:
Unfortunately I don't think this is going to help; we use `cargo metadata` in other CI jobs, so this'll just kick the can down the road and presumably those jobs will start failing instead. I'd also like to keep using `cargo metadata` since it (combined with `jq`) is more robust than grep, especially in the later jobs where we're doing more complex queries.

I'm gonna close this for now, but feel free to re-open if you think there's something I've missed.

PR #302, opened 2023-08-31
Cache the dependencies

### LINKED ISSUE #295: CI step "Set toolchain version" is flaky due to network timeouts

zoo868e:
Fixes #295

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,35 +78,6 @@ jobs:
     steps:
     - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
 
-    - name: Check cache
-      uses: actions/cache@v3
-      with:
-        path: |
-          ~/.cargo/bin/
-          ~/.cargo/registry/index/
-          ~/.cargo/registry/cache/
-          ~/.cargo/git/db/
-          target/
-        key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
-    - name: Get the packages
-      if: steps.cache-check.outputs.cache-hit != 'true'
-      run: |
-        cargo metadata
-    - name: Save cache
-      if: steps.cache-check.outputs.cache-hit != 'true'
-      uses: actions/cache@v3
-      with:
-        path: |
-          ~/.cargo/bin/
-          ~/.cargo/registry/index/
-          ~/.cargo/registry/cache/
-          ~/.cargo/git/db/
-          target/
-        key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
-
-    - name: Clean build artifacts
-      run: |
-        cargo clean
     # We use toolchain descriptors ("msrv", "stable", and "nightly") in the
     # matrix. This step converts the current descriptor to a particular
     # toolchain version by looking up the corresponding key in `Cargo.toml`. It
@@ -380,3 +351,24 @@ jobs:
               | tee -a $GITHUB_STEP_SUMMARY >&2
             exit 1
           fi
+  generate_cache:
+    runs-on: ubuntu-latest
+    name: Generate rust package cache
+    steps:
+      - name: Check cache
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
+      - name: Get the packages
+        if: steps.cache-check.outputs.cache-hit != 'true'
+        run: |
+          cargo metadata
+      - name: Clean build artifacts
+        run: |
+          cargo clean
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,7 +78,8 @@ jobs:
     steps:
     - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
 
-    - uses: actions/cache@v3
+    - name: Check cache
+      uses: actions/cache@v3
       with:
         path: |
           ~/.cargo/bin/
@@ -87,6 +88,22 @@ jobs:
           ~/.cargo/git/db/
           target/
         key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
+    - name: Get the packages
+      if: steps.cache-check.outputs.cache-hit != 'true'
+      run: |
+        cargo metadata
+    - name: Save cache
+      if: steps.cache-check.outputs.cache-hit != 'true'
+      uses: actions/cache@v3
+      with:
+        path: |
+          ~/.cargo/bin/
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+          ~/.cargo/git/db/
+          target/
+        key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
+
     - name: Clean build artifacts
       run: |
         cargo clean
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,6 +78,18 @@ jobs:
     steps:
     - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
 
+    - uses: actions/cache@v3
+      with:
+        path: |
+          ~/.cargo/bin/
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+          ~/.cargo/git/db/
+          target/
+        key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
+    - name: Clean build artifacts
+      run: |
+        cargo clean
     # We use toolchain descriptors ("msrv", "stable", and "nightly") in the
     # matrix. This step converts the current descriptor to a particular
     # toolchain version by looking up the corresponding key in `Cargo.toml`. It
@@ -351,12 +363,15 @@ jobs:
               | tee -a $GITHUB_STEP_SUMMARY >&2
             exit 1
           fi
+
+
+
   generate_cache:
     runs-on: ubuntu-latest
-    name: Generate rust package cache
+    name: Generate cache
     steps:
-      - name: Check cache
-        uses: actions/cache@v3
+      - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
+      - uses: actions/cache@v3
         with:
           path: |
             ~/.cargo/bin/
@@ -364,11 +379,7 @@ jobs:
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
             target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
-      - name: Get the packages
-        if: steps.cache-check.outputs.cache-hit != 'true'
-        run: |
-          cargo metadata
+          key: ${{ runner.os }}-cargo-test-CI-${{ hashFiles('**/Cargo.toml') }}
       - name: Clean build artifacts
         run: |
           cargo clean
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -363,8 +363,6 @@ jobs:
             exit 1
           fi
 
-
-
   generate_cache:
     runs-on: ubuntu-latest
     name: Generate cache
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,7 @@ env:
 jobs:
   build_test:
     runs-on: ubuntu-latest
+    needs: generate_cache
 
     strategy:
       # By default, this is set to `true`, which means that a single CI job
@@ -81,12 +82,10 @@ jobs:
     - uses: actions/cache@v3
       with:
         path: |
-          ~/.cargo/bin/
           ~/.cargo/registry/index/
           ~/.cargo/registry/cache/
-          ~/.cargo/git/db/
-          target/
-        key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
+        key: ${{ runner.os }}-cargo-test-CI2-${{ hashFiles('**/Cargo.toml') }}
+
     - name: Clean build artifacts
       run: |
         cargo clean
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -78,6 +78,18 @@ jobs:
...
+        path: |
+          ~/.cargo/bin/
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+          ~/.cargo/git/db/
+          target/
+        key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
+    - name: Clean build artifacts
+      run: |
+        cargo clean
```
joshlf, line 92: Can you add a trailing newline here (like we do between other steps)?

.github/workflows/ci.yml
```diff
@@ -78,6 +78,18 @@ jobs:
...
     - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
 
+    - uses: actions/cache@v3
+      with:
+        path: |
+          ~/.cargo/bin/
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+          ~/.cargo/git/db/
+          target/
```
joshlf, line 88: Is it necessary to cache these in order for `cargo metadata` to work?

.github/workflows/ci.yml
zoo868e: Sure, I will add it.

.github/workflows/ci.yml
zoo868e: I'm utilizing the provided [example](https://github.com/actions/cache/blob/main/examples.md#rust---cargo) directly. I'm uncertain about whether the command `cargo metadata` will be used or not. As far as I understand, this command checks if the package has not been installed yet. If the package has been installed, it will be stored in the `~/.cargo/registry` directory. Given this, it seems plausible that we could potentially cache only the `~/.cargo/registry/index` and the `~/.cargo/registry/cache` directory.

zoo868e: I've tested this [here](https://github.com/zoo868e/zerocopy/actions/runs/6042393394/job/16397422883), and it appears that only caching the `~/.cargo/registry/index` is not sufficient for the `cargo metadata` command.

.github/workflows/ci.yml
```diff
@@ -78,6 +79,16 @@ jobs:
     steps:
     - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
 
+    - uses: actions/cache@v3
+      with:
+        path: |
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+        key: ${{ runner.os }}-cargo-test-CI2-${{ hashFiles('**/Cargo.toml') }}
```
zoo868e, line 87: I will remove the `-test-CI2-` when the PR going to merge.

.github/workflows/ci.yml
```diff
@@ -351,3 +362,24 @@ jobs:
...
+    name: Generate cache
+    steps:
+      - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
+
+      - uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+          key: ${{ runner.os }}-cargo-test-CI2-${{ hashFiles('**/Cargo.toml') }}
```
zoo868e, line 377: I will remove the `-test-CI2-` when the PR going to merge.

.github/workflows/ci.yml
```diff
@@ -351,3 +362,24 @@ jobs:
               | tee -a $GITHUB_STEP_SUMMARY >&2
             exit 1
           fi
+
+  generate_cache:
```
joshlf, line 366: Why do we need this separate job? Will the caching not happen automatically the first time a job/step is run that uses `cargo metadata` (ie, the "Set toolchain version" step)?

.github/workflows/ci.yml
joshlf: Looks like the latest version of the PR only includes `~/.cargo/registry/index/` and `~/.cargo/registry/cache`; I assume your testing revealed that just caching these two is sufficient?

.github/workflows/ci.yml
zoo868e: Because Git actions start the jobs simultaneously and the cache is written back after the build and test processes, the jobs can't utilize the cache until it's been written back, potentially leading to network timeouts. To address this, I've modified the job `build and test` while also waiting for the completion of the `cargo metadata` operation and the successful writing of the cache to GitHub. This adjustment generates the cache before the jobs start, allowing them to retrieve it and thus avoiding any potential network timeout issues.

.github/workflows/ci.yml
joshlf: If we're just caching global Cargo files, is including the hashes of `Cargo.toml` files in the key unnecessary? Presumably the final version of this key could just be `${{ runner.os }}-cargo`?

.github/workflows/ci.yml
joshlf: Ah I see. I was assuming that the cache would be kept around between PR job runs, so while the first run of jobs might still have problems, subsequent runs would be fine. It looks like that [might be how it works](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows), but now that I'm thinking about it more, it strikes me as cleaner to have a separate "generate cache" step even if it's technically unnecessary.

.github/workflows/ci.yml
```diff
@@ -351,3 +362,24 @@ jobs:
...
+            ~/.cargo/registry/cache/
+          key: ${{ runner.os }}-cargo-test-CI2-${{ hashFiles('**/Cargo.toml') }}
+
+      - name: Cargo metadata
+        run: |
+          cargo metadata
+
+      - name: Clean build artifacts
+        run: |
+          cargo clean
```
joshlf, line 385: Might be a bit overkill to have these in two separate steps. Maybe just:

```yaml
      - name: Populate cache
        run: |
          cargo metadata >/dev/null
          cargo clean
```

.github/workflows/ci.yml
```diff
@@ -351,3 +362,24 @@ jobs:
...
+      - uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+          key: ${{ runner.os }}-cargo-test-CI2-${{ hashFiles('**/Cargo.toml') }}
+
+      - name: Cargo metadata
+        run: |
+          cargo metadata
```
joshlf, line 381: Can you redirect the output of `cargo metadata` to `/dev/null`?

.github/workflows/ci.yml
joshlf, line 385: Since we don't cache `target` anymore, is running `cargo clean` necessary?

.github/workflows/ci.yml
joshlf: (Commenting here to turn this into a review comment)

.github/workflows/ci.yml
joshlf: (Commenting here so this turns into a review comment and we make sure to do it before I approve the PR)

.github/workflows/ci.yml
```diff
@@ -31,6 +31,7 @@ env:
 jobs:
   build_test:
     runs-on: ubuntu-latest
+    needs: generate_cache
```
joshlf, line 34: Can you add a comment here something to this effect?

> Generate and populate the global Cargo registry and cache first. Each job in the matrix runs in parallel, so without populating the cache first, most jobs would duplicate the work of downloading crates from the internet. Populating the cache first ensures that this work only happens once.

.github/workflows/ci.yml
joshlf: If we change the key below, make sure to update it here too:

> If we're just caching global Cargo files, is including the hashes of `Cargo.toml` files in the key unnecessary? Presumably the final version of this key could just be `${{ runner.os }}-cargo`?

.github/workflows/ci.yml
joshlf: Same comment as below: Is running `cargo clean` necessary here?

joshlf (changes requested):
Okay, I think this is in a good state other than a few small cleanup items. Thanks so much for your work on this, @zoo868e! Very excited to have more reliable jobs - they've been failing a _lot_ recently and causing a lot of headache.

.github/workflows/ci.yml
zoo868e: I agree with that 😄

.github/workflows/ci.yml
zoo868e: Sure

.github/workflows/ci.yml
zoo868e: It's not necessary now!

.github/workflows/ci.yml
zoo868e: Sure

.github/workflows/ci.yml
zoo868e: Yes, I tested it on my forked repository, and it only requires these two directories to prevent downloading the crates every time.

joshlf: Sounds good!

.github/workflows/ci.yml
zoo868e: I believe it is necessary because we need to regenerate the cache whenever the `Cargo.toml` has been modified. Here's an example of the final version of this key: `Linux-cargo-a7354d5361cd9a3c83f18de0f397f087d1b2569d3df9d9b4f7489b77d7559185`.

.github/workflows/ci.yml
zoo868e: Sure

.github/workflows/ci.yml
joshlf: You mean because a modification to `Cargo.toml` could introduce new dependencies?

.github/workflows/ci.yml
zoo868e: I'd removing all instances of `cargo clean` as it is currently unnecessary.

.github/workflows/ci.yml
zoo868e: Indeed, when we modify the `Cargo.toml` file, it typically involves incorporating new crates or adjusting the version dependencies of existing ones. In such cases, it is advisable to regenerate the cache.

joshlf: Gotcha. In that case, could you add the following? I realized that while this job probably downloads all of our dependencies in practice (`cargo manifest`'s dependencies are likely a superset of ours), that's not guaranteed to always hold.

```shell
# Ensure all crate dependencies are downloaded. There is no support
# in Cargo for downloading dependencies without also building the
# current crate [1], so we check all packages with --tests as a workaround.
# We may get errors other than network errors (ie, build errors), so we
# discard stderr. This entire job is best-effort since it's just a performance
# optimization, so discarding stderr is fine.
#
# [1] https://stackoverflow.com/a/42139535/836390
cargo check --workspace --tests >/dev/null 2>/dev/null
```

.github/workflows/ci.yml
joshlf, line 381: Can you add `set -eo pipefail` at the top of this script?

joshlf (changes requested):

.github/workflows/ci.yml
zoo868e: Certainly, should it be inserted before the `cargo metadata` command within the `generate cache` section?

.github/workflows/ci.yml
zoo868e: Sure

.github/workflows/ci.yml
joshlf: Sure, and actually, while we're at it, I think that we can just adjust that comment to document everything:

```shell
# Ensure all dependencies are downloaded - both for our crates and for tools
# we use in CI. We don't care about these tools succeeding for two reasons:
# First, this entire job is best-effort since it's just a performance optimization.
# Second, there may be failures due to issues other than failing to download
# dependencies (e.g., `cargo metadata` called with a malformed `Cargo.toml`,
# build failure in our own crate or in dependencies, etc). For those reasons,
# we discard stderr and ignore status codes.
#
# For downloading our crates' dependencies in particular, note that there is
# no support for doing this directly [1], so we just check all crates using --tests.
#
# [1]  https://stackoverflow.com/a/42139535/836390
cargo check --workspace --tests >/dev/null 2>/dev/null || true
cargo manifest >/dev/null 2>/dev/null || true
```

Let's also not set `-eo pipefail`. That technically makes the first `|| true` redundant, but I'm 80% sure the second one is necessary (I assume that GitHub actions treat the status of the final command as the status of the entire job), so we may as well have both for symmetry.

.github/workflows/ci.yml
joshlf: (Nevermind - see my other comment)

.github/workflows/ci.yml
zoo868e: I see, I think we can use `&>` for redirection, as it will redirect all the output to `/dev/null`.

zoo868e: I'm not familiar with the `cargo manifest` command and couldn't locate its manual. Could you please assist me in comprehending it?

joshlf: Sorry, typo on my part - I meant `cargo metadata`. I've updated the original comment.

joshlf (changes requested):
Okay, just looks like we need to remove `-test-CI2-` from the cache keys and then this should be good to go!

joshlf (approved):

joshlf:
Awesome, thanks so much for all your work on this, @zoo868e! At least based on how the jobs ran on this PR, I expect this is going to almost entirely fix the issue (perhaps with the exception of cases where populating the cache itself fails, which should presumably be much less common than at least one of 149 jobs failing for the same reason).

Issue #29, opened 2022-10-07
Support `KnownLayout` trait and custom DSTs


joshlf:
# Overview

The `KnownLayout` trait encodes enough information about a type's layout to be able to synthesize pointers to values of that type at runtime. While we can already do this for sized types and slice types, `KnownLayout` generalizes this to "slice DSTs" - types with leading fixed-sized fields followed by a trailing slice. 

# Progress

- [x] Encode the layout of unsized types in a value that can be operated on in `const` code
  - https://github.com/google/zerocopy/pull/347
- [x] Implement a runtime check to validate pointer type casts and compute pointer metadata
  - https://github.com/google/zerocopy/pull/349
  - https://github.com/google/zerocopy/pull/376
  - https://github.com/google/zerocopy/pull/393
  - https://github.com/google/zerocopy/pull/402
- [x] Support deriving `KnownLayout` for sized types
  - https://github.com/google/zerocopy/pull/514
- [x] Support deriving `KnownLayout` for unsized types
  - [ ] Use built-in Rust APIs to query the size an alignment of an unsized type:
    - [ ] Be able to determine the alignment of an unsized type; either:
      - [ ] Remove the `T: Sized` bound from [`align_of<T>()`](https://doc.rust-lang.org/core/mem/fn.align_of.html)
      - [ ] Stabilize [`offset_of!`](https://doc.rust-lang.org/core/mem/macro.offset_of.html) and expand it to support unsized types
      - [ ] Stabilize [`align_of_val_raw`](https://doc.rust-lang.org/core/mem/fn.align_of_val_raw.html)
    - [ ] Be able to calculate the offset of the trailing field of a potentially-unsized type; either:
      - [ ] Stabilize [`offset_of!`](https://doc.rust-lang.org/core/mem/macro.offset_of.html) and expand it to support unsized types
      - [ ] Stabilize [`size_of_val_raw`](https://doc.rust-lang.org/core/mem/fn.size_of_val_raw.html)
  - [x] Alternatively, manually implement Rust's `repr(C)` layout algorithm
    - https://github.com/google/zerocopy/pull/633
    - https://github.com/google/zerocopy/pull/638
  - [x] Implement the derive
    - https://github.com/google/zerocopy/pull/643
- [x] Support `KnownLayout` in `Ref`, and use this to support `KnownLayout` in other places
[...]

PR #312, opened 2023-09-03
Add `MaybeUninit` type

joshlf:
The standard library's `MaybeUninit` type does not currently support wrapping unsized types. This commit introduces a polyfill with the same behavior as `MaybeUninit` which does support wrapping unsized types.

In this commit, the only supported types are sized types and slice types. Later (as part of #29), we will add the ability to derive the `AsMaybeUninit` trait, which will extend support to custom DSTs.

Makes progress on #29


joshlf:
This was accidentally closed when renaming a branch. It's reopened in https://github.com/google/zerocopy/pull/319.

Issue #5, opened 2022-09-19
Support `TryFromBytes` - conditional conversion analogous to `FromBytes`


joshlf:
*Co-authored with @jswrenn.*

## Overview

Add a `TryFromBytes` trait, which supports byte-to-type conversions for non-`FromBytes` types by performing runtime validation. Add a custom derive which generates this validation code automatically.

*Many thanks to @kupiakos and @djkoloski for providing invaluable feedback and input on this design.*

## Progress

- [x] Add `TryFromBytes` trait definition
  - https://github.com/google/zerocopy/pull/641
- [x] Implement `TryFromBytes` for existing `FromBytes` types
  - https://github.com/google/zerocopy/pull/644
- [x] Add `try_from_ref` method; impl for `bool`
  - https://github.com/google/zerocopy/pull/647
- [x] Implement derive for structs
  - https://github.com/google/zerocopy/pull/665
- [x] Implement for slices
  - https://github.com/google/zerocopy/pull/666
- [x] Implement for arrays
  - https://github.com/google/zerocopy/pull/799
- [x] Allow deriving on `repr(packed)` structs
  - https://github.com/google/zerocopy/pull/794
- [x] Allow deriving on unions
  - See for context: https://github.com/google/zerocopy/discussions/696
  - https://github.com/google/zerocopy/pull/800
- [x] Allow deriving on field-less enums with primitive reprs (`u8`, `i16`, etc)
  - https://github.com/google/zerocopy/pull/803
- [x] Allow deriving on field-less enums with `repr(C)` by treating the discriminant type as `[u8; size_of::<Self>()]`
  - https://github.com/google/zerocopy/pull/806
- [x] https://github.com/google/zerocopy/issues/873
- [x] Implement `TryFromBytes` for `fn()` and `extern "C" fn()` types
  - #904 
- [x] Implement `TryFromBytes` for `UnsafeCell<T>`
  - #905
- [x] Make `TryFromBytes` a super-trait of `FromZeros`
  - #962
- [x] Remove `#[doc(hidden)]` from all items which are intended to be public
  - https://github.com/google/zerocopy/pull/976
- [x] Add to `TryFromBytes` docs to explain that you can't always round trip `T -> [u8] -> T` (notably for pointer types), which could be confusing given that, for `TryFromBytes`, the failure would show up at runtime
  - #860
- [x] Rename methods consistent with #1095
  - https://github.com/google/zerocopy/pull/1119
[...]

PR #315, opened 2023-09-03
Add `MaybeValid` type

joshlf:
`MaybeValid<T>` is a `T` which might not be valid. It is similar to
`MaybeUninit<T>`, but it is slightly more strict: any byte in `T` which
is guaranteed to be initialized is also guaranteed to be initialized in
`MaybeValid<T>` (see the doc comment for a more precise definition).

`MaybeValid` is a building block of the `TryFromBytes` design outlined
in https://github.com/google/zerocopy/issues/5.

Makes progress on https://github.com/google/zerocopy/issues/5

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -173,7 +173,7 @@ use core::{
     fmt::{self, Debug, Display, Formatter},
     hash::{Hash, Hasher},
     marker::PhantomData,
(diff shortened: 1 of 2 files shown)


joshlf:
This was accidentally closed when renaming a branch. It's been replaced by https://github.com/google/zerocopy/pull/315.

PR #316, opened 2023-09-03
Minimal implementation of TryFromBytes

joshlf:
This PR is a more minimal version of #279. Most notably, it doesn't rely on field projection, which was a significant source of complexity. A future PR which adds support for deriving `TryFromBytes` will need to tackle field projection.

Makes progress on #5.


joshlf:
This was accidentally closed when renaming a branch. It's replaced by https://github.com/google/zerocopy/pull/317.

PR #329, opened 2023-09-05
Simplify unaligned Ref constructors

joshlf:
These constructors don't currently make use of the `T: Unaligned` bound internally - they just use it to document that an alignment check is guaranteed to be elided. The alignment check in the non-`Unaligned` constructors is trivially optimizable when it's unneeded since it's just a check that `x % 1 == 0`. For that reason, in this commit, we remove these methods' internals and simply have them be wrappers around the equivalent non-`Unaligned` constructors.


src/lib.rs
```diff
@@ -1947,10 +1947,7 @@ where
     /// constructs a new `Ref`. If the check fails, it returns `None`.
     #[inline]
     pub fn new_unaligned(bytes: B) -> Option<Ref<B, T>> {
-        if bytes.len() != mem::size_of::<T>() {
-            return None;
-        }
-        Some(Ref(bytes, PhantomData))
+        Ref::new(bytes)
     }
```
jswrenn, line 1951: Since these methods just become aliases for their non-unaligned counterparts, what's the rationale for keeping them? If it's just to minimize immediate breakage, I'd suggest that they are mark deprecated and removed from a future release.

src/lib.rs
```diff
@@ -1962,11 +1959,7 @@ where
     /// caller. If the length check fails, it returns `None`.
     #[inline]
```
jswrenn, line 1960: Since these are just wrappers around single function calls, I'd suggest going further and marking these as:
```suggestion
    #[inline(always)]
```

jswrenn (dismissed):

src/lib.rs
joshlf: It's to minimize breakage, yeah. I don't want to mark them as deprecated yet for two reasons:
- Without deprecating them, this is a breaking change [1]
- I'm not convinced we'll actually want to get rid of them; there's more work to be done around how our API expresses what is and isn't validated at runtime, and I could see a future in which we keep these and have them return a different error type than the constructors that need to check alignment

I've filed https://github.com/google/zerocopy/issues/338 to track this.

[1] At some point I want to come up with a more unified story around how we keep track of "pending breaking changes". It'd be good to be able to say "this is a breaking change we'll make as soon as we're ready to release e.g. 0.8." We have that loosely tracked in issues right now, but it's neither consistent nor enforced.

src/lib.rs
joshlf: Good call! Done here and elsewhere.

joshlf:
Looks like pushing new changes removed your approval, sorry. Can you re-approve?

jswrenn (approved):

PR #339, opened 2023-09-06
[CI] Use generate_cache job in more places

joshlf:
Have most jobs depend on the `generate_cache` job. Have that job also cache ~/.cargo/bin and install the `cargo readme` tool.


joshlf:
[Asked here](https://users.rust-lang.org/t/get-cargo-install-to-fail-silently-if-the-binary-already-exists/99507) about the CI job failure.

PR #347, opened 2023-09-07
Add KnownLayout trait

joshlf:
In its initial form, the `KnownLayout` trait encodes type layout information slightly more complex than can be gleaned from any arbitrary `T: ?Sized`. This allows it to support not just sized and slice types, but also "custom DSTs" (those with fixed-size fields followed by a trailing slice type). This is the first step to supporting various operations on arbitrary custom DSTs.

Makes progress on #29

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -173,6 +173,7 @@ pub use crate::wrappers::*;
 pub use zerocopy_derive::*;
 
 use core::{
+    alloc::Layout,
     cell::{self, RefMut},
     cmp::Ordering,
     fmt::{self, Debug, Display, Formatter},
@@ -191,9 +192,8 @@ use core::{
 extern crate alloc;
 #[cfg(feature = "alloc")]
 use {
-    alloc::boxed::Box,
-    alloc::vec::Vec,
-    core::{alloc::Layout, ptr::NonNull},
+    alloc::{boxed::Box, vec::Vec},
+    core::ptr::NonNull,
 };
 
 // This is a hack to allow zerocopy-derive derives to work in this crate. They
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -204,6 +203,110 @@ mod zerocopy {
...
+/// trailing slice type, but the type system isn't aware of it).
+#[doc(hidden)]
+#[allow(missing_debug_implementations, missing_copy_implementations)]
+#[cfg_attr(test, derive(Debug, PartialEq, Eq))]
+pub struct DstLayout {
+    /// For sized types, `size_of::<T>()`. For DSTs, the size of the type when
+    /// the trailing slice field contains 0 elements.
+    _fixed_prefix_size: usize,
+    /// The alignment of the type.
+    _align: NonZeroUsize,
```
jswrenn, line 220: These fields could be collapsed into a single `core::alloc::Layout` field:
```suggestion
    /// For sized types, `Layout::new::<T>()`. For DSTs, the it's the layout of
    /// the type when the trailing slice field contains 0 elements.
    _base_layout: core::alloc::Layout,
```
Doing so would simplify the definition of `DstLayout::for_type`, and give us a space-optimized representation of alignment.

src/lib.rs
```diff
@@ -204,6 +203,110 @@ mod zerocopy {
...
+            },
+            _trailing_slice_elem_size: Some(mem::size_of::<T>()),
+        }
+    }
+}
+
+/// A trait which carries information about a type's layout that is used by the
+/// internals of this crate.
+///
+/// This trait is not meant for consumption by code outsie of this crate. While
```
jswrenn, line 263:
```suggestion
/// This trait is not meant for consumption by code outside of this crate. While
```

src/lib.rs
```diff
@@ -204,6 +203,110 @@ mod zerocopy {
...
+/// the normal semver stability guarantees apply with respect to which types
+/// implement this trait and which trait implementations are implied by this
+/// trait, no semver stability guarantees are made regarding its internals; they
+/// may change at any time, and code which makes use of them may break.
```
jswrenn, line 267: I think this is fine for now, but I'm going to want to bikeshed this when `doc(hidden)` is removed. I think this is threading a very fine needle, and the subtleties of what's being said here aren't going to be appreciated.

jswrenn (dismissed):

src/lib.rs
joshlf: I'd be happy to take a suggested alternative wording in this PR if you have one. Also happy to punt until later.

jswrenn: Let's punt until this actually gets exposed — I think I'll have better feedback then.

src/lib.rs
joshlf: Great idea, done!

joshlf:
There's a bit of different logic to handle `Layout`s now that could use a glance before you approve again.

jswrenn (approved):

Issue #325, opened 2023-09-04
Statically prevent ZST slices rather than panicking at runtime?


joshlf:
https://github.com/google/zerocopy/blob/07854bc5c12ce4a4a92ebcb44129da3a9cc485fe/src/lib.rs#L1731-L1735

Alternatively, maybe we could just give these types well-defined semantics (e.g., always synthesize a pointer with 0 slice elements or `isize::MAX` slice elements). Need to be very careful that none of our operations can be round-tripped losslessly; this might break that.

There are also [reasons to be skeptical](https://github.com/google/zerocopy/pull/349#discussion_r1319275140) of wanting to prevent this at all:

> In theory I like the idea of being able to have code like this fail at compile time. However, what do you do about a user who wants to write code like this?
> 
> ```rust
> pub fn do_thing<T: ?Sized + KnownLayout>() {
>     if T::LAYOUT.trailing_slice_elem_type == Some(0) {
>         // Do workaround thing
>     } else {
>         // Do other thing
>     }
> }
> ```
> 
> As in, you can't say that it's a compile-time error to have code that _could_ trigger this panic because the user might have their own logic that ensures it'll never be encountered at runtime.

As of this writing, per #1149, we intend to use post-monomorphization errors for ZST slice conversions. We can always make our restrictions looser in the future. If we do decide to loosen restrictions, note the following caveats:
- We cannot support "exact" conversions since those promise to produce a reference to the entire input byte sequence
- We can support prefix and suffix conversions, but we will need to relax their semantics. Currently, prefix and suffix conversions promise to produce the "largest possible" value that will fit in the given byte sequence. While we could in principle argue that a slice with the maximum number of elements satisfies this promise, it clearly violates the spirit. It would likely make more sense to explicitly carve out ZST slices in the documentation.

See also #202, #284, https://github.com/google/zerocopy/pull/349#discussion_r1319185450, #1125, and #1149

jswrenn:
I'd quite like to see this code work:
```rust
fn main() {
    use zerocopy::*;
    use zerocopy_derive::*;
    
    #[derive(FromBytes, Immutable, KnownLayout)]
    #[repr(C)]
    struct Zesty {
        leading_sized: u8,
        trailing_dst: [()],
    }

    let bytes = [0u8; 100];

    for count in 0..256 {
        let (zesty, _) = Zesty::from_prefix_with_trailing_elements(bytes.as_slice(), count).unwrap();
[...]

PR #349, opened 2023-09-07
[WIP] KnownLayout::validate_size_align

joshlf:
TODO: Tests


src/lib.rs
```diff
@@ -251,6 +256,133 @@ impl DstLayout {
...
+            return None;
+        };
+
+        let (elems, self_bytes) = if let Some(elem_size) = self._trailing_slice_elem_size {
+            // TODO(#67): Once our MSRV is 1.65, use let-else:
+            // https://blog.rust-lang.org/2022/11/03/Rust-1.65.0.html#let-else-statements
+            let elem_size = if let Some(elem_size) = NonZeroUsize::new(elem_size) {
+                elem_size
+            } else {
+                panic!("attempted to cast to slice type with zero-sized element");
```
jswrenn, line 300: Why the panic here? Is the caller also responsible for ensuring that `_trailing_slice_elem_size` is non-zero?

joshlf: Yeah, good point - I'll need to add that to the doc comment. The idea is that the calling code (in zerocopy) just gets to do `T::LAYOUT.validate_cast(...)`, and it is guaranteed to panic if the user (outside of zerocopy) provided an unsupported type. This is consistent with the documented behavior on the various `Ref` constructors.

I have https://github.com/google/zerocopy/issues/325 filed to track an alternate approach at some point in the future.

jswrenn: If ZST slices are generally unsupported, should `DstLayout.trailing_slice_elem_size` have type `Option<NonZeroUsize>` instead?

joshlf: In theory I like the idea of being able to have code like this fail at compile time. However, what do you do about a user who wants to write code like this?

```rust
pub fn do_thing<T: ?Sized + KnownLayout>() {
    if T::LAYOUT.trailing_slice_elem_type == Some(0) {
        // Do workaround thing
    }
    // Do other thing
}
```

As in, you can't say that it's a compile-time error to have code that _could_ trigger this panic because the user might have their own logic that ensures it'll never be encountered at runtime.

joshlf: Just wanted to call out that I moved this logic to the top of the method so that it unconditionally panics. Previously, if certain runtime errors were encountered first, the method would return `None` before encountering the 0-length-slice-element and panicking.

joshlf:
@jswrenn Another thing to call out: I still need to determine whether there are `isize` overflow issues. The function itself doesn't have any, of course, but will it ever be used in a context where it needs to either a) handle checking for `isize` overflow or, b) require that the caller make `isize`-related guarantees?

I think the answer is no: Given an address/length of a sound allocation (ie, one that safe Rust could produce):
- The length cannot overflow `isize`
- The address plus length cannot overflow `usize`

Those are the preconditions to the pointer [`add`](https://doc.rust-lang.org/std/primitive.pointer.html#method.add) method. Note that the address + length sum doesn't have to fit in an `isize`, just in a `usize`. The `isize` overflow issues only apply to _offsets_ within the allocation, not to _addresses_.

That's certainly consistent with the use in the [`TryFromBytes` PR](https://github.com/google/zerocopy/pull/353). Just wanted to put it in your head in case you see something that I'm missing.

src/lib.rs
```diff
@@ -251,6 +260,217 @@ impl DstLayout {
...
+    /// Note that this method does *not* ensure that a pointer constructed from
+    /// its return values will be a valid pointer. In particular, this method
+    /// does not reason about `isize` overflow, which is a requirement of many
+    /// Rust pointer APIs, and may at some point be determined to be a validity
+    /// invariant of pointer types themselves. This should never be a problem so
+    /// long as the arguments to this method are derived from a known-valid
+    /// pointer (e.g., one derived from a safe Rust reference), but it is
+    /// nonetheless the caller's responsibility to justify that pointer
+    /// arithmetic will not overflow based on a safety argument *other than* the
+    /// mere fact that this method returned successfully.
```
joshlf, line 315: @jswrenn Calling this out for review, as I added it after we spoke offline.

src/lib.rs
```diff
@@ -251,6 +260,217 @@ impl DstLayout {
...
+    ///
+    /// # Panics
+    ///
+    /// If `addr + bytes_len` overflows `usize`, `validate_cast` may panic, or
+    /// it may return incorrect results. No guarantees are made about when
+    /// `validate_cast` will panic. The caller should not rely on
+    /// `validate_cast` panicking in any particular condition, even if
+    /// `debug_assertions` are enabled.
+    const fn _validate_cast(
```
joshlf, line 324: TODO: Bikeshed this name

src/lib.rs
```diff
@@ -2748,6 +2968,199 @@ mod tests {
...
+
+        // Unfortunately, these constants cannot easily be used in the
+        // implementation of `validate_cast`, since `panic!` consumes a string
+        // literal, not an expression.
+        //
+        // It's important that these messages be in a separate module. If they
+        // were at the function's top level, we'd pass them to `test!` as, e.g.,
+        // `Err(TRAILING)`, which would run into a subtle Rust footgun - the
+        // `TRAILING` identifier would be treated as a pattern to match rather
+        // than a value to check for equality.
```
joshlf, line 3081: @jswrenn Added this after we spoke offline.

src/lib.rs
```diff
@@ -2748,6 +2968,199 @@ mod tests {
...
+            }
+        }
+
+        let layouts = itertools::iproduct!(0..8, [1, 2, 4, 8], (1..8).map(Some).chain([None]))
+            .filter(|(size, align, trailing_elem_size)| {
+                size % align == 0 && trailing_elem_size.unwrap_or(*align) % align == 0
+            })
+            .map(|(s, a, t)| layout(s, a, t));
+        itertools::iproduct!(layouts, 0..8, 0..8, [_CastType::_Prefix, _CastType::_Suffix])
+            .for_each(validate_behavior);
```
joshlf, line 3161: @jswrenn Slightly refactored this after we spoke offline (including changing the name and signature of `validate_behavior`).

src/lib.rs
joshlf: @jswrenn A few more thoughts:
- `validate_cast_and_convert_metadata` / `validate_cast_and_convert_metadata_no_prefix_suffix`
- `try_compute_cast_metadata` / `try_compute_cast_metadata_no_prefix_suffix`

jswrenn: Given that it's an internal API, I'm partial the first pair of options. They're _very_ descriptive.

joshlf: Awesome, I'll do that.

joshlf: Done!

jswrenn (approved):

Issue #352, opened 2023-09-08
Traits should be clear about stability guarantees regarding `UnsafeCell`


joshlf:
Currently, `FromZeroes`, `FromBytes`, and `AsBytes` require that implementing types contain no `UnsafeCell`s. However, we may [lift that restriction](https://github.com/google/zerocopy/issues/251) in the future.

If any code takes our ban on `UnsafeCell`s to mean that they can use, e.g. `T: FromZeroes` as a guarantee that `T` contains no `UnsafeCell`s, then if we were to relax this requirement, we could allow their code to become silently unsound.

We should update our docs to be very precise and very loud about what external code may take as stability guarantees. We should also guide users towards the zerocopy API itself and discourage using our traits to gate user-written unsafe code.

PR #353, opened 2023-09-08
Add TryFromBytes trait

joshlf:
`TryFromBytes` can be implemented for types which are not `FromZeroes` or `FromBytes`; it supports performing a runtime check to determine whether a given byte sequence contains a valid instance of `Self`.

This is the first step of #5. Future commits will add support for a custom derive and for implementing `TryFromBytes` on unsized types.

Makes progress on #5

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -23,12 +23,14 @@
 //!
 //! # Overview
 //!
-//! Zerocopy provides four core marker traits, each of which can be derived
+//! Zerocopy provides five core marker traits, each of which can be derived
 //! (e.g., `#[derive(FromZeroes)]`):
 //! - [`FromZeroes`] indicates that a sequence of zero bytes represents a valid
 //!   instance of a type
 //! - [`FromBytes`] indicates that a type may safely be converted from an
 //!   arbitrary byte sequence
+//! - [`TryFromBytes`] supports non-`FromBytes` types by providing the ability
+//!   to check the validity of a conversion at runtime
 //! - [`AsBytes`] indicates that a type may safely be converted *to* a byte
 //!   sequence
 //! - [`Unaligned`] indicates that a type's alignment requirement is 1
@@ -288,6 +290,8 @@ use {
     core::alloc::Layout,
 };
 
+pub use crate::util::ptr::Ptr;
+
 // For each polyfill, as soon as the corresponding feature is stable, the
 // polyfill import will be unused because method/function resolution will prefer
 // the inherent method/function over a trait method/function. Thus, we suppress
@@ -387,8 +391,8 @@ impl SizeInfo {
 #[derive(Copy, Clone)]
 #[cfg_attr(test, derive(Debug))]
 #[allow(missing_debug_implementations)]
-pub enum _CastType {
-    _Prefix,
+pub enum CastType {
+    Prefix,
     _Suffix,
 }
 
@@ -503,11 +507,11 @@ impl DstLayout {
     /// `validate_cast_and_convert_metadata` will panic. The caller should not
     /// rely on `validate_cast_and_convert_metadata` panicking in any particular
     /// condition, even if `debug_assertions` are enabled.
-    const fn _validate_cast_and_convert_metadata(
+    const fn validate_cast_and_convert_metadata(
         &self,
         addr: usize,
         bytes_len: usize,
-        cast_type: _CastType,
+        cast_type: CastType,
     ) -> Option<(usize, usize)> {
         // `debug_assert!`, but with `#[allow(clippy::arithmetic_side_effects)]`.
         macro_rules! __debug_assert {
@@ -553,8 +557,8 @@ impl DstLayout {
             // address for a suffix cast (`addr + bytes_len`) is not aligned,
             // then no valid start address will be aligned either.
             let offset = match cast_type {
-                _CastType::_Prefix => 0,
-                _CastType::_Suffix => bytes_len,
+                CastType::Prefix => 0,
+                CastType::_Suffix => bytes_len,
             };
 
             // Addition is guaranteed not to overflow because `offset <=
@@ -580,7 +584,7 @@ impl DstLayout {
                 // multiple of the alignment, or will be larger than
                 // `bytes_len`.
                 let max_total_bytes =
-                    util::_round_down_to_next_multiple_of_alignment(bytes_len, self._align);
+                    util::round_down_to_next_multiple_of_alignment(bytes_len, self._align);
                 // Calculate the maximum number of bytes that could be consumed
                 // by the trailing slice.
                 //
@@ -634,14 +638,14 @@ impl DstLayout {
         __debug_assert!(self_bytes <= bytes_len);
 
         let split_at = match cast_type {
-            _CastType::_Prefix => self_bytes,
+            CastType::Prefix => self_bytes,
             // Guaranteed not to underflow:
             // - In the `Sized` branch, only returns `size` if `size <=
             //   bytes_len`.
             // - In the `SliceDst` branch, calculates `self_bytes <=
             //   max_toatl_bytes`, which is upper-bounded by `bytes_len`.
             #[allow(clippy::arithmetic_side_effects)]
-            _CastType::_Suffix => bytes_len - self_bytes,
+            CastType::_Suffix => bytes_len - self_bytes,
         };
 
         Some((elems, split_at))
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -844,6 +879,214 @@ pub unsafe trait FromBytes: FromZeroes {
...
+/// # Safety
+///
+/// On its own, `T: TryFromBytes` does not make any guarantees about the layout
+/// or representation of `T`. It merely provides the ability to perform a
+/// validity check at runtime via methods like [`try_from_ref`].
+///
+/// TODO(#352): UnsafeCell? We need to specify that it's unsound to implement
+/// `TryFromBytes` for types which contain them, but we also don't want to allow
+/// code to use `TryFromBytes` to mean "has no `UnsafeCell`" in case we relax
+/// that requirement in the future.
```
joshlf, line 904: Need to address this before merging.

src/lib.rs
```diff
@@ -3886,9 +4302,180 @@ mod tests {
...
+                        _f(case.as_bytes());
+                    )*)?
+                }
+            };
+        }
+
+        // SAFETY: All failure cases use a type with:
+        // - the same size as `Self`
+        // - the same alignment as `Self`
+        // - for unsized types, the same slice element size as `Self`
```
joshlf, line 4390: There are no longer any safety preconditions on the callbacks passed to `with_{passing,failing}_test_cases`, so remove this comment.

src/lib.rs
```diff
@@ -3886,9 +4302,180 @@ mod tests {
...
+        }
+
+        // Implements `TryFromBytesTestable`.
+        //
+        // # Safety
+        //
+        // All failure cases must have a valid size and alignment for `Self`.
+        // For unsized types, the failure case's type's trailing slice element
+        // must have the same size as `Self`'s trailing slice element so that
+        // `as` casting a raw pointer preserves length.
```
joshlf, line 4347: There are no longer any safety preconditions on the callbacks passed to `with_{passing,failing}_test_cases`, so remove this section.

src/macros.rs
```diff
@@ -29,13 +29,40 @@ macro_rules! safety_comment {
 }
 
 /// Unsafely implements trait(s) for a type.
+///
+/// # Safety
+///
+/// The trait impl must be sound.
+///
+/// TODO: Update these for `as_bit_valid` taking raw pointer.
```
joshlf, line 37: This is done now; remove this TODO

src/macros.rs
```diff
@@ -66,36 +93,66 @@ macro_rules! unsafe_impl {
...
+            let $candidate = unsafe { NonNull::new_unchecked(candidate.as_ptr() as *mut $repr) };
+            $is_bit_valid
+        }
+    };
+    (@method TryFromBytes) => { #[inline(always)] unsafe fn is_bit_valid(_: NonNull<Self>) -> bool { true } };
+    (@method $trait:ident) => {
+        #[allow(clippy::missing_inline_in_public_items)]
+        fn only_derive_is_allowed_to_implement_this_trait() {}
+    };
+    // TODO: Test this with trybuild.
```
joshlf, line 152: Nevermind, that's overkill. Even if this were buggy, you'd just end up with a worse compiler error message - either a  missing `is_bit_valid` impl (for `TryFromBytes`) or an unexpected `is_bit_valid` impl (for any other trait)

src/lib.rs
joshlf: Maybe, given the following warning about not providing stability guarantees and about making code silently become unsound, we can just omit mentioning this until we address #352 as a whole.

src/lib.rs
```diff
@@ -884,6 +921,215 @@ pub unsafe trait FromBytes: FromZeroes {
...
+pub unsafe trait TryFromBytes: KnownLayout {
+    /// Does a given memory range contain a valid instance of `Self`?
+    ///
+    /// # Safety
+    ///
+    /// ## Preconditions
+    ///
+    /// `candidate` must be a validly-aligned pointer, and it must be [valid]
+    /// for reads. The total length encoded by the pointer must not overflow an
+    /// `isize`. The memory addressed by `c` must fall within a single
```
joshlf, line 966: s/`c`/`candidate`

src/lib.rs
```diff
@@ -884,6 +921,215 @@ pub unsafe trait FromBytes: FromZeroes {
...
+    /// `*candidate` contains a valid `Self`.
+    ///
+    /// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
+    #[doc(hidden)]
+    unsafe fn is_bit_valid(candidate: NonNull<Self>) -> bool;
+
+    /// Attempts to interpret a byte slice as a `Self`.
+    ///
+    /// `try_from_ref` validates that `bytes` contains a valid `Self` as defined
+    /// by [`is_bit_valid`]. If it does, then `bytes` is reinterpreted as a
```
joshlf, line 1008: `is_bit_valid` is `doc(hidden)`; don't mention it here.

src/lib.rs
```diff
@@ -884,6 +921,215 @@ pub unsafe trait FromBytes: FromZeroes {
...
+    /// `Self`.
+    ///
```
joshlf, line 1010: TODO: Maybe document that we guarantee that our validity is a subset of Rust's? Ie, we might reject a type which Rust thinks is valid, but we'll never do the inverse.

src/lib.rs
```diff
@@ -447,15 +456,28 @@ impl DstLayout {
...
         Some((elems, split_at))
     }
+
+    /// Calls `validate_exact_cast` and ensures that all of `bytes_len` are
+    /// used.
+    ///
+    /// Passes all arguments to `validate_cast`, and passes `CastType::Prefix`.
+    /// If the cast would leave any suffix bytes left over,
+    /// `validate_exact_cast` returns `None`.
+    const fn validate_exact_cast(&self, addr: usize, bytes_len: usize) -> Option<usize> {
```
joshlf, line 475: Consider renaming this:
- `validate` (as opposed to `is_valid`) implies side effects
- `validate` implies that it _only_ performs validation, but it also computes metadata

Some options brainstormed with @jswrenn:
- `try_cast_metadata` / `try_cast_metadata_no_prefix_suffix`

src/lib.rs
```diff
@@ -1197,6 +1458,33 @@ safety_comment! {
...
+    ///   sound implementation of `TryFromBytes::is_bit_valid`.
+    ///
+    /// Note that it might be slightly simpler to treat `candidate` as a `[u8;
+    /// 4]` - it would make the code and the safety argument a bit simpler.
+    /// However, using `&MaybeUninit<char>` ensures that the compiler preserves
+    /// alignment information, which it may be able to use to produce smaller or
+    /// better-performing assembly.
+    ///
+    /// [1] https://doc.rust-lang.org/std/primitive.char.html#method.from_u32
+    unsafe_impl!(char: TryFromBytes; |candidate: &MaybeUninit<char>| {
```
joshlf, line 1480: Leave a TODO that, once `char` is guaranteed to have the same alignment as `u32`, we can use that instead of `MaybeUninit<char>`.

src/lib.rs
```diff
@@ -884,6 +921,215 @@ pub unsafe trait FromBytes: FromZeroes {
...
+///
+/// [`try_from_ref`]: TryFromBytes::try_from_ref
+pub unsafe trait TryFromBytes: KnownLayout {
+    /// Does a given memory range contain a valid instance of `Self`?
+    ///
+    /// # Safety
+    ///
+    /// ## Preconditions
+    ///
+    /// `candidate` must be a validly-aligned pointer, and it must be [valid]
```
djkoloski, line 964: Suggest: "`candidate` must be properly aligned for `Self``

joshlf, line 964: Specify that this is the alignment of `Self`.

src/lib.rs
```diff
@@ -884,6 +921,215 @@ pub unsafe trait FromBytes: FromZeroes {
...
+    /// for reads. The total length encoded by the pointer must not overflow an
+    /// `isize`. The memory addressed by `c` must fall within a single
+    /// allocation.
+    ///
+    /// `candidate`'s referent is not required to contain a valid `Self`.
```
joshlf, line 969: Rephrase: "`candidate` is not required to refer to a valid `Self`."

src/lib.rs
```diff
@@ -884,6 +921,215 @@ pub unsafe trait FromBytes: FromZeroes {
...
+        // - Since `Self` is not allowed to contain any `UnsafeCell`s:
+        //   - The caller cannot use the `&Self` to perform interior mutation on
+        //     a byte range that `bytes` views as not containing `UnsafeCell`s
+        //   - The caller cannot use the `&Self` to write invalid values to
+        //     `bytes` (namely, uninitialized bytes, as `[u8]` has no other bit
+        //     validity constraints)
+        // - Since `[u8]` does not contain any `UnsafeCell`s, we are guaranteed
+        //   that, having verified that `slf` currently contains a valid `Self`,
+        //   code with access to `bytes` cannot cause it to no longer contain a
+        //   valid `Self` in the future
```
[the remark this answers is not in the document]
joshlf: This doesn't affect us right now: We only permit transmuting from `&[u8]`. There is no memory model under which it would be sound for us to allow `&[u8]` -> `&T` where `T` contains `UnsafeCell`s.* If we support `&T` -> `&U` at some point in the future, then this will become relevant.

\* In theory, this could be valid in a world where UnsafeCells only cause problems when _used_, but even then, it would only be valid for types which we define and so can guarantee never allow the interior mutability to be exercised.

src/lib.rs
```diff
@@ -13,12 +13,14 @@
...
-//! This is enabled by four core marker traits, each of which can be derived
-//! (e.g., `#[derive(FromZeroes)]`):
+//! This is enabled by five core traits, each of which can be derived (e.g.,
+//! `#[derive(FromZeroes)]`):
 //! - [`FromZeroes`] indicates that a sequence of zero bytes represents a valid
 //!   instance of a type
 //! - [`FromBytes`] indicates that a type may safely be converted from an
 //!   arbitrary byte sequence
+//! - [`TryFromBytes`] supports non-`FromBytes` types by providing the ability
+//!   to check the validity of a conversion at runtime
```
joshlf, line 23: I don't love this wording. @jswrenn any suggestions?

src/lib.rs
```diff
@@ -901,6 +944,268 @@ pub unsafe trait FromBytes: FromZeroes {
...
+    /// Unsafe code may assume that, if `is_bit_valid(candidate)` returns true,
+    /// `*candidate` contains a valid `Self`.
+    ///
+    /// # Panics
+    ///
+    /// `is_bit_valid` may panic. Callers are responsible for ensuring that any
+    /// `unsafe` code remains sound even in the face of `is_bit_valid`
+    /// panicking. (We support user-defined validation routines; so long as
+    /// these routines are not required to be `unsafe`, there is no way to
+    /// ensure that these do not generate panics.)
```
joshlf, line 1054: cc @jswrenn @djkoloski Calling this out; I've updated all the safety comments in this PR to address panics.

src/lib.rs
```diff
@@ -1354,6 +1697,40 @@ safety_comment! {
...
+    /// 4]` - it would make the code and the safety argument a bit simpler.
+    /// However, using `&MaybeUninit<char>` ensures that the compiler preserves
+    /// alignment information, which it may be able to use to produce smaller or
+    /// better-performing assembly.
+    ///
+    /// [1] https://doc.rust-lang.org/std/primitive.char.html#method.from_u32
+    ///
+    /// TODO(https://github.com/rust-lang/reference/pull/1401): Use `&u32`
+    /// instead once it's guaranteed that `align_of::<u32>() ==
+    /// align_of::<char>()`.
```
joshlf, line 1725: This has landed; update this PR.

PR #356, opened 2023-09-08
DRAFT for adding Parser type

sivadeilra:
I use the `zerocopy` crate extensively in my work.  I find myself writing this `Parser` type over and over in different variations in different projects. It seems like the best home for `Parser` is in the `zerocopy` crate itself.

This is a DRAFT PR.  If you agree with the general idea of this PR, I'll write better doc comments, make the API more consistent, etc.  I think the core idea is sound: `Parser` is a type that decodes values and references from a `&[u8]`, in a form that is much more convenient than repeated calls to methods on `Ref::new_slice_from...`.

It looks like my editor has chopped off some whitespace at the end of lines.  I'll revert that before submitting real PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,7 @@ derive = ["zerocopy-derive"]
 alloc = []
 simd = []
 simd-nightly = ["simd"]
+std = []
 # This feature depends on all other features that work on the stable compiler.
 # We make no stability guarantees about this feature; it may be modified or
 # removed at any time.
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -1,0 +1,229 @@
+//! Provides a way to easily decode types from a `&[u8]`.
+
+use crate::{FromBytes, Ref, Unaligned};
+use core::mem::size_of;
+
+/// Parses values and types
+#[derive(Debug, Clone)]
+pub struct Parser<'a> {
+    /// The bytes that have not yet been decoded.
+    rest: &'a [u8],
+}
+
+/// This type is returned by Parser methods.
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+pub struct ParserError;
+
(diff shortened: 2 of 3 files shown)


sivadeilra:
Hmmm, it's showing my CLA as being invalid.  I had a CLA on file in 2021.  I'll check on getting that refreshed.  (I've made contributions to this repo before.)

Cargo.toml
```diff
@@ -49,6 +49,7 @@ derive = ["zerocopy-derive"]
 alloc = []
 simd = []
 simd-nightly = ["simd"]
+std = []
```
sivadeilra, line 52: This is here because I added a `ParserError` type.  When `std` is available, `ParserError` implements the `std::error::Error` trait.

src/parser.rs
```diff
@@ -0,0 +1,229 @@
...
+/// Parses values and types
+#[derive(Debug, Clone)]
+pub struct Parser<'a> {
+    /// The bytes that have not yet been decoded.
+    rest: &'a [u8],
+}
+
+/// This type is returned by Parser methods.
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+pub struct ParserError;
```
sivadeilra, line 15: This `ParserError` type exists so that users of the crate can define their own `From<ParserError>` implementations, which convert from `ParserError` to their own error types.  Because `ParserError` implements `std::error::Error`, it can also be converted to `anyhow::Error`.

Both of these are ergonomic benefits, and they would be lost if these functions returned `Option` instead.

joshlf:
Discussed offline; I'm going to close this for the time being, but for those following along at home: We have active work ongoing to add support for a higher-level buffer/parsing framework built around zerocopy. If you're interested in participating in development or have suggestions or requested use cases, please reach out.

PR #361, opened 2023-09-09
[derive] Simplify code, remove obsolete features

joshlf:
Clean up the implementation, especially in `fn impl_block`. Make the following notable changes:
- Previously, `syn` didn't support parsing macro invocations in const generics without the `full` feature enabled. To avoid the compile-time overhead of that feature, we worked around it by constructing AST nodes manually. `syn` has since added support for this without requiring the `full` feature, so we make use of it.
- We used to need to split types into those that transatively depended upon type generics (like `[T; 2]`) and those that didn't (like `[u8; 2]`). We made a change in #119 that made this distinction irrelevant, but we never removed the code to perform the split. In this commit, we remove that code. That code was the only reason we needed to enable `syn`'s `visit` feature, so we are also able to remove that feature dependency.

### DIFF
--- a/zerocopy-derive/src/lib.rs
+++ b/zerocopy-derive/src/lib.rs
@@ -30,10 +30,9 @@ mod repr;
 use {
     proc_macro2::Span,
     quote::quote,
-    syn::visit::{self, Visit},
     syn::{
-        parse_quote, punctuated::Punctuated, token::Comma, Data, DataEnum, DataStruct, DataUnion,
-        DeriveInput, Error, Expr, ExprLit, GenericParam, Ident, Lifetime, Lit, Type, TypePath,
+        parse_quote, Data, DataEnum, DataStruct, DataUnion, DeriveInput, Error, Expr, ExprLit,
+        GenericParam, Ident, Lit,
     },
 };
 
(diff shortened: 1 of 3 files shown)


zerocopy-derive/src/lib.rs
```diff
@@ -610,77 +564,37 @@ fn impl_block<D: DataExt>(
...
-    let use_concrete = if input.generics.params.is_empty() {
-        Some(quote! {
-            const _: () = {
-                fn must_implement_trait<T: zerocopy::#trait_ident + ?::core::marker::Sized>() {}
-                let _ = must_implement_trait::<#type_ident>;
-            };
-        })
-    } else {
-        None
-    };
```
joshlf, line 670: @djkoloski Calling this out: This is the only change here that affects the generated code. It doesn't seem to be necessary now that we add bounds for field types to the `where` clause, but I wanted to make sure to get scrutiny on this. I'm already surprised that we can put bounds on non-generic field types (like `u8`) in the `where` clause because I would assume that requires trivial_bounds ([which is unstable](https://github.com/rust-lang/rust/issues/48214)), but it does work:

```rust
trait Foo {}
impl Foo for u8 {}
impl Foo for () where u8: Foo {} // not trivial_bounds apparently??
```

joshlf: Discussed offline: We're not sure why this is valid, but it's definitely valid. It's been compiling successfully since 1.6.

PR #371, opened 2023-09-11
[WIP] TryFromBuild derive support for enums

joshlf:
TODO:
- Clean up repr computation
- Support enums with fields
- If we don't support enums with fields, enforce that
- Should enum variants be considered "used" in the sense of dead code? It's possible to never construct an enum variant from Rust's perspective even though it is constructed via TryFromBytes.

### DIFF
--- a/zerocopy-derive/src/lib.rs
+++ b/zerocopy-derive/src/lib.rs
@@ -28,11 +28,11 @@ mod ext;
 mod repr;
 
 use {
-    proc_macro2::Span,
+    proc_macro2::{Span, TokenStream},
     quote::quote,
     syn::{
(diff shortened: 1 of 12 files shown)


zerocopy-derive/src/lib.rs
```diff
@@ -597,6 +607,66 @@ fn impl_block<D: DataExt>(
...
+                    .iter()
+                    .scan(Discriminant::default(), |disc, var| {
+                        let disc_expr = disc.update_and_generate_expr(&var.discriminant);
+
+                        let variant_fields =
+                            var.fields.fields().into_iter().map(|(name, ty)| quote!(#name: #ty));
+
+                        let cases = var.fields.fields().into_iter().map(foo);
+                        Some(quote!(|| {
+                            struct Variant {
```
joshlf, line 657: Give this an explicit repr -- oops!

PR #372, opened 2023-09-11
[WIP][derive] Support custom TryFromBytes validator

joshlf:
TODO:
- Cleaner way to pass name of validator to `impl_block`?
- Cleaner way to parse validator attribute?
- Safety comment in emitted call to `validate`
- Tests (especially including tests for the error message resulting from passing a validator with the wrong type signature - test for both argument types and return types)
  - Also test for invalid zerocopy attributes
- Other misc TODO comments in code


zerocopy-derive/src/lib.rs
```diff
@@ -616,10 +653,38 @@ fn impl_block<D: DataExt>(
...
+fn parse_zerocopy_attrs(attrs: &[syn::Attribute]) -> Result<Option<ZcValidatorAttr>, Error> {
+    let mut attrs = attrs.iter().filter(|attr| attr.path().is_ident("zerocopy"));
+    // TODO: Any way to do this unwrap-or-return-Ok(None) using a combinator?
+    let attr = if let Some(attr) = attrs.next() {
+        attr
+    } else {
+        return Ok(None);
+    };
+    if let Some(attr) = attrs.next() {
+        return Err(Error::new_spanned(attr, "multiple attributes are unsupported"));
```
joshlf, line 671: Change this to "multiple zerocopy attributes"

zerocopy-derive/src/lib.rs
```diff
@@ -587,7 +596,33 @@ fn impl_block<D: DataExt>(
...
+                    //                          ^^^^^^^^^^
+                    //                               |
+                    //                               |
+                    //   (type errors attributed to this span)
+                    //
+                    // TODO: Validate that assigning to this function pointer has no
+                    // performance impact. Presumably it's trivially
+                    // constant-foldable, but since it's a function pointer, maybe
+                    // the compiler is worse at optimizing it?
+                    let validate: fn(&Self) -> bool = Self::#validator;
```
joshlf, line 619: Support functions that are not methods; the user can themselves write `Self::validator` if need be.

zerocopy-derive/src/lib.rs
```diff
@@ -587,7 +596,33 @@ fn impl_block<D: DataExt>(
...
+                    // SAFETY: TODO
+                    validate(unsafe { candidate.as_ref() })
```
joshlf, line 621: Remember: This must be sound in the face of `validate` panicking.

zerocopy-derive/src/lib.rs
joshlf: Better yet, support arbitrary expressions:

```rust
#[zerocopy(validator = |x| x.foo())]
```

Emit code in a wrapped function to prevent validator from messing with environment:

```rust
fn is_bit_valid(candidate: NonNull<Self>) -> bool {
    let c: &Self = {
        // unsafe validation stuff
    };

    // Even if hygiene is broken, this can't access anything from inside
    // the preceding block.
    let validate: fn(&Self) -> bool = #validator;
    validate(c)
}
```

Issue #378, opened 2023-09-13
Use formal modeling/verification to test or prove the correctness of core algorithms


joshlf:
While zerocopy is conceptually complex, the amount of code that executes in order to perform a particular operation is often very small. This makes it a perfect target for formal modeling and verification techniques.

This issue tracks testing or proving the correctness of our core algorithms using these tools.

Currently, we use formal modeling or verification in the following places (TODO: better way to keep this list up-to-date automatically?):
- #394
- #402

jswrenn:
I left a comment about our need for alternative targets here: https://github.com/model-checking/kani/issues/2402#issuecomment-1726613332

PR #393, opened 2023-09-18
Add round_down_to_next_multiple_of_alignment

### LINKED ISSUE #390: Compute "round down to next multiple of alignment" more efficiently
```rust
fn round_down_to_next_multiple_of_alignment(n: usize, align: NonZeroUsize) -> usize {
    let mask = !(align.get() - 1);
    n & mask
}
```

joshlf:
This function implements the operation more efficiently using bitmasking instead of division and multiplication. While LLVM is probably smart enough in practice to produce this sort of code, Kani (#378) has trouble with multiplication, and so this code is hopefully more Kani-friendly.

Closes #390

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -5,7 +5,7 @@
 #[path = "third_party/rust/layout.rs"]
 pub(crate) mod core_layout;
 
-use core::mem;
+use core::{mem, num::NonZeroUsize};
 
 pub(crate) trait AsAddress {
     fn addr(self) -> usize;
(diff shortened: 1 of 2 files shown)


src/util.rs
```diff
@@ -106,3 +127,25 @@ pub(crate) mod testutil {
...
+        for align in [1, 2, 4, 8, 16] {
+            for n in 0..256 {
+                let align = NonZeroUsize::new(align).unwrap();
+                let want = alt_impl(n, align);
+                let got = _round_down_to_next_multiple_of_alignment(n, align);
+                assert_eq!(got, want, "round_down_to_next_multiple_of_alignment({n}, {align})");
+            }
+        }
+    }
+}
```
jswrenn, line 151: The testing here is sufficiently convincing, but I'd also eventually like to see this proved correct with kani, if possible.

jswrenn (approved):

jswrenn:
Looks like the test is a little to much for miri to handle. Reduce the iterations, and I'll re-approve.

src/util.rs
joshlf: Agreed; I'll track this in #378

PR #398, opened 2023-09-19
[clippy] Be more permissive with safety comments

joshlf:
Set the following Clippy configuration variables to `true`:
- `accept-comment-above-statement`
- `accept-comment-above-attributes`

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,0 +1,6 @@
+# Copyright 2023 The Fuchsia Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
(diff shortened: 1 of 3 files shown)


jswrenn (dismissed):

clippy.toml
joshlf, line 1: This should be unnecessary as of [this change](https://github.com/rust-lang/rust-clippy/pull/11170). Now we just need to wait for a nightly that a) includes that change and, b) supports all of our CI targets.

joshlf: The `clippy.toml` seems to still be necessary. We can remove it later if it's no longer needed.

jswrenn (approved):

PR #402, opened 2023-09-19
Prove correctness of layout len utilities with kani

jswrenn:
Uses kani to prove that `round_down_to_next_multiple_of_alignment` and `padding_needed_for` are equivalent to straightforward model implementations.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -149,3 +149,53 @@ mod tests {
         }
     }
 }
+
+#[cfg(kani)]
+mod proofs {
+    use super::*;
+
+    #[kani::proof]
+    fn prove_round_down_to_next_multiple_of_alignment() {
+        fn model_impl(n: usize, align: NonZeroUsize) -> usize {
+            assert!(align.get().is_power_of_two());
+            let mul = n / align.get();
+            mul * align.get()
+        }
+
+        let align: NonZeroUsize = kani::any();
+        kani::assume(align.get().is_power_of_two());
+        let n: usize = kani::any();
+
+        let expected = model_impl(n, align);
+        let actual = _round_down_to_next_multiple_of_alignment(n, align);
+        assert_eq!(expected, actual, "round_down_to_next_multiple_of_alignment({n}, {align})");
+    }
+
+    // Restricted to nightly since we use the unstable `usize::next_multiple_of`
+    // in our model implementation.
+    #[cfg(__INTERNAL_USE_ONLY_NIGHLTY_FEATURES_IN_TESTS)]
+    #[kani::proof]
+    fn prove_padding_needed_for() {
+        fn model_impl(len: usize, align: NonZeroUsize) -> usize {
+            let padded = len.next_multiple_of(align.get());
+            let padding = padded - len;
+            padding
+        }
+
+        let align: NonZeroUsize = kani::any();
+        kani::assume(align.get().is_power_of_two());
+        let len: usize = kani::any();
+        // Constrain `len` to valid Rust lengths, since our model implementation
+        // isn't robust to overflow.
+        kani::assume(len <= isize::MAX as usize);
+        kani::assume(align.get() < 1 << 29);
+
+        let expected = model_impl(len, align);
+        let actual = core_layout::_padding_needed_for(len, align);
+        assert_eq!(expected, actual, "padding_needed_for({len}, {align})");
+
+        let padded_len = actual + len;
+        assert_eq!(padded_len % align, 0);
+        assert!(padded_len / align >= len / align);
+    }
+}
--- a/src/third_party/rust/layout.rs
+++ b/src/third_party/rust/layout.rs
@@ -14,6 +14,8 @@ use core::num::NonZeroUsize;
 /// # Panics
 ///
 /// May panic if `align` is not a power of two.
+//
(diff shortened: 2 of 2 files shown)


src/util.rs
```diff
@@ -149,3 +149,48 @@ mod tests {
...
 }
+
+#[cfg(kani)]
+mod proofs {
+    use super::*;
+
+    #[kani::proof]
+    fn prove_round_down_to_next_multiple_of_alignment() {
+        fn model_impl(n: usize, align: NonZeroUsize) -> usize {
+            debug_assert!(align.get().is_power_of_two());
```
joshlf, line 160: Since we're in tests, might as well make this a non-debug assert:

```suggestion
            assert!(align.get().is_power_of_two());
```

src/util.rs
```diff
@@ -149,3 +149,48 @@ mod tests {
...
+        let align: NonZeroUsize = kani::any();
+        kani::assume(align.get().is_power_of_two());
+        let n: usize = kani::any();
+
+        let expected = model_impl(n, align);
+        let actual = _round_down_to_next_multiple_of_alignment(n, align);
+        assert_eq!(expected, actual, "round_down_to_next_multiple_of_alignment({n}, {align})");
+    }
+
+    // Restricted to nightly, since we use the unstable
```
joshlf, line 174:
```suggestion
    // Restricted to nightly since we use the unstable
```

src/util.rs
```diff
@@ -149,3 +149,48 @@ mod tests {
...
+            let padding = padded - len;
+            padding
+        }
+
+        let align: NonZeroUsize = kani::any();
+        kani::assume(align.get().is_power_of_two());
+        let len: usize = kani::any();
+        // Constrain `len` to valid Rust lengths, since our model implementation
+        // isn't robust to overflow.
+        kani::assume(len < isize::MAX as usize);
```
joshlf, line 190: Maybe we could also do a version of this test that validates padding by reference to properties that must be true of the return value rather than by reference to a reference implementation in order to cover this part of the input space?

joshlf (changes requested):
Sweet!!

src/util.rs
jswrenn: I've added two additional assertions that aren't model-dependent, but I don't think I quite follow your point.

Perhaps I misunderstand, but I'd argue we *don't* want to cover *that* part of the input space. Rust lengths absolutely cannot exceed `isize::MAX`, and alignments cannot exceed 2<sup>29</sup>. Since it's nonsensical to pad a Rust type such that it would exceed `isize::MAX`, it does not matter what our optimized implementation does in this range of inputs. If we could come up with an even more optimized implementation at the cost of doing something weird outside of this range, it'd be worth it.

jswrenn: I'd almost argue that this isn't quite a model test:
1. The "model" here is a verbatim (almost literate) expression of the contract of `padding_needed_for`.
2. Due to the domain restriction, we're *not* actually testing that the model and actual impl are exactly identical.

If I were writing a traditional, non-model property based test, I'd more or less do so by inlining `model`.

joshlf: My perspective is that if we test for the whole input space (all `len: usize` and all power-of-two `align: NonZeroUsize`), then we don't need to say that, in order for a call to `padding_needed_for` to be valid, its arguments need to be restricted to a subset of what the types allow. Of course, we never expect to be called with that part of the input space, but IMO it makes the correctness argument for usage sites easier if we don't have to also justify that inputs are in this subset of the input space.



> If we could come up with an even more optimized implementation at the cost of doing something weird outside of this domain, it'd be worth it.

Definitely agreed, and IMO that would justify changing what we test for here.

joshlf: This would be a longer-term effort, but I could also see a witness type: `UsizeThatDoesntOverflowIsize(usize)`. If we thread its use through the codebase, it might make it easier for us to reason about these invariants locally rather than having to reason about them as a viral property.

jswrenn: If we are expanding the meaningful domain of the function, I think we need to do one of:
1. Retain the optimized implementation, and rename/redocument `padding_needed_for`. If the function does not do what the name says for *most* of its inputs, it probably shouldn't have that name.
2. Retain the optimized implementation, but panic or return `None` for out-of-domain inputs, and document that.
3. Keep the name, but change the type signature to return `u128`, allowing us correctly handle larger lengths.

It's not clear to me which of these will simplify the correctness argument at use-sites. Already, the presence of lengths greater than `isize::MAX` is a soundness issue *anywhere* we manipulate allocation lengths, and none of these options totally eliminate that. The second option comes closest, in that it provides an extra check.

We could perhaps also create invariant-checked abstraction netypes around allocation length and alignment, and use those newtypes as the arguments to `padding_needed_for` and anywhere else we manipulate allocation lengths.

jswrenn: Whoops — only just seeing your comment about `UsizeThatDoesntOverflowIsize`. I agree that's probably the way to go, but it's a very big undertaking.

joshlf: I'm okay approving this as-is, but can you add a TODO comment to `_padding_needed_for` (and other relevant locations while you're at it) to use that approach? I filed an issue to track: https://github.com/google/zerocopy/issues/419.

joshlf (approved):

PR #406, opened 2023-09-20
Initial commit of `Ptr` type

joshlf:
`Ptr` is like `NonNull`, but has many restrictions which make it so that using a `Ptr` in unsafe code requires much simpler soundness proofs. Notably, it supports a `try_cast_into<U>` method where `U: ?Sized + KnownLayout`, which is a building block of `TryFromBytes`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -247,7 +247,8 @@ use core::{
         NonZeroU16, NonZeroU32, NonZeroU64, NonZeroU8, NonZeroUsize, Wrapping,
     },
     ops::{Deref, DerefMut},
-    ptr, slice,
+    ptr::{self, NonNull},
+    slice,
 };
 
 #[cfg(feature = "alloc")]
@@ -255,8 +256,17 @@ extern crate alloc;
 #[cfg(feature = "alloc")]
 use {
     alloc::{boxed::Box, vec::Vec},
-    core::{alloc::Layout, ptr::NonNull},
+    core::alloc::Layout,
 };
+
+// For each polyfill, as soon as the corresponding feature is stable, the
+// polyfill import will be unused because method/function resolution will prefer
+// the inherent method/function over a trait method/function. Thus, we suppress
+// the `unused_imports` warning.
+//
+// See the documentation on `util::polyfills` for more information.
+#[allow(unused_imports)]
+use crate::util::polyfills::NonNullExt as _;
 
 // This is a hack to allow zerocopy-derive derives to work in this crate. They
 // assume that zerocopy is linked as an extern crate, so they access items from
@@ -352,8 +362,11 @@ impl SizeInfo {
     }
 }
 
-#[cfg_attr(test, derive(Copy, Clone, Debug))]
-enum _CastType {
+#[doc(hidden)]
+#[derive(Copy, Clone)]
+#[cfg_attr(test, derive(Debug))]
+#[allow(missing_debug_implementations)]
+pub enum _CastType {
     _Prefix,
     _Suffix,
 }
@@ -457,6 +470,9 @@ impl DstLayout {
     /// mere fact that this method returned successfully.
     ///
     /// # Panics
+    ///
+    /// `validate_cast_and_convert_metadata` will panic if `self` describes a
+    /// DST whose trailing slice element is zero-sized.
     ///
     /// If `addr + bytes_len` overflows `usize`,
     /// `validate_cast_and_convert_metadata` may panic, or it may return
@@ -625,12 +641,27 @@ impl DstLayout {
 pub unsafe trait KnownLayout: sealed::KnownLayoutSealed {
     #[doc(hidden)]
     const LAYOUT: DstLayout;
+
+    /// SAFETY: The returned pointer has the same address and provenance as
+    /// `bytes`. If `Self` is a DST, the returned pointer's referent has `elems`
+    /// elements in its trailing slice. If `Self` is sized, `elems` is ignored.
+    #[doc(hidden)]
+    fn raw_from_ptr_len(bytes: NonNull<u8>, elems: usize) -> NonNull<Self>;
 }
 
 impl<T: KnownLayout> sealed::KnownLayoutSealed for [T] {}
 // SAFETY: Delegates safety to `DstLayout::for_slice`.
 unsafe impl<T: KnownLayout> KnownLayout for [T] {
     const LAYOUT: DstLayout = DstLayout::for_slice::<T>();
+
+    // SAFETY: `.cast` preserves address and provenance. The returned pointer
+    // refers to an object with `elems` elements by construction.
+    #[inline(always)]
+    fn raw_from_ptr_len(data: NonNull<u8>, elems: usize) -> NonNull<Self> {
+        // TODO(#67): Remove this allow. See NonNullExt for more details.
+        #[allow(unstable_name_collisions)]
+        NonNull::slice_from_raw_parts(data.cast::<T>(), elems)
+    }
 }
 
 #[rustfmt::skip]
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -573,12 +577,65 @@ impl DstLayout {
...
+    }
+
+    /// TODO
+    ///
+    /// # Safety
+    ///
+    /// TODO: Mention isize overflow, address space wraparound.
+    #[doc(hidden)]
+    #[inline(always)]
+    fn try_cast_from_no_prefix_suffix(bytes: NonNull<[u8]>) -> Option<NonNull<Self>> {
```
jswrenn, line 612: Okay, I've *finally* thought of something better than "no_affix":
```suggestion
    fn try_cast_from_no_excess(bytes: NonNull<[u8]>) -> Option<NonNull<Self>> {
```

joshlf: I still feel like `prefix_suffix`, though a bit more verbose, is the most straightforward and descriptive.

jswrenn: Maybe? I'm a little concerned that prefix and suffix are getting a bit overloaded. A prefix cast isn't one with no prefix, and a suffix cast isn't one with no suffix. If anything, this is a cast in which prefix == suffix.

src/util.rs
```diff
@@ -5,7 +5,245 @@
...
+    // - `ptr` has the same provenance as `A`
+    // - `ptr` addresses a byte range which is entirely contained in `A`
+    // - `ptr` addresses a byte range which is not longer than `isize::MAX`
+    // - `ptr` addresses a byte range which does not wrap around the address
+    //   space
+    // - `ptr` is validly-aligned for `T`
+    // - `A` is guaranteed to live for at least `'a`
+    // - `T: 'a`
+    ptr: NonNull<T>,
+    _marker: PhantomData<&'a ()>,
```
jswrenn, line 52: In cases where a marker field can be given a more specific name, I prefer to use that name:
```suggestion
    _lifetime: PhantomData<&'a ()>,
```

src/util.rs
```diff
@@ -5,7 +5,245 @@
...
+/// - It must derive from a valid allocation
+/// - It must reference a byte range which is contained inside the allocation
+///   from which it derives
+///   - As a consequence, the byte range it references must have a size which
+///     does not overflow `isize`
+/// - It must satisfy `T`'s alignment requirement
+///
+/// Thanks to these restrictions, it is easier to prove the soundness of some
+/// operations using `Ptr`s.
+pub(crate) struct Ptr<'a, T: 'a + ?Sized> {
```
jswrenn, line 40: Like `NonNull`, we should probably document the variance of the parameters. Here, both `'a` and `T` are covariant.

jswrenn (changes requested):

src/util.rs
```diff
@@ -130,7 +423,138 @@ pub(crate) mod testutil {
 
 #[cfg(test)]
 mod tests {
+    use core::mem::MaybeUninit;
+
     use super::*;
+    use crate::{util::testutil::AU64, FromBytes};
+
+    #[test]
+    fn test_ptr_try_cast_into_soundness() {
```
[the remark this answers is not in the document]
joshlf: Discussed offline; we're going to just disable symbolic alignment checking because the alternative introduces more complexity and CI test time than it's worth for a marginal benefit.

src/lib.rs
joshlf: Discussed offline; we're going with `try_cast_from_no_leftover`.

src/util.rs
joshlf: Done.

src/util.rs
joshlf: Done.

src/util.rs
```diff
@@ -5,7 +5,280 @@
...
+    ///
+    /// Panics if `U` is a DST whose trailing slice element is zero-sized.
+    pub(crate) fn _try_cast_into<U: 'a + ?Sized + KnownLayout>(
+        &self,
+        cast_type: _CastType,
+    ) -> Option<(Ptr<'a, U>, usize)> {
+        // PANICS: By invariant, the byte range addressed by `self.ptr` does not
+        // wrap around the address space. This implies that the sum of the
+        // address (represented as a `usize`) and length do not overflow
+        // `usize`, as required by `validate_cast_and_convert_metadata`.
```
jswrenn, line 160: Considering that "SAFETY" is used to introduce a comment that explains why code is actually safe, it's a bit odd that "PANICS" is introducing a comment about why something *doesn't* panic. I think you could just append "and thus `validate_cast_and_convert_metadata` won't panic.

src/util.rs
```diff
@@ -5,7 +5,280 @@
...
+        // Once it's documented, reference the guarantee that `NonNull::from`
+        // preserves provenance.
+        //
+        // TODO(#429),
+        // TODO(https://github.com/rust-lang/unsafe-code-guidelines/issues/465):
+        // - [1] Where does the reference document that allocations fit in
+        //   `isize`?
+        // - [2] Where does the reference document that allocations don't wrap
+        //   around the address space?
+        Ptr { ptr: NonNull::from(t), _lifetime: PhantomData }
```
jswrenn, line 273: It's regrettable that we don't yet have unsafe fields, or unsafe implicit instructors. What do you think about renaming the `ptr` field to `unsafe_ptr`, just to hammer the point home that this field has safety invariants at its use-sites?

src/util.rs
```diff
@@ -77,6 +350,61 @@ pub(crate) const fn _round_down_to_next_multiple_of_alignment(
...
+    // TODO(#67): Once our MSRV is 1.63, remove this.
+    pub(crate) trait NonNullSliceExt<T> {
+        fn len(&self) -> usize;
+    }
+
+    impl<T> NonNullSliceExt<T> for NonNull<[T]> {
+        #[inline(always)]
+        fn len(&self) -> usize {
+            #[allow(clippy::as_conversions)]
+            let slc = self.as_ptr() as *const [()];
```
jswrenn, line 395: This is subtle. Could you cite the reference that fat pointer conversions preserve length?

joshlf: So fun discovery; there is no such guarantee as of today. I've put up a PR to add one: https://github.com/rust-lang/reference/pull/1417. We'll need to block this PR until that lands.

src/macros.rs
```diff
@@ -204,11 +204,21 @@ macro_rules! impl_known_layout {
...
-        }
+        const _: () = {
+            use core::ptr::NonNull;
+
+            impl<$(const $constvar : $constty,)? $($tyvar $(: ?$optbound)?)?> sealed::KnownLayoutSealed for $ty {}
+            // SAFETY: Delegates safety to `DstLayout::for_type`.
+            unsafe impl<$(const $constvar : $constty,)? $($tyvar $(: ?$optbound)?)?> KnownLayout for $ty {
+                const LAYOUT: DstLayout = DstLayout::for_type::<$ty>();
+
+                // SAFETY: `.cast` preserves address and provenance.
```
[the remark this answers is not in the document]
joshlf: We don't; I've added a TODO.

src/util.rs
joshlf: Done.

src/util.rs
joshlf: Another option would be to split this PR so that all of the bits that transitively rely on this method are in a follow-up PR. That would include the `try_from` method, so it'd be a good chunk of this PR that we'd need to save for a follow-up.

joshlf: Discussed offline. I've split this and transitive dependencies into https://github.com/google/zerocopy/pull/498.

src/util.rs
joshlf: I'm worried that if we do that, then for consistency's sake we'll need to do it on almost every field in the crate. I feel like if you're working on this crate, you've gotta already be aware that most things have safety invariants (and certainly code reviewers know that), so the risk of missing an invariant is low. Another angle is that the sort of person who wouldn't know to check whether `ptr` has a safety invariant is the same sort of person who wouldn't get the hint from the name `unsafe_ptr`.

joshlf: Yeah fair enough; I've moved it into a submodule.

jswrenn (approved):

PR #430, opened 2023-09-29
Rename `FromZeroes` to `FromZeros`

jswrenn:
Pretend you didn't see this.

`FromZeroes` is retained as a deprecated, `doc(hidden)` re-export.

Additionally, for consistency, `LayoutVerified` is changed from a type alias to `pub use ... as ...` reexport.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -24,8 +24,8 @@
 //! # Overview
 //!
 //! Zerocopy provides four core marker traits, each of which can be derived
-//! (e.g., `#[derive(FromZeroes)]`):
-//! - [`FromZeroes`] indicates that a sequence of zero bytes represents a valid
+//! (e.g., `#[derive(FromZeros)]`):
+//! - [`FromZeros`] indicates that a sequence of zero bytes represents a valid
 //!   instance of a type
 //! - [`FromBytes`] indicates that a type may safely be converted from an
 //!   arbitrary byte sequence
@@ -76,7 +76,7 @@
 //!   ```
 //!
 //! - **`simd`**   
-//!   When the `simd` feature is enabled, `FromZeroes`, `FromBytes`, and
+//!   When the `simd` feature is enabled, `FromZeros`, `FromBytes`, and
 //!   `AsBytes` impls are emitted for all stable SIMD types which exist on the
 //!   target platform. Note that the layout of SIMD types is not yet stabilized,
 //!   so these impls may be removed in the future if layout changes make them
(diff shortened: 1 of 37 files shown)


src/lib.rs
```diff
@@ -1725,7 +1730,7 @@ pub struct Ref<B, T: ?Sized>(B, PhantomData<T>);
 /// Deprecated: prefer [`Ref`] instead.
 #[deprecated(since = "0.7.0", note = "LayoutVerified has been renamed to Ref")]
 #[doc(hidden)]
-pub type LayoutVerified<B, T> = Ref<B, T>;
+pub use Ref as LayoutVerified;
```
jswrenn, line 1733: ^

src/lib.rs
```diff
@@ -858,6 +858,11 @@ pub unsafe trait FromZeroes {
     }
 }
 
+/// Deprecated: prefer [`FromZeros`] instead.
+#[deprecated(since = "0.7.6", note = "FromZeroes has been renamed to FromZeros")]
+#[doc(hidden)]
+pub use FromZeros as FromZeroes;
```
jswrenn, line 864: ^

src/lib.rs
```diff
@@ -858,6 +858,11 @@ pub unsafe trait FromZeroes {
     }
 }
 
+/// Deprecated: prefer [`FromZeros`] instead.
+#[deprecated(since = "0.7.6", note = "FromZeroes has been renamed to FromZeros")]
```
joshlf, line 862: Deprecating is a breaking change for users who deny warnings in their dependencies; we need to wait for 0.8 to mark this as deprecated.

src/lib.rs
```diff
@@ -858,6 +858,10 @@ pub unsafe trait FromZeroes {
     }
 }
 
+/// To-be-deprecated: prefer [`FromZeros`] instead.
+#[doc(hidden)]
+pub use FromZeros as Fromzeros;
```
joshlf, line 863: s/Fromzeroes/FromZeroes

joshlf (changes requested):

joshlf (changes requested):
There are some files with the `zeroes` spelling in the name that need to be changed, but otherwise LGTM!

joshlf (changes requested):
Also, can you update the commit message?
- Capitalize "rename" in the summary line
- Mention that "zeroes" is a valid spelling, but that "zeros" is more commonly accepted, possibly citing [this StackExchange answer](https://english.stackexchange.com/questions/3824/what-is-the-plural-form-of-zero/3825#3825)
- Add `Closes #431` at the end

joshlf (changes requested):
In the commit message:

> Although both spelling are acceptable

Should be "spellings" plural.

joshlf (approved):

src/lib.rs
```diff
@@ -1617,6 +1617,11 @@ pub unsafe trait FromZeroes {
     }
 }
 
+/// Deprecated: prefer [`FromZeros`] instead.
+#[deprecated(since = "0.8.0", note = "`FromZeroes` was renamed to `FromZeros`")]
+#[doc(hidden)]
+pub use FromZeros as FromZeroes;
```
jswrenn, line 1623: This is going to use-rename both the trait *and* the proc macro (if it's in scope from the `derive`).

Issue #429, opened 2023-09-29
Annotate all `unsafe` blocks with a safety comment and prevent regressions


joshlf:
As part of #61, we need to make sure that all `unsafe` code is proven to be sound. Currently, all `unsafe` code is either documented with a safety comment (`// SAFETY: ...`), or is marked with a TODO that references this issue or #61. Our goal is to reach 100% safety comment coverage, and to not regress once we've reached 100%. To that end, we enforce the [`clippy::undocumented_unsafe_blocks`](https://rust-lang.github.io/rust-clippy/master/index.html#/undocumented_unsafe_blocks) lint to prevent regressions.

*A note on linting: It'd be nice to be able to replace the top-level `#![deny(clippy::undocumented_unsafe_blocks)]` with a `forbid` once all TODOs are burned down, but unfortunately our `safety_comment!` macro relies on the ability to use `#[allow(clippy::undocumented_unsafe_blocks)]`, so we have to settle for a deny.*

In order to ensure that our soundness is forwards-compatible, safety comments must satisfy the following criteria:
- Safety comments must constitute a (possibly informal) proof that all of Rust's soundness rules are upheld
- Safety comments must not rely on any statements which do not appear in the stable versions of the [Rust reference](https://doc.rust-lang.org/reference/) or standard library documentation (ie, the docs for [core](https://doc.rust-lang.org/stable/core/), [alloc](https://doc.rust-lang.org/stable/alloc/), and [std](https://doc.rust-lang.org/stable/std/)); arguments which rely on text from the beta or nightly versions of these documents are not considered complete
- All statements from the reference or standard library documentation which are relied upon for soundness must be quoted in the safety comment. This ensures that there is no ambiguity as to what aspect of the text is being cited. This is especially important in cases where the text of these documents changes in the future. Such changes are of course required to be backwards-compatible, but may change the manner in which a particular guarantee is explained.

## Mentoring instructions

- [ ] Easy tasks
  - [ ] Find any safety comment which links to a version of the Rust reference or stdlib docs that uses a beta or nightly link, and replace it with a link to the stable version of the same docs
- [ ] Medium tasks
  - [ ] Find any safety comment which cites the reference or stdlib docs without quoting the text it is citing. Fix this by quoting any text on which the safety comment relies for soundness.
- [ ] Hard tasks
[...]

PR #433, opened 2023-09-29
Add TODOs related to #429 and quotes to safety comments.

samuelselleck:
While reading through some of the code I added some quotes and TODOs here and there related to #429.


src/lib.rs
```diff
@@ -1403,12 +1418,14 @@ safety_comment! {
...
     /// SAFETY:
+    /// TODO(#429): Add quotes
     /// `Wrapping<T>` is guaranteed by its docs [1] to have the same layout as
     /// `T`. Also, `Wrapping<T>` is `#[repr(transparent)]`, and has a single
     /// field, which is `pub`. Per the reference [2], this means that the
     /// `#[repr(transparent)]` attribute is "considered part of the public ABI".
     ///
-    /// [1] https://doc.rust-lang.org/nightly/core/num/struct.Wrapping.html#layout-1
+    ///  references nightly?:
+    /// [1] TODO(https://doc.rust-lang.org/nightly/core/num/struct.Wrapping.html#layout-1):
```
joshlf, line 1428:
```suggestion
    /// [1] TODO(https://doc.rust-lang.org/nightly/core/num/struct.Wrapping.html#layout-1):
    ///     Reference this documentation once it's available on stable.
```

joshlf (changes requested):
Awesome, thank you so much for these! This is exactly what we're looking for. A few minor style nits, but the content is good.

samuelselleck:
Thank you for your feedback! Don't worry about being picky, that's the kind of thing I'm looking to learn :)

I think I've addressed all of your comments, if there is anything else do let me know!

**Question**: The code still contains safety comments (starting with //SAFETY:) that are not wrapped in the safety_comment! macro. Is this intentional and if so what is the purpose of the differentiation?

(I had some git problems, but think they are fixed? (accidentally merged upstream changes through the github gui). If anything is off do let me know and I'll try to figure it out.)

joshlf:
> I think I've addressed all of your comments, if there is anything else do let me know.

Awesome, thanks! I'll try to get to these soon.

> **Question**: The code still contains safety comments (starting with //SAFETY:) that are not wrapped in the safety_comment! macro. Is this intentional and if so what is the purpose of the differentiation?

Yeah, that's intentional. Normally you'd just put a `// SAFETY:` comment on each `unsafe` block. The reason for the `safety_comment!` macro is that we often have a lot of `unsafe` trait impls whose soundness is based on identical or very similar arguments, and so we'd _like_ to be able to write a single `// SAFETY:` comment to annotate multiple such impls. Since there's no way to do this that is legible to Clippy, we use the `safety_comment!` macro instead, which under the hood just annotates each macro invocation with `#[allow(clippy::undocumented_unsafe_blocks)]`:

https://github.com/google/zerocopy/blob/9ce05f8fa5c4761020c082ce18f071007125902f/src/macros.rs#L24-L29

> (I had some git problems, but think they are fixed? (accidentally merged upstream changes through the github gui). If anything is off do let me know and I'll try to figure it out.)

Seems fine to me 🤷‍♂️  We have this repo configured to squash all commits into a single one when a PR is merged, so it probably doesn't make much of a difference (perhaps with the exception that the resulting commit message will include a summary of each commit that was included in the PR). Personally my flow is to always just edit the commit and then force push, but whatever flow works for you is fine.

samuelselleck:
Thanks for the explanation of the safety_comment macro, that makes sense! I'll adopt the same push -f method for edits in the future :)

src/lib.rs
```diff
@@ -655,6 +655,7 @@ safety_comment! {
     /// `str` and `ManuallyDrop<[T]>` have the same representations as `[u8]`
     /// and `[T]` repsectively. `str` has different bit validity than `[u8]`,
     /// but that doesn't affect the soundness of this impl.
+    /// TODO(#429): Add references to docs and quotes.
```
joshlf, line 658:
```suggestion
    ///
    /// TODO(#429): Add references to docs and quotes.
```

src/lib.rs
```diff
@@ -762,6 +763,7 @@ pub unsafe trait FromZeroes {
         //   as required by `u8`.
         // - Since `Self: FromZeroes`, the all-zeroes instance is a valid
         //   instance of `Self.`
+        // TODO(#429): Add references to docs and quotes.
```
joshlf, line 766:
```suggestion
        //
        // TODO(#429): Add references to docs and quotes.
```

src/lib.rs
```diff
@@ -1161,6 +1163,7 @@ pub unsafe trait AsBytes {
         // - The total size of the resulting slice is no larger than
         //   `isize::MAX` because no allocation produced by safe code can be
         //   larger than `isize::MAX`.
+        // TODO(#429): Add references to docs and quotes.
```
joshlf, line 1166:
```suggestion
        //
        // TODO(#429): Add references to docs and quotes.
```

src/lib.rs
```diff
@@ -1196,6 +1199,7 @@ pub unsafe trait AsBytes {
...
```
joshlf, line 1202:
```suggestion
        //
        // TODO(#429): Add references to docs and quotes.
```

src/lib.rs
```diff
@@ -1332,11 +1340,12 @@ safety_comment! {
 
 safety_comment! {
     /// SAFETY:
-    /// - `FromZeroes`: Per the reference [1], 0x00 is a valid bit pattern for
-    ///   `bool`.
-    /// - `AsBytes`: Per the reference [1], `bool` always has a size of 1 with
-    ///   valid bit patterns 0x01 and 0x00, so the only byte of the bool is
-    ///   always initialized
+    /// - `FromZeroes`: Valid since "The value false has the bit pattern
```
joshlf, line 1343:
```suggestion
    /// - `FromZeroes`: Valid since "[t]he value false has the bit pattern
```

src/lib.rs
```diff
@@ -1332,11 +1340,12 @@ safety_comment! {
...
+    ///   0x00" [1].
+    /// - `AsBytes`: Since "the boolean type has a size and alignment of 1 each"
+    ///   and "The value false has the bit pattern 0x00 and the value true has
```
joshlf, line 1346:
```suggestion
    ///   and "[t]he value false has the bit pattern 0x00 and the value true has
```

src/lib.rs
```diff
@@ -1346,24 +1355,28 @@ safety_comment! {
 }
 safety_comment! {
     /// SAFETY:
-    /// - `FromZeroes`: Per the reference [1], 0x0000 is a valid bit pattern for
-    ///   `char`.
-    /// - `AsBytes`: `char` is represented as a 32-bit unsigned word (`u32`)
-    ///   [1], which is `AsBytes`. Note that unlike `u32`, not all bit patterns
-    ///   are valid for `char`.
+    /// - `FromZeroes`: Per reference [1], "A value of type char is a Unicode
```
joshlf, line 1358:
```suggestion
    /// - `FromZeroes`: Per reference [1], "[a] value of type char is a Unicode
```

src/lib.rs
```diff
@@ -1511,6 +1535,7 @@ safety_comment! {
     ///   code can only ever access a `ManuallyDrop` with all initialized bytes.
     /// - `Unaligned`: `ManuallyDrop` has the same layout (and thus alignment)
     ///   as `T`, and `T: Unaligned` guarantees that that alignment is 1.
+    /// TODO(#429): Add references to docs and quotes.
```
joshlf, line 1538:
```suggestion
    ///
    /// TODO(#429): Add references to docs and quotes.
```

joshlf (changes requested):
Okay, this looks great! Just a few tiny nits, but as soon as these are resolved, I'll approve it 🙂

samuelselleck (commented):
Added your changes! I'll make sure to keep spacing and use [] when quoting in the middle of a sentence. merged all commits into one now as well.

src/lib.rs
```diff
@@ -656,12 +656,14 @@ safety_comment! {
     /// `[u8]` and `[T]` repsectively. `str` has different bit validity than
     /// `[u8]`, but that doesn't affect the soundness of this impl.
     ///
-    /// [1] Per https://doc.rust-lang.org/nightly/core/mem/struct.ManuallyDrop.html:
-    ///
     ///   `ManuallyDrop<T>` is guaranteed to have the same layout and bit
     ///   validity as `T`
     ///
-    /// TODO(#429): Once this text (added in
+    /// [1] Per https://doc.rust-lang.org/nightly/core/mem/struct.ManuallyDrop.html:
```
joshlf, line 662: Looks like these got swapped by accident?

joshlf (changes requested):

joshlf (approved):
Thanks again for all your work on this! I cleaned a few things up (deleting spurious extra newlines and such) and it's ready to merge.

samuelselleck:
Apologies for not getting back to you sooner, thanks for taking the time!

joshlf:
> Apologies for not getting back to you sooner, thanks for taking the time!

No worries at all! I was just going back and cleaning up old PRs and I figured it was silly to ask you to do a few obvious edits when I was doing it already 🙂

PR #442, opened 2023-10-02
Make sure generate cache without needs

### LINKED ISSUE #340: [CI] Deduplicate `actions/cache` directives
We have a `generate_cache` job in CI which pre-populates a cache that is re-used by other jobs:

https://github.com/google/zerocopy/blob/3bb9a54ff458e104f7eb24beda7f8fb09794e1a5/.github/workflows/ci.yml#L410-L442

In each dependent job, we do the following:
- Use `needs: generate_cache`
- Use the following `actions/cache` action configuration:
  https://github.com/google/zerocopy/blob/3bb9a54ff458e104f7eb24beda7f8fb09794e1a5/.github/workflows/ci.yml#L77-L81

As of this writing, we copy that exact text five times. It'd be great if we could reduce the duplication.

## Mentoring instructions

As [proposed here](https://github.com/google/zerocopy/issues/340#issuecomment-1747242751) by @mgeisler:
- [ ] Create a separate `cache.yml` file in `.github/workflows`
- [ ] Populate it with a job that performs these steps:
  
  https://github.com/google/zerocopy/blob/3bb9a54ff458e104f7eb24beda7f8fb09794e1a5/.github/workflows/ci.yml#L77-L81
- [ ] Replace uses of that pattern in `ci.yml` with an invocation of that file, ie:
  ```yaml
  - name: Populate cache
    uses: `.github/workflows/cache.yml`
  ```

<details>
<summary>Old/obsolete mentoring instructions</summary>

As discussed in [this StackOverflow comment](https://stackoverflow.com/questions/77053736/give-alias-to-github-actions-uses-step?noredirect=1#comment135837054_77053736), 

> You can extract your own actions: [docs.github.com/en/actions/creating-actions](https://docs.github.com/en/actions/creating-actions). YAML's own support for this isn't available in GHA though: [github.com/actions/runner/issues/1182](https://github.com/actions/runner/issues/1182).

The task is to figure out whether there's a way to use that technique to define a new action in `ci.yml` that we can re-use in other steps. If there's an easy, simple syntax for this, we should do it. If it would require a lot of complexity, then it's not worth it.

</details>

zoo868e:
Fixes #340 

Use the `workflow_run` event to ensure that jobs are triggered only when the cache already exists. The new `Generate Cache Workflow` is triggered by:

1. A pull request (PR).
2. A branch push.
3. A merge operation within a group.

Once the `Generate Cache Workflow` is completed, trigger the `Build & Tests` workflow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,12 +1,10 @@
 name: Build & Tests
 
 on:
-  pull_request:
-  push:
-    branches:
-      - main
-      - v0.6.x
-  merge_group:
+  workflow_run:
+    workflows: [Generate Cache Workflow]
+    types:
+      - completed
 
 permissions: read-all
 
@@ -27,7 +25,6 @@ jobs:
     # first, most jobs would duplicate the work of downloading crates from
     # the internet. Populating the cache first ensures that this work only
     # happens once.
-    needs: generate_cache
 
     strategy:
       # By default, this is set to `true`, which means that a single CI job
@@ -230,7 +227,6 @@ jobs:
   kani:
     runs-on: ubuntu-latest
     name: Run tests under Kani
-    needs: generate_cache
     steps:
       - uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608 # v4.1.0
       - name: Configure ZC_TOOLCHAIN environment variable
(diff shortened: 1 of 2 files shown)


jswrenn:
It appears this didn't actually trigger the Build & Tests workflow, unfortunately.

PR #446, opened 2023-10-03
Roll stable and nightly toolchains separately

joshlf:
This should allow the stable toolchain to roll successfully sooner after it's released. Previously, both toolchains were rolled together, which meant that when the nightly roll failed (frequent thanks to most nightlies missing at least one target architecture we use in CI), the stable roll would fail too.


joshlf:
Tested this by first pushing a version that is triggered on pull requests. That run generated these PRs for [stable](https://github.com/google/zerocopy/pull/447) and [nightly](https://github.com/google/zerocopy/pull/448).

jswrenn (approved):

PR #457, opened 2023-10-04
[ci] invert the condition on all-jobs-succeeded

jswrenn:
Previously, a failing job could cause all-jobs-succeeded to be skipped, and branch protections interpreted a skipped job as being vacuously successful. This failure mode caused #454 to be merged to main, despite failing many CI checks.

This PR inverts the `if` condition of this job: It only runs on failure, and upon failure, it fails.


joshlf (approved):

joshlf:
Looks like `success()` isn't available in the context it's used here. I put up a slight modification of this idea in https://github.com/google/zerocopy/pull/460.

PR #472, opened 2023-10-06
[ci][kani] Use kani-github-action

### LINKED ISSUE #423: Use `kani-github-action` for Kani in CI.
Replace our hand-rolled Kani CI job:
https://github.com/google/zerocopy/blob/3bb9a54ff458e104f7eb24beda7f8fb09794e1a5/.github/workflows/ci.yml#L229-L250
...with the official Kani github action: https://model-checking.github.io/kani/install-github-ci.html

Agnik7:
Removes the hand-rolled Kani CI job and replaces it with the official `kani-github-action`.

Resolves #423


Agnik7:
@joshlf requesting your review in this PR. Apologies for any inconvenience caused

.github/workflows/ci.yml
```diff
@@ -241,27 +241,20 @@ jobs:
...
-          cargo kani setup
-          RUSTFLAGS="$ZC_NIGHTLY_RUSTFLAGS" ./cargo.sh +nightly kani                   \
-            --package zerocopy --all-features --output-format terse --randomize-layout \
-            --memory-safety-checks --overflow-checks --undefined-function-checks       \
-            --unwinding-checks
+    on:
+      pull_request:
+      push:
+    jobs:
+      run-kani:
```
joshlf, line 248: IIUC, you should be able to mirror the structure of jobs like `check_fmt`, where the top-level keys would look something like this:

```yaml
  kani:
    runs-on: ubuntu-latest
    name: Run tests under Kani
    steps:
```

In other words, you don't need a separate `jobs` key; you can just have the `steps` directly at the top level.

.github/workflows/ci.yml
```diff
@@ -241,27 +241,20 @@ jobs:
...
+        runs-on: ubuntu-latest
+        steps:
+          - name: 'Checkout your code.'
+            uses: actions/checkout@v3
```
joshlf, line 252: It's probably overkill to name this step, and you can also pin to a specific version of `checkout` like we do in other jobs:

```yaml
- uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608 # v4.1.0
```

.github/workflows/ci.yml
```diff
@@ -241,27 +241,20 @@ jobs:
...
+    
+          - name: 'Run Kani on your code.'
```
joshlf, line 254: I'd name the whole job rather than this step (see my first review comment), so you can just remove the `name:` directive here.

.github/workflows/ci.yml
```diff
@@ -241,27 +241,20 @@ jobs:
...
+            uses: model-checking/kani-github-action@v4.1
```
joshlf, line 255: Maybe I'm missing something, but isn't the latest version 1.0? I'm going based off of this page: https://github.com/marketplace/actions/kani-rust-verifier

joshlf (changes requested):
Thanks for putting this up! I think we can simplify this some, but this is definitely on the right track.

.github/workflows/ci.yml
Agnik7: Oh alright, I am fixing this right now.

.github/workflows/ci.yml
Agnik7: Alright, fixing this right now

.github/workflows/ci.yml
Agnik7: Changing this right away

.github/workflows/ci.yml
Agnik7: Alright, doing it right now

Agnik7:
@joshlf I have made the changes. Kindly review.

Agnik7:
@joshlf requesting for a follow up. Apologies for any inconvenience caused.

.github/workflows/ci.yml
```diff
@@ -241,27 +241,17 @@ jobs:
...
-          toolchain: ${{ env.ZC_TOOLCHAIN }}
-      - run: |
-          set -eo pipefail
-          cargo install --locked kani-verifier
-          cargo kani setup
-          RUSTFLAGS="$ZC_NIGHTLY_RUSTFLAGS" ./cargo.sh +nightly kani                   \
-            --package zerocopy --all-features --output-format terse --randomize-layout \
-            --memory-safety-checks --overflow-checks --undefined-function-checks       \
-            --unwinding-checks
+          args: "--package zerocopy --all-features --output-format=terse --randomize-layout --memory-safety-checks --overflow-checks --undefined-function-checks --unwinding-checks"
```
joshlf, line 254: A few more changes, and then this should be good!
- Can you move the `name` field to the top level?
- Can you use the current name for this job ("Run tests under Kani")?
- Can you remove the `on:` directive?
- Can you move the `runs-on:` directive to the top level?

In total, with those changes, it should look like this:

```yaml
  kani:
    runs-on: ubuntu-latest
    name: Run tests under Kani
    steps:
      - uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608 # v4.1.0
      - uses: model-checking/kani-github-action@v1.0
        with:
          args: "--package zerocopy --all-features --output-format=terse --randomize-layout --memory-safety-checks --overflow-checks --undefined-function-checks --unwinding-checks"
```

joshlf (changes requested):

Agnik7:
@joshlf I have made the changes. Kindly review

Agnik7:
@joshlf I have already rebased the branch, so there should be no conflict.

joshlf:
> @joshlf I have already rebased the branch, so there should be no conflict.

Hmm interesting. In that case, I have no idea why [this job](https://github.com/google/zerocopy/actions/runs/6460038768/job/17536959163?pr=472) is failing. I'm going to try rebasing through the UI again to see if that magically fixes things 🤷‍♂️

Agnik7:
@joshlf the jobs have passed!! Thanks a lot for your help and guidance

joshlf (approved):
Okay, looks like it was a bug on `main` that your PR just exposed. I've fixed the bug and rebased this PR on that fix, and now everything works!

joshlf:
Thanks again, @Agnik7!

Agnik7:
Thanks a lot @joshlf for the opportunity and the guidance you provided!!

PR #484, opened 2023-10-10
document zerocopy's relationship to Project Safe Transmute

### LINKED ISSUE #480: Document relationship to Project Safe Transmute
@briansmith, author of *ring*, [wrote the following](https://github.com/google/zerocopy/issues/98#issuecomment-1748140023):

> > These users are extremely wary of taking external dependencies, and only take dependencies when they absolutely need to or when they have a high degree of trust in an external software artifact.
> 
> This describes me well. I considered multiple times using zerocopy and extending zerocopy. Pulling it out of Fuchsia and making accessible here on GitHub was a big step towards that. But then the formation of the "safe transmute" working group in the rust-lang organization made it seem like it would be just better to wait until the most important functionality is in libstd and guaranteed by the language/lib spec more explicitly.
> 
> [...]
> 
> Ultimately, I think people just would rather have as much of the functinoality w.r.t. transmuting and casting in the standard library as possible. I think if this is seen as a prototype that helps that happen, more people will be eager to adopt it.

It is problematic for both Project Safe Transmute and crates like zerocopy that the announcement of Project Safe Transmute has had a chilling effect on the adoption of safer abstractions over transmutation. Project Safe Transmute relies on the adoption of these third-party crates to inform its design process. Moreover, it is unlikely that the work of Project Safe Transmute will supplant the need for these third-party crates. 

The current plan of record for Project Safe Transmute will produce a low-level feature which supports the soundness analysis currently implemented by our custom derives. Eventually, crates like zerocopy will be able to transparently replace their bespoke soundness analyses with the one provided by Project Safe Transmute.

However, Project Safe Transmute won't replace most of the higher-level abstractions in zerocopy itself. For the foreseeable future, crates like zerocopy will still be required in order to provide these higher-level abstractions on top of the building block provided by Project Safe Transmute. For many higher-level use cases, Project Safe Transmute will provide the foundation on which users can prove the soundness of their `unsafe` code, but it will not obviate the need to write `unsafe` code itself.

Our documentation should clearly describe the relationship of zerocopy to Project Safe Transmute, noting:
1. the different abstraction levels of zerocopy and Project Safe Transmute
2. how the progression of Project Safe Transmute will affect zerocopy (see #481)

jswrenn:
Fixes #480

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,29 @@ memory model, and *any future memory model*. We ensure this by:
 [Miri]: https://github.com/rust-lang/miri
 [Kani]: https://github.com/model-checking/kani
 
+## Relationship to Project Safe Transmute
+
+[Project Safe Transmute] is an official initiative of the Rust Project to
+develop language-level support for safer transmutation. The Project consults
+with crates like zerocopy to identify aspects of safer transmutation that
+would benefit from compiler support, and has developed an [experimental,
+compiler-supported analysis][mcp-transmutability] which determines whether,
+for a given type, any value of that type may be soundly transmuted into
+another type. Once this functionality is sufficiently mature, zerocopy
+intends to replace its internal transmutability analysis (implemented by our
+custom derives) with the compiler-supported one. This change will likely be
+an implementation detail that is invisible to zerocopy's users.
+
+Project Safe Transmute will not replace the need for most of zerocopy's
+higher-level abstractions. The experimental compiler analysis is a tool for
+checking the soundness of `unsafe` code, not a tool to avoid writing
+`unsafe` code altogether. For the foreseeable future, crates like zerocopy
+will still be required in order to provide higher-level abstractions on top
+of the building block provided by Project Safe Transmute.
+
+[Project Safe Transmute]: https://rust-lang.github.io/rfcs/2835-project-safe-transmute.html
+[mcp-transmutability]: https://github.com/rust-lang/compiler-team/issues/411
+
 ## Disclaimer
 
 Disclaimer: Zerocopy is not an officially supported Google product.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -105,6 +105,29 @@
 //!
 //! [Miri]: https://github.com/rust-lang/miri
 //! [Kani]: https://github.com/model-checking/kani
+//!
+//! # Relationship to Project Safe Transmute
+//!
+//! [Project Safe Transmute] is an official initiative of the Rust Project to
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -107,6 +107,29 @@ memory model, and *any future memory model*. We ensure this by:
...
 [Kani]: https://github.com/model-checking/kani
 
+## Relationship to Project Safe Transmute
+
+[Project Safe Transmute] is an official initiative of the Rust Project to
+develop language-level support for safer transmutation. The Project consults
+with crates like zerocopy to identify aspects of safer transmutation that
+would benefit from compiler support, and has developed an [experimental,
+compiler-supported analysis][mcp-transmutability] of when a value of one
+type is soundly transmutable into another. Once this functionality is
```
joshlf, line 117: Maybe:

> ...compiler-supported analysis which determines whether, for a given type, any value of that type may be soundly transmuted into another type.

README.md
```diff
@@ -107,6 +107,29 @@ memory model, and *any future memory model*. We ensure this by:
...
+sufficiently mature, zerocopy intends to replace its internal
+transmutability analysis with the compiler-supported one. This change will
```
joshlf, line 119: Maybe:

> ...transmutability analysis (implemented in zerocopy-derive)

Or:

> ...transmutability analysis (implemented by our custom derives)

...or similar.

README.md
```diff
@@ -107,6 +107,29 @@ memory model, and *any future memory model*. We ensure this by:
...
+likely be an implementation detail that is invisible to zerocopy's users.
+
+However, Project Safe Transmute will not replace the need for most of
```
joshlf, line 122: Given the last sentence of the previous paragraph, it probably makes sense to remove "However".

README.md
```diff
@@ -107,6 +107,29 @@ memory model, and *any future memory model*. We ensure this by:
...
+zerocopy's higher-level abstractions over transmutability. The experimental
```
joshlf, line 123: Maybe remove "...over transmutability".

README.md
```diff
@@ -107,6 +107,29 @@ memory model, and *any future memory model*. We ensure this by:
...
+compiler analysis is a tool for checking the soundness of `unsafe` code—not
```
joshlf, line 124: Might make a sense for the separator to be a comma rather than an emdash.

joshlf (changes requested):

joshlf (approved):

PR #498, opened 2023-10-12
Add `Ptr::try_cast_into` method

joshlf:
`try_cast_into` attempts to cast a `Ptr<[u8]>` into a `Ptr<U>` where `U: ?Sized + KnownLayout`, and will be built upon in future commits as a building block of `TryFromBytes`.

Because `try_cast_into` performs a runtime check to validate alignment, this requires disabling Miri's "symbolic alignment check" feature. While it would be possible to run both with and without that feature in order to still test other code using symbolic alignment checking, I don't think that the benefit is worth a) the complexity of passing the information necessary for certain tests to not run under symbolic alignment checking and, b) the extra CI time to run Miri tests twice.

Makes progress on #29


joshlf:
Implemented in #406

PR #499, opened 2023-10-12
[ci] Don't test on MIPS

joshlf:
MIPS targets have been consistently unavailable recently, which has blocked our ability to roll the nightly toolchain.


jswrenn (approved):
Alternatively, we could do `-Z build-std`, but if there isn't a clear reason why we should be running CI on MIPS (e.g., it has uniquely interesting properties), then cutting it out is fine.

joshlf:
> Alternatively, we could do `-Z build-std`, but if there isn't a clear reason why we should be running CI on MIPS (e.g., it has uniquely interesting properties), then cutting it out is fine.

Yeah it was just a "let's test on as many targets as possible" thing; no reason to test on mips in particular.

Issue #288, opened 2023-08-25
Improve safety documentation for unsafe traits


joshlf:
Currently, the safety documentation for `FromZeroes`, `FromBytes`, and `AsBytes` specifies a) what unsafe code may assume about types which implement those traits and, b) what must hold in order for our custom derives to accept a type as soundly implementing those traits (although this is misleadingly presented as the conditions that must hold in order for any impl to be sound).

We should change this layout to the following:
- What unsafe code is allowed to assume about the trait (this is the raison d'etre for the trait)
- What properties a type must satisfy in order to allow the preceding to be sound. Unlike what's currently in the documentation, this is more abstract and doesn't refer to specific types. For example, for `AsBytes`, this might be:
  - No bytes that can be uninitialized
  - No `UnsafeCell`
- Rules for specific types (structs, enums, unions, etc) that, if met, satisfy the preceding properties
- Rules for specific types that are sufficient for our custom derive to recognize the type as soundly implementing the trait

We may want to combine the last two if it makes sense (e.g., inline, call out specific cases where the derive is less powerful).

It may also be useful to call out certain notable types which don't satisfy the requirements (e.g., `MaybeUninit` might be uninitialized).

Finally, this would be a good opportunity to consider ways that we could make the derives smarter.

This issue was inspired by the following PR comments by @kupiakos:
- https://github.com/google/zerocopy/pull/283#discussion_r1305974101
- https://github.com/google/zerocopy/pull/283#discussion_r1305983622
- https://github.com/google/zerocopy/pull/283#discussion_r1305984043
- https://github.com/google/zerocopy/pull/283#discussion_r1305985093

joshlf:
Done, mostly by @jswrenn.

Issue #32, opened 2022-10-07
Documentation should describe the broad view of zerocopy's use cases


joshlf:
Our documentation currently focuses on the parsing and serialization use cases that zerocopy was originally designed for. However, we've long since realized that the techniques enabled by zerocopy are general, and not constrained to parsing and serialization. Our documentation should instead describe zerocopy's purpose as enabling general-purpose safe memory access and memory modification. (We should also figure out exactly how to express that linguistically - it's not clear that the phrasing of the previous sentence is the phrasing we want to use.)

tim-seoss:
Whilst this is a big task, a couple of examples[^1] would probably go a long way to helping newcomers to zerocopy.

At its very simplest, it could take the form of a series of links to other repositories, each of which shows usage zerocopy in a common/canonical way.

If you could point me to an example or any notes or other resources for packet serialisation I'd be happy to try and knock something up and crate a PR.

[^1]:  OK, there are already a couple of examples, but they are two slight variants of the same code - one in `src/byteorder.rs` at line 33 and the other in `src/lib.rs` at line 1482,  and the example is UDP packet parsing/deserialisation - there are no examples at all of the sort of overview of code flow which is suitable for serialisation e.g. packet assembly example with header+body+footer

joshlf:
Hey @tim-seoss, still don't have a ton of time to devote to mentoring this, but I realized that I could at least point you to some locations that have a lot of zerocopy uses you could take a look at. If you're interested in writing these up, I'd suggest at least pinging this thread first to confirm that a given example uses a pattern/style we'd like to encourage.

These are both in Fuchsia; the latter is specifically a packet parsing/serializing library, so it's got a lot of more gnarly uses of zerocopy. I'd just use the search function to search for "zerocopy" or the names of individual types, and that'll be a good starting point.
- https://cs.opensource.google/fuchsia/fuchsia/+/main:src/
- https://cs.opensource.google/fuchsia/fuchsia/+/main:src/connectivity/lib/packet-formats/

PR #505, opened 2023-10-13
improve `FromZeroes` documentation

jswrenn:
This PR does essentially three things:
1. Provides examples of `#[derive(FromZeroes)]` being used
2. Provides examples of using many of `FromZeroes`'s methods.
3. Separates the documentation of the safety conditions of `FromZeroes`, from the documentation of how `derive(FromZeroes)` performs its analysis.

If the model of this PR looks good, I'll also apply it to `FromBytes` and `AsBytes`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -699,44 +699,54 @@ safety_comment! {
     unsafe_impl_known_layout!(T: ?Sized + KnownLayout => #[repr(T)] ManuallyDrop<T>);
 }
 
-/// Types for which a sequence of bytes all set to zero represents a valid
-/// instance of the type.
+/// Analyzes whether a type is [`FromZeroes`].
 ///
-/// WARNING: Do not implement this trait yourself! Instead, use
-/// `#[derive(FromZeroes)]` (requires the `derive` Cargo feature).
+/// This derive analyzes, at compile time, whether the annotated type satisfies
+/// the [safety conditions] of `FromZeroes` and implements `FromZeroes` if it is
+/// sound to do so. This derive can be applied to structs, enums, and unions;
+/// e.g.:
 ///
-/// Any memory region of the appropriate length which is guaranteed to contain
-/// only zero bytes can be viewed as any `FromZeroes` type with no runtime
-/// overhead. This is useful whenever memory is known to be in a zeroed state,
-/// such memory returned from some allocation routines.
+/// ```
+/// # use zerocopy_derive::FromZeroes;
+/// #[derive(FromZeroes)]
+/// struct MyStruct {
+/// # /*
+///     ...
+/// # */
+/// }
 ///
-/// `FromZeroes` is ignorant of byte order. For byte order-aware types, see the
-/// [`byteorder`] module.
+/// #[derive(FromZeroes)]
+/// #[repr(u8)]
+/// enum MyEnum {
+/// #   Variant0,
+/// # /*
+///     ...
+/// # */
+/// }
 ///
-/// # Safety
+/// #[derive(FromZeroes)]
+/// union MyUnion {
+/// #   variant: u8,
+/// # /*
+///     ...
+/// # */
+/// }
+/// ```
 ///
-/// *This section describes what is required in order for `T: FromZeroes`, and
-/// what unsafe code may assume of such types. `#[derive(FromZeroes)]` only
-/// permits types which satisfy these requirements. If you don't plan on
-/// implementing `FromZeroes` manually, and you don't plan on writing unsafe
-/// code that operates on `FromZeroes` types, then you don't need to read this
-/// section.*
+/// [safety conditions]: trait@FromZeroes#safety
 ///
-/// If `T: FromZeroes`, then unsafe code may assume that:
-/// - It is sound to treat any initialized sequence of zero bytes of length
-///   `size_of::<T>()` as a `T`.
-/// - Given `b: &[u8]` where `b.len() == size_of::<T>()`, `b` is aligned to
-///   `align_of::<T>()`, and `b` contains only zero bytes, it is sound to
-///   construct a `t: &T` at the same address as `b`, and it is sound for both
-///   `b` and `t` to be live at the same time.
+/// # Analysis
 ///
-/// If a type is marked as `FromZeroes` which violates this contract, it may
-/// cause undefined behavior.
+/// *This section describes, roughly, the analysis performed by this derive to
+/// determine whether it is sound to implement `FromZeroes` for a given type.
+/// Unless you are modifying the implementation of this derive, or attempting to
+/// manually implement `FromZeroes` for a type yourself, you don't need to read
+/// this section.*
 ///
-/// If a type has the following properties, then it is sound to implement
+/// If a type has the following properties, then this derive can implement
 /// `FromZeroes` for that type:
-/// - If the type is a struct, all of its fields must satisfy the requirements
-///   to be `FromZeroes` (they do not actually have to be `FromZeroes`).
+///
+/// - If the type is a struct, all of its fields must be `FromZeroes`.
 /// - If the type is an enum, it must be C-like (meaning that all variants have
 ///   no fields) and it must have a variant with a discriminant of `0`. See [the
 ///   reference] for a description of how discriminant values are chosen.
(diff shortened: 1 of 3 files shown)


joshlf:
Awesome! I'll take a look when I get a chance.

In case it's useful to you, there are some ideas for ways to improve these docs: https://github.com/google/zerocopy/issues/288

src/lib.rs
```diff
@@ -659,42 +659,53 @@ safety_comment! {
...
-///   construct a `t: &T` at the same address as `b`, and it is sound for both
-///   `b` and `t` to be live at the same time.
+/// # Analysis
 ///
-/// If a type is marked as `FromZeroes` which violates this contract, it may
-/// cause undefined behavior.
+/// *This section describes, roughly, the analysis performed by this derive to
+/// determine whether it is sound to implement `FromZeroes` for a given type.
+/// Unless you are modifying the implementation of this derive, or attempting to
+/// manually-implement `FromZeroes` for a type yourself, you don't need to read
```
joshlf, line 703:
```suggestion
/// manually implement `FromZeroes` for a type yourself, you don't need to read
```

src/lib.rs
```diff
@@ -659,42 +659,53 @@ safety_comment! {
...
+/// this section.*
 ///
-/// If a type has the following properties, then it is sound to implement
+/// If a type has the following properties, then this derive can implement
 /// `FromZeroes` for that type:
+///
 /// - If the type is a struct, all of its fields must satisfy the requirements
 ///   to be `FromZeroes` (they do not actually have to be `FromZeroes`).
```
joshlf, line 710: Right now, the derive *does* require that field types be `FromZeroes`. If this is meant to document the behavior of the derive, then we might want to update this. It's also worth considering whether we should separately document the theoretical soundness requirements separately from documenting the (somewhat more conservative) behavior of the derive.

https://github.com/google/zerocopy/issues/288 contains some relevant ideas here.

src/lib.rs
```diff
@@ -707,19 +718,26 @@ safety_comment! {
...
 ///
-/// [the reference]: https://doc.rust-lang.org/reference/items/enumerations.html#custom-discriminant-values-for-fieldless-enumerations
-/// [`UnsafeCell`]: core::cell::UnsafeCell
+/// This analysis is subject to change. Unsafe code may *only* rely on the
+/// documented [safety conditions] of `FromZeroes`, and must *not* rely on the
+/// implementation details of this derive.
 ///
-/// # Rationale
+/// [the reference]:
+///     https://doc.rust-lang.org/reference/items/enumerations.html#custom-discriminant-values-for-fieldless-enumerations
```
joshlf, line 726:
```suggestion
/// [the reference]: https://doc.rust-lang.org/reference/items/enumerations.html#custom-discriminant-values-for-fieldless-enumerations
```

src/lib.rs
```diff
@@ -707,19 +718,26 @@ safety_comment! {
...
+/// that structs are marked with `#[repr(C)]`.
+///
 /// Per the [Rust reference](reference),
 ///
 /// > The representation of a type can change the padding between fields, but
 /// does not change the layout of the fields themselves.
 ///
-/// [reference]: https://doc.rust-lang.org/reference/type-layout.html#representations
+/// [reference]:
+///     https://doc.rust-lang.org/reference/type-layout.html#representations
```
joshlf, line 740:
```suggestion
/// [reference]: https://doc.rust-lang.org/reference/type-layout.html#representations
```

src/lib.rs
```diff
@@ -751,6 +846,34 @@ pub unsafe trait FromZeroes {
     /// Self::new_zeroed()`, it differs in that `zero` does not semantically
     /// drop the current value and replace it with a new one - it simply
     /// modifies the bytes of the existing value.
+    ///
```
joshlf, line 849:
```suggestion
    ///
    /// # Examples
    ///
```

src/lib.rs
```diff
@@ -751,6 +846,34 @@ pub unsafe trait FromZeroes {
...
+    /// ```
+    /// # mod derive { pub use zerocopy_derive::FromZeroes; }
+    /// # use zerocopy::FromZeroes;
+    /// # use derive::*;
+    /// #
+    /// # #[derive(FromZeroes)]
+    /// # #[repr(C)]
+    /// # struct PacketHeader {
```
joshlf, line 857: So long as we're using a network example, we might as well use a real one.

```suggestion
    /// # struct UdpHeader {
```

src/lib.rs
```diff
@@ -751,6 +846,34 @@ pub unsafe trait FromZeroes {
...
+    /// #     src_port: [u8; 2],
+    /// #     dst_port: [u8; 2],
+    /// #     length: [u8; 2],
+    /// #     checksum: [u8; 2],
+    /// # }
```
joshlf, line 862: Do we maybe want to include this in the example so that users see the definition of the type?

src/lib.rs
```diff
@@ -751,6 +846,34 @@ pub unsafe trait FromZeroes {
...
+    ///     length: 300u16.to_be_bytes(),
+    ///     checksum: 400u16.to_be_bytes(),
+    /// };
+    ///
+    /// header.zero();
+    ///
+    /// assert_eq!([0, 0], header.src_port);
+    /// assert_eq!([0, 0], header.dst_port);
+    /// assert_eq!([0, 0], header.length);
+    /// assert_eq!([0, 0], header.checksum);
```
joshlf, line 875: To be consistent with [stdlib examples](https://doc.rust-lang.org/stable/std/ptr/struct.NonNull.html#examples-3), swap the order here:

```suggestion
    /// assert_eq!(header.src_port, [0, 0]);
    /// assert_eq!(header.dst_port, [0, 0]);
    /// assert_eq!(header.length, [0, 0]);
    /// assert_eq!(header.checksum, [0, 0]);
```

src/lib.rs
```diff
@@ -766,6 +889,30 @@ pub unsafe trait FromZeroes {
...
+    ///     length: [u8; 2],
+    ///     checksum: [u8; 2],
+    /// }
+    ///
+    /// let header: PacketHeader = FromZeroes::new_zeroed();
+    ///
+    /// assert_eq!([0, 0], header.src_port);
+    /// assert_eq!([0, 0], header.dst_port);
+    /// assert_eq!([0, 0], header.length);
+    /// assert_eq!([0, 0], header.checksum);
```
joshlf, line 914: To be consistent with [stdlib examples](https://doc.rust-lang.org/stable/std/ptr/struct.NonNull.html#examples-3), swap the order here:

```suggestion
    /// assert_eq!(header.src_port, [0, 0]);
    /// assert_eq!(header.dst_port, [0, 0]);
    /// assert_eq!(header.length, [0, 0]);
    /// assert_eq!(header.checksum, [0, 0]);
```

joshlf (changes requested):
Love this! Some minor nits, but I think the general approach is the right one.

Also, can you reference #32 in the commit message body? For things that aren't closing an issue, I usually do something like:

```text
Makes progress on #32
```

src/lib.rs
jswrenn: The name is intentionally obfuscated, as our doc examples aren't meant to be a piecemeal guide for implementing a UDP (de)serializer — they're meant only to provide a gist of how to use the documented function. There are surely going to be complexities in parsing `UdpHeader`s that we might want to elide for these examples. For instance, in this case, I'm not using the byteorder types.

src/lib.rs
jswrenn: This example hides the definition itself for brevity, as it's apparent from its initialization:
```rust
let mut header = PacketHeader {
     src_port: 100u16.to_be_bytes(),
     dst_port: 200u16.to_be_bytes(),
     length: 300u16.to_be_bytes(),
     checksum: 400u16.to_be_bytes(),
};
```

In the other example, I *don't* hide the definition, because that example doesn't use the implicit constructor.

src/lib.rs
joshlf: Discussed offline; I'm on board with leaving this as-is.

src/lib.rs
```diff
@@ -751,6 +844,36 @@ pub unsafe trait FromZeroes {
     /// Self::new_zeroed()`, it differs in that `zero` does not semantically
     /// drop the current value and replace it with a new one - it simply
     /// modifies the bytes of the existing value.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # mod derive { pub use zerocopy_derive::FromZeroes; }
+    /// # use zerocopy::FromZeroes;
+    /// # use derive::*;
```
joshlf, line 853:
```suggestion
    /// # use zerocopy::FromZeroes;
    /// # use zerocopy_derive::*;
```

Same for other examples.

joshlf (changes requested):

src/lib.rs
joshlf: Also filed an issue based on this: https://github.com/google/zerocopy/issues/539

src/lib.rs

joshlf (changes requested):
Can you also bump the version to 0.7.12 so I can release these docs?

jswrenn:
Done!

joshlf (approved):

PR #506, opened 2023-10-13
Fix Ref::new_slice_from_suffix()

dmitryash:
`new_slice_from_suffix()` has a bug and implemented exactly as `new_slice_from_prefix()`. Now part of buffer is chosen right.

Tests couldn't have caught because buffer was just zeros. Now buffer is not just zeros so that tests can prove correctness of `new_slice_from_suffix()`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2051,10 +2051,8 @@ where
             Some(len) => len,
             None => return None,
         };
-        if bytes.len() < expected_len {
-            return None;
-        }
-        let (bytes, suffix) = bytes.split_at(expected_len);
+        let position = bytes.len().checked_sub(expected_len)?;
+        let (bytes, suffix) = bytes.split_at(position);
         Self::new_slice(suffix).map(move |l| (bytes, l))
     }
 }
@@ -4236,8 +4234,9 @@ mod tests {
 
         {
             buf.set_default();
+            buf.t[8..].fill(0xFF);
             let (r, suffix) = Ref::<_, [AU64]>::new_slice_from_prefix(&mut buf.t[..], 1).unwrap();
-            assert_eq!(suffix, [0; 8]);
+            assert_eq!(suffix, [0xFF; 8]);
             test_new_helper_slice(r, 1);
         }
         {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -2051,10 +2051,8 @@ where
             Some(len) => len,
             None => return None,
         };
-        if bytes.len() < expected_len {
-            return None;
-        }
-        let (bytes, suffix) = bytes.split_at(expected_len);
+        let position = bytes.len().checked_sub(expected_len)?;
+        let (bytes, suffix) = bytes.split_at(position);
```
joshlf, line 2055: To be consistent with the implementation of `new_from_suffix`:

```suggestion
        let split_at = bytes.len().checked_sub(expected_len)?;
        let (bytes, suffix) = bytes.split_at(split_at);
```

joshlf (changes requested):
Awesome, thanks so much for catching this! One small change to request. I want to give some more thought to whether we should do more to beef up our tests in this PR to catch bugs like this in the future; I'll have time to do that and give more detailed feedback early this coming week.

joshlf:
Not sure what's up with those failing Miri test cases. Maybe try rebasing on `main`? If that doesn't work, we can investigate in more depth.

joshlf:
I've uploaded #511, which follows on this to beef up `test_new_aligned_sized` a bit. Once CI passes for this, I'll approve and merge and then rebase #511 on top of this.

joshlf:
Note that we plan to follow up on this consistent with #512.

joshlf (approved):

joshlf:
It appears that Miri is failing nondeterministically. I've reported the issue here: https://github.com/rust-lang/miri/issues/3125. This appears to be an issue with Miri rather than with our code, so I'll try a few more times and maybe we'll get lucky with this making it through the merge queue. Obviously we'll need to fix this issue fairly soon, but there's no reason for resolving that issue to block this PR.

joshlf:
@dmitryash Out of curiosity, how did you stumble upon this issue? We're always curious to hear how zerocopy is getting used in the wild!

dmitryash:
@joshlf I needed to decode some data that has "footer" and I tried from_suffix variant but found that it did not work correctly.

joshlf:
> @joshlf I needed to decode some data that had "footer" and I tried from_suffix variant but found that it did not work correctly.

Interesting, can you say more about why it didn't work? I'd like to be able to support your use case if possible.

PR #526, opened 2023-10-19
Add bytes to type reference helper methods to `FromBytes`

kupiakos:
One common complaint I receive is how onerous it is to convert from bytes to type references in zerocopy for simple cases. I fully agree; we don't use the `Ref` type in most of our code. `AsBytes::as_bytes` makes it easy to go from `&T` to `&[u8]`, so why is there such a dance to go the other way?

These helper methods intend to improve the developer experience. They follow the same naming and error handling scheme as `read_from`. I would rather these stay named as such until a refactor of `read_from` and `write_to` occurs to match - I find that consistency valuable.

These avoid a tuple return in `from_[suffix|prefix]`  - again for developer UX since we usually threw these out.

```rust
    fn ref_from(bytes: &[u8]) -> Option<&Self> where Self: Sized;
    fn ref_from_prefix(bytes: &[u8]) -> Option<&Self> where Self: Sized;
    fn ref_from_suffix(bytes: &[u8]) -> Option<&Self> where Self: Sized;

    fn mut_from(bytes: &mut [u8]) -> Option<&mut Self> where Self: Sized + AsBytes;
    fn mut_from_prefix(bytes: &mut [u8]) -> Option<&mut Self> where Self: Sized + AsBytes;
    fn mut_from_suffix(bytes: &mut [u8]) -> Option<&mut Self> where Self: Sized + AsBytes;
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1009,6 +1009,94 @@ pub unsafe trait FromBytes: FromZeroes {
     where
         Self: Sized;
 
+    /// Interprets the given `bytes` as a `&Self` without copying.
+    ///
+    /// If `bytes.len() != size_of::<T>()` or `bytes` is not aligned to
+    /// `align_of::<T>()`, this returns `None`.
+    #[inline]
+    fn ref_from(bytes: &[u8]) -> Option<&Self>
+    where
+        Self: Sized,
+    {
+        Ref::<&[u8], Self>::new(bytes).map(Ref::into_ref)
+    }
+
+    /// Interprets the prefix of the given `bytes` as a `&Self` without copying.
+    ///
+    /// `ref_from_prefix` returns a reference to the first `size_of::<Self>()`
+    /// bytes of `bytes`. If `bytes.len() != size_of::<T>()` or `bytes` is not
+    /// aligned to `align_of::<T>()`, this returns `None`.
+    ///
+    /// To also access the prefix bytes, use [`Ref::new_from_prefix`]. Then,
+    /// use [`Ref::into_ref`] to get a `&Self` with the same lifetime.
+    #[inline]
+    fn ref_from_prefix(bytes: &[u8]) -> Option<&Self>
+    where
+        Self: Sized,
+    {
+        Ref::<&[u8], Self>::new_from_prefix(bytes).map(|(r, _)| r.into_ref())
+    }
+
+    /// Interprets the suffix of the given `bytes` as a `&Self` without copying.
+    ///
+    /// `ref_from_suffix` returns a reference to the last `size_of::<Self>()`
+    /// bytes of `bytes`. If `bytes.len() != size_of::<T>()` or the suffix of
+    /// `bytes` is not aligned to `align_of::<T>()`, this returns `None`.
+    ///
+    /// To also access the suffix bytes, use [`Ref::new_from_suffix`]. Then,
+    /// use [`Ref::into_ref`] to get a `&Self` with the same lifetime.
+    #[inline]
+    fn ref_from_suffix(bytes: &[u8]) -> Option<&Self>
+    where
+        Self: Sized,
+    {
+        Ref::<&[u8], Self>::new_from_suffix(bytes).map(|(_, r)| r.into_ref())
+    }
+
+    /// Interprets the given `bytes` as a `&mut Self` without copying.
+    ///
+    /// If `bytes.len() != size_of::<T>()` or `bytes` is not aligned to
+    /// `align_of::<T>()`, this returns `None`.
+    #[inline]
+    fn mut_from(bytes: &mut [u8]) -> Option<&mut Self>
+    where
+        Self: Sized + AsBytes,
+    {
+        Ref::<&mut [u8], Self>::new(bytes).map(Ref::into_mut)
+    }
+
+    /// Interprets the prefix of the given `bytes` as a `&mut Self` without copying.
+    ///
+    /// `mut_from_prefix` returns a reference to the first `size_of::<Self>()`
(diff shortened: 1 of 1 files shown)


joshlf:
Awesome, thanks for this! Taking a look now...

Do y'all actually make use of `from_[suffix|prefix]`? I'm curious what your use cases are since I generally have found that those APIs have very few users.

src/lib.rs
```diff
@@ -1000,6 +1000,82 @@ pub unsafe trait FromBytes: FromZeroes {
     where
         Self: Sized;
 
+    /// Interprets the given `bytes` as a `&Self` without copying.
+    ///
+    /// If `bytes.len() != size_of::<T>()` or `bytes` is
+    /// not aligned to `align_of::<T>()`, this returns `None`.
```
joshlf, line 1006: Can you reflow this and other comments to 80 columns?

kupiakos:
> Do y'all actually make use of `from_[suffix|prefix]`?

We don't use `suffix` like ever, but we do use `prefix`. We rarely check the exact size, just that it's large _enough_.

src/lib.rs
```diff
@@ -1000,6 +1000,82 @@ pub unsafe trait FromBytes: FromZeroes {
...
+        Self: Sized,
+    {
+        Ref::<&[u8], Self>::new(bytes).map(Ref::into_ref)
+    }
+
+    /// Interprets the prefix of the given `bytes` as a `&Self` without copying.
+    ///
+    /// `ref_from_prefix` returns a reference to the first `size_of::<Self>()` bytes of `bytes`.
+    /// If `bytes.len() < size_of::<T>()` or `bytes` is
+    /// not aligned to `align_of::<T>()`, this returns `None`.
```
joshlf, line 1019: Can you mention in this doc comment that users who need to get access to the suffix can use `Ref::new_from_prefix(bytes).map(Ref::into_ref)`? Same in doc comments for other prefix/suffix methods.

src/lib.rs
```diff
@@ -4425,6 +4500,69 @@ mod tests {
...
     }
 
+    #[test]
+    fn test_ref_from_mut_from() {
+        // Test `FromBytes::{ref_from, mut_from}{,_prefix,_suffix}` success cases
+        // Exhaustive coverage for these methods is covered by the `Ref` tests above,
+        // which these helper methods defer to.
+
+        let mut buf =
+            Align::<[u8; 16], AU64>::new([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]);
```
joshlf, line 4510: Can you make this 24 bytes long to avoid bugs like the one discovered in https://github.com/google/zerocopy/pull/520?

joshlf:
> > Do y'all actually make use of `from_[suffix|prefix]`?
> 
> We don't use `suffix` like ever, but we do use `prefix`. We rarely check the exact size, just that it's large _enough_. I'm supportive of removing `_suffix`, but I do like it there for consistency.

No objection to supporting both; I'm just curious what your use cases are.

joshlf (changes requested):
Looks good! A few small nits.

I'm going to hold off on approving until we resolve the discussion in https://github.com/google/zerocopy/discussions/524, and I also want to wait until some other folks have a chance to chime in who aren't working this week. I expect we can have this all resolved and merged early next week.

src/lib.rs
kupiakos: It's not clear to me how making it 24 bytes would improve anything with respect to the bug described. `suffix` acting the same as `prefix` would be caught by these tests as written - which they actually did catch during writing because I made a copy paste error.

src/lib.rs
kupiakos: Done.

src/lib.rs
kupiakos: Done, though not with an exact code snippet, but rather hyperlinks.

kupiakos:
> No objection to supporting both; I'm just curious what your use cases are.

There are probably valid use cases for `_from_suffix`, maybe for dynamic data, but it's just never come up for us from what I've seen.

It's common for a subsystem like TPM to receive a buffer that is larger than the message it contains when the request buffer is shared for the response. `_from_prefix` is particularly useful there.

src/lib.rs
joshlf: Sorry, I should have linked to #506 and #511; #520 was just me backporting to 0.6.x. The bug was in part that we used the wrong value as the location to split the byte slice in two. When the length of the object is exactly half the length of the byte slice, then the length of the object and the location to split at are identical, and so it can hide the bug.

kupiakos: Again, I don't see how changing to 24 bytes improves the test as written. The two improvements described:

- Using non-zero values, catching an incorrect position
- Using a non-half amount of the full buffer, catching an incorrect length

are both implemented already. `<[u8; 9]>::ref_from_suffix` and `<[u8; 9]>::ref_from_prefix` both check the non-zero values of the buffer they are in, and do so with a non-half amount of the full buffer.

joshlf: Oh gotcha, I didn't catch the `[u8; 9]`. Maybe leave a comment on those calling them out and explaining why the array length is important? I agree that this covers the bug I was concerned about.

joshlf (approved):
We had another request for this, so I'm going to land it and follow up on the outstanding comments in a follow-up PR.

jswrenn (approved):

joshlf:
This has been published in 0.7.18.

src/lib.rs
kupiakos: Done in #559

PR #537, opened 2023-10-24
[doc] Test `cargo doc` under `--no-default-features`

joshlf:


src/lib.rs
```diff
@@ -137,6 +137,7 @@
 //! See our [MSRV policy].
 //!
 //! [MSRV policy]: https://github.com/google/zerocopy/blob/main/POLICIES.md#msrv
+#![cfg_attr(not(feature = "byteorder"), doc = "[`byteorder`]: #")]
```
joshlf, line 141: Rustfmt wants to get rid of this newline, which IMO is a bug. TODO: Accept rustfmt's opinion? Wait until this bug is fixed? Something else?

https://github.com/rust-lang/rustfmt/issues/5947

jswrenn:
```suggestion
// Without this comment, rustfmt incorrectly tries to get rid of the following newline. 

```

joshlf: Good idea! Done.

jswrenn (approved):

PR #540, opened 2023-10-24
[macro_util] Add trailing_field_offset! macro

joshlf:
This macro will be used by the derive of KnownLayout for slice DSTs.

Makes progress on #29

### DIFF
--- a/src/macro_util.rs
+++ b/src/macro_util.rs
@@ -19,6 +19,11 @@
 
 use core::{marker::PhantomData, mem::ManuallyDrop};
 
+// TODO(#29), TODO(https://github.com/rust-lang/rust/issues/69835): Remove this
+// `cfg` when `size_of_val_raw` is stabilized.
+#[cfg(__INTERNAL_USE_ONLY_NIGHLTY_FEATURES_IN_TESTS)]
+use core::ptr::{self, NonNull};
+
 /// A compile-time check that should be one particular value.
 pub trait ShouldBe<const VALUE: bool> {}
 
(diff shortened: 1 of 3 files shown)


src/macro_util.rs
```diff
@@ -61,6 +65,133 @@ impl<T, U> MaxAlignsOf<T, U> {
...
+            // [1] TODO(#429): Citation for this?
+            unsafe {
+                #[allow(clippy::as_conversions)]
+                ::core::mem::size_of_val_raw(zero_elems as *const $ty)
+            }
+        };
+
+        assert!(min_size <= _64K);
+
+        let ptr = ALIGNED_64K_ALLOCATION.as_ptr().cast::<$ty>();
```
jswrenn, line 142: `NonNull::cast<U>` requires `U: Sized`; we need to do an `as` cast here instead if we want to support DSTs. I'll make the fix.

src/macro_util.rs
```diff
@@ -276,6 +407,32 @@ mod tests {
...
+                test!($(#[$cfgs])* ($($ts),* ; $trailing_field_ty) => $expect);
+            };
+            (@offset ; $_trailing:ty) => { trailing_field_offset!(Test, 0) };
+            (@offset $_t:ty ; $_trailing:ty) => { trailing_field_offset!(Test, 1) };
+        }
+
+        test!(#[repr(C)] #[repr(transparent)] #[repr(packed)] #[repr(C)] (; u8) => Some(0));
+        test!(#[repr(C)] #[repr(packed)] #[repr(C)] (u8; u8) => Some(1));
+        test!(#[repr(C)] (u8; [AU64]) => Some(8));
+        test!(#[repr(C)] (u8; Nested<u8, AU64>) => Some(8));
```
joshlf, line 431: Some other test ideas:
- `(u8; Nested<u8, [AU64]>)`
- `(; u8)`
- `(; [u8])`
- `(; AU64)`
- `(; [AU64])`
- `(; Nested<u8; [AU64]>)`

I'd also test with variants of `#[repr(packed(N))]` and `#[repr(align(N))]`, maybe for values of `N` in `{1, 2, 8, 16}`. I'm specifically interested in making sure that if the padding is modified by `repr(packed)`, the macro detects it. `repr(align)` shouldn't modify padding, and we should confirm that too.

src/macro_util.rs
```diff
@@ -61,6 +65,134 @@ impl<T, U> MaxAlignsOf<T, U> {
...
+#[repr(C, align(65536))]
+struct Aligned64kAllocation([u8; _64K]);
+
+/// A pointer to an aligned allocation of size 2^16.
+///
+/// # Safety
+///
+/// `ALIGNED_64K_ALLOCATION` is guaranteed to point to the entirety of an
+/// allocation with size and alignment 2^16, and to have valid provenance.
+// remove this `cfg` when `size_of_val_raw` is stabilized
```
joshlf, line 81:
```suggestion
// TODO(#29), TODO(https://github.com/rust-lang/rust/issues/69835): Remove this `cfg` when
// `size_of_val_raw` is stabilized.
```

(but reflowed to 80 columns, which I assume my suggestion isn't)

Same elsewhere in this file.

src/macro_util.rs
```diff
@@ -61,6 +65,134 @@ impl<T, U> MaxAlignsOf<T, U> {
     }
 }
 
+const _64K: usize = 1 << 16;
+
+// remove this `cfg` when `size_of_val_raw` is stabilized
+#[cfg(__INTERNAL_USE_ONLY_NIGHLTY_FEATURES_IN_TESTS)]
+#[repr(C, align(65536))]
+struct Aligned64kAllocation([u8; _64K]);
```
joshlf, line 73: Can we add a test that `assert_eq!(align_of::< Aligned64kAllocation>(), _64K);`?

src/macro_util.rs
```diff
@@ -61,6 +65,134 @@ impl<T, U> MaxAlignsOf<T, U> {
...
+            //     fail due to "mismatched vtable kinds", and compilation will
+            //     fail.
+            //   - If `$ty` is a slice DST, the safety requirement is that "the
+            //     length of the slice tail must be an initialized integer, and
+            //     the size of the entire value (dynamic tail length +
+            //     statically sized prefix) must fit in isize." The length is
+            //     initialized to 0 above, and Rust guarantees that no type's
+            //     minimum size may overflow `isize`. [1]
+            //
+            // [1] TODO(#429): Citation for this?
```
joshlf, line 133:
```suggestion
            // [1] TODO(#429), TODO(https://github.com/rust-lang/unsafe-code-guidelines/issues/465#issuecomment-1782206516): Citation for this?
```

src/macro_util.rs
```diff
@@ -61,6 +65,134 @@ impl<T, U> MaxAlignsOf<T, U> {
...
+///
+/// `trailing_field_offset!` produces code which is valid in a `const` context.
+// remove this `cfg` when `size_of_val_raw` is stabilized
+#[cfg(__INTERNAL_USE_ONLY_NIGHLTY_FEATURES_IN_TESTS)]
+#[doc(hidden)] // `#[macro_export]` bypasses this module's `#[doc(hidden)]`.
+#[macro_export]
+macro_rules! trailing_field_offset {
+    ($ty:ty, $trailing_field_name:tt) => {{
+        let min_size = {
+            let zero_elems: *const [()] = ::core::ptr::slice_from_raw_parts(
```
joshlf, line 116: Use `$crate::macro_util::core_reexport` here and elsewhere in this macro?

src/macro_util.rs
```diff
@@ -61,6 +65,134 @@ impl<T, U> MaxAlignsOf<T, U> {
...
+        //   is guaranteed to remain in-bounds of its allocation.
+        // - Because the minimum size is no larger than `_64K` bytes, and
+        //   because an object's size must always be a multiple of its alignment
+        //   [1], we know that `$ty`'s alignment is no larger than `_64K`. The
+        //   allocation addressed by `ALIGNED_64K_ALLOCATION` is guaranteed to
+        //   be aligned to `_64K`, so `ptr` is guaranteed to satisfy `$ty`'s
+        //   alignment.
+        //
+        //   Note that, as of [2], this requirement is technically unnecessary,
+        //   but no harm in guaranteeing it anyway.
```
joshlf, line 159: Maybe also mention that our MSRV precedes [2], and so this is actually required until our MSRV is updated?

jswrenn (approved):

PR #541, opened 2023-10-24
Initial commit of DstLayout::for_dst

joshlf:
This will be used by the custom derive of `KnownLayout` to compute the `DstLayout` for unsized types.

Makes progress on #29

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -226,7 +226,7 @@
 #![cfg_attr(doc_cfg, feature(doc_cfg))]
 #![cfg_attr(
     __INTERNAL_USE_ONLY_NIGHLTY_FEATURES_IN_TESTS,
-    feature(layout_for_ptr, strict_provenance)
+    feature(const_size_of_val_raw, layout_for_ptr, strict_provenance)
 )]
 
 // This is a hack to allow zerocopy-derive derives to work in this crate. They
@@ -335,19 +335,22 @@ const _: () = {
 /// [the reference]: https://doc.rust-lang.org/reference/type-layout.html
 #[doc(hidden)]
 #[allow(missing_debug_implementations, missing_copy_implementations)]
-#[cfg_attr(test, derive(Copy, Clone, Debug, PartialEq, Eq))]
+#[cfg_attr(any(test, kani), derive(Copy, Clone, Debug, PartialEq, Eq))]
+#[cfg_attr(kani, derive(kani::Arbitrary))]
 pub struct DstLayout {
     _align: NonZeroUsize,
     _size_info: SizeInfo,
 }
 
-#[cfg_attr(test, derive(Copy, Clone, Debug, PartialEq, Eq))]
+#[cfg_attr(any(test, kani), derive(Copy, Clone, Debug, PartialEq, Eq))]
+#[cfg_attr(kani, derive(kani::Arbitrary))]
 enum SizeInfo<E = usize> {
     Sized { _size: usize },
     SliceDst(TrailingSliceLayout<E>),
 }
 
-#[cfg_attr(test, derive(Copy, Clone, Debug, PartialEq, Eq))]
+#[cfg_attr(any(test, kani), derive(Copy, Clone, Debug, PartialEq, Eq))]
+#[cfg_attr(kani, derive(kani::Arbitrary))]
 struct TrailingSliceLayout<E = usize> {
     // The offset of the first byte of the trailing slice field. Note that this
     // is NOT the same as the minimum size of the type. For example, consider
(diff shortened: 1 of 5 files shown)


src/byteorder.rs
```diff
@@ -74,6 +74,9 @@ use core::{
     num::TryFromIntError,
 };
 
+// Ensures proc-macro derive works in the context of this crate.
+use crate as zerocopy;
+
```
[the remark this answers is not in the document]
jswrenn: I'm stumped about this. `KnownLayout` isn't the only trait that depends on zerocopy items, but it's the only trait here causing an issue.

src/lib.rs
```diff
@@ -439,6 +442,64 @@ impl DstLayout {
...
+
+                    let padding = util::core_layout::_padding_needed_for(without_padding, align);
+
+                    let _size = match without_padding.checked_add(padding) {
+                        Some(_size) => _size,
+                        None => return None,
+                    };
+
+                    // SAFETY: TODO
+                    SizeInfo::Sized { _size }
```
jswrenn, line 484: I'm not convinced this is sound for non-`repr(C)` types. Rust reserves the right to add excess padding.

src/byteorder.rs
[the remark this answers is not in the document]
jswrenn: Resolved in #604.

src/lib.rs
joshlf: Oh yeah very good point. Maybe rename to `for_repr_c_dst` or something? And add a safety precondition?

src/lib.rs
```diff
@@ -3908,6 +3965,80 @@ mod tests {
         }
     }
 
+    mod for_dst {
```
joshlf, line 3968: What's the thinking behind having these in a nested module as opposed to just naming them `test_dst_layout_for_sized` etc?

src/lib.rs
```diff
@@ -5677,3 +5808,91 @@ mod tests {
         }
     }
 }
+
+#[cfg(kani)]
+mod proofs {
```
joshlf, line 5813: ❤️

src/macro_util.rs
```diff
@@ -146,10 +146,16 @@ macro_rules! trailing_field_offset {
             }
         };
 
-        assert!(min_size <= _64K);
+        if min_size > $crate::macro_util::_64K {
+            panic!(concat!(
+                "cannot compute layout of `",
+                stringify!($ty),
+                "`; type is larger than zerocopy supports"
```
joshlf, line 153: Leaving a TODO so we don't forget: We should have an issue link here so there's a way for users to report that they need support for larger types.

zerocopy-derive/src/lib.rs
```diff
@@ -548,11 +578,19 @@ fn impl_block<D: DataExt>(
...
+            field_types.iter().map(|ty| parse_quote!(#ty: ::zerocopy::#trait_ident)).collect()
+        }
+        RequireBoundedFields::Trailing => {
+            if let Some(trailing) = field_types.last() {
+                vec![parse_quote!(#trailing: ::zerocopy::#trait_ident)]
+            } else {
+                vec![]
+            }
+        }
+    };
```
joshlf, line 593: If you want to be really cute. I swapped the order of `Yes` and `No` so that the two `Trailing` branches are grouped. Whether this is a good idea is highly dependent upon personal preference. Feel free to take all, some (ie, just the `bound_tt` idea), or none of this suggestion.

```suggestion
    let bound_tt = |ty| parse_quote!(#ty: ::zerocopy::#trait_ident);
    let field_type_bounds: Vec<_> = match (require_trait_bound_on_field_types, field_types) {
        (RequireBoundedFields::Yes, _) => field_types.iter().map(bound_tt).collect(),
        (RequireBoundedFields::No, _) | (RequireBoundedFields::Trailing, []) => vec![],
        (RequireBoundedFields::Trailing, [.., last]) => vec![bound_tt(last)],
    };
```

src/lib.rs
jswrenn: - reduces repetition in the names
- makes it possible to code-collapse the entire group in an editor

My preference (not in this PR) would be even to go a little further and drop `test_` from all tests. All of the tests are already located in a module called "tests".

zerocopy-derive/src/lib.rs
jswrenn: I *do* love being cute.

src/lib.rs
jswrenn: What should the UX of the derive be? Some possibilities:
- unconditionally require `repr(C)` on structs.
- don't unconditionally require `repr(C)`, but expand to `DstLayout::for_type::<Self>()` on non-`repr(C)` types, and `DstLayout::for_repr_c_dst` on `repr(C)` types.

joshlf: Generally I'd prefer the latter, but how can we make the UX of the error message good when the user tries to derive on an unsized, non-`repr(C)` type?

Issue #202, opened 2023-07-28
Prevent panics statically


joshlf:
Issues like #200 demonstrate that the compiler is sometimes not smart enough to optimize out panics that we can prove statically won't happen. It would be good if we could refactor our code to only use operations which don't include panicking in their call graph so that we don't have to rely on optimization to detect these cases and elide them for us.

jettr:
Here is a discussion on another project about having a no panic policy with some good information: https://github.com/orgs/tpm-rs/discussions/5

I think it could help the zerocopy library if it were interested in some automated CI checks to ensure no panics are present in "tier 1" tool chains.

joshlf:
Awesome, thanks for that breadcrumb! I'd like to go even further than that, and ensure that panics aren't just optimized out, but never emitted in the first place. Do you know if that technique would work in conjunction with `opt-level = 0`, and would that suffice to guarantee that we're never emitting panics? I want to make sure we don't get into the situation in #200 in which panics are optimized out in a toy example but not in a larger application.

jettr:
I like the idea of using `opt-level=0` for the "tier 1" tool chains that CI checks against to ensure there are no panics. I don't know of a way to ensure panics are never emitted with 100% certainty though. I think your idea is about as close as you can get with currently tooling.

jrvanwhy:
I'm starting to work on this. I suspect I'll run out of time to send a PR this week but I hope to send one next week.

joshlf:
Awesome, thanks! I'll assign it to you. Feel free to comment here if you have any questions.

jrvanwhy:
It turns out that [`redpen`](https://crates.io/crates/redpen-linter)  does not catch every possible panic path. For example, it does not error on this program:

```rust
pub trait MaybePanic { fn maybe_panic(&self); }

pub struct WillPanic;
impl MaybePanic for WillPanic {
    fn maybe_panic(&self) { panic!(); }
}

#[redpen::dont_panic]
pub fn will_this_panic<M: MaybePanic>(m: M) {
    m.maybe_panic();
}

#[redpen::dont_panic]
pub fn main() {
    will_this_panic(WillPanic);
}
```

So we have a few tools at our disposal, all of which are incomplete:

1. Linking-based checks: Build code that uses zerocopy in a way where we can look at the emitted symbols/links to determine whether a panic could have happened. [`no_panic`](https://crates.io/crates/no_panic) is an example of this approach, but I think it could also be done using e.g. a custom #[panic_handler].
[...]

Issue #368, opened 2023-09-11
Shrink the size of a `Ref` in memory for some `B: ByteSlice`


joshlf:
Currently `Ref` is defined as:

https://github.com/google/zerocopy/blob/52ffa4d1cbd81d0570276d20ddac783ba2a5ecf2/src/lib.rs#L1417

This has the consequence that, even if `T: Sized`, when in principle `Ref` could be a single-word thin pointer, `Ref` is still as large as `B`. Usually this is two words (e.g. for `&[u8]`), but sometimes it's more (e.g. for [`core::cell::Ref`](https://doc.rust-lang.org/std/cell/struct.Ref.html), which has a reference count).

For some types, this is unavoidable. E.g., no matter what we do, we need to store the entire `core::cell::Ref` - if we didn't, we'd be leaking memory. But for some types - such as `&[u8]` - we can reconstruct the original `B: ByteSlice` just from a pointer to `T`.

Thus, I propose the following addition to `ByteSlice`:

```rust
trait ByteSlice {
    // Name to be bikeshedded later. We may not want to tie the name to the `Ref`
    // type since we plan on maybe making `ByteSlice` unsealed at some point.
    //
    // For `&[u8]`/`&mut [u8]`, this is just `NonNullFoo<T>`. For `core::cell::Ref`/
    // `core::cell::RefMut`, this is just `Self`.
    type Foo<T: ?Sized>: AsRef<[u8]> + AsMut<[u8]>;

    // Either discards `foo`, constructing `Foo` from `self` (e.g. for `core::cell::Ref`),
    // or discards `self`, constructing `Foo` from `foo` (e.g. for `&[u8]`).
    //
    // SAFETY: Caller promises that `ptr` references a sequence of bytes which are
    // all initialized. That ensures that we can implement `AsRef<[u8]>`/`AsMut<[u8]>`
    // without a `T: AsBytes` bound.
    unsafe fn into_foo<T: ?Sized>(self, foo: NonNullFoo<T>) -> Self::Foo;
}

struct NonNullFoo<T> {
    // Invariant: Always references a fully-initialized sequences of bytes. This is fine
    // because this is always constructed from a `B: ByteSlice` during `Ref` construction.
    inner: NonNull<T>,
}

// NOTE: Caller needs to ensure the `NonNullFoo` doesn't live for too long since
// these methods are not `usnafe`. Might need to be a safety requirement on
// `ByteSlice::into_foo`.
[...]

PR #544, opened 2023-10-24
Replace a panic with an error branch

jrvanwhy:
This enables Clippy's [`expect_used`](https://rust-lang.github.io/rust-clippy/master/index.html#/expect_used) lint, which errors on calls to `Option::expect` and `Result::{expect, expect_err}`. The intent of enabling this lint is to reduce the number of panics emitted in `zerocopy`'s code (issue #202).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3399,10 +3399,10 @@ pub unsafe trait IntoBytes {
         Self: Immutable,
     {
         let start = bytes.len().checked_sub(mem::size_of_val(self))?;
-        bytes
-            .get_mut(start..)
-            .expect("`start` should be in-bounds of `bytes`")
-            .copy_from_slice(self.as_bytes());
+        // get_mut() should never return None here. We use ? rather than
+        // .unwrap() because in the event the branch is not optimized away,
+        // returning None is generally lighter-weight than panicking.
+        bytes.get_mut(start..)?.copy_from_slice(self.as_bytes());
         Some(())
     }
 


jrvanwhy:
Question: is `expect_used` worth the false positives? It triggered twice:

1. True positive on `AsBytes::write_to_suffix` on a dead branch. I think this `expect` would have been optimized away.
2. False positive on `Ref::new_slice`, which is intended to panic (as indicated by its documentation).

jrvanwhy:
Question: This branch is already out-of-date and needs changes, do you prefer that I do a merge or that I do a rebase?

joshlf:
> Question: This branch is already out-of-date and needs changes, do you prefer that I do a merge or that I do a rebase?

Rebase please, thanks!

joshlf:
Maybe we could also enable `clippy::unwrap_used` in this PR and get more coverage?

Also, we'll probably want some way of disabling these in testing code, since we do want to be able to unwrap and expect in tests. Maybe we can just put these behind a `cfg_attr(not(test), ...)`?

jrvanwhy:
> Maybe we could also enable `clippy::unwrap_used` in this PR and get more coverage?

`clippy::unwrap_used` is already enabled in zerocopy :-)

> Also, we'll probably want some way of disabling these in testing code, since we do want to be able to unwrap and expect in tests. Maybe we can just put these behind a `cfg_attr(not(test), ...)`?

Good point. There is already a `cfg_attr` block that allows `clippy::unwrap_used` and a few other lints in the `test` and `kani` environments. I added `clippy::expect_used` to that block to accept it in test environments.

src/lib.rs
```diff
@@ -2992,10 +2995,7 @@ pub unsafe trait AsBytes {
         Self: NoCell,
     {
         let start = bytes.len().checked_sub(mem::size_of_val(self))?;
-        bytes
-            .get_mut(start..)
-            .expect("`start` should be in-bounds of `bytes`")
-            .copy_from_slice(self.as_bytes());
+        bytes.get_mut(start..)?.copy_from_slice(self.as_bytes());
```
joshlf, line 2998: Can you add a comment here explaining why we're using `?` rather than unwrapping? IIUC, the idea is that we know this will always succeed, so in principle unwrapping should be fine, but we don't want to risk the compiler not being able to optimize out the panic in all cases. Is that right?

Also, can you leave a TODO comment linking to #202 to improve this eventually? See other TODO comments for the syntax.

joshlf (changes requested):

src/lib.rs
jrvanwhy: > Also, can you leave a TODO comment linking to #202 to improve this eventually? See other TODO comments for the syntax.

I don't understand what you're looking for here. Is there a further improvement that should be made to `write_to_suffix`'s implementation? Or are you referring to better checking (e.g. redpen)?

joshlf: Yeah, I'd like to be able to go back to panicking once we can prove that the panic is optimized out, or ideally to some other proven-with-the-type-system approach that gives us a guarantee that there will be no panic. The problem with `?` is that it won't panic, but might introduce a new branch, and that's harder for us to detect. By contrast, if we can detect panics with something like redpen, then we can have higher confidence that the method is well-optimized.

jrvanwhy: Okay, after https://github.com/google/zerocopy/issues/202#issuecomment-1979344700 I added a comment indicating that we are using `?` rather than `unwrap`/`expect` because it is lighter-weight.

jrvanwhy:
Updated statistics for `clippy::expect_used`:

We have one function that is documented as panicing as part of its public API (`FromZeros::new_box_slice_zeroed`).

We have one function that called `expect` which we can replace with a lighter-weight branch (`IntoBytes::write_to_suffix`).

We have four functions (`Ref::into_ref`, `Ref::into_mut`, `Ref::deref`, `Ref::deref_mut`) with "should never be hit" panics that we cannot replace easily. I don't think we can rely on the optimizer to reliably remove these panics, as their unreachability depends on invariants of `Ref`. The only thing we could do to remove these panics without changing the public API is to replace the panic with something else undesirable, such as `loop {}` or `unreachable_unchecked!`.

One useful warning versus five useless warnings is not a great ratio. I'll let @joshlf decide whether you think that `clippy::expect_used` is worth the noise.

joshlf:
> Updated statistics for `clippy::expect_used`:
> 
> We have one function that is documented as panicing as part of its public API (`FromZeros::new_box_slice_zeroed`).
> 
> We have one function that called `expect` which we can replace with a lighter-weight branch (`IntoBytes::write_to_suffix`).
> 
> We have four functions (`Ref::into_ref`, `Ref::into_mut`, `Ref::deref`, `Ref::deref_mut`) with "should never be hit" panics that we cannot replace easily. I don't think we can rely on the optimizer to reliably remove these panics, as their unreachability depends on invariants of `Ref`. The only thing we could do to remove these panics without changing the public API is to replace the panic with something else undesirable, such as `loop {}` or `unreachable_unchecked!`.

IIUC these will be resolved by https://github.com/google/zerocopy/issues/368, at least for byte slice types (smart pointer types like `core::cell::Ref` will still need to have panic paths).

> One useful warning versus five useless warnings is not a great ratio. I'll let @joshlf decide whether you think that `clippy::expect_used` is worth the noise.

I'm leaning no, which is a shame because you've put a lot of work into this PR, but I agree that it's a lot of noise for not much benefit.

That said, maybe you could land the refactor of `write_to_suffix` (either just simplifying this PR to contain only that change or in a new PR)? Even without the clippy lint, that's a useful change.

jrvanwhy:
> > One useful warning versus five useless warnings is not a great ratio. I'll let @joshlf decide whether you think that `clippy::expect_used` is worth the noise.
> 
> I'm leaning no, which is a shame because you've put a lot of work into this PR, but I agree that it's a lot of noise for not much benefit.

Not really -- most of the work on this PR has been on further refining what exactly it means to "prevent panics" in zerocopy, and what sort of tradeoffs we want to make in achieving that goal. I decided to stick with one-Clippy-lint-per-PR so that "we don't want lint X" is an easy decision to make.

> That said, maybe you could land the refactor of `write_to_suffix` (either just simplifying this PR to contain only that change or in a new PR)? Even without the clippy lint, that's a useful change.

I followed the "simplify this PR" route so it only lands that one change, and renamed the PR accordingly.

src/lib.rs
```diff
@@ -3373,10 +3373,10 @@ pub unsafe trait IntoBytes {
         Self: NoCell,
     {
         let start = bytes.len().checked_sub(mem::size_of_val(self))?;
-        bytes
-            .get_mut(start..)
-            .expect("`start` should be in-bounds of `bytes`")
-            .copy_from_slice(self.as_bytes());
+        // get_mut() should never return None here. We use ? rather than
+        // .unwrap() because in the event the branch is not optimized away,
+        // returning None is generally lighter-weight than panicing.
```
joshlf, line 3378:
```suggestion
        // returning None is generally lighter-weight than panicking.
```

joshlf (changes requested):
Sorry this has taken so long to get to.

Also, see https://github.com/google/zerocopy/issues/1125 and feel free to chime in there.

src/lib.rs
jrvanwhy: Fixed.

joshlf (approved):

PR #550, opened 2023-10-26
Add `transmute_mut!` macro

jswrenn:
This macro is like the existing `transmute!`, but it transmutes mutable references rather than values.

Issue #159

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1980,7 +1980,7 @@ macro_rules! transmute {
 ///   |
 ///   = note: source type: `AlignOf<[u8; 2]>` (8 bits)
 ///   = note: target type: `MaxAlignsOf<[u8; 2], u16>` (16 bits)
-///   = note: this error originates in the macro `zerocopy::transmute_ref` (in Nightly builds, run with -Z macro-backtrace for more info)
+///   = note: this error originates in the macro `$crate::assert_align_gt_eq` which comes from the expansion of the macro `transmute_ref` (in Nightly builds, run with -Z macro-backtrace for more info)
 /// ```
 ///
 /// This is saying that `max(align_of::<T>(), align_of::<U>()) !=
@@ -2011,7 +2011,8 @@ macro_rules! transmute_ref {
             transmute(e)
         } else if false {
             // This branch, though never taken, ensures that `size_of::<T>() ==
-            // size_of::<U>()`.
+            // size_of::<U>()` and that that `align_of::<T>() >=
+            // align_of::<U>()`.
 
             // `t` is inferred to have type `T` because it's assigned to `e` (of
             // type `&T`) as `&t`.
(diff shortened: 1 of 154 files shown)


joshlf (changes requested):
Just skimming; I'll take more of a look later once it's ready for review.

src/macro_util.rs
```diff
@@ -101,6 +101,58 @@ macro_rules! union_has_padding {
...
+            $t = align_of.into_t();
+            // `max_aligns` is inferred to have type `MaxAlignsOf<T, U>` because
+            // of the inferred types of `t` and `u`.
+            let mut max_aligns = $crate::macro_util::MaxAlignsOf::new($t, $u);
+
+            // This transmute will only compile successfully if
+            // `align_of::<T>() == max(align_of::<T>(), align_of::<U>())` - in
+            // other words, if `align_of::<T>() >= align_of::<U>()`.
+            //
+            // SAFETY: This code is never run.
```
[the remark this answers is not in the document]
jswrenn: It's not a safety condition of the entire macro — just that one `transmute`. It's satisfied by nestling that `transmute` in an `if false { ... }`.

The macro docs do say that the macro, as a whole, should be invoked in dead code, but that's *not* a safety condition — memory safety isn't violated if the macro is invoked in a live path.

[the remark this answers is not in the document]
joshlf: Ooops, you're right. I missed that.

src/lib.rs
```diff
@@ -2020,13 +2020,10 @@ macro_rules! transmute_ref {
...
-            let u = unsafe {
-                 // Clippy: It's okay to transmute a type to itself.
-                #[allow(clippy::useless_transmute)]
-                $crate::macro_util::core_reexport::mem::transmute(t)
-            };
+            let u;
+
+            $crate::assert_size_eq!(t, u);
+
             &u
```
joshlf, line 2027: Slight stylistic preference for grouping these together, but feel free to ignore this.

Same in the next branch.

```suggestion
            let u;
            $crate::assert_size_eq!(t, u);
            &u
```

src/lib.rs
```diff
@@ -2067,11 +2051,139 @@ macro_rules! transmute_ref {
...
+    ($e:expr) => {{
+        // NOTE: This must be a macro (rather than a function with trait bounds)
+        // because there's no way, in a generic context, to enforce that two
+        // types have the same size or alignment.
+
+        // Reborrow so that mutable references are supported too.
+        //
+        // In the rest of the comments, we refer only to `&T` since this
+        // reborrow ensures that `e` is an immutable reference.
+        let e = &mut *$e;
```
joshlf, line 2125: Is this necessary for `transmute_mut!`?

src/lib.rs
```diff
@@ -2067,11 +2051,139 @@ macro_rules! transmute_ref {
...
+            // type of this macro expression is `&mut U` where `U: 'u + Sized +
+            // AsBytes + FromBytes`.
+            fn transmute<'t, T, U>(_t: &'t mut T) -> &'t mut U
+            where
+                T: 't + Sized + $crate::AsBytes + $crate::FromBytes,
+                U: 't + Sized + $crate::AsBytes + $crate::FromBytes,
+            {
+                loop {}
+            }
+            transmute(&mut *e)
```
joshlf, line 2140: Can this not just be `transmute(e)`?

src/lib.rs
```diff
@@ -2067,11 +2051,139 @@ macro_rules! transmute_ref {
...
+            let mut t = unreachable!();
+            e = &mut t;
+
+            // `u` is inferred to have type `U` because it's used as `&mut u` as
+            // the value returned from this branch.
+            let u;
+
+            $crate::assert_size_eq!(t, u);
+
+            &mut u
```
joshlf, line 2156: Same here, also optional.

```suggestion
            let u;
            $crate::assert_size_eq!(t, u);
            &mut u
```

src/lib.rs
```diff
@@ -2067,11 +2051,139 @@ macro_rules! transmute_ref {
...
+            let mut t = unreachable!();
+            e = &mut t;
+
+            // `u` is inferred to have type `U` because it's used as `&mut u` as
+            // the value returned from this branch.
+            let mut u = unreachable!();
+
+            $crate::assert_align_gt_eq!(t, u);
+
+            &mut u
```
joshlf, line 2172: Same here, also optional.

```suggestion
            let mut u = unreachable!();
            $crate::assert_align_gt_eq!(t, u);
            &mut u
```

joshlf, line 2172: Could this branch be collapsed with the previous one, and call both macros on the same `t` and `u`? Same in `transmute_ref!`.

joshlf (changes requested):

src/lib.rs

joshlf (dismissed):

joshlf (approved):

joshlf (approved):

PR #559, opened 2023-10-27
Improve docs for #526

kupiakos:
This corrects a docstring mistake made during editing, and adds a comment for the `[u8; 9]` as requested.

See #526. Sorry for not responding - I've been out with COVID!


joshlf (approved):
Good catch, thanks!

PR #562, opened 2023-10-27
[macro_util] Simplify core_reexport

joshlf:

### DIFF
--- a/src/macro_util.rs
+++ b/src/macro_util.rs
@@ -153,9 +153,7 @@ macro_rules! assert_size_eq {
     }};
 }
 
-pub mod core_reexport {
-    pub use core::*;
-}
+pub use core as core_reexport;
 
 #[cfg(test)]
 mod tests {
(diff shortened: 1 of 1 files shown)


src/macro_util.rs
```diff
@@ -154,9 +154,7 @@ macro_rules! assert_size_eq {
 }
 
 pub mod core_reexport {
-    pub mod mem {
-        pub use core::mem::transmute;
-    }
+    pub use core::*;
 }
```
jswrenn, line 158: Instead of all this, why not:
```
pub use ::core as core_reexport;
```

jswrenn (changes requested):

src/macro_util.rs
joshlf: Good call, done.

jswrenn (approved):

jswrenn:
I believe the semver-checks failure is a false-positive, and I've filed a bug report: https://github.com/obi1kenobi/cargo-semver-checks/issues/573

joshlf:
Closed in favor of https://github.com/google/zerocopy/pull/586.

Issue #554, opened 2023-10-27
Reduce our MSRV as much as possible


joshlf:
Our MSRV is currently 1.61, which is higher than the MSRVs of many crates. Here are MSRVs of the subset of the top 100 most-recently-downloaded crates that contain `unsafe` code:

TODO

We should work to reduce our MSRV as much as possible so that more crates can replace `unsafe` code with zerocopy.

## Mentoring instructions

- Repeat the following steps as many times as possible:
  - [ ] Determine what is preventing our MSRV from being one version earlier
  - [ ] Determine whether there's a workaround for the same behavior that is compatible with an earlier MSRV

polarathene:
I'm a little curious about the impact of this (_as a general goal, not specifically the impact to `zerocopy`_).
- As you raise PRs to iteratively lower the MSRV, you need to make more sacrifices that affects the maintenance of `zerocopy`?
- And then eventually raising MSRV again to undo these changes?

Perhaps it's more feasible for `zerocopy` due to minimal dependencies, so most of the concerns below can be disregarded 😅 

---

In https://github.com/google/zerocopy/issues/557 you reference libc MSRV and targets of `1.30` (8% compatible) and `1.31` (19% compatible), with [MSRV compatibility of current crates.io ecosystem](https://lib.rs/stats#rustc):

![image](https://github.com/google/zerocopy/assets/5098581/85c635b8-644d-4d69-90ef-68b6cc43bf28)

Your current target reached (`1.57`) is 59% compatible:

![image](https://github.com/google/zerocopy/assets/5098581/fa08ef81-ee98-457d-8861-e02ad023bf22)

Once you pass `1.56.0` you've lost support for depending on any crates using `edition = 2021`, while some crates for example require `1.60` due to usage of new `Cargo.toml` syntax introduced then but those releases are completely invisible/ignored by the resolver on earlier toolchains it seems. The `rust-version` field was also introduced in `1.56.0` (_which I see `zerocopy` leverages_).

In contrast of that full ecosystem, the `lib.rs` stats link also shows the MSRV compatibility of more recently updated crates:

![image](https://github.com/google/zerocopy/assets/5098581/612eb56b-d086-4e08-a428-2d307f2f8d2c)

Note the widening "broken deps" compatibility, along with the steep "incompatible" increase after going below `edition = "2021"` compatibility with `1.55`.

Possibly relevant: https://github.com/mitsuhiko/insta/issues/289#issuecomment-1484309066

[...]

PR #574, opened 2023-10-31
Lower MSRV to 1.56

joshlf:
In order to do this, we make a few changes:
- In 1.56, panicking is not supported in `const fn`s. In order to
  support 1.56 while still panicking in `const fn`s in later versions,
  we make a few changes:
  - We introduce a `build.rs` script which detects the Rust toolchain
    version and emits the cfg `zerocopy_panic_in_const_fn` if the Rust
    version is at least 1.57.
  - In some `const fn`s, we put debug assertions behind
    `#[cfg(zerocopy_panic_in_const_fn)]`.
  - In some `const fn`s, we replace `unreachable!()` with non-panicking
    code when `#[cfg(not(zerocopy_panic_in_const_fn))]`.
  - In one case, we compile a method as either a `const fn` or a
    non-`const fn` depending on `zerocopy_panic_in_const_fn`.
- We replace `&*dst` with `transmute(dst)`.
- We make sure that, in `impl` blocks, `const` parameters always come
  last.
- We make the `byteorder` feature/crate dependency off by default. This
  is a breaking change, and so we bump the version number to
  0.8.0-alpha.

Release 0.8.0-alpha.

Makes progress on https://github.com/google/zerocopy/issues/554

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,12 +15,12 @@
 [package]
 edition = "2018"
 name = "zerocopy"
-version = "0.7.21"
+version = "0.8.0-alpha"
 authors = ["Joshua Liebow-Feeser <joshlf@google.com>"]
 description = "Utilities for zero-copy parsing and serialization"
 license = "BSD-2-Clause OR Apache-2.0 OR MIT"
 repository = "https://github.com/google/zerocopy"
-rust-version = "1.60.0"
+rust-version = "1.56.0"
 
 exclude = [".*"]
 
@@ -36,8 +36,6 @@ pinned-nightly = "nightly-2023-10-31"
 features = ["__internal_use_only_features_that_work_on_stable"]
 
 [features]
-default = ["byteorder"]
-
 alloc = []
 derive = ["zerocopy-derive"]
 simd = []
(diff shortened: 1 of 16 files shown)


testutil/src/lib.rs
```diff
@@ -40,10 +40,10 @@ impl PinnedVersions {
             .ok_or("failed to find msrv: no `rust-version` key present")?
             .to_string();
```
joshlf, line 41: We'll need to find a way to work around https://github.com/oli-obk/cargo_metadata/issues/253.

Cargo.toml
```diff
@@ -15,12 +15,12 @@
...
 edition = "2018"
 name = "zerocopy"
-version = "0.7.21"
+version = "0.8.0-alpha"
 authors = ["Joshua Liebow-Feeser <joshlf@google.com>"]
 description = "Utilities for zero-copy parsing and serialization"
 license = "BSD-2-Clause OR Apache-2.0 OR MIT"
 repository = "https://github.com/google/zerocopy"
-rust-version = "1.60.0"
+rust-version = "1.56.0"
```
joshlf, line 23: We may need to update our testing infrastructure to support testing on more stable toolchains other than one MSRV and one stable. In particular, we'll want to test on every toolchain that enables some functionality in `build.rs`. If our MSRV is A, our `build.rs` emits a particular cfg for versions B and above (B > A), and our pinned stable is C (C > B), then if we only test on A and C in CI, it's possible for behavior which is gated behind the cfg to be broken on any version in the range [B, C). Thus, we should test on B as well. In the general case, we need the ability to test on the full list of toolchain versions that invoke distinct `build.rs` behavior.

joshlf:
Superseded by #855.

PR #576, opened 2023-10-31
add `align_of!` for computing alignment of DSTs

jswrenn:
The `align_of!` macro uses `trailing_field_offset!` to compute the alignment of DSTs with minimum alignments up to 65536. By contrast, `core::mem::align_of` does not support DSTs.

We can use this for supplying the alignment argument required by `DstLayout::for_dst` (introduced by #541).

### DIFF
--- a/src/macro_util.rs
+++ b/src/macro_util.rs
@@ -205,6 +205,46 @@ macro_rules! trailing_field_offset {
     }};
 }
 
+/// Computes alignment of `$ty: ?Sized`.
+///
+/// `align_of!` produces code which is valid in a `const` context.
+// TODO(#29), TODO(https://github.com/rust-lang/rust/issues/69835): Remove this
+// `cfg` when `size_of_val_raw` is stabilized.
+#[cfg(__INTERNAL_USE_ONLY_NIGHLTY_FEATURES_IN_TESTS)]
+#[doc(hidden)] // `#[macro_export]` bypasses this module's `#[doc(hidden)]`.
+#[macro_export]
+macro_rules! align_of {
+    ($ty:ty) => {{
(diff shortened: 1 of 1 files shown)


jswrenn:
This probably could be implemented directly in terms of `align_of_val_raw`, but writing it in terms of `trailing_field_offset!` means we:
- *only* need to wait for `size_of_val_raw` to be stabilized
- don't need to relitigate the safety preconditions of using `{size,align}_of_val_raw`

src/macro_util.rs
```diff
@@ -205,6 +205,26 @@ macro_rules! trailing_field_offset {
...
+///
+/// `align_of!` produces code which is valid in a `const` context.
+// TODO(#29), TODO(https://github.com/rust-lang/rust/issues/69835): Remove this
+// `cfg` when `size_of_val_raw` is stabilized.
+#[cfg(__INTERNAL_USE_ONLY_NIGHLTY_FEATURES_IN_TESTS)]
+#[doc(hidden)] // `#[macro_export]` bypasses this module's `#[doc(hidden)]`.
+#[macro_export]
+macro_rules! align_of {
+    ($t:ty) => {{
+        #[repr(C)]
```
joshlf, line 218: Can you add a safety comment here that proves that this macro computes the correct value?

src/macro_util.rs
```diff
@@ -205,6 +205,26 @@ macro_rules! trailing_field_offset {
     }};
 }
 
+/// Computes alignment of `$ty`.
```
joshlf, line 208: Maybe also note that `$ty` doesn't need to be `Sized`?

joshlf (changes requested):

joshlf (approved):

PR #577, opened 2023-10-31
[cargo.sh] Auto-install toolchains and components

joshlf:
`cargo.sh` detects missing Rust toolchains and the `rust-src` component. When missing, it offers to install them automatically via `rustup`.

### DIFF
--- a/cargo.sh
+++ b/cargo.sh
@@ -63,6 +63,19 @@ function get-rustflags {
   [ "$1" == nightly ] && echo "--cfg __INTERNAL_USE_ONLY_NIGHLTY_FEATURES_IN_TESTS"
 }
 
+function prompt {
+  PROMPT="$1"
+  YES="$2"
+  while true; do
+    read -p "$PROMPT " yn
+    case "$yn" in
+      [Yy]) $YES; return $?; ;;
+      [Nn])       return 1;  ;;
+      *)          break;     ;;
+    esac
+  done
+}
+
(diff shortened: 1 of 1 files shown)


jswrenn (commented):
You can also just simply do:
```
rustup toolchain install $TOOLCHAIN -c rust-src
```
If the toolchain isn't installed, it installs it. If the component isn't installed, it installs it. If the toolchain and component are installed, it does nothing.

joshlf:
> You can also just simply do:
> 
> ```
> rustup toolchain install $TOOLCHAIN -c rust-src
> ```
> 
> If the toolchain isn't installed, it installs it. If the component isn't installed, it installs it. If the toolchain and component are installed, it does nothing.

I'm not sure we should install without user confirmation. Is there a way to have rustup confirm first?

joshlf:
> > You can also just simply do:
> > ```
> > rustup toolchain install $TOOLCHAIN -c rust-src
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If the toolchain isn't installed, it installs it. If the component isn't installed, it installs it. If the toolchain and component are installed, it does nothing.
> 
> I'm not sure we should install without user confirmation. Is there a way to have rustup confirm first?

Discussed offline; this still allows us to simplify the code even while prompting the user.

jswrenn (approved):

PR #578, opened 2023-10-31
Enable all stable features on the Rust playground

joshlf:
Release 0.7.21.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@
 [package]
 edition = "2018"
 name = "zerocopy"
-version = "0.7.20"
+version = "0.7.21"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -32,6 +32,9 @@ all-features = true
 pinned-stable = "1.73.0"
 pinned-nightly = "nightly-2023-10-30"
 
+[package.metadata.playground]
+features = ["__internal_use_only_features_that_work_on_stable"]
```
jswrenn, line 36: I expected this list to be longer. Do you not need to list all of the features you'd like enabled?

jswrenn: Whoops, I see what's going on.

jswrenn (approved):

PR #580, opened 2023-11-01
[cargo.sh] Set CARGO_TARGET_DIR for cargo metadata

### LINKED ISSUE #547: Use a different build directory for `cargo metadata` in `cargo.sh`
Currently, `cargo.sh` invokes `cargo metadata` using the default build directory. Since we don't pass a particular toolchain version, it is run using the default toolchain. Once `cargo metadata` builds and runs, we run the actual `cargo` command with a specific toolchain which is almost always not the default one, causing a full rebuild and a clearing of the build cache. Thus, the next time `cargo.sh` is run, `cargo metadata` is built from scratch again.

If we used a different build directory for `cargo metadata`, this wouldn't happen.

joshlf:
Closes #547

### DIFF
--- a/cargo.sh
+++ b/cargo.sh
@@ -37,7 +37,14 @@ function print-usage-and-exit {
 [[ $# -gt 0 ]] || print-usage-and-exit
 
 function pkg-meta {
-  cargo metadata --format-version 1 | jq -r ".packages[] | select(.name == \"zerocopy\").$1"
+  # NOTE(#547): We set `CARGO_TARGET_DIR` here because `cargo metadata`
+  # sometimes causes the `cargo-metadata` crate to be rebuilt from source using
+  # the default toolchain. This has the effect of clobbering any existing build
+  # artifacts from whatever toolchain the user has specified (e.g., `+nightly`),
+  # causing the subsequent `cargo` invocation to rebuild unnecessarily. By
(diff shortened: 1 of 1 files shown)


cargo.sh
```diff
@@ -37,7 +37,14 @@ function print-usage-and-exit {
 [[ $# -gt 0 ]] || print-usage-and-exit
 
 function pkg-meta {
-  cargo metadata --format-version 1 | jq -r ".packages[] | select(.name == \"zerocopy\").$1"
+  # NOTE(#547): We set `CARGO_TARGET_DIR` here because `cargo metadata`
```
joshlf, line 40: TODO: Test on the machine which exhibited this behavior to make sure it fixes the problem.

joshlf:
The machine on which this problem originally existed seems to have stopped exhibiting the problematic behavior. IMO it's still worth merging this because it's obviously correct, and will help if the problem resurfaces. In the worst case, it just won't do anything.

cargo.sh
joshlf: The machine on which this problem originally existed seems to have stopped exhibiting the problematic behavior. IMO it's still worth merging this because it's obviously correct, and will help if the problem resurfaces. In the worst case, it just won't do anything.

jswrenn (approved):

PR #583, opened 2023-11-01
[byteorder] Remove byteorder feature and crate dep

### LINKED ISSUE #438: Remove dependency on `byteorder` crate, make `byteorder` type methods `const`
# Summary

Remove our dependency on the `byteorder` crate and remove the `byteorder` feature, making the `byteorder` module available unconditinoally.

# Progress

In progress in https://github.com/google/zerocopy/pull/583.

- [x] Update [#583](https://github.com/google/zerocopy/pull/583) to have `const fn`s no longer be `const` so that it works with our MSRV
- [ ] Support functions which are conditionally `const` depending on Rust version (as prototyped in [#574](https://github.com/google/zerocopy/pull/574))
- [ ] Make byteorder functions conditionally `const`

# Description

Some potential users have expressed concerns over our dependency on the byteorder crate. I have not seen any uses of the types in our `byteorder` module which mix those types with other uses of the byteorder crate, which implies that we could just define our own `ByteOrder` trait rather than using the one from the byteorder crate and this wouldn't cause problems for users.

While we're at it, we could use a technique like the following to allow the methods on our types to be `const fn`s:

```rust
use core::marker::PhantomData;

pub trait ByteOrder {
    #[doc(hidden)]
    const ORDER: Order;
}

#[doc(hidden)]
pub enum Order {
    BigEndian,
    LittleEndian,
}

#[repr(transparent)]
pub struct U64<O>([u8; core::mem::size_of::<u64>()], PhantomData<O>);

impl<O: ByteOrder> U64<O> {
    pub const fn new(u: u64) -> U64<O> {
        let bytes = match O::ORDER {
            Order::BigEndian => u.to_be_bytes(),
            Order::LittleEndian => u.to_le_bytes(),
        };
        
        U64(bytes, PhantomData)
    }
}
```

joshlf:
Remove the `byteorder` feature and the dependency on the `byteorder` crate. Replace the `ByteOrder` trait and types which implement it from the `byteorder` crate with our own native implementations. This allows us to make some of our functions and methods `const`.

Add `Usize` and `Isize` byte order-aware types.

Closes #438

### DIFF
--- a/src/byteorder.rs
+++ b/src/byteorder.rs
@@ -71,15 +71,94 @@ use core::{
     convert::{TryFrom, TryInto},
     fmt::{self, Binary, Debug, Display, Formatter, LowerHex, Octal, UpperHex},
     marker::PhantomData,
+    mem,
     num::TryFromIntError,
 };
 
-// We don't reexport `WriteBytesExt` or `ReadBytesExt` because those are only
-// available with the `std` feature enabled, and zerocopy is `no_std` by
-// default.
-pub use ::byteorder::{BigEndian, ByteOrder, LittleEndian, NativeEndian, NetworkEndian, BE, LE};
-
 use super::*;
+
+/// A type-level representation of byte order.
+///
+/// This type is implemented by [`BigEndian`] and [`LittleEndian`], which
+/// represent big-endian and little-endian byte order respectively. This module
+/// also provides a number of useful aliases for those types: [`NativeEndian`],
+/// [`NetworkEndian`], [`BE`], and [`LE`].
+///
+/// `ByteOrder` types can be used to specify the byte order of the types in this
+/// module - for example, [`U32<BigEndian>`] is a 32-bit integer stored in
+/// big-endian byte order.
+///
+/// [`U32<BigEndian>`]: U32
+pub trait ByteOrder: Copy + Clone + Debug + Display + Eq + PartialEq + Ord + PartialOrd {
+    #[doc(hidden)]
+    const ORDER: Order;
+}
+
+#[allow(missing_copy_implementations, missing_debug_implementations)]
+#[doc(hidden)]
+pub enum Order {
+    BigEndian,
+    LittleEndian,
+}
+
+/// Big-endian byte order.
+///
+/// See [`ByteOrder`] for more details.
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -301,6 +288,9 @@ pub use crate::util::ptr::Ptr;
 #[allow(unused_imports)]
 use crate::util::polyfills::NonNullExt as _;
 
+#[cfg(feature = "byteorder")]
+compile_error!("the `byteorder` feature has been removed; its functionality is now available regardless of feature set");
```
joshlf, line 292: This will cause problems for anyone using `--all-features`, and that includes us: our Kani tests are currently failing for this reason.

One option would be to have this emit a deprecation warning rather than an error. In our CI tests, we could `-Dwarnings` (as we do today) but also add `-Adeprecated` so that we can test any of our own deprecated items such as this one.

joshlf: Discussed offline with @jswrenn. Decided to just remove this feature entirely on the theory that most users aren't relying on it directly (since, on 0.7.X, it's a default feature) and for those that are, it should be obvious what to do.

src/byteorder.rs
```diff
@@ -498,37 +594,74 @@ define_type!(
...
+    [],
+    [],
+    [],
+    []
+);
+define_type!(
+    A,
+    Usize,
+    usize,
+    mem::size_of::<usize>() * 8,
```
joshlf, line 627: The rustdoc CI failures seem to come from the fact that, in `define_type!`, we use `stringify!` on this `expr` in order to generate the doc comment.

I think the simplest thing to do here might just be to introduce a new macro argument that is used in the summary sentence of the doc comment instead of doing `stringify!($bits)`. It would have the text, "A 16-bit unsigned integer" , "A word-sized signed integer", etc. i.e.:

### Old

```rust
        doc_comment! {
            concat!("A ", stringify!($bits), "-bit ", $number_kind,
            " stored in a given byte order.
```

### New

```rust
        doc_comment! {
            concat!($description, " stored in a given byte order.
```

This would also allow us to get rid of the `$number_kind` argument.

jswrenn (approved):

PR #586, opened 2023-11-02
[macro_util] Add comment to core_reexport

joshlf:
Add a comment explaining why we don't change core_reexport to `pub use core as core_reexport`.


jswrenn (changes requested):
Nit: We don't need to *release* a semver-breaking version, we merely need to bump the version in our `Cargo.toml` to `0.8` in `main`.

joshlf:
> Nit: We don't need to _release_ a semver-breaking version, we merely need to bump the version in our `Cargo.toml` to `0.8` in `main`.

Done.

jswrenn (approved):

PR #593, opened 2023-11-03
Add slice-based methods to FromBytes

newpavlov:
Right now slice conversion methods are somewhat buried in `Ref`. For better ergonomics and discoverability I think it's worth to have the slice-based methods on the `FromBytes` trait.


src/lib.rs
```diff
@@ -1317,6 +1317,132 @@ pub unsafe trait FromBytes: FromZeroes {
         Ref::<&mut [u8], Self>::new_from_suffix(bytes).map(|(_, r)| r.into_mut())
     }
 
+    /// Interprets the given `bytes` as a `&[Self]` without copying.
+    ///
+    /// If `bytes.len() % size_of::<T>() != 0` or `bytes` is not aligned to
+    /// `align_of::<T>()`, this returns `None`.
+    ///
+    /// If you need to convert a specific number of slice elements, see
+    /// [`FromBytes::slice_from_prefix`] or [`FromBytes::slice_from_suffix`].
```
joshlf, line 1326: Sorry, uber nit: Can you have these omit the trait name and then use markdown out-of-line link syntax? Given that these will generate intra-page doc links, I think the trait name prefix is redundant. I.e.:

```rust
/// ... [`slice_from_prefix`] ...
///
/// [`slice_from_prefix`]: FromBytes::slice_from_prefix
```

newpavlov:
Done! I've used ```[`FromBytes::slice_from_prefix`]``` instead of ```[`slice_from_prefix`]```, but I can replace it to ```[`slice_from_prefix`](FromBytes::slice_from_prefix)``` if you prefer it.

joshlf:
> Done! I've used `` [`FromBytes::slice_from_prefix`] `` instead of `` [`slice_from_prefix`] ``, but I can replace it to `` [`slice_from_prefix`](FromBytes::slice_from_prefix) `` if you prefer it.

Oops, just saw this. Yeah, that'd be preferable, thanks!

joshlf:
Okay, looks good! Could you update `Cargo.toml` and `zerocopy-derive/Cargo.toml` to bump the version number to 0.7.24 so I can release this once it merges to `main`? There are four locations you'll need to update in `Cargo.toml` and one in `zerocopy-derive/Cargo.toml`.

newpavlov:
Done!

joshlf (approved):
Sweet, thanks!

joshlf:
[Published](https://crates.io/crates/zerocopy/0.7.24)!

PR #620, opened 2023-11-15
Update safety comments for `NonZeroXxx` types

joshlf:
Makes progress on #429


src/lib.rs
```diff
@@ -1832,26 +1832,28 @@ safety_comment! {
...
-    ///   `Option<NonZeroU8>` and `Option<NonZeroI8>` both have size 1. [1] [2]
-    ///   This is worded in a way that makes it unclear whether it's meant as a
-    ///   guarantee, but given the purpose of those types, it's virtually
-    ///   unthinkable that that would ever change. `Option` cannot be smaller
-    ///   than its contained type, which implies that, and `NonZeroX8` are of
-    ///   size 1 or 0. `NonZeroX8` can represent multiple states, so they cannot
-    ///   be 0 bytes, which means that they must be 1 byte. The only valid
-    ///   alignment for a 1-byte type is 1.
+    /// `NonZeroXxx` has the same layout and bit validity as its associated
+    /// primitive with the exception that 0 is not a valid instance. [1]
```
jswrenn, line 1836: Can you also cite documentation that primitive number types are represented in two's complement? In principle, the value `0u8` in a programming language does not *necessarily* need to be represented as `0b0000_0000`, but that's exactly what matters here.

joshlf: I can't actually find where this is documented (e.g. of `u16` or `i16`). Are you able to find any documentation for it?

joshlf: Bump. Are you comfortable just merging as-is?

PR #633, opened 2023-11-21
Add `DstLayout::extend`

jswrenn:
`DstLayout::extend` is comparable to `alloc::Layout`, except that it additionally handles DSTs. This is the first of two steps needed for `DstLayout` to support an API in the form of:

```rust
struct ReprCStruct<T, U, V> {
    foo: Foo<T>,
    bar: Bar<U>,
    baz: Baz<V>,
}

    const LAYOUT: DstLayout = {
        use ::zerocopy::DstLayout;

        let min_align = NonZeroUsize::new(2); // or `None`, w/o `align`
        let max_align = NonZeroUsize::new(4); // or `None`, w/o `packed`

        DstLayout::for_type::<()>()
            .extend(<Foo<T> as KnownLayout>::LAYOUT, max_align)
            .extend(<Bar<T> as KnownLayout>::LAYOUT, max_align)
            .extend(<Baz<T> as KnownLayout>::LAYOUT, max_align)
            .pad_to_align(min_align, max_align)
    };
}
```
(The next step is implementing `DstLayout::pad_to_align`). 

Despite the similarities between `DstLayout::extend` and `Layout::extend`, the former cannot be implemented in terms of the latter because `Layout::extend` is not `const`. This introduces a risk that our computations here diverge from those in `Layout::extend`, but this PR also includes Kani proofs that `DstLayout::extend` behaves comparably to `Layout::extend`.

Makes progress towards #29.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -283,10 +283,10 @@ use core::{
 #[cfg(feature = "alloc")]
 extern crate alloc;
 #[cfg(feature = "alloc")]
-use {
-    alloc::{boxed::Box, vec::Vec},
-    core::alloc::Layout,
-};
+use alloc::{boxed::Box, vec::Vec};
+
+#[cfg(any(feature = "alloc", kani))]
+use core::alloc::Layout;
 
 // For each polyfill, as soon as the corresponding feature is stable, the
 // polyfill import will be unused because method/function resolution will prefer
@@ -305,6 +305,9 @@ const _: () = {
     #[warn(deprecated)]
     _WARNING
 };
+
+/// The target pointer width, counted in bits.
+const POINTER_WIDTH_BITS: usize = mem::size_of::<usize>() * 8;
 
 /// The layout of a type which might be dynamically-sized.
 ///
@@ -335,19 +338,19 @@ const _: () = {
 /// [the reference]: https://doc.rust-lang.org/reference/type-layout.html
 #[doc(hidden)]
 #[allow(missing_debug_implementations, missing_copy_implementations)]
-#[cfg_attr(test, derive(Copy, Clone, Debug, PartialEq, Eq))]
+#[cfg_attr(any(kani, test), derive(Copy, Clone, Debug, PartialEq, Eq))]
 pub struct DstLayout {
     _align: NonZeroUsize,
     _size_info: SizeInfo,
 }
 
-#[cfg_attr(test, derive(Copy, Clone, Debug, PartialEq, Eq))]
+#[cfg_attr(any(kani, test), derive(Copy, Clone, Debug, PartialEq, Eq))]
 enum SizeInfo<E = usize> {
     Sized { _size: usize },
     SliceDst(TrailingSliceLayout<E>),
 }
 
-#[cfg_attr(test, derive(Copy, Clone, Debug, PartialEq, Eq))]
+#[cfg_attr(any(kani, test), derive(Copy, Clone, Debug, PartialEq, Eq))]
 struct TrailingSliceLayout<E = usize> {
     // The offset of the first byte of the trailing slice field. Note that this
     // is NOT the same as the minimum size of the type. For example, consider
@@ -393,6 +396,29 @@ pub enum _CastType {
 }
 
 impl DstLayout {
+    /// The maximum theoretic possible alignment of a type.
+    ///
+    /// For compatibility with future Rust versions, this is defined as the
+    /// maximum power-of-two that fits into a `usize`. See also
+    /// [`DstLayout::CURRENT_MAX_ALIGN`].
+    const THEORETICAL_MAX_ALIGN: NonZeroUsize =
+        match NonZeroUsize::new(1 << (POINTER_WIDTH_BITS - 1)) {
+            Some(max_align) => max_align,
+            None => unreachable!(),
+        };
+
+    /// The current, documented max alignment of a type \[1\].
+    ///
+    /// \[1\] Per <https://doc.rust-lang.org/reference/type-layout.html#the-alignment-modifiers>:
+    ///
+    ///   The alignment value must be a power of two from 1 up to
+    ///   2<sup>29</sup>.
+    #[cfg(not(kani))]
+    const CURRENT_MAX_ALIGN: NonZeroUsize = match NonZeroUsize::new(1 << 28) {
+        Some(max_align) => max_align,
+        None => unreachable!(),
+    };
+
     /// Constructs a `DstLayout` which describes `T`.
     ///
     /// # Safety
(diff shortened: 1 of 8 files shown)


src/lib.rs
```diff
@@ -393,6 +393,17 @@ pub enum _CastType {
 }
 
 impl DstLayout {
+    /// The maximum theoretic possible alignment of a type (1).
+    ///
+    /// (1) Per https://doc.rust-lang.org/reference/type-layout.html#the-alignment-modifiers:
```
joshlf: Assuming that your use of `(1)` is to avoid `[1]` being formatted as Markdown:

```suggestion
    /// The maximum theoretic possible alignment of a type \[1\].
    ///
    /// \[1\] Per https://doc.rust-lang.org/reference/type-layout.html#the-alignment-modifiers:
```

Alternatively, since we never actually rustdoc this, you could just change this to be a non-doc comment (and also you could change `2<sup>29</sup>` to `2^29`, which is more readable in the source code).

src/lib.rs
```diff
@@ -439,6 +450,98 @@ impl DstLayout {
         }
     }
 
+    /// Extends `self` with the layout of `field`, adding whatever padding is
+    /// needed to satisfy the field's natural alignment or `max_align`
+    /// (whichever comes first). If `self` corresponds to a type marked with
```
joshlf, line 455:
```suggestion
    /// (whichever is lower). If `self` corresponds to a type marked with
```

src/lib.rs
```diff
@@ -439,6 +450,98 @@ impl DstLayout {
...
+    /// `repr(packed(N))`, then `max_align` should be `Some(N)`, otherwise
+    /// `None`.
+    ///
+    /// This method does **not** add trailing padding needed to satisfy the
+    /// type's overall alignment requirements.
+    ///
+    ///
```
joshlf, line 462:
```suggestion
```

src/lib.rs
```diff
@@ -439,6 +450,98 @@ impl DstLayout {
...
+    #[allow(dead_code)]
+    pub(crate) const fn extend(
+        self,
+        field: DstLayout,
+        packed_modifier: Option<NonZeroUsize>,
```
joshlf, line 467: Feel free to take or leave this; it strikes me that this is a bit more self-explanatory for programmers who are reasoning about this method in terms of its use in a derive.

```suggestion
        repr_packed: Option<NonZeroUsize>,
```

Alternatively, if you prefer this not be tied to its intended use, maybe this might be better:

```suggestion
        max_align: Option<NonZeroUsize>,
```

src/lib.rs
```diff
@@ -439,6 +450,98 @@ impl DstLayout {
...
+
+        // If `max_align` is `None`, there are no alignment constraints, and the
+        // value can be defaulted to `MAX_ALIGN`.
+        let max_align = match packed_modifier {
+            Some(max_align) => max_align,
+            None => Self::MAX_ALIGN,
+        };
+
+        debug_assert!(max_align.is_power_of_two());
+        debug_assert!(self._align.get() <= max_align.get());
```
joshlf: Also assert that `field._align.get() <= Self::MAX_ALIGN`?

src/lib.rs
```diff
@@ -439,6 +450,98 @@ impl DstLayout {
...
+
+        // The field's alignment is clamped by `max_align` (i.e., the `packed`
```
joshlf, line 481:
```suggestion
        // The field's alignment is clamped by `max_align` (i.e., the `repr(packed)`
```

src/lib.rs
```diff
@@ -439,6 +450,98 @@ impl DstLayout {
         }
     }
 
+    /// Extends `self` with the layout of `field`, adding whatever padding is
```
joshlf, line 453: This comment should probably call out somewhere that the result of this method is only valid if the struct's field order is guaranteed. It might be worth writing a `# Safety` section documenting what correctness guarantees are provided, and with what preconditions.

src/lib.rs
```diff
@@ -439,6 +450,98 @@ impl DstLayout {
...
+            SizeInfo::Sized { _size: preceeding_size } => {
+                // Compute the minimum amount of inter-field padding needed to
+                // satisfy the field's alignment, and offset of the trailing
+                // field. [1]
+                //
+                // [1] Per https://doc.rust-lang.org/reference/type-layout.html#the-alignment-modifiers:
+                //
+                //  Inter-field padding is guaranteed to be the minimum required
+                //  in order to satisfy each field's (possibly altered)
+                //  alignment.
```
joshlf, line 509: Indent this by two spaces like you do above?

src/lib.rs
```diff
@@ -439,6 +450,98 @@ impl DstLayout {
...
+                let padding = _padding_needed_for(preceeding_size, field_align);
+                #[allow(clippy::arithmetic_side_effects)]
+                let offset = preceeding_size + padding;
```
joshlf, line 512: Comment proving that this won't panic?

src/lib.rs
```diff
@@ -439,6 +450,98 @@ impl DstLayout {
...
+
+                match field._size_info {
+                    SizeInfo::Sized { _size: field_size } => {
+                        // If the trailing field is sized, the resulting layout
+                        // will be sized. Its size will be the sum of the
+                        // preceeding layout, the size of the new field, and the
+                        // size of interfield padding between the two.
```
joshlf, line 519:
```suggestion
                        // size of inter-field padding between the two.
```

src/lib.rs
```diff
@@ -439,6 +450,98 @@ impl DstLayout {
...
+                        #[allow(clippy::arithmetic_side_effects)]
+                        SizeInfo::Sized { _size: offset + field_size }
+                    }
+                    SizeInfo::SliceDst(TrailingSliceLayout {
+                        _offset: trailing_offset,
+                        _elem_size,
+                    }) => {
+                        // If the trailing field is dynamically sized, so to
+                        // will the resulting layout. The offset of the trailing
```
joshlf, line 528:
```suggestion
                        // If the trailing field is dynamically sized, so too
                        // is the resulting layout. The offset of the trailing
```

src/lib.rs
```diff
@@ -439,6 +450,98 @@ impl DstLayout {
...
+                        // slice component is the sum of the offset of the
+                        // trailing field and the trailing offset within that
```
joshlf, line 530:
```suggestion
                        // trailing field and the trailing slice offset within that
```

src/lib.rs
```diff
@@ -439,6 +450,98 @@ impl DstLayout {
...
+                        // field.
+                        #[allow(clippy::arithmetic_side_effects)]
+                        SizeInfo::SliceDst(TrailingSliceLayout {
+                            _offset: offset + trailing_offset,
```
joshlf, line 534: Comment proving that this won't panic?

joshlf (changes requested):

src/lib.rs
jswrenn: I didn't mean to assert `debug_assert!(self._align.get() <= max_align.get());`, and revised the PR accordingly. I don't think we need to peek into `self` and `field` and make sure they're valid — that's the job of `DstLayout`'s constructors.

src/lib.rs
joshlf: Following up on this, we may need to be very pedantic about the preconditions. It's roughly that the type needs to be `repr(C)`, but even that's not technically sufficient - e.g., a `repr(u8)` enum has a layout in which each of its variants is laid out _as if_ it was a `repr(C)` struct. Maybe instead of making reference to something like `repr(C)`, we can just *describe* the algorithm (ie, fields are laid out in order and separated by the minimum possible padding) and require that the type's layout follow that algorithm? Regardless of what approach we take, it will take some very careful and precise language to articulate when this method's result is correct.

Then, in the comments in the body, we may need to be more pedantic in order to articulate why the implementation is correct in terms of the pre- and post-conditions.

src/lib.rs
```diff
@@ -3918,6 +4021,92 @@ mod tests {
         }
     }
 
+    mod dst_layout {
```
joshlf, line 4024: IMO it'd be cleaner if you just put these in the parent module and called added a `test_dst_layout` prefix to their names. If we want to consider a broader overhaul of how these tests are structured, that's fine, but for now we should be consistent with the existing convention - IMO consistency is more important.

joshlf, line 4024: These tests should probably also exercise calling `extend` with a maximum alignment argument.

src/lib.rs
```diff
@@ -5703,3 +5892,213 @@ mod tests {
...
+            // `DstLayout` that can correspond to actual Rust types. We use
+            // `Layout` to verify that our `DstLayout` satisfies the validity
+            // conditions of Rust layouts.
+            kani::assume(
+                match size_info {
+                    SizeInfo::Sized { _size } => Layout::from_size_align(_size, align.get()),
+                    SizeInfo::SliceDst(TrailingSliceLayout { _offset, _elem_size }) => {
+                        // `SliceDst`` cannot encode an exact size, but we know
+                        // it is at least `_offset` bytes.
+                        Layout::from_size_align(_offset, align.get())
```
joshlf, line 5918: This is too strict - it's valid to have types whose trailing field offset does not satisfy the overall type's alignment. E.g.:

```rust
#[repr(C)]
struct Foo {
    a: u64,   // at byte offset 0
    b: u8,    // at byte offset 8
    c: [u16], // at byte offset 10
}
```

This type has alignment 8, but the trailing slice offset is 10, which is not a multiple of 8.

src/lib.rs
```diff
@@ -5703,3 +5892,213 @@ mod tests {
...
+            let offset: usize = kani::any();
+
+            kani::assume(elem_size < isize::MAX as _);
+            kani::assume(offset < isize::MAX as _);
+
+            TrailingSliceLayout { _elem_size: elem_size, _offset: offset }
+        }
+    }
+
+    mod dst_layout {
```
joshlf, line 5957: Same here - I'd advocate for being consistent w/ our existing conventions and changing them wholesale later if we decide to do that.

src/lib.rs
```diff
@@ -5703,3 +5892,213 @@ mod tests {
...
+            let base: DstLayout = kani::any();
+            let field: DstLayout = kani::any();
+            let packed: Option<NonZeroUsize> = kani::any();
+
+            if let Some(max_align) = packed {
+                kani::assume(max_align.is_power_of_two());
+                kani::assume(base._align <= max_align);
+            }
+
+            // the base can only be extended if it's sized
```
joshlf, line 5974: To be consistent with our CONTRIBUTING.md style guidelines:

```suggestion
            // The base can only be extended if it's sized.
```

src/lib.rs
```diff
@@ -5703,3 +5892,213 @@ mod tests {
...
+            assert_eq!(composite._align, max(base._align, field_align));
+
+            // Compute the minimum amount of inter-field padding needed to
+            // satisfy the field's alignment, and offset of the trailing field.
+            // [1]
+            //
+            // [1] Per https://doc.rust-lang.org/reference/type-layout.html#the-alignment-modifiers:
+            //
+            //  Inter-field padding is guaranteed to be the minimum required in
+            //  order to satisfy each field's (possibly altered) alignment.
```
joshlf, line 6006: Indent by two spaces to be consistent with other comments?

joshlf (changes requested):

src/lib.rs
joshlf: We don't currently document anywhere that `DstLayout`'s `align` field is guaranteed to have a particular maximum value - that's just a consequence of Rust's type rules. IMO we should either explicitly document that as an invariant and add comments at construction sites proving that it's upheld, or we should assert it here out of an abundance of caution. As this code is currently written, it could break in the future if Rust raises the maximum alignment.

src/lib.rs
jswrenn: `Layout::from_size_align` does not enforce that size satisfies alignment, since it's intended to be used for representing fragments of layouts.

src/lib.rs
jswrenn: I've made three changes:
1. I've verified, with Kani, that this method is robust to future increases of Rust's maximum-allowed alignment.
2. I've added debug assertions that fire if we see an alignment exceeding Rust's current maximum-allowed alignment.
3. I've replaced the `+` operations with explicit checked adds, and panic messages explaining what went wrong.

src/lib.rs
```diff
@@ -439,6 +465,145 @@ impl DstLayout {
...
+    ///
+    /// We make no guarantees to the behavior of this method if these fragments
+    /// cannot appear in a valid Rust type (e.g., the concatenation of the
+    /// layouts would lead to a size larger than `isize::MAX`).
+    #[allow(dead_code)]
+    pub(crate) const fn extend(self, field: DstLayout, repr_packed: Option<NonZeroUsize>) -> Self {
+        use util::{core_layout::_padding_needed_for, max, min};
+
+        // If `repr_packed` is `None`, there are no alignment constraints, and
+        // the value can be defaulted to `MAX_ALIGN`.
```
joshlf, line 498:
```suggestion
        // the value can be defaulted to `THEORETICAL_MAX_ALIGN`.
```

src/lib.rs
```diff
@@ -393,6 +396,29 @@ pub enum _CastType {
...
+        };
+
+    /// The current, documented max alignment of a type \[1\].
+    ///
+    /// \[1\] Per <https://doc.rust-lang.org/reference/type-layout.html#the-alignment-modifiers>:
+    ///
+    ///   The alignment value must be a power of two from 1 up to
+    ///   2<sup>29</sup>.
+    #[cfg(not(kani))]
+    const CURRENT_MAX_ALIGN: NonZeroUsize = match NonZeroUsize::new(1 << 28) {
```
joshlf, line 417: Could we add a compile-fail test to verify that this is accurate (ie, a type with `repr(align(2^30))`)? I worry that we won't detect issues in practice because we're using debug (rather than non-debug) assertions.

src/lib.rs
```diff
@@ -439,6 +465,145 @@ impl DstLayout {
...
+        // The struct's alignment is the maximum of its previous alignment and
+        // `field_align`.
+        let align = max(self._align, field_align);
+
+        let size_info = match self._size_info {
+            // If the layout is already a DST, we panic; DSTs cannot be extended
+            // with additional fields.
+            SizeInfo::SliceDst(..) => panic!("Cannot extend a DST with additional fields."),
+
+            SizeInfo::Sized { _size: preceeding_size } => {
```
joshlf, line 537:
```suggestion
            SizeInfo::Sized { _size: preceding_size } => {
```

src/lib.rs
```diff
@@ -439,6 +465,145 @@ impl DstLayout {
...
+                        //
+                        // This will not panic (and is proven with Kani to not
+                        // panic) if the layout components can correspond to a
+                        // leading layout fragment of a valid Rust type, but may
+                        // panic otherwise (e.g., combining or aligning the
+                        // components would create a size exceeding
+                        // `isize::MAX`).
+                        let size = match offset.checked_add(field_size) {
+                            Some(size) => size,
+                            None => panic!("`field` cannot be appended without the total size overflowing `isize::MAX`"),
```
joshlf, line 574: This case is only triggered if the total overflows `usize`, not `isize`, right?

This raises an interesting question: Should we be doing more to guarantee that we don't overflow `isize`? Maybe validate using Kani that the output size fits in `isize`?

joshlf (changes requested):

src/lib.rs
jswrenn: I've fixed the panic message.

> Should we be doing more to guarantee that we don't overflow `isize`?

Maybe, though I'd argue doing so doesn't buy us much. We currently prove with kani that across `Layout::extend`'s domain, `DstLayout::extend` behaves analogously (among other correctness properties that guard against the possibility of a bug arising in `Layout::extend`).

We prove this by constructing `Layout` analogs to the `DstLayout` constituents, combining them with `Layout::extend`, and then verifying that the resulting `Layout` composite matches the `DstLayout` composite; e.g.:
```rust
if let Ok((actual_composite, actual_offset)) = base_analog.extend(field_analog)
{
    assert_eq!(actual_offset, offset);
    assert_eq!(actual_composite.size(), composite_size);
    assert_eq!(actual_composite.align(), composite._align.get());
} else {
    // An error here reflects that composite of `base`
    // and `field` cannot correspond to a real Rust type
    // fragment, because such a fragment would violate
    // the basic invariants of a valid Rust layout. At
    // the time of writing, `DstLayout` is a little more
    // permissive than `Layout`, so we don't assert
    // anything in this branch (e.g., unreachability).
}
```

This is enough to support the *Safety* section in `DstLayout::extend`'s documentation (and our use-case for this method):

> If a (potentially hypothetical) valid `repr(C)` Rust type begins with fields whose layout are `self`, and those fields are immediately followed by a field whose layout is `field`, then unsafe code may rely on `self.extend(field, repr_packed)` producing a layout that correctly encompasses those two components.
>
> We make no guarantees to the behavior of this method if these fragments cannot appear in a valid Rust type (e.g., the concatenation of the layouts would lead to a size larger than `isize::MAX`).

We *could* establish safety invariants on `DstLayout`'s fields (e.g., alignment is valid) and rigorously enforce those conditions, but we would need to weigh that against the maintenance costs of the potential for extra unsafety/unsoundness.

Given our currently limited use of `DstLayout` — only describing real Rust types — I think that we should punt on adding these safety conditions until we actually need them.

src/lib.rs
```diff
@@ -393,6 +396,29 @@ pub enum _CastType {
 }
 
 impl DstLayout {
+    /// The maximum theoretic possible alignment of a type.
+    ///
+    /// For compatibility with future Rust versions, this is defined as the
+    /// maximum power-of-two that fits into a `usize`. See also
+    /// [`DstLayout::CURRENT_MAX_ALIGN`].
+    const THEORETICAL_MAX_ALIGN: NonZeroUsize =
+        match NonZeroUsize::new(1 << (POINTER_WIDTH_BITS - 1)) {
```
jswrenn, line 405: This can be expressed more simply as:
```suggestion
        match NonZeroUsize::new(isize::MIN as usize) {
```
...which would let us avoid introducing the `POINTER_WIDTH_BITS` constant. 

Is this too magical, @joshlf?

joshlf: If you're hoping to avoid the constant, maybe just inline its definition? It'd accomplish that goal while keeping this pretty self-explanatory.

jswrenn: I did that initially, but it looked a little too messy to my eyes.

src/lib.rs
```diff
@@ -3918,6 +4083,103 @@ mod tests {
         }
     }
 
+    #[allow(clippy::decimal_literal_representation)]
+    #[test]
+    fn test_dst_layout_extend_sized_with_sized() {
```
joshlf, line 4088: Comment explaining what this test is doing?

src/lib.rs
```diff
@@ -3918,6 +4083,103 @@ mod tests {
...
+        test_align_is_size!(16777216);
+        test_align_is_size!(33554432);
+        test_align_is_size!(67108864);
+        test_align_is_size!(33554432);
+        test_align_is_size!(134217728);
+        test_align_is_size!(268435456);
+    }
+
+    #[test]
+    fn test_dst_layout_extend_sized_with_dst() {
```
joshlf, line 4145: Comment explaining what this test is doing?

src/lib.rs
joshlf: Ack.

src/lib.rs
joshlf: Up to you, but IMO `isize::MIN as usize` is too magical.

joshlf (approved):

PR #638, opened 2023-11-24
Add `DstLayout::pad_to_align`

jswrenn:
`DstLayout::pad_to_align` is comparable to `alloc::Layout::pad_to_align`, except that it additionally handles DSTs. This is the second of two steps needed for `DstLayout` to support an API in the form of:

```rust
struct ReprCStruct<T, U, V> {
    foo: Foo<T>,
    bar: Bar<U>,
    baz: Baz<V>,
}

    const LAYOUT: DstLayout = {
        use ::zerocopy::DstLayout;

        let min_align = NonZeroUsize::new(2); // or `None`, w/o `align`
        let max_align = NonZeroUsize::new(4); // or `None`, w/o `packed`

        DstLayout::new(min_align)
            .extend(<Foo<T> as KnownLayout>::LAYOUT, max_align)
            .extend(<Bar<T> as KnownLayout>::LAYOUT, max_align)
            .extend(<Baz<T> as KnownLayout>::LAYOUT, max_align)
            .pad_to_align()
    };
}
```
(The first step was implementing `DstLayout::extend` in #633). 

Despite the similarities between `DstLayout::pad_to_align` and `Layout::pad_to_align`, the former cannot be implemented in terms of the latter because `Layout::pad_to_align` is not `const`. This introduces a risk that our computations here diverge from those in `Layout::pad_to_align`, but this PR also includes Kani proofs that `DstLayout::pad_to_align` behaves comparably to `Layout::pad_to_align`.

Makes progress towards #29.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -604,6 +604,55 @@ impl DstLayout {
         DstLayout { _align: align, _size_info: size_info }
     }
 
+    /// Like `Layout::pad_to_align`, this routine rounds the size of this layout
+    /// up to the nearest multiple of this type's alignment or `repr_packed`
+    /// (whichever is less). This method leaves DST layouts unchanged, since the
+    /// trailing padding of DSTs is computed at runtime.
+    ///
+    /// In order to match the layout of a `#[repr(C)]` struct, this method
+    /// should be invoked after the invocations of [`DstLayout::extend`]. If
+    /// `self` corresponds to a type marked with `repr(packed(N))`, then
+    /// `repr_packed` should be set to `Some(N)`, otherwise `None`.
+    ///
+    /// This method cannot be used to match the layout of a record with the
+    /// default representation, as that representation is mostly unspecified.
+    ///
+    /// # Safety
+    ///
+    /// If a (potentially hypothetical) valid `repr(C)` type begins with fields
+    /// whose layout are `self` followed only by zero or more bytes of trailing
+    /// padding (not included in `self`), then unsafe code may rely on
+    /// `self.pad_to_align(repr_packed)` producing a layout that correctly
+    /// encapsulates the layout of that type.
+    ///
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -604,6 +604,55 @@ impl DstLayout {
         DstLayout { _align: align, _size_info: size_info }
     }
 
+    /// Like `Layout::pad_to_align`, this routine rounds the size of this layout
+    /// up to the nearest multiple of this type's alignment or `repr_packed`
+    /// (whichever is less). This method leaves DST layouts unchanged, since the
+    /// trailing padding of DSTs is computed at runtime.
+    ///
+    /// In order to match the layout of a `#[repr(C)]` struct, this method
+    /// should invoked after the invocations of [`DstLayout::extend`]. If `self`
```
joshlf, line 613:
```suggestion
    /// should be invoked after the invocations of [`DstLayout::extend`]. If `self`
```

src/lib.rs
```diff
@@ -604,6 +604,55 @@ impl DstLayout {
...
+                    Some(size) => size,
+                    None => panic!("Adding padding caused size to overflow `usize`."),
+                };
+                SizeInfo::Sized { _size: size }
+            }
+            // For DST layouts, trailing padding depends on the length of the
+            // trailing DST and is computed at runtime. This does not alter the
+            // offset or element size of the layout, so we leave `size_info`
+            // unchanged.
+            SizeInfo::SliceDst(..) => self._size_info,
```
joshlf, line 650: To lean a bit more into the whole ADT recursive destructuring thing:

```suggestion
            size_info @ SizeInfo::SliceDst(..) => size_info,
```

joshlf (changes requested):
Two tiny nits, but otherwise looks great!

joshlf (approved):

PR #643, opened 2023-11-27
Support `derive(KnownLayout)` on DSTs

jswrenn:
Makes progress towards #29. Supersedes #541.

For types marked `#[repr(C)]`, this derive requires that the trailing field is `KnownLayout`:
```rust
struct ReprCStruct<T, U, V> {
    foo: Foo<T>,
    bar: Bar<U>,
    baz: Baz<V>,
}

    const LAYOUT: DstLayout = {
        use ::zerocopy::{DstLayout, KnownLayout};

        let repr_align = NonZeroUsize::new(2); // or `None`, w/o `align`
        let repr_packed = NonZeroUsize::new(4); // or `None`, w/o `packed`

        DstLayout::new(repr_align)
            .extend(DstLayout::for_type::<Foo<T>>(), repr_packed)
            .extend(DstLayout::for_type::<Bar<U>>(), repr_packed)
            .extend(<Baz<V> as KnownLayout>::LAYOUT, repr_packed)
            .pad_to_align()
    };
}
```

For non-`repr(C)` layouts, it's only required that `Self: Sized`.


src/lib.rs
```diff
@@ -419,6 +425,30 @@ impl DstLayout {
...
+    /// # Panics
+    ///
+    /// This function panics if the supplied `repr_align` is not a power of two.
+    ///
+    /// # Safety
+    ///
+    /// Unsafe code may assume that the contract of this function satisfied.
+    #[doc(hidden)]
+    #[inline]
+    pub const fn new(repr_align: Option<NonZeroUsize>) -> DstLayout {
```
joshlf, line 441:
```suggestion
    pub const fn new_zst(repr_align: Option<NonZeroUsize>) -> DstLayout {
```

src/lib.rs
```diff
@@ -419,6 +425,30 @@ impl DstLayout {
...
+    /// alignment (or 1). If `repr_align` is provided, then it must be a power
+    /// of two.
+    ///
+    /// # Panics
+    ///
+    /// This function panics if the supplied `repr_align` is not a power of two.
+    ///
+    /// # Safety
+    ///
+    /// Unsafe code may assume that the contract of this function satisfied.
```
joshlf, line 438:
```suggestion
    /// Unsafe code may assume that the contract of this function is satisfied.
```

zerocopy-derive/src/ext.rs
```diff
@@ -6,12 +6,18 @@
...
 
 pub trait DataExt {
     /// Extract the types of all fields. For enums, extract the types of fields
     /// from each variant.
     fn field_types(&self) -> Vec<&Type>;
+
+    /// Extract the single trailing field of a struct, if any.
+    fn trailing_field(&self) -> Option<(syn::Member, &Type)> {
+        None
+    }
```
joshlf, line 20: Discussed offline: Maybe we can remove this and other changes to this file?

zerocopy-derive/src/lib.rs
```diff
@@ -62,10 +73,112 @@ use {crate::ext::*, crate::repr::*};
...
+                Some(reprs)
+            } else {
+                None
+            }
+        }
+        Data::Enum(..) | Data::Union(..) => None,
+    };
+
+    let fields = ast.data.field_types();
+    let fields = fields.split_last();
```
joshlf, line 90: Maybe a more descriptive name for this variable or inline this into its usage site below?

zerocopy-derive/src/lib.rs
```diff
@@ -62,10 +73,112 @@ use {crate::ext::*, crate::repr::*};
...
+                    use ::zerocopy::{DstLayout, KnownLayout};
+
+                    let repr_align = #repr_align;
+                    let repr_packed = #repr_packed;
+
+                    DstLayout::new(repr_align)
+                        #(.extend(DstLayout::for_type::<#leading_fields>(), repr_packed))*
+                        .extend(<#trailing_field as KnownLayout>::LAYOUT, repr_packed)
+                        .pad_to_align()
+                }),
```
joshlf, line 130: Not sure exactly what it should say, but some kind of safety comment here?

zerocopy-derive/src/lib.rs
```diff
@@ -62,10 +73,112 @@ use {crate::ext::*, crate::repr::*};
...
+            )
+        } else {
+            // For enums, unions, and non-`repr(C)` structs, we require that `Self`
+            // is sized.
```
joshlf, line 134:
```suggestion
            // For enums, unions, and non-`repr(C)` structs, we require that `Self`
            // is sized, and as a result don't need to reason about the internals of the type.
```

zerocopy-derive/src/lib.rs
```diff
@@ -62,10 +73,112 @@ use {crate::ext::*, crate::repr::*};
...
+            // it preserves provenance.
+            #[inline(always)]
+            fn raw_from_ptr_len(
+                bytes: ::core::ptr::NonNull<u8>,
+                _elems: usize,
+            ) -> ::core::ptr::NonNull<Self> {
+                bytes.cast::<Self>()
+            }
+        )
+    });
```
joshlf, line 154:
```suggestion
            // SAFETY:
            // - The recursive call to `raw_from_ptr_len` preserves both address and provenance.
            // - The `as` cast preserves both address and provenance.
            // - `NonNull::new_unchecked` preserves both address and provenance.
            #[inline(always)]
            fn raw_from_ptr_len(
                bytes: ::core::ptr::NonNull<u8>,
                elems: usize,
            ) -> ::core::ptr::NonNull<Self> {
                let trailing = <#TRAILING_FIELD_TYPE_HERE as KnownLayout>::raw_from_ptr_len(bytes, elems);
                let slf = trailing.as_ptr() as *mut Self;
                // SAFETY: Constructed from `trailing`, which is non-null.
                unsafe { NonNull::new_unchecked(slf) }
            }
        )
    });
```

zerocopy-derive/src/lib.rs
```diff
@@ -565,6 +694,8 @@ fn impl_block<D: DataExt>(
         )
     });
 
+    let self_sized_bound = Some(parse_quote!(Self: Sized)).filter(|_| require_self_sized);
```
joshlf, line 697: Maybe make this a bit less clever with a normal if-else?

joshlf (changes requested):

joshlf (changes requested):
Need to add tests, notably:
- [ ]  In `src/lib.rs`, the impl tests for types like `Unalign` should test with both sized and unsized types
- UI failure tests for the following cases:
  - [x] Type is `repr(C)`, trailing field is a concrete type (not a type parameter), and trailing field doesn't implement `KnownLayout`
  - [x] Type is not `repr(C)`, trailing field is a concrete type (not a type parameter), and trailing field is unsized
- Non-UI tests for all combinations of the following axes:
  - [x] `repr(C)`/not `repr(C)`
  - [x] Trailing field is concrete/generic
  - [x] Trailing field does/does not implement `KnownLayout`
  - [x] Trailing field is sized/unsized

zerocopy-derive/tests/ui-nightly/struct.rs
```diff
@@ -23,15 +22,31 @@ fn main() {}
...
+// |          N |        N |              Y |        N |      KL02 |
+#[derive(KnownLayout)]
+struct KL02(u8, [u8]);
 
+// | `repr(C)`? | generic? | `KnownLayout`? | `Sized`? | Type Name |
+// |          Y |        N |              N |        N |      KL08 |
 #[derive(KnownLayout)]
-struct KnownLayout2<T: ?Sized>(T);
+#[repr(C)]
+struct KL08(u8, NotKnownLayoutDst);
```
joshlf, line 43: Since this succeeds, should it be in a non-UI test file?

zerocopy-derive/tests/ui-nightly/struct.rs
```diff
@@ -23,15 +22,31 @@ fn main() {}
...
 
-assert_impl_all!(KnownLayout2<[u8]>: KnownLayout);
+// | `repr(C)`? | generic? | `KnownLayout`? | `Sized`? | Type Name |
+// |          Y |        N |              N |        Y |      KL09 |
+#[derive(KnownLayout)]
+#[repr(C)]
+struct KL09(NotKnownLayout, NotKnownLayout);
```
joshlf, line 50: Since this succeeds, should it be in a non-UI test file?

zerocopy-derive/tests/ui-nightly/late_compile_pass.rs
```diff
@@ -72,3 +73,17 @@ struct Unaligned2 {
...
+
+// | `repr(C)`? | generic? | `KnownLayout`? | `Sized`? | Type Name |
+// |          Y |        Y |              N |        Y |      KL13 |
+#[derive(KnownLayout)]
+#[repr(C)]
+struct KL13<T>(u8, T);
+
+fn test_kl13<T>(t: T) -> impl KnownLayout {
+    KL13(0u8, t)
+}
```
joshlf, line 89: Why does this type only generate errors on the nightly toolchain?

joshlf (changes requested):
For success cases, can we also test that we get the expected value for `KnownLayout::LAYOUT`?

It's probably also worth adding a test along these lines:

```rust
#[repr(C)]
struct Foo<A, B, C>(A, B, C);

macro_rules! test_foo {
    ($a:ty, $b:ty, $c:ty => $layout:expr) => {
        assert_eq!(<Foo<$a, $b, $c> as KnownLayout>::LAYOUT, $layout);
    };
}

// TODO: A bunch of different combinations of types with different sizes and alignments
```

zerocopy-derive/tests/ui-nightly/struct.rs
jswrenn: It doesn't succeed. It produces this error:
```
error[E0277]: the trait bound `NotKnownLayout: KnownLayout` is not satisfied
  --> tests/ui-nightly/struct.rs:47:10
   |
47 | #[derive(KnownLayout)]
   |          ^^^^^^^^^^^ the trait `KnownLayout` is not implemented for `NotKnownLayout`
   |
   = help: the following other types implement trait `KnownLayout`:
             bool
             char
             isize
             i8
             i16
             i32
             i64
             i128
           and $N others
   = help: see issue #48214
   = help: add `#![feature(trivial_bounds)]` to the crate attributes to enable
   = note: this error originates in the derive macro `KnownLayout` (in Nightly builds, run with -Z macro-backtrace for more info)
```

joshlf: Ah gotcha; I was cmd+f-ing for `KL09`, which doesn't appear in that error output. My bad.

zerocopy-derive/tests/ui-nightly/late_compile_pass.rs
jswrenn: More compiler pass weirdness. I've move the test to `mid_compile_pass.rs`, and it seems to produce an error on all toolchains. It's unfortunate how fragile this all is. We really should revisit the `ui-test` crate.

zerocopy-derive/tests/ui-nightly/struct.rs
jswrenn: This also has a failure:
```
error[E0277]: the trait bound `NotKnownLayoutDst: KnownLayout` is not satisfied
  --> tests/ui-msrv/struct.rs:41:10
   |
41 | #[derive(KnownLayout)]
   |          ^^^^^^^^^^^ the trait `KnownLayout` is not implemented for `NotKnownLayoutDst`
   |
   = help: see issue #48214
   = note: this error originates in the derive macro `KnownLayout` (in Nightly builds, run with -Z macro-backtrace for more info)
```

zerocopy-derive/tests/ui-nightly/late_compile_pass.rs
joshlf: Agreed.

src/lib.rs
```diff
@@ -5100,6 +5132,486 @@ mod tests {
...
+        #[derive(KnownLayout)]
+        #[repr(C)]
+        struct KLTU<T, U: ?Sized>(T, U);
+
+        assert_eq!(
+            <KLTU<(), ()> as KnownLayout>::LAYOUT,
+            DstLayout {
+                align: NonZeroUsize::new(1).unwrap(),
+                size_info: SizeInfo::Sized { _size: 0 }
+            }
```
joshlf, line 5413: Given how many of these you have, it might be helpful to factor into a functions or something:

```rust
let sized_layout = |size, align| ...;
let unsized_layout = |offset, trailing_size, align| ...;
```

src/lib.rs
```diff
@@ -5100,6 +5132,486 @@ mod tests {
...
+        assert_eq!(
+            <KLTU<(), u8> as KnownLayout>::LAYOUT,
+            DstLayout {
+                align: NonZeroUsize::new(1).unwrap(),
+                size_info: SizeInfo::Sized { _size: 1 }
+            }
+        );
+
+        assert_eq!(
+            <KLTU<(), u16> as KnownLayout>::LAYOUT,
```
joshlf, line 5425: We should use a type whose alignment is guaranteed. `u16` has alignment 2 on most platforms, but it isn't guaranteed. Maybe use the existing `AU64` type?

joshlf (changes requested):

src/lib.rs
```diff
@@ -5100,6 +5132,267 @@ mod tests {
...
+        // |          Y |        N |              Y |        Y |      KL11 |
+        // |          Y |        Y |              N |        N |      KL12 |
+        // |          Y |        Y |              N |        Y |      KL13 |
+        // |          Y |        Y |              Y |        N |      KL14 |
+        // |          Y |        Y |              Y |        Y |      KL15 |
+        // +------------+----------+----------------+----------+-----------+
+
+        struct NotKnownLayout<T = ()> {
+            _t: T,
+        }
```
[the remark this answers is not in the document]
jswrenn: Discussed offline: It *could* use a `repr(transparent)`, but it would be sufficiently interesting if this test started failing on a future Rust version, that it'd be nice to get the heads up.

joshlf (approved):

PR #645, opened 2023-11-27
[util] Ptr doesn't track byte length separately

joshlf:
Since raw pointer slice casts are now guaranteed to preserve element count [1], we no longer need to separately store the byte length of a `Ptr`. Instead, we can compute it from the raw pointer field.

[1] https://github.com/rust-lang/reference/pull/1417

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -15,7 +15,6 @@ pub(crate) mod ptr {
     use core::{
         fmt::{Debug, Formatter},
         marker::PhantomData,
-        mem,
         ptr::NonNull,
     };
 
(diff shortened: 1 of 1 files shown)


src/util.rs
```diff
@@ -252,12 +225,47 @@ pub(crate) mod ptr {
...
-                Some((slf, split_at)) if split_at == self._bytes_len => Some(slf),
+                Some((slf, split_at)) if split_at == self._len() => Some(slf),
                 Some(_) | None => None,
             }
         }
     }
 
+    impl<'a, T> Ptr<'a, [T]> {
+        /// The number of slice elements referenced by `self`.
+        fn _len(&self) -> usize {
```
jswrenn, line 236: I'd like to suggest that we start getting into the practice of using `allow(dead_code)` at definition sites rather than `_name` at use-sites. It's a major chore to clean those use-sites up, especially when they appear in macros and comments and cannot be renamed automatically.

jswrenn (changes requested):
See comments.

src/util.rs
```diff
@@ -252,12 +225,47 @@ pub(crate) mod ptr {
...
+            //   of the pointer's referent (e.g., casting `*const [u16]` to
+            //   `*const [u8]` will result in a raw pointer which refers to an
+            //   object of half the size of the original). The same holds for
+            //   `str` and any compound type whose unsized tail is a slice type,
+            //   such as struct `Foo(i32, [u8])` or `(u64, Foo)`.
+            //
+            // TODO(#429),
+            // TODO(https://github.com/rust-lang/reference/pull/1417): Once this
+            // text is available on the Stable docs, cite those instead of the
+            // Nightly docs.
```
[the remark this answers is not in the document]
joshlf: Discussed offline; this is consistent with our safety comment policy: https://github.com/google/zerocopy/blob/main/POLICIES.md#exceptions-to-our-safety-comment-policy

src/util.rs
```diff
@@ -252,12 +225,47 @@ pub(crate) mod ptr {
...
+        fn _len(&self) -> usize {
+            #[allow(clippy::as_conversions)]
+            let slc = self.ptr.as_ptr() as *const [()];
+            // SAFETY:
+            // - `()` has alignment 1, so `slc` is trivially aligned.
+            // - `slc` was derived from a non-null pointer.
+            // - The size is 0 regardless of the length, so it is sound to
+            //   materialize a reference regardless of location.
+            // - Pointer provenance may be an issue, but we never dereference.
+            let slc = unsafe { &*slc };
```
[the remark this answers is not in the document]
joshlf: Updated; turns out it does have valid provenance and I was just lazy in writing this comment.

src/util.rs
joshlf: Discussed offline: we'll transition to this in a separate PR.

jswrenn (approved):

src/util.rs
joshlf: Uploaded a PR for this: https://github.com/google/zerocopy/pull/646

PR #652, opened 2023-11-28
[proofs] Initial commit

joshlf:
Add axioms and lemmas which are useful in proving the soundness of some trait impls.

Makes progress on #429


src/lib.rs
```diff
@@ -2303,17 +2304,18 @@ safety_comment! {
 }
 safety_comment! {
     /// SAFETY:
-    /// `Wrapping<T>` is guaranteed by its docs [1] to have the same layout as
-    /// `T`. Also, `Wrapping<T>` is `#[repr(transparent)]`, and has a single
-    /// field, which is `pub`. Per the reference [2], this means that the
-    /// `#[repr(transparent)]` attribute is "considered part of the public ABI".
+    /// `Wrapping<T>` is `#[repr(transparent)]` and has a single `T` field,
+    /// which is `pub`. [1] Per axiom-repr-transparent-layout-validity, we may
```
jswrenn, line 2308: Should `axiom-repr-transparent-layout-validity` perhaps be a intradoc link? It'd give us a modest defense against proof rot.

src/proof_utils.rs
```diff
@@ -0,0 +1,100 @@
...
+//!   `size_of_val_raw(t) == size_of_val_raw(u)`.
+//! - For all `u: *const U`, `let t = *const T` is valid and `size_of_val_raw(u)
+//!   == size_of_val_raw(t)`.
+//! - For all `(t, u): (*const T, *const U)` where `size_of_val_raw(t) ==
+//!   size_of_val_raw(u)`:
+//!   - `t` and `u` refer to `UnsafeCell`s at the same byte ranges.
+//!   - If `*t` contains a valid `T`, that implies that `*u` contains a valid
+//!     `U`.
+//!   - If `*u` contains a valid `U`, that implies that `*t` contains a valid
+//!     `T`.
```
jswrenn, line 29: Concretely:
```rust
/// A type, `T`, has the property `transparent-layout-validity(U)` if the
/// following all hold:
/// - `T` and `U` have the same alignment
/// - For all `t: *const T`, `let u = t as *const U` is valid and
///   `size_of_val_raw(t) == size_of_val_raw(u)`.
/// - For all `u: *const U`, `let t = *const T` is valid and `size_of_val_raw(u)
///   == size_of_val_raw(t)`.
/// - For all `(t, u): (*const T, *const U)` where `size_of_val_raw(t) ==
///   size_of_val_raw(u)`:
///   - `t` and `u` refer to `UnsafeCell`s at the same byte ranges.
///   - If `*t` contains a valid `T`, that implies that `*u` contains a valid
///     `U`.
///   - If `*u` contains a valid `U`, that implies that `*t` contains a valid
///     `T`.
#[allow(unsused)] // This only exists as an anchor for intradoc links.
const TRANSPARENT_LAYOUT_VALIDIDTY: () = ();
```

src/proof_utils.rs
```diff
@@ -0,0 +1,100 @@
...
+//! type's fields. In practice, phrases like the above are only ever used in
+//! contexts where the following additional properties also hold:
+//! - `T` and `U` have the same vtable kinds. `T`'s and `U`'s pointer metadata
+//!   is such that raw pointer casts preserve size and field placement.
+//! - `T` and `U` have `UnsafeCell`s at the same byte ranges.
+//!
+//! ## axiom-repr-transparent-layout-validity
+//!
+//! Given types `T` and `U`, if `T` is a `#[repr(transparent)]` struct with a
+//! `pub` field of type `U`, and `T` does not contain any other fields, then
```
jswrenn, line 54: Is "`T` does not contain any other fields" a necessary precondition? A `repr(transparent)` type may contain at most one field with non-trivial size and alignment. The addition of fields with trivial size (`0`) and alignment (`1`) cannot  impact the conditions of `TRANSPARENT_LAYOUT_VALIDIDTY`.

jswrenn (changes requested):

src/lib.rs
joshlf: Oooh good idea. I'll also need to make sure that our safety doc comments are actually being parsed by rustdoc.

src/proof_utils.rs
joshlf: My only concern is that formally defining /which/ field is non-zero-sized would be annoying, and I'm not sure how to articulate it. Obviously you can just say "the only non-zero-sized field", but that kicks the can down the road: how do you know, at the proof site, which fields are zero sized? Right now, we only use this for `Wrapped`, which satisfies this more restricted precondition.

PR #658, opened 2023-11-29
Improve `FromBytes` documentation

jswrenn:
This PR adds examples to most methods of `FromBytes`, and separately documents `derive(FromBytes)`. Theoretical soundness requirements are placed with trait documentation, and the details of the derive's analysis is placed in the derive documentation.

This PR follows the pattern of #505.


src/lib.rs
```diff
@@ -1585,38 +1586,70 @@ pub unsafe trait FromZeroes {
...
+/// Unless you are modifying the implementation of this derive, or attempting to
+/// manually implement `FromBytes` for a type yourself, you don't need to read
+/// this section.*
 ///
-/// If a type has the following properties, then it is sound to implement
+/// If a type has the following properties, then this derive can implement
 /// `FromBytes` for that type:
+///
 /// - If the type is a struct, all of its fields must satisfy the requirements
 ///   to be `FromBytes` (they do not actually have to be `FromBytes`)
```
joshlf, line 1654: This isn't true of the derive. While it would be sound (the old wording), the derive is in practice more strict: it requires that all field types actually are `FromBytes`.

src/lib.rs
```diff
@@ -1661,6 +1699,101 @@ pub unsafe trait FromZeroes {
...
+///   `size_of::<T>()` as a `T`.
+/// - Given `b: &[u8]` where `b.len() == size_of::<T>()`, `b` is aligned to
+///   `align_of::<T>()` it is sound to construct a `t: &T` at the same address
+///   as `b`, and it is sound for both `b` and `t` to be live at the same time.
+///
+/// If a type is marked as `FromBytes` which violates this contract, it may
+/// cause undefined behavior.
+///
+/// `#[derive(FromBytes)]` only permits [types which satisfy these
+/// requirements][derive-analysis].
```
joshlf, line 1785: Maybe note here or somewhere else within this section that the derive may not be complete (ie, there may be types which satisfy these safety preconditions but which the derive still rejects)?

src/lib.rs
```diff
@@ -1715,8 +1920,34 @@ pub unsafe trait FromBytes: FromZeroes {
...
+    ///
+    /// // These bytes encode a `PacketHeader`.
+    /// let bytes = &mut [0, 1, 2, 3, 4, 5, 6, 7][..];
+    ///
+    /// let header = PacketHeader::mut_from(bytes).unwrap();
+    ///
+    /// assert_eq!(header.src_port, [0, 1]);
+    /// assert_eq!(header.dst_port, [2, 3]);
+    /// assert_eq!(header.length, [4, 5]);
+    /// assert_eq!(header.checksum, [6, 7]);
```
joshlf, line 1949: Maybe also perform mutation to demonstrate that the memory is shared? Something like:

```rust
header.checksum = [0, 0];

assert_eq!(bytes, &[0, 1, 2, 3, 4, 5, 0, 0]);
```

Same below with other `mut` methods.

joshlf (changes requested):

joshlf:
Can you also mention #32 in the commit message?

src/lib.rs
jswrenn: Maaaaybe. I'm hesitant to add any text that might induce people to write their own impls.

joshlf: Ack.

src/lib.rs
jswrenn: Done!

src/lib.rs
```diff
@@ -1585,40 +1586,72 @@ pub unsafe trait FromZeroes {
...
+/// manually implement `FromBytes` for a type yourself, you don't need to read
+/// this section.*
 ///
-/// If a type has the following properties, then it is sound to implement
+/// If a type has the following properties, then this derive can implement
 /// `FromBytes` for that type:
+///
 /// - If the type is a struct, all of its fields must satisfy the requirements
-///   to be `FromBytes` (they do not actually have to be `FromBytes`)
+///   to be `FromBytes`.
```
joshlf, line 1654: IMO this should be stricter: "all of its fields must implement `FromBytes`" or something to that effect.

jswrenn: Yep, agreed.

joshlf (dismissed):
Feel free to also bump the version number, and I can publish.

jswrenn:
Version bumped!

joshlf (approved):

PR #660, opened 2023-11-30
Improve `AsBytes` documentation

jswrenn:
This PR adds examples to most methods of `AsBytes`, and separately documents `derive(AsBytes)`. Theoretical soundness requirements are placed with trait documentation, and the details of the derive's analysis is placed in the derive documentation.

This PR follows the pattern of #505.

Makes progress towards #32.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -254,7 +254,7 @@ pub use crate::wrappers::*;
 
 #[cfg(any(feature = "derive", test))]
 #[cfg_attr(doc_cfg, doc(cfg(feature = "derive")))]
-pub use zerocopy_derive::{AsBytes, Unaligned};
+pub use zerocopy_derive::Unaligned;
 
 // `pub use` separately here so that we can mark it `#[doc(hidden)]`.
 //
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -2633,6 +2796,50 @@ pub unsafe trait AsBytes {
...
+    /// let mut bytes = [0, 0, 0, 0, 0, 0, 0, 0];
+    ///
+    /// header.write_to(&mut bytes[..]);
+    ///
+    /// assert_eq!(bytes, [0, 1, 2, 3, 4, 5, 6, 7]);
+    /// ```
+    ///
+    /// The size of `Self` and length of the target byte slice differ, If
+    /// insufficient target bytes are provided, `write_to` returns `None` and
+    /// leaves the target bytes unmodified:
```
joshlf, line 2831:
```suggestion
    /// If insufficient target bytes are provided, `write_to` returns `None`
    /// and leaves the target bytes unmodified:
```

joshlf (changes requested):

src/lib.rs
jswrenn: I opted for:
```
/// If too many or too few target bytes are provided, `write_to` returns
/// `None` and leaves the target bytes unmodified:
```

joshlf (approved):
Want to also release a new version? Or are you planning on putting up more doc PRs today?

PR #665, opened 2023-12-01
[derive] Support `derive(TryFromBytes)` for structs

jswrenn:
Supersedes #370.

Makes progress on #5.


src/util.rs
```diff
@@ -139,6 +139,65 @@ pub(crate) mod ptr {
             // - `U: 'a`
             Ptr { ptr, _lifetime: PhantomData }
         }
+
+        /// Project a field from `self`.
```
joshlf, line 143:
```suggestion
        /// Projects a field from `self`.
```

src/util.rs
```diff
@@ -139,6 +139,65 @@ pub(crate) mod ptr {
...
+        /// ## Preconditions
+        ///
+        /// The caller promises that `projector` projects a well-aligned field
+        /// of its argument. Its argument will be `self` casted to a raw
+        /// pointer.
+        ///
+        /// ## Postconditions
+        ///
+        /// If the preconditions of this function are met, this function will
+        /// return a `Ptr` the field projected from `self` by `projector`.
```
joshlf, line 156:
```suggestion
        /// return a `Ptr` to the field projected from `self` by `projector`.
```

src/util.rs
```diff
@@ -139,6 +139,65 @@ pub(crate) mod ptr {
...
+        ) -> Ptr<'a, U> {
+            // SAFETY: `projector` is provided with `self` casted to a raw
+            // pointer.
+            let field = projector(self.ptr.as_ptr());
+
+            // SAFETY: The caller of `project` promises that `projector` is
+            // provided with `self` casted to a raw pointer. By invariant,
+            // `self` is non-null. By contract, `projector` is a field
+            // projection of its argument. The field projected by `projector`
+            // will, too, be non-null.
```
joshlf, line 171: There's a bit in here that is off or insufficient IMO:
- The caller doesn't promise that `projector` is provided with a raw pointer of `self`, we do
- We should explicitly call out that field projection can't wrap around the address space to null. You can cite the field invariant on `ptr` and could also leave a TODO to cite [these docs](https://github.com/rust-lang/rust/pull/116675) once they're merged.

zerocopy-derive/src/ext.rs
```diff
@@ -6,48 +6,74 @@
...
-use syn::{Data, DataEnum, DataStruct, DataUnion, Type};
+use proc_macro2::TokenStream;
+use quote::ToTokens;
+use syn::{Data, DataEnum, DataStruct, DataUnion, Field, Index, Type};
 
 pub trait DataExt {
-    /// Extract the types of all fields. For enums, extract the types of fields
-    /// from each variant.
-    fn field_types(&self) -> Vec<&Type>;
+    /// Extract the names and types of all fields. For enums, extract the names
```
joshlf, line 14:
```suggestion
    /// Extracts the names and types of all fields. For enums, extracts the names
```

zerocopy-derive/src/lib.rs
```diff
@@ -304,6 +322,68 @@ pub fn derive_unaligned(ts: proc_macro::TokenStream) -> proc_macro::TokenStream
     .into()
 }
 
+// A struct is `TryFromBytes` if:
+// - all fields are `TryFromBytes`
```
joshlf, line 326: Update this to mention `repr(packed)`?

zerocopy-derive/src/lib.rs
```diff
@@ -304,6 +322,68 @@ pub fn derive_unaligned(ts: proc_macro::TokenStream) -> proc_macro::TokenStream
...
+            // SAFETY: We use `is_bit_valid` to validate that each field is
+            // bit-valid, and only return `true` if all of them are. The bit
+            // validity of a struct is just the composition of the bit
+            // validities of its fields, so this is a sound implementation of
+            // `is_bit_valid`.
+            unsafe fn is_bit_valid(candidate: zerocopy::Ptr<Self>) -> bool {
+                true #(&& {
+                    let project = |slf: *mut Self| ::core::ptr::addr_of_mut!((*slf).#field_names);
+
+                    // SAFETY: `projector` is a field projection of `candidate`.
```
joshlf, line 354:
```suggestion
                    // SAFETY: `project` is a field projection of `candidate`.
```

zerocopy-derive/src/lib.rs
```diff
@@ -304,6 +322,68 @@ pub fn derive_unaligned(ts: proc_macro::TokenStream) -> proc_macro::TokenStream
...
+                    // The projected field will be well-aligned, because this
```
joshlf, line 355:
```suggestion
                    // The projected field will be well-aligned because this
```

joshlf (changes requested):
We also need tests which actually execute the derived `is_bit_valid` implementation. Our existing `TryFromBytes` tests in zerocopy only exercise it for hand-rolled implementations.

src/util.rs
jswrenn: Done.

zerocopy-derive/tests/struct_try_from_bytes.rs
```diff
@@ -0,0 +1,143 @@
...
+fn two_bad() {
+    // TODO(#5): Use `try_transmute` in this test once it's available.
+    let candidate = zerocopy::Ptr::from(&[2][..]);
+    // SAFETY:
+    // - The cast `cast(p)` is implemented exactly as follows: `|p: *mut T| p as
+    //   *mut U`.
+    // - The size of the object referenced by the resulting pointer is equal to
+    //   the size of the object referenced by `self`.
+    // - The alignment of `Unsized` is equal to the alignment of `[u8]`.
+    let candidate = unsafe { candidate.cast_unsized(|p| p as *mut Two) };
```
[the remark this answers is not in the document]
jswrenn: Lol, I genuinely forgot `try_from_ref` existed. Actually, though, I think I prefer the current approach, since it allows us to focus the test on `is_bit_valid` (which is the item actually generated by `derive(TryFromBytes)`) and not take a dependency on `KnownLayout`.

PR #672, opened 2023-12-04
[derive] Fix AsBytes for `#[repr(C, packed(N))]`

maurer:
`packed(N)` does not gaurantee no padding, but it doesn't prevent it either. This was previously supported.

This regressed in #643 (left out of the commit message because of the requirement not to link to review sites).

### DIFF
--- a/zerocopy-derive/src/repr.rs
+++ b/zerocopy-derive/src/repr.rs
@@ -106,23 +106,24 @@ pub trait KindRepr: 'static + Sized + Ord {
 // etc), and provide implementations of `KindRepr`, `Ord`, and `Display`, and
 // those traits' super-traits.
 macro_rules! define_kind_specific_repr {
-    ($type_name:expr, $repr_name:ident, $($repr_variant:ident),*) => {
+    ($type_name:expr, $repr_name:ident, [ $($repr_variant:ident),* ] , [ $($repr_variant_aligned:ident),* ]) => {
         #[derive(Copy, Clone, Debug, Eq, PartialEq)]
         pub enum $repr_name {
             $($repr_variant,)*
-            Align(u64),
+            $($repr_variant_aligned(u64),)*
         }
 
         impl KindRepr for $repr_name {
             fn is_align(&self) -> bool {
                 match self {
-                    $repr_name::Align(_) => true,
+                    $($repr_name::$repr_variant_aligned(_) => true,)*
                     _ => false,
                 }
             }
 
             fn is_align_gt_one(&self) -> bool {
                 match self {
+                    // `packed(n)` only lowers alignment
                     $repr_name::Align(n) => n > &1,
                     _ => false,
                 }
@@ -131,7 +132,7 @@ macro_rules! define_kind_specific_repr {
             fn parse(meta: &Meta) -> syn::Result<$repr_name> {
                 match Repr::from_meta(meta)? {
                     $(Repr::$repr_variant => Ok($repr_name::$repr_variant),)*
-                    Repr::Align(u) => Ok($repr_name::Align(u)),
+                    $(Repr::$repr_variant_aligned(u) => Ok($repr_name::$repr_variant_aligned(u)),)*
                     _ => Err(Error::new_spanned(meta, concat!("unsupported representation for deriving FromBytes, AsBytes, or Unaligned on ", $type_name)))
                 }
             }
@@ -155,16 +156,19 @@ macro_rules! define_kind_specific_repr {
             fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
                 match self {
                     $($repr_name::$repr_variant => Repr::$repr_variant,)*
-                    $repr_name::Align(u) => Repr::Align(*u),
+                    $($repr_name::$repr_variant_aligned(u) => Repr::$repr_variant_aligned(*u),)*
                 }.fmt(f)
             }
         }
     }
 }
 
-define_kind_specific_repr!("a struct", StructRepr, C, Transparent, Packed);
+define_kind_specific_repr!("a struct", StructRepr, [C, Transparent, Packed], [Align, PackedN]);
 define_kind_specific_repr!(
-    "an enum", EnumRepr, C, U8, U16, U32, U64, Usize, I8, I16, I32, I64, Isize
+    "an enum",
+    EnumRepr,
+    [C, U8, U16, U32, U64, Usize, I8, I16, I32, I64, Isize],
+    [Align]
 );
 
 // All representations known to Rust.
--- a/zerocopy-derive/tests/struct_as_bytes.rs
+++ b/zerocopy-derive/tests/struct_as_bytes.rs
@@ -80,6 +80,18 @@ struct CPacked {
 assert_impl_all!(CPacked: AsBytes);
 
 #[derive(AsBytes)]
+#[repr(C, packed(2))]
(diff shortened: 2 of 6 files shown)


joshlf:
Woah, great catch, and thanks for such a thorough PR! We'll review this soon. Out of curiosity, how did you find this issue?

maurer:
crosvm uses `#[repr(C, packed(4))]` in a correct but not load bearing (compared to `#[repr(C, packed)]`) in one of their tests. This hit us when updating zerocopy.

joshlf:
Okay cool, thanks!

jswrenn:
What happened here is subtle, so I'm attempting to figure it out and put it in writing:

## Parsing before #643

In 7b98c7fb7877b6d29d1496226cf3dc31605bae30 (the commit before #643), the regression test provided by @maurer is accepted by zerocopy:
```rust
#[derive(AsBytes)]
#[repr(C, packed(2))]
// The same caveats as for CPacked apply - we're assuming u64 is at least
// 4-byte aligned by default. Without packed(2), this should fail, as there
// would be padding between a/b assuming u64 is 4+ byte aligned.
struct CPacked2 {
    a: u16,
    b: u64,
}

static_assertions::assert_impl_all!(CPacked2: AsBytes);
```
This was accepted possibly accidentally. To see why, let's trace the expansion of `derive(AsBytes)`. We start at `derive_as_bytes`: https://github.com/google/zerocopy/blob/7b98c7fb7877b6d29d1496226cf3dc31605bae30/zerocopy-derive/src/lib.rs#L95-L99

Then enter `derive_as_bytes_struct`: https://github.com/google/zerocopy/blob/7b98c7fb7877b6d29d1496226cf3dc31605bae30/zerocopy-derive/src/lib.rs#L270-L273

The definition of `STRUCT_UNION_ALLOWED_REPR_COMBINATIONS` in `STRUCT_UNION_AS_BYTES_CFG` would suggest that `packed(N)` isn't permitted: https://github.com/google/zerocopy/blob/7b98c7fb7877b6d29d1496226cf3dc31605bae30/zerocopy-derive/src/lib.rs#L128-L133

So what happens when we call `valid_reprs`? The first thing is that we call `reprs` to parse the metas: https://github.com/google/zerocopy/blob/7b98c7fb7877b6d29d1496226cf3dc31605bae30/zerocopy-derive/src/repr.rs#L51-L52

...which calls `StructRepr::parse`: https://github.com/google/zerocopy/blob/7b98c7fb7877b6d29d1496226cf3dc31605bae30/zerocopy-derive/src/repr.rs#L131-L136

...which calls `Repr::from_meta`:
https://github.com/google/zerocopy/blob/7b98c7fb7877b6d29d1496226cf3dc31605bae30/zerocopy-derive/src/repr.rs#L190-L220

**Here, we find the bug.** If the meta has arguments, it is *unconditionally* parsed as `Repr::Align(..)`: https://github.com/google/zerocopy/blob/7b98c7fb7877b6d29d1496226cf3dc31605bae30/zerocopy-derive/src/repr.rs#L213-L215

`validate_reprs` then filters this out: https://github.com/google/zerocopy/blob/7b98c7fb7877b6d29d1496226cf3dc31605bae30/zerocopy-derive/src/repr.rs#L68-L71

...so the apparent `repr` of `CPacked2` becomes just `repr(C)`. Since this is in the `allowed_combinations` of `STRUCT_UNION_ALLOWED_REPR_COMBINATIONS`, the `repr` is accepted. 

**Fortunately, this is not a soundness bug.** If the example is modified to use `repr(C, packed(4))` instead, an error is produced. This is because the expansion of `AsBytes` includes a const-eval padding check for padding: https://github.com/google/zerocopy/blob/7b98c7fb7877b6d29d1496226cf3dc31605bae30/zerocopy-derive/src/lib.rs#L559-L566

## Parsing after #643

In #643, `from_repr` is modified to parse `packed(n)` as `Repr::PackedN(n)`, rather than `Repr::Align(n)`. Since `PackedN` is not in the allowed combinations of `STRUCT_UNION_ALLOWED_REPR_COMBINATIONS`, the `repr` is rejected by `validate_reprs`.

## Parsing with #672

TODO.

jswrenn (changes requested):
This PR partially repairs a longstanding oddity in our `repr` parsing: the special treatment of `Align`. `align` does not appear in our list of allowed combinations:
https://github.com/google/zerocopy/blob/7b98c7fb7877b6d29d1496226cf3dc31605bae30/zerocopy-derive/src/lib.rs#L128-L133
...nor the invocation of `define_kind_specific_repr`: https://github.com/google/zerocopy/blob/7b98c7fb7877b6d29d1496226cf3dc31605bae30/zerocopy-derive/src/repr.rs#L165

Rather, it is defined implicitly during macro expansion: https://github.com/google/zerocopy/blob/7b98c7fb7877b6d29d1496226cf3dc31605bae30/zerocopy-derive/src/repr.rs#L111-L114

This is because, unlike the other modifiers, `align` *always* takes an argument. This PR makes `define_kind_specific_repr` a *little* less magical: It extends the macro with the ability to generate provided argument-taking modifiers.

However, the way it fixes the issue is still a little magical. Recall that `validate_reprs` *filters out* the `Align` repr (because it would be a pain to enumerate these in `allowed_combinations`): https://github.com/google/zerocopy/blob/7b98c7fb7877b6d29d1496226cf3dc31605bae30/zerocopy-derive/src/repr.rs#L68-L71

This PR leverages that behavior by having  `PackedN(n).is_align()` return `true`:

https://github.com/google/zerocopy/blob/8d3c3fccd61a0e0e3cb83575a970a715d4fcdbcb/zerocopy-derive/src/repr.rs#L117-L122

I'm not entirely comfortable with this. Not only is the name misleading now, but what happens when a new argument-consuming is added (particularly one that doesn't modify alignment)?

@joshlf, I think we need to reconsider our repr parsing framework, and reduce the the special handling of `Align` and `PackedN`. My understanding is we need this special handling because we can't write out argument-bearing modifiers in `allowed_combinations`. Perhaps we can abstract the argument component so we can specify either concrete or pattern arguments; e.g.:
```rust
// ditto for `StructRepr`, etc.
enum ReprKind<A: ReprArgument> {
    C,
    Transparent,
    Packed,
    PackedN(A::PackedN),
    Align(A::Align),
}

trait ReprArgument {
    type PackedN;
    type Align,
}

struct Pattern;
struct Concrete;

impl ReprArgument for Pattern {
    type PackedN = u64;
    type Align = u64;
}

impl ReprArgument for Concrete {
    type PackedN = RangeFull;
    type Align = RangeFull;
}

const STRUCT_UNION_ALLOWED_REPR_COMBINATIONS: &[&[StructRepr<Pattern>]] = &[
    &[StructRepr::C],
    &[StructRepr::Transparent],
    &[StructRepr::Packed],
    &[StructRepr::PackedN(..)],
    &[StructRepr::Align],
    &[StructRepr::C, StructRepr::Packed],
    &[StructRepr::C, StructRepr::Packed(..)],
    &[StructRepr::C, StructRepr::Align],
];
```

joshlf (approved):
@jswrenn and I discussed this offline. We're going to follow up with the changes he suggested, but in the meantime we'll merge this and publish a new version. Thanks again for catching this and for submitting a fix!

joshlf (approved):

joshlf:
Published as 0.7.29.

joshlf:
Yanked 0.7.27 and 0.7.28, both of which contain the regression.

Issue #251, opened 2023-08-11
Separate "no `UnsafeCell`" property into separate `Immutable` trait; allow `FromZeros`, `FromBytes`, and `AsBytes` on types with `UnsafeCell`s


joshlf:
## Summary

Introduce a `Immutable` trait which indicates that a type contains no `UnsafeCell`s. Allow `FromZeros`, `FromBytes`, and `AsBytes` on types with `UnsafeCell`s and use `Immutable` as a bound on individual methods/functions.

This enables by-value transmutations such as `transmute!` on types containing `UnsafeCell`s.

See also https://github.com/google/zerocopy/issues/694.

## Progress

- [x] Add `Immutable` trait
  - https://github.com/google/zerocopy/pull/656
- [x] Allow deriving `Immutable`
  - https://github.com/google/zerocopy/pull/667
- [x] Allow `FromZeros`, `FromBytes`, and `AsBytes` on `UnsafeCell`; add `Immutable` bound on methods as needed
  - https://github.com/google/zerocopy/pull/682
- [x] https://github.com/google/zerocopy/issues/695
- [x] Implement `TryFromBytes`, `FromZeros`, and `FromBytes` for unsized `UnsafeCell`. See the discussion below for why we originally limited these impls to sized `UnsafeCell`s only.
  - #1619
- [x] Allow `TryFromBytes` on `UnsafeCell`; add `Immutable` bound on `try_from_ref` method
  - https://github.com/google/zerocopy/pull/890
- [x] Remove `#[doc(hidden)]` from `Immutable` and `pub use zerocopy_derive::Immutable`
  - https://github.com/google/zerocopy/pull/977
- [x] Consider renaming to be consistent with whatever name Rust chooses for [what is currently called Freeze](https://github.com/rust-lang/rust/pull/121501) since we will likely treat `Immutable` as merely a [polyfill for that trait](https://github.com/google/zerocopy/issues/941)
- [x] Loosen documentation to promise only "no interior mutability" (TODO: how do we define this formally?) and say that it is *sufficient* to prove that a type (recursively) contains no `UnsafeCell`s, but not *necessary*, and that callers cannot assume that `T: Immutable` implies that `T` contains no `UnsafeCell`s. Note that *our* code will need to rely on this implication, and all safety comments should be updated to point at an issue that tracks this change.
- Non-breaking changes
[...]

PR #682, opened 2023-12-06
Traits other than `NoCell` permit `UnsafeCell`s

joshlf:
Previously, `FromZeros`, `FromBytes`, and `AsBytes` could not be
implemented for types containing `UnsafeCell`s. This is a soundness
precondition for some types of reference transmutations (notably
transmuting either direction between `&[u8]` and `&T`). However, some of
our machinery operates only on values (e.g. `transmute!`), and that
machinery should in principle be able to support types which contain
`UnsafeCell`s.

In this commit, we remove the "no `UnsafeCell`" restriction from
`FromZeros`, `FromBytes`, and `AsBytes`. We use the recently-added
`NoCell` trait as a bound on individual functions and methods where
`UnsafeCell`s would be unsound. This permits some APIs to support
`UnsafeCell`s which could not previously support them.

Closes https://github.com/google/zerocopy/issues/251

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -279,7 +279,7 @@ pub use zerocopy_derive::KnownLayout;
 pub use zerocopy_derive::NoCell;
 
 use core::{
-    cell::{self, RefMut},
+    cell::{self, RefMut, UnsafeCell},
     cmp::Ordering,
     fmt::{self, Debug, Display, Formatter},
     hash::Hasher,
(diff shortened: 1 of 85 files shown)


src/lib.rs
```diff
@@ -2884,7 +2849,10 @@ pub unsafe trait AsBytes {
     /// assert_eq!(excessive_bytes, [0u8; 128]);
     /// ```
     #[inline]
-    fn write_to(&self, bytes: &mut [u8]) -> Option<()> {
+    fn write_to(&self, bytes: &mut [u8]) -> Option<()>
+    where
+        Self: NoCell,
```
joshlf, line 2854: Ideally we wouldn't need this bound, but I'm not sure how to implement this method soundly without it. Thankfully this isn't a regression (in 0.7.X, `AsBytes` forbids `UnsafeCell`s full stop). However, at a minimum, we may want to file an issue to track eventually relaxing this bound (and the `NoCell` bound on other `write_to_xxx` methods).

joshlf: Filed https://github.com/google/zerocopy/issues/688 to track.

jswrenn (changes requested):

src/lib.rs
```diff
@@ -982,22 +982,29 @@ impl_known_layout!(const N: usize, T => [T; N]);
 
 safety_comment! {
     /// SAFETY:
-    /// `str` and `ManuallyDrop<[T]>` [1] have the same representations as
-    /// `[u8]` and `[T]` repsectively. `str` has different bit validity than
-    /// `[u8]`, but that doesn't affect the soundness of this impl.
+    /// `str`, `ManuallyDrop<[T]>` [1], and `UnsafeCell<T>` [2] have the same
+    /// representations as `[u8]`, `T`, and `T` repsectively. `str` has
```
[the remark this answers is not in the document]

jswrenn (approved):

Issue #238, opened 2023-08-08
Document yanked releases


joshlf:
Document releases which have been yanked (and possibly subsequently un-yanked) along with an explanation for why each release was yanked/un-yanked. Per https://github.com/google/zerocopy/issues/238#issuecomment-1843636007, a good place for this would be a `CHANGELOG.md` file. This would also have the advantage of being a default location that users might look expecting to find release notes, and so we could include a reference to our [GitHub Releases](https://github.com/google/zerocopy/releases) in that file.

## Old text

Keep a changelog (e.g. in a `CHANGELOG.md` file). It would be easy for it to become stale, so it'd be good if there were some way of ensuring it is kept up-to-date. My current thinking:
- In CI, enforce that the most recent commit alters the changelog
- Allow an annotation in the commit message to opt out of this for changes that don't warrant a changelog update

Combined, these ought to ensure that the author has considered updating the changelog and either done it or consciously decided not to.

UPDATE: Consider using [this GitHub Action](https://github.com/marketplace/actions/changelog-enforcer), which does all of the above for us.

Open question: How does this play with the fact that we keep 0.6.x releases on a separate branch? When we make a new change we expect to backport to 0.6.x, what do we put in the changelog? What about when we cherry-pick that commit on the [v0.6.x](https://github.com/google/zerocopy/tree/v0.6.x) branch? Do we expect to keep the same changelog file on both branches, or have that file reflect what's on the branch in question?

joshlf:
Having a changelog would have improved the user experience in https://github.com/google/zerocopy/issues/679 (emphasis added):

> My project at work was affected by the yanking of 0.7.28 yesterday. I was curious about the reason (**which wasn't trivial to find, since there don't seem to be any release notes**) and ran into your new policy from https://github.com/google/zerocopy/pull/677:

jswrenn:
I'm a changelog skeptic. Maintaining a `CHANGELOG.md` can be a significant maintenance burden for underwhelming benefits. When changelog updates are made by individual feature PRs, those PRs will nearly always have a merge conflict that must be resolved. When the changes are made in batch by the maintainer, new releases are blocked on keeping the changelog updated. Frictionless merges and releases are a productivity multiplier, and we should be extremely wary of compromising these aspects of our workflow. I mostly regret keeping changelogs of the crates for which I maintain changelogs.

[...]

PR #692, opened 2023-12-08
[readme] Link to GitHub Releases

joshlf:


jswrenn (approved):
LGTM, but I still think we should have a `CHANGELOG.md` (or something like it) to document yanks.

joshlf:
> LGTM, but I still think we should have a `CHANGELOG.md` (or something like it) to document yanks.

Updated https://github.com/google/zerocopy/issues/238 to track this.

PR #699, opened 2023-12-08
Invariant-parameterize `Ptr` and make `is_bit_valid` safe

### LINKED ISSUE #715: Invariant-parameterize `Ptr`
In #406, we introduced the `Ptr` type to describe pointers that lie somewhere between the invariants of `NonNull<T>` and `&T`. This type has both reduced the labor required to track invariants in our documentation, and revealed further opportunities for improvement. Namely: do actually *do* many things with a `Ptr`, you must do so through unsafe methods that have further unencapsulated requirements.

I propose that we extend `Ptr` to handle these conditional invariants, by encoding them as const generic parameters; e.g.:
```rust
/// A raw pointer with more restrictions.
///
/// `Ptr<T>` is similar to `NonNull<T>`, but it is more restrictive in the
/// following ways:
/// - It must derive from a valid allocation
/// - It must reference a byte range which is contained inside the
///   allocation from which it derives
///   - As a consequence, the byte range it references must have a size
///     which does not overflow `isize`
/// - `ptr` conforms to the aliasing invariant of `ASSUME_ALIASING`
/// - If `ASSUME_ALIGNMENT`, it must satisfy `T`'s alignment requirement
/// - If `ASSUME_VALIDITY`, it must satisfy `T`'s bit validity requirement
///
/// Thanks to these restrictions, it is easier to prove the soundness of
/// some operations using `Ptr`s.
///
/// `Ptr<'a, T>` is [covariant] in `'a` and `T`.
///
/// [covariant]: https://doc.rust-lang.org/reference/subtyping.html
pub struct Ptr<
    'a,
    T,
    const ASSUME_ALIASING: invariant::Aliasing,
    const ASSUME_ALIGNMENT: bool,
    const ASSUME_VALIDITY: bool,
> where
    T: 'a + ?Sized,
{
    /// INVARIANTS:
    /// 0. `ptr` is derived from some valid Rust allocation, `A`
    /// 1. `ptr` has the same provenance as `A`
    /// 2. `ptr` addresses a byte range which is entirely contained in `A`
    /// 3. `ptr` addresses a byte range whose length fits in an `isize`
    /// 4. `ptr` addresses a byte range which does not wrap around the address
    ///     space
    /// 5. `A` is guaranteed to live for at least `'a`
    /// 6. `T: 'a`
    /// 7. `ptr` conforms to the aliasing invariant of `ASSUME_ALIASING`
    /// 8. If `ASSUME_ALIGNMENT`, `ptr` is validly-aligned for `T`
    /// 9. If `ASSUME_VALIDITY`, `ptr` is validly-initialized for `T`
    ptr: NonNull<T>,
    _lifetime: PhantomData<&'a ()>,
}

// TODO: Replace this module with enums, once it becomes possible to do so.
// Use a unique integer for each invariant kind.
pub mod invariant {
    pub type Aliasing = u8;

    /// No invariants.
    pub const None: Aliasing = 0;
    
    /// The pointer adheres to the aliasing rules of a `&T`.
    pub const Shared: Aliasing = 1;
    
    /// The pointer adheres to the aliasing rules of a `&mut T`. 
    pub const Unique: Aliasing = 2;
}
```

We would then provide methods for asserting invariants; e.g.:
```rust
/// Assumes that `Ptr` adheres to the aliasing requirements of a `&T` reference.
///
/// # Safety
///
/// The caller promises that `Ptr` adheres to the aliasing requirements of a
/// `&T` reference.
impl<
        'a,
        T,
        const ASSUME_ALIASING: invariant::Aliasing,
        const ASSUME_ALIGNMENT: bool,
        const ASSUME_VALIDITY: bool,
    > Ptr<'a, T, ASSUME_ALIASING, ASSUME_ALIGNMENT, ASSUME_VALIDITY>
where
    T: 'a + ?Sized,
{
    pub unsafe fn assume_unique(
        self,
    ) -> Ptr<'a, T, {invariant::Unique}, ASSUME_ALIGNMENT, ASSUME_VALIDITY> {
        // SAFETY: The caller has promised that `Ptr` adheres to the aliasing
        // requirements of a `&T` reference.
        unsafe { Ptr::new(self.ptr) }
    }
}
```
...forgetting invariants; e.g.:
```rust
impl<
        'a,
        T,
        const ASSUME_ALIASING: invariant::Aliasing,
        const ASSUME_ALIGNMENT: bool,
        const ASSUME_VALIDITY: bool,
    > Ptr<'a, T, ASSUME_ALIASING, ASSUME_ALIGNMENT, ASSUME_VALIDITY>
where
    T: 'a + ?Sized,
{
    /// Forgets that `Ptr` has a validity invariant.
    pub fn forget_valid(self) -> Ptr<'a, T, ASSUME_ALIASING, ASSUME_ALIGNMENT, false> {
        // SAFETY: The invariants of `Ptr` when `ASSUME_VALIDITY = false` are a
        // strict subset of the obligations when `ASSUME_VALIDITY = true`.
        unsafe { Ptr::new(self.ptr) }
    }
}
```
...and **safe** methods conditioned on those invariants:
```rust
impl<'a, T> Ptr<'a, T, { invariant::Shared }, true, true>
where
    T: 'a + ?Sized,
{
    pub fn as_ref(self) -> &'a T {
        // SAFETY:
        // By invariant on `Ptr`:
        // 0. `ptr` is derived from some valid Rust allocation, `A`
        // 1. `ptr` has the same provenance as `A`
        // 2. `ptr` addresses a byte range which is entirely contained in `A`
        // 3. `ptr` addresses a byte range whose length fits in an `isize`
        // 4. `ptr` addresses a byte range which does not wrap around the address
        //     space
        // 5. `A` is guaranteed to live for at least `'a`
        // 6. `T: 'a`
        // 7. `ptr` conforms to the aliasing invariant of a shared reference,
        //    because `ASSUME_ALIASING = invariant::Shared`
        // 8. `ptr` is validly-aligned for `T` because `ASSUME_ALIGNMENT = true`
        // 9. `ptr` is validly-initialized for `T` because `ASSUME_VALIDITY = true`
        unsafe { self.ptr.as_ref() }
    }
}
```

## Why should we do this?

Doing this will help us maintain and document the safety invariants on `Ptr` and its methods. The above proposal has enough fidelity that we will be able to remove caller-enforced safety conditions on `TryFromBytes::is_bit_valid` to type invariants on a `Ptr<'_, Self, ASSUME_ALIASING = {invariants::Shared}, ASSUME_ALIGNMENT = false, ASSUME_VALIDITY = false>` and make `is_bit_valid` safe.

## Unresolved Questions

How should we name things? Does `ASSUME_VALIDITY` have enough fidelity? How should we express maybe-invalid-but-has-unsafecells-in-the-right-places?

## Prototype

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f4e7f9df5377bda62ab7e17c0287128a

jswrenn:
Closes #715.


src/util.rs
```diff
@@ -209,7 +229,7 @@ pub(crate) mod ptr {
         }
     }
 
-    impl<'a> Ptr<'a, [u8]> {
+    impl<'a> Ptr<'a, [u8], true> {
```
jswrenn, line 232: I think this PR is just about ready for review. The last blocker to supporting unaligned types is removing the alignment requirements from the items in this `impl` block.

src/util.rs
```diff
@@ -98,7 +101,18 @@ pub(crate) mod ptr {
             //   validly-intialized `T`.
             unsafe { self.ptr.as_ref() }
         }
+    }
 
+    impl<'a, T: ?Sized, const ASSUME_ALIGNED: bool> Ptr<'a, T, ASSUME_ALIGNED> where T: Copy {}
```
joshlf, line 106: Remove this?

src/util.rs
```diff
@@ -142,15 +156,21 @@ pub(crate) mod ptr {
             Ptr { ptr, _lifetime: PhantomData }
         }
 
+        /// Forget that `Ptr` is known to be validly-aligned for `T`.
```
joshlf, line 159:
```suggestion
        /// Forgets that `Ptr` is known to be validly-aligned for `T`.
```

src/util.rs
```diff
@@ -142,15 +156,21 @@ pub(crate) mod ptr {
...
+        pub(crate) fn forget_aligned(self) -> Ptr<'a, T, false> {
+            Ptr { ptr: self.ptr, _lifetime: self._lifetime }
```
joshlf, line 161: Safety comment? I know this is silly, but something as basic as "the invariants when `ASSUME_ALIGNED = false` are a subset of the invariants when `ASSUME_ALIGNED = true`" should be fine.

src/util.rs
```diff
@@ -472,7 +492,58 @@ pub(crate) mod ptr {
...
+        /// INVARIANTS:
+        /// 1. `ptr` is validly-initialized for `T`
+        ptr: Ptr<'a, T, false>,
+    }
+
+    impl<'a, T> MaybeAligned<'a, T>
+    where
+        T: 'a + ?Sized,
+    {
+        /// Read the value from `MaybeAligned`.
```
joshlf, line 516:
```suggestion
        /// Reads the value from this `MaybeAligned`.
```

src/util.rs
```diff
@@ -472,7 +492,58 @@ pub(crate) mod ptr {
...
+        }
+    }
+
+    /// A reference to a validly-initialized `T` of lifetime `'a` that might
+    /// *not* be validly aligned for `T`.
+    #[doc(hidden)]
+    #[allow(unreachable_pub)] // false positive?
+    pub struct MaybeAligned<'a, T: ?Sized> {
+        /// INVARIANTS:
+        /// 1. `ptr` is validly-initialized for `T`
```
joshlf, line 508:
```suggestion
        /// INVARIANT: `*ptr` is validly-initialized for `T`
```

I assume you're not planning on adding other invariants? Also, it's not `ptr` that's initialized, but `*ptr` (you could also say "`ptr`'s referent", etc).

src/util.rs
```diff
@@ -472,7 +492,58 @@ pub(crate) mod ptr {
...
+        #[inline]
+        pub fn read_unaligned(self) -> T
+        where
+            T: Copy,
+        {
+            let raw = self.ptr.ptr.as_ptr();
+            // SAFETY: By invariant on `MaybeAligned`, `raw` contains
+            // validly-initialized data for `T`. The value is safe to read and
+            // return, because `T` is copy.
+            unsafe { core::ptr::read_unaligned(raw) }
```
joshlf, line 528: It might be nice to add a `read` method to `Ptr`. That'd give us a more local spot to write down the safety preconditions on that method and then to justify the implementation in terms of `Ptr`'s internal invariants. It'd also presumably mean that the method's safety preconditions were pretty minimal.

src/util.rs
```diff
@@ -472,7 +492,58 @@ pub(crate) mod ptr {
...
+        fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+            self.ptr.fmt(f)
+        }
+    }
+
+    /// A reference to a validly-initialized `T` of lifetime `'a` that might
+    /// *not* be validly aligned for `T`.
+    #[doc(hidden)]
+    #[allow(unreachable_pub)] // false positive?
+    pub struct MaybeAligned<'a, T: ?Sized> {
```
joshlf, line 506: Can this go in its own module? Having these methods reach into the internals of `Ptr` wigs me out given how complex its internal invariants are.

src/util.rs
```diff
@@ -472,7 +492,58 @@ pub(crate) mod ptr {
...
+            T: Copy,
+        {
+            let raw = self.ptr.ptr.as_ptr();
+            // SAFETY: By invariant on `MaybeAligned`, `raw` contains
+            // validly-initialized data for `T`. The value is safe to read and
+            // return, because `T` is copy.
+            unsafe { core::ptr::read_unaligned(raw) }
+        }
+
+        /// View the value as an aligned reference.
```
joshlf, line 531:
```suggestion
        /// Views the value as an aligned reference.
```

src/util.rs
```diff
@@ -37,30 +37,33 @@ pub(crate) mod ptr {
...
+        /// 7. `A` is guaranteed to live for at least `'a`
+        /// 8. `T: 'a`
         ptr: NonNull<T>,
         _lifetime: PhantomData<&'a ()>,
     }
 
-    impl<'a, T: ?Sized> Copy for Ptr<'a, T> {}
-    impl<'a, T: ?Sized> Clone for Ptr<'a, T> {
+    /// The argument to `TryFromBytes::is_bit_valid`.
+    pub type MaybeValid<'a, T> = Ptr<'a, T, false>;
```
joshlf, line 56: IMO the name `MaybeValid` implies that the only thing that's not guaranteed is validity, but here lifetime and aliasing are both not guaranteed. I think we should just use `Ptr` directly since it's more clear that the user should go look at the type rather than guess its semantics.

src/util.rs
```diff
@@ -37,30 +37,33 @@ pub(crate) mod ptr {
...
+
+    impl<'a, T: ?Sized, const ASSUME_ALIGNED: bool> Copy for Ptr<'a, T, ASSUME_ALIGNED> {}
+    impl<'a, T: ?Sized, const ASSUME_ALIGNED: bool> Clone for Ptr<'a, T, ASSUME_ALIGNED> {
         #[inline]
         fn clone(&self) -> Self {
             *self
```
joshlf, line 62: While you're here, safety comment on this? We should have had one from the beginning.

src/util.rs
```diff
@@ -472,7 +492,58 @@ pub(crate) mod ptr {
...
+        ///
+        /// This is only available if `T` is [`Unaligned`].
+        #[inline]
+        pub fn as_ref(self) -> &'a T
+        where
+            T: Unaligned,
+        {
+            // SAFETY: By invariant on `MaybeAligned`, `ptr` is
+            // validly-initialized for `T`. Since `T` is `Unaligned`, it is also
+            // validly-aligned for `T`.
```
joshlf, line 541: What about aliasing? `Ptr` doesn't guarantee that other code accesses the same memory.

joshlf (changes requested):
Some invariants are now invalid - e.g., in `as_ref`, the safety comment assumes that `self.ptr` is aligned even though it's inside a block in which `ASSUME_ALIGN` can be either true or false.

src/util.rs
jswrenn: `Ptr` does guarantee lifetime (item 8):
```rust
/// INVARIANTS:
/// 1. `ptr` is derived from some valid Rust allocation, `A`
/// 2. `ptr` has the same provenance as `A`
/// 3. `ptr` addresses a byte range which is entirely contained in `A`
/// 4. `ptr` addresses a byte range whose length fits in an `isize`
/// 5. `ptr` addresses a byte range which does not wrap around the address
///     space
/// 6. If `ASSUME_ALIGNED`, `ptr` is validly-aligned for `T`
/// 7. `A` is guaranteed to live for at least `'a`
/// 8. `T: 'a`
```
...but not aliasing. However, this type *only* appears as an argument to `is_bit_valid` (and is documented as such), and `is_bit_valid`, as a pre-condition, guarantees the appropriate aliasing requirements. The name `MaybeValid`, then, does convey the salient quality of the argument.

src/pointer/ptr.rs
```diff
@@ -0,0 +1,1007 @@
...
+    /// - [`AnyAliasing`]
+    /// - [`Shared`]
+    /// - [`Unique`]
+    pub type Aliasing = u8;
+
+    /// No aliasing invariant.
+    pub const AnyAliasing: Aliasing = 1 << 0;
+
+    /// The pointer adheres to the aliasing rules of a `&T`.
+    pub const Shared: Aliasing = 1 << 2;
```
jswrenn, line 180: @joshlf Previously, you commented that you'd like to see the aliasing rules of `&T` inlined here. Are those rules crisply documented anywhere? I imagine we could write something to the effect of:

> The referent of a shared-aliased `Ptr` may be referenced by any number of shared aliased `Ptr`s or references. Safe code may assume that the referent will not be structurally muted, except through interior `UnsafeCell`s (if any).

How does that sound to you?

src/pointer/ptr.rs
```diff
@@ -0,0 +1,1010 @@
...
```
jswrenn, line 180: Consistency: I've skipped a number here.

src/pointer/ptr.rs
```diff
@@ -0,0 +1,1010 @@
...
+    /// - [`AnyValidity`]
+    /// - [`AsInitialized`]
+    /// - [`Valid`]
+    pub type Validity = u16;
+
+    /// No validity invariant.
+    pub const AnyValidity: Validity = 1 << 6;
+
+    /// The byte ranges initialized in `T` are also initialized in `Ptr`'s
+    /// referent.
```
joshlf, line 210: I don't believe this description is sufficiently precise, especially when it comes to enums (where the required initialization state of a given byte is dynamic, and depends upon preceding enum discriminants). IIUC, this basically needs to contain the entire text of the old safety precondition from `is_bit_valid`.

joshlf (changes requested):

src/pointer.rs
```diff
@@ -0,0 +1,88 @@
+//! Abstractions over raw pointers.
+
+mod ptr;
+
+pub use maybe_aligned::MaybeAligned;
+pub use ptr::{invariant, Ptr};
+
+/// A shorthand for a maybe-valid, maybe-aligned reference. Used as the argument
+/// to [`TryFromBytes::is_bit_valid`][crate::TryFromBytes::is_bit_valid].
+pub type MaybeValid<'a, T> =
```
joshlf, line 10: Can we call this `MaybeAlignedValid` or something? I was confused by the naming when reading `lib.rs` until I came and looked at this doc comment.

src/pointer.rs
```diff
@@ -0,0 +1,88 @@
...
+    /// A reference to a validly-initialized `T` of lifetime `'a` that might
+    /// *not* be validly aligned for `T`.
+    //  NOTE: Although this type is called `MaybeAligned`, that name is only
+    //  appropriate in its sole place of use: `TryFromBytes::is_bit_valid`.
+    //  There, the preconditions on the function's argument ensure that
+    #[doc(hidden)]
+    #[allow(unreachable_pub)] // false positive?
+    pub struct MaybeAligned<'a, T: ?Sized> {
+        ptr: Ptr<'a, T, { invariant::Shared }, { invariant::AnyAlignment }, { invariant::Valid }>,
+    }
```
joshlf, line 29: Why can't this just be a type alias like `MaybeValid`?

src/pointer/ptr.rs
joshlf: Yeah that sounds roughly good, except:
- Let's also call out that it may *not* be aliased by any mutable references? Even if those references are held by `unsafe` code and never actually used to perform mutation, the validity invariants are up in the air and it might be considered UB at some point in the future. We should probably also ban aliasing by mutable `Ptr`s for symmetry/abundance of caution regarding future memory model changes.
- s/muted/mutated
- What does "structurally" mean?

src/lib.rs
```diff
@@ -3624,17 +3569,7 @@ safety_comment! {
...
         // returning.
-        c.iter().all(|elem|
-            // SAFETY: We uphold the safety contract of `is_bit_valid(elem)`, by
-            // precondition on the surrounding call to `is_bit_valid`. The
-            // memory referenced by `elem` is contained entirely within `c`, and
-            // satisfies the preconditions satisfied by `c`. By axiom, we assume
-            // that `Iterator:all` does not invalidate these preconditions
-            // (e.g., by writing to `elem`.) Since `elem` is derived from `c`,
-            // it is only possible for uninitialized bytes to occur in `elem` at
-            // the same bytes they occur within `c`.
```
joshlf, line 3635: ❤️

src/lib.rs
```diff
@@ -3600,7 +3545,7 @@ safety_comment! {
     unsafe_impl!(const N: usize, T: Unaligned => Unaligned for [T; N]);
     assert_unaligned!([(); 0], [(); 1], [u8; 0], [u8; 1]);
     unsafe_impl!(T: NoCell => NoCell for [T]);
-    unsafe_impl!(T: TryFromBytes => TryFromBytes for [T]; |c: Ptr<[T]>| {
+    unsafe_impl!(T: TryFromBytes => TryFromBytes for [T]; |c: MaybeValid<[T]>| {
         // SAFETY: Assuming the preconditions of `is_bit_valid` are satisfied,
```
joshlf, line 3549: Is this entire safety comment unnecessary now?

src/macros.rs
```diff
@@ -97,93 +95,88 @@ macro_rules! unsafe_impl {
...
             // - The caller has promised that the cast results in an object of
             //   equal or lesser size.
-            // - The caller has promised that `$repr`'s alignment is less than
-            //   or equal to `Self`'s alignment.
             #[allow(clippy::as_conversions)]
             let $candidate = unsafe { candidate.cast_unsized::<$repr, _>(|p| p as *mut _) };
+
+            // SAFETY: The caller has promised that the referenced memory region
+            // will contain a valid `$repr`.
+            let $candidate = unsafe { $candidate.assume_as_initialized() };
```
joshlf, line 171: Isn't it the case that, when the caller provides a closure which takes a `MaybeValid`, they're *not* promising that any byte sequence is a valid `Self`?  IIUC, the only safety requirement here is that "the memory satisfies the initialization condition for `Self`" implies "the memory satisfies the initialization condition for `$repr`". I'm being vague about "initialization condition" because it's a pain to write down and we have another outstanding comment thread about the precise details.

src/pointer.rs
```diff
@@ -0,0 +1,88 @@
...
+mod maybe_aligned {
+    use super::{ptr::invariant, Ptr};
+    use crate::Unaligned;
+    use core::fmt::{Debug, Formatter};
+
+    /// A reference to a validly-initialized `T` of lifetime `'a` that might
+    /// *not* be validly aligned for `T`.
+    //  NOTE: Although this type is called `MaybeAligned`, that name is only
+    //  appropriate in its sole place of use: `TryFromBytes::is_bit_valid`.
+    //  There, the preconditions on the function's argument ensure that
```
joshlf, line 24: Presumably there's more here or this was meant to be deleted?

src/pointer.rs
```diff
@@ -0,0 +1,88 @@
...
+            Self { ptr }
+        }
+
+        /// Reads the value from `MaybeAligned`.
+        ///
+        /// This is only available if `T` is [`Copy`].
+        #[inline]
+        pub fn read_unaligned(self) -> T
+        where
+            T: Copy,
```
joshlf, line 58: We might want to figure out a way to relax this bound. One of the use cases for `Ptr` is to go from `[u8]` to `T`, and we support `T: FromBytes` even when `T: !Copy`. In principle, we should support reading a `T` from a byte slice even if `T: !Copy`. I grant that, in this particular case, it would be unsound to omit `T: Copy` because the `Ptr` could have been constructed from a `&Box<...>` or something.

src/pointer/ptr.rs
```diff
@@ -0,0 +1,1010 @@
+use core::ptr::NonNull;
+
+use crate::{util::AsAddress, KnownLayout, _CastType};
+
+/// Module used to gate access to [`Ptr`]'s fields.
+mod def {
```
joshlf, line 6: Are these submodules useful now that all of this code has been moved to `ptr` (as opposed to living directly in `util`?

src/pointer/ptr.rs
```diff
@@ -0,0 +1,1010 @@
...
+        ///    [`ALIGNMENT_INVARIANT`](invariant::Alignment); namely:
+        ///    - If [`invariant::Aligned`], the pointer's referent is
+        ///      well-aligned.
+        /// 9. `ptr` conforms to the validity invariant of
+        ///    [`VALIDITY_INVARIANT`](invariant::Validity); namely:
+        ///    - If [`invariant::AsInitialized`], the pointer's referent is
+        ///      initialized (but not necessarily validly so) in all of the
+        ///      initialized ranges of `T`.
+        ///    - If [`invariant::Valid`], the pointer's referent is bit-valid.
+        // SAFETY: `NonNull<T>` is covariant over `T` [1].
```
joshlf, line 68:
```suggestion
        // SAFETY: `NonNull<T>` is covariant over `T`. [1]
```

src/pointer/ptr.rs
```diff
@@ -0,0 +1,1010 @@
...
+        //
+        // [1]: https://doc.rust-lang.org/std/ptr/struct.NonNull.html
+        ptr: NonNull<T>,
+        // SAFETY: `&'a ()` is covariant over `'a` [1].
```
joshlf, line 72:
```suggestion
        // SAFETY: `&'a ()` is covariant over `'a`. [1]
```

src/pointer/ptr.rs
```diff
@@ -0,0 +1,1010 @@
...
+            // by-contract by the source `Ptr`.
+            unsafe { Self::new(ptr.as_non_null()) }
+        }
+
+        /// Converts this `Ptr<T>` to a [`NonNull<T>`].
+        ///
+        /// Note that this method does not consume `self`. Callers should be
+        /// aware that safe code may assume that the aliasing invariants of
+        /// `self` are not violated, and that the `NonNull` may be used in ways
+        /// that violate those assumptions.
```
joshlf, line 151: Did you mean that *unsafe* code may assume that the aliasing invariants aren't violated? Also, did you mean to warn that it's possible to use the returned `NonNull` in a way that violates those invariants (as opposed to warning that "safe code may assume... that the `NonNull` may be used in ways that violate those assumptions")?

src/pointer/ptr.rs
```diff
@@ -0,0 +1,1010 @@
...
+    }
+}
+
+pub use def::Ptr;
+
+/// The parameterized invariants of a [`Ptr`].
+//
+// These items are named according to the conventions for enums, should be
+// treated as if they are enums (i.e., only referenced by name), and should be
+// re-defined as enums as soon as it is possible to do so.
```
joshlf, line 164: Is there a Rust issue for adding support for enums in const that we could link to?

src/pointer/ptr.rs
```diff
@@ -0,0 +1,1010 @@
...
+    pub const Valid: Validity = 1 << 8;
+}
+
+/// External trait implementations on [`Ptr`].
+mod _external {
+    use super::*;
+    use core::fmt::{Debug, Formatter};
+
+    // SAFETY: Shared pointers are safely `Copy`. We do not implement `Copy` for
+    // unique pointers, since at most one may exist at a time.
```
joshlf, line 223: Can you also mention that all of the other invariants are unaffected by the number of copies of a pointer that exist? Same for `Clone` below.

src/pointer/ptr.rs
```diff
@@ -0,0 +1,1010 @@
...
+    }
+
+    /// `Ptr<'a, T>` → `&'a T`
+    impl<'a, T> Ptr<'a, T, { invariant::Shared }, { invariant::Aligned }, { invariant::Valid }>
+    where
+        T: 'a + ?Sized,
+    {
+        /// Converts the `Ptr` to a shared reference.
+        // This consumes `self`, not `&self`, because `self` is, logically, a
+        // pointer.
```
joshlf, line 312:
```suggestion
        // This consumes `self`, not `&self`, because `self` is, logically, a
        // pointer. Since this method is only available for `invariant::Shared`,
        // `Self: Copy`, and so this doesn't prevent the caller from still
        // using the pointer after calling `as_ref`.
```

src/pointer/ptr.rs
```diff
@@ -0,0 +1,1010 @@
...
+        > Ptr<'a, T, ALIASING_INVARIANT, ALIGNMENT_INVARIANT, VALIDITY_INVARIANT>
+    where
+        T: 'a + ?Sized,
+    {
+        /// Assumes that `Ptr`'s referent is validly-aligned for `T`.
+        ///
+        /// # Safety
+        ///
+        /// The caller promises that `Ptr`'s referent conforms to the alignment
+        /// invariants of `T`.
```
joshlf, line 365:
```suggestion
        /// invariant of `T`.
```

src/pointer/ptr.rs
joshlf: Update: Just saw that you moved the safety comment to `assume_as_initialized`. Maybe we can move those here and have `assume_as_initialized` reference these docs?

[TODO for self: Go through and audit uses to make sure safety comments cite this safety precondition and justify it]

src/pointer/ptr.rs
```diff
@@ -0,0 +1,1010 @@
...
+        /// uninitialized bytes in a subset of the uninitialized ranges in `T`.
+        ///
+        /// In other words, the `Ptr` is not required to refer to a valid `T`.
+        /// However, it must satisfy the requirement that uninitialized bytes
+        /// may only be present where it is possible for them to be present in
+        /// `T`. This is a dynamic property: if, at a particular byte offset, a
+        /// valid enum discriminant is set, the subsequent bytes may only have
+        /// uninitialized bytes as specificed by the corresponding enum.
+        ///
+        /// Formally, given `len = size_of_val_raw(candidate)`, at every byte
```
joshlf, line 389:
```suggestion
        /// Formally, given `len = size_of_val_raw(self)`, at every byte
```

In general, update the rest of this comment to make sure it's consistent with the names used here.

src/pointer/ptr.rs
```diff
@@ -0,0 +1,1010 @@
...
+            self,
+        ) -> Ptr<'a, T, ALIASING_INVARIANT, ALIGNMENT_INVARIANT, { invariant::AsInitialized }>
+        {
+            // SAFETY: The caller promises that `self`'s referent only contains
+            // uninitialized bytes in a subset of the uninitialized ranges in
+            // `T`. for `T`.
+            unsafe { Ptr::from_ptr(self) }
+        }
+
+        /// Assumes that `Ptr`'s referent is validly-initialized for `T`.
```
joshlf, line 421:
```suggestion
        /// Assumes that `Ptr`'s referent is validly initialized for `T`.
```

src/pointer/ptr.rs
```diff
@@ -0,0 +1,1010 @@
...
+        ///
+        /// # Safety
+        ///
+        /// The caller promises that `Ptr`'s referent conforms to the
+        /// bit-validity invariants on `T`.
```
joshlf, line 426:
```suggestion
        /// bit validity invariants on `T`.
```

src/pointer/ptr.rs
```diff
@@ -0,0 +1,1010 @@
...
+            self,
+            cast: F,
+        ) -> Ptr<'a, U, ALIASING_INVARIANT, { invariant::AnyAlignment }, { invariant::AnyValidity }>
+        where
+            U: 'a,
+        {
+            let ptr = cast(self.as_non_null().as_ptr());
+
+            // SAFETY: Caller promises that `cast` is just an `as` cast. We call
+            // `cast` on `self.ptr.as_ptr()`, which is non-null by construction.
```
joshlf, line 495:
```suggestion
            // `cast` on `self.as_non_null().as_ptr()`, which is non-null by construction.
```

src/pointer/ptr.rs
```diff
@@ -0,0 +1,1010 @@
...
+            let ptr = unsafe { NonNull::new_unchecked(ptr) };
+
+            // SAFETY:
+            // 0. By invariant on `self` and contract on `cast`, `ptr` is
+            //    derived from some valid Rust allocation, `A`.
+            // 1. By invariant on `self` and contract on `cast`, `ptr` has valid
+            //    provenance for `A`.
+            // 2. By invariant on `self` and contract on `cast`, `ptr` addresses
+            //    a byte range which is entirely contained in `A`.
```
joshlf, line 504: Maybe call out explicitly that this is a result of the fact that `cast` cannot produce a pointer which addresses a larger-sized object, and so the addressed bytes are a subset of the bytes addressed by `self`, and thus are also contained in `A`? Same with points 3 and 4.

src/pointer/ptr.rs
```diff
@@ -0,0 +1,1010 @@
...
+        ) -> Option<(
+            Ptr<'a, U, ALIASING_INVARIANT, { invariant::Aligned }, { invariant::AnyValidity }>,
+            usize,
+        )> {
+            // PANICS: By invariant, the byte range addressed by `self.ptr` does
+            // not wrap around the address space. This implies that the sum of
+            // the address (represented as a `usize`) and length do not overflow
+            // `usize`, as required by `validate_cast_and_convert_metadata`.
+            // Thus, this call to `validate_cast_and_convert_metadata` won't
+            // panic.
```
joshlf, line 564:
```suggestion
            // Thus, this call to `validate_cast_and_convert_metadata` will
            // only panic if `U` is a DST whose trailing slice element is zero-sized.
```

src/pointer/ptr.rs
```diff
@@ -0,0 +1,1010 @@
...
+            //   `'a`.
+            // - `U: 'a` by trait bound.
+            // - `ptr` conforms to the aliasing invariant of
+            //   `ALIASING_INVARIANT` because casting does not impact the
+            //   aliasing invariant
+            // - `ptr` conforms to the alignment invariant of `Aligned`, because
+            //   `validate_cast_and_convert_metadata` returns `None` if the cast
+            //   would leave `U` misaligned
+            // - `ptr`, trivially, conforms to the validity invariant of
+            //   `AnyValidity`
```
joshlf, line 629: Update this to use numbers rather than bullets like you've done elsewhere?

src/pointer/ptr.rs
```diff
@@ -0,0 +1,1010 @@
...
+        T: 'a + ?Sized,
+    {
+        /// Projects a field from `self`.
+        ///
+        /// # Safety
+        ///
+        /// ## Preconditions
+        ///
+        /// The caller promises that `projector` projects a field of its
+        /// argument. Its argument will be `self` casted to a raw pointer.
```
joshlf, line 682: We probably also need to mention that it doesn't follow references or perform other indirection or whatever. Alternatively, we could require that the bytes referenced by `projector(p)` are a subset of those referenced by `p`.

src/pointer/ptr.rs
```diff
@@ -0,0 +1,1010 @@
...
+
+#[cfg(test)]
+mod tests {
+    use core::mem::{self, MaybeUninit};
+
+    use super::*;
+    use crate::{util::testutil::AU64, FromBytes};
+
+    #[test]
+    fn test_ptrtry_cast_into_soundness() {
```
joshlf, line 890:
```suggestion
    fn test_ptr_try_cast_into_soundness() {
```

joshlf (changes requested):

src/pointer.rs
joshlf: Gotcha. Would we ever want to expose other `Ptr` functionality to callers? I worry that doing this sends us down a road in which we'll end up duplicating methods from `Ptr`.

Alternatively, is there no way for us to coax Rust into emitting better error messages while still using an alias?

src/pointer/ptr.rs
joshlf: Ack.

src/lib.rs
joshlf: Ack.

src/pointer/ptr.rs
joshlf: Ack.

src/pointer.rs
jswrenn: Like this:
```rust
macro_rules! define_invariants {
    ($($(#[$set_attr:meta])* $set:ident {
        $( $(#[$elem_attr:meta])* $elem:ident,)*
    })*) => {
        mod sealed {
            $(  
                #[allow(non_snake_case)]
                pub(super) mod $set {
                    pub trait Sealed {}
                    
                    $(impl Sealed for super::super::$elem {})*
                }
            )*
        }
        $(
            $(#[$set_attr])*
            pub trait $set: sealed::$set::Sealed {}

            $(
                $(#[$elem_attr])*
                pub enum $elem {}
                
                impl $set for $elem {}
            )*
        )*
    }
}

pub mod invariants {
    define_invariants! {
        /// The aliasing invariant of a [`Ptr`][super::Ptr].
        Aliasing {
            /// No aliasing invariant.
            AnyAliasing,
            
            /// The `Ptr<'a, T>` adheres to the aliasing rules of a `&'a T`.
            ///
            /// The referent of a shared-aliased `Ptr` may be concurrently
            /// referenced by any number of shared-aliased `Ptr` or `&T`
            /// references, and may not be concurrently referenced by any
            /// exclusively-aliased `Ptr`s or `&mut T` references.
            ///
            /// Unsafe code must assume that safe code will assume that the
            /// referent of such a `Ptr` will not be mutated, except through
            /// interior [`UnsafeCell`]s (if any).
            ///
            /// [`UnsafeCell`]: core::cell::UnsafeCell
            Shared,
            
            /// The `Ptr<'a, T>` adheres to the aliasing rules of a `&'a mut T`.
            ///
            /// The referent of an exclusively-aliased `Ptr` may not be
            /// concurrently referenced by any other `Ptr`s or references.
            /// Owners of an exclusively-aliased `Ptr` may safely (contigent on
            /// upholding the referent's original validity invariants) mutate
            /// its referent.
            Exclusive,
        }

        /// The alignment invariant of a [`Ptr`][super::Ptr].
        Alignment {
            /// The referent is not necessarily aligned.
            AnyAlignment,
            /// THe referent is aligned.
            Aligned,
        }

        /// The validity invariant of a [`Ptr`][super::Ptr].
        Validity {
            /// The referent is not necessarily initialized.
            AnyValidity,
            
            /// The byte ranges initialized in `T` are also initialized in the
            /// referent.
            ///
            /// Formally: uninitialized bytes may only be present in `Ptr<T>`'s
            /// referent where it is possible for them to be present in `T`.
            /// This is a dynamic property: if, at a particular byte offset, a
            /// valid enum discriminant is set, the subsequent bytes may only
            /// have uninitialized bytes as specificed by the corresponding
            /// enum.
            ///
            /// Formally, given `len = size_of_val_raw(ptr)`, at every byte
            /// offset, `b`, in the range `[0, len)`:
            /// - If, in all instances `t: T` of length `len`, the byte at
            ///   offset `b` in `t` is initialized, then the byte at offset `b`
            ///   within `*ptr` must be initialized.
            /// - Let `c` be the contents of the byte range `[0, b)` in `*ptr`.
            ///   Let `S` be the subset of valid instances of `T` of length
            ///   `len` which contain `c` in the offset range `[0, b)`. If, for
            ///   all instances of `t: T` in `S`, the byte at offset `b` in `t`
            ///   is initialized, then the byte at offset `b` in `*ptr` must be
            ///   initialized.
            ///
            ///   Pragmatically, this means that if `*ptr` is guaranteed to
            ///   contain an enum type at a particular offset, and the enum
            ///   discriminant stored in `*ptr` corresponds to a valid variant
            ///   of that enum type, then it is guaranteed that the appropriate
            ///   bytes of `*ptr` are initialized as defined by that variant's
            ///   bit validity (although note that the variant may contain
            ///   another enum type, in which case the same rules apply
            ///   depending on the state of its discriminant, and so on
            ///   recursively).
            AsInitialized,
            
            /// The referent is bit-valid for `T`.
            Valid,
        }
    }
}
```

joshlf (changes requested):

src/pointer/ptr.rs
```diff
@@ -158,104 +133,156 @@ mod def {
...
+    }) => {
+        mod sealed {
+            pub trait Sealed {}
 
-    /// No aliasing invariant.
-    pub const AnyAliasing: Aliasing = 1 << 0;
+            impl<$($set,)*> Sealed for ($($set,)*)
+            where
+                $($set: super::$set,)*
+            {}
```
joshlf, line 149: We could also just put a single `Sealed` trait in `util::sealed` or something. We already have a top-level `sealed` module which we could move to `util::sealed`.

joshlf (changes requested):

src/pointer.rs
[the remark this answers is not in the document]
jswrenn: I used the pattern I did because we're using edition 2018, but I'm in full agreement. I prefer the `mod.rs` structuring.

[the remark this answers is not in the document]
joshlf: Gotcha. IMO we should prioritize our own style preferences here over (hypothetical) community expectations since this isn't exposed in our API - it only affects contributors.

src/pointer.rs
jswrenn: Renamed to `Maybe<T>`. >:]

I'm signing off for the night, but feel free to bikeshed the name without me!

src/pointer.rs
jswrenn: I've implemented the above strategy, and made `MaybeAligned` merely a alias of `Ptr`.

src/util.rs
jswrenn: Now that `MaybeAligned` is just an alias on `Ptr`, this is technically now the case! However, the invariant bounds are overly restrictive.  We currently define `read_unaligned` on `invariant::Shared`, and we could also define it on `invariant::Unique` — but what about `invariant::AnyAliasing`?

src/pointer/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+        // issues, as we have not generated any invalid state which we need to
+        // fix before returning.
+        if T::is_bit_valid(candidate.forget_aligned()) {
+            // SAFETY: If `Self::is_bit_valid`, code may assume that `candidate`
+            // contains a bit-valid instance of `Self`.
+            let candidate = unsafe { candidate.assume_valid() };
+            Some(candidate)
+        } else {
+            None
+        }
```
joshlf, line 48:
```suggestion
        // This call may panic. If that happens, it doesn't cause any soundness
        // issues, as we have not generated any invalid state which we need to
        // fix before returning.
        if T::is_bit_valid(self.forget_aligned()) {
            // SAFETY: If `T::is_bit_valid`, code may assume that `self`
            // contains a bit-valid instance of `Self`.
            let candidate = unsafe { self.assume_valid() };
            Some(self)
        } else {
            None
        }
```

src/lib.rs
```diff
@@ -1295,30 +1252,19 @@ pub unsafe trait TryFromBytes {
...
-            return None;
-        }
+        let candidate = candidate.check_valid();
 
-        // SAFETY:
-        // - `is_bit_valid` guarantees that `*maybe_self` contains a valid
-        //   `Self`.
-        // - `TryFromBytes` types may not contain any `UnsafeCell`s.
-        Some(unsafe { maybe_self.into_ref_unchecked() })
+        candidate.map(MaybeAligned::as_ref)
```
joshlf, line 1267: Can you coalesce these?

src/pointer/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+    Ptr<'a, T, (invariant::Shared, Alignment, invariant::AsInitialized)>;
+
+// These methods are defined on the type alias, `Maybe`, so as to bring them to
+// the forefront of the rendered rustdoc.
+impl<'a, T, Alignment> Maybe<'a, T, Alignment>
+where
+    T: 'a + ?Sized + TryFromBytes,
+    Alignment: invariant::Alignment,
+{
+    /// Check that `Ptr`'s referent is validly initialized for `T`.
```
joshlf, line 29:
```suggestion
    /// Checks that `Ptr`'s referent is validly initialized for `T`.
```

src/util.rs
joshlf: Can you leave a TODO comment on `read_unaligned` to consider loosening the requirements at some point in the future? In particular, I could see us introducing a trait which represents "at least shared" aliasing, implementing it for both `invariant::Shared` and `invariant::Exclusive`, and using that trait (rather than a particular concrete type) as the aliasing bound.

src/pointer.rs
joshlf: Can we add a separate `#[doc(hidden)]` attribute here (so that it remains even if we remove the outer one on the module) and leave a TODO comment to bikeshed this name if we're planning on making this public?

src/pointer/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+// These methods are defined on the type alias, `MaybeAligned`, so as to bring
+// them to the forefront of the rendered rustdoc for that type alias.
+impl<'a, T, Alignment> MaybeAligned<'a, T, Alignment>
+where
+    T: 'a + ?Sized,
+    Alignment: invariant::Alignment,
+{
+    /// Reads the value from `MaybeAligned`.
+    ///
+    /// This is only available if `T` is [`Copy`].
```
joshlf, line 66: IMO this is redundant from the perspective of the rustdoc, which will render the `T: Copy` bound.

```suggestion
    /// Reads the value from `MaybeAligned`.
```

src/pointer/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+    #[inline]
+    pub fn read_unaligned(self) -> T
+    where
+        T: Copy,
+    {
+        let raw = self.as_non_null().as_ptr();
+        // SAFETY: By invariant on `MaybeAligned`, `raw` contains
+        // validly-initialized data for `T`. The value is safe to read and
+        // return, because `T` is copy.
```
joshlf, line 75: Can you add a bit more detail here? As discussed in [this thread](https://github.com/google/zerocopy/pull/699#discussion_r1437070945), you might naively assume that this is just permitting `&[u8] -> T`, and so `T: Copy` is unnecessary. It's specifically the fact that we support constructing a `Ptr` from a `&T` that makes this bound necessary.

src/pointer.rs
joshlf: Maybe let's open an issue to track having `Ptr` store the "original" type of its data? Probably not worth holding up this PR for that, and I suspect there are a lot of subtleties.

src/pointer/mod.rs
jswrenn: Discussed offline. Decided it's a non-issue.

src/macros.rs
```diff
@@ -52,16 +52,14 @@ macro_rules! safety_comment {
 ///     $ty>` which satisfies the preconditions of
 ///     `TryFromBytes::<$ty>::is_bit_valid`, it must be guaranteed that the
 ///     memory referenced by that `Ptr` always contains a valid `$repr`.
-///   - The alignment of `$repr` is less than or equal to the alignment of
-///     `$ty`.
 ///   - The impl of `is_bit_valid` must only return `true` for its argument
 ///     `Ptr<$repr>` if the original `Ptr<$ty>` refers to a valid `$ty`.
 macro_rules! unsafe_impl {
```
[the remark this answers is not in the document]
jswrenn: Non-issue, now that `Maybe` is an alias.

src/lib.rs
jswrenn: <3

src/lib.rs
jswrenn: Discussed offline. Decided against doing so, because the panic comment is specific to `check_valid`.

joshlf (approved):

Issue #187, opened 2023-05-27
Switch from `trybuild` to `ui-test`


joshlf:
See also: https://github.com/google/zerocopy/issues/560

The [trybuild](https://crates.io/crates/trybuild) crate keeps track of expected compiler errors in a separate `.stderr` file - one `.stderr` file per `.rs` file. This makes it difficult to audit updates to these files, which tend to be very cluttered and verbose, and makes issues like https://github.com/google/zerocopy/pull/706 / https://github.com/google/zerocopy/issues/707 more likely to slip through.

By contrast, the [ui-test](https://crates.io/crates/ui-test) crate allows annotating expected errors in an inline comment, which is much more readable and thus much less error-prone. It will also make it feasible for us to combine multiple tests in a single file without worrying about it increasing the likelihood of errors slipping through.

joshlf:
This would likely have prevented https://github.com/google/zerocopy/pull/706.

joshlf:
Another reason to migrate off of `trybuild`: https://github.com/dtolnay/trybuild/issues/241 is preventing us from implementing a test which would detect a regression of https://github.com/google/zerocopy/issues/716.

PR #706, opened 2023-12-09
[tests] Fix UI tests with missing imports

joshlf:
Previously, some UI tests had missing imports which caused the corresponding `.stderr` files to be populated with unhelpful "cannot find ..." messages. This is another point in favor of #187 (switching to the `ui-test` crate), which presumably would have made it much less likely for this to slip under the radar.

### DIFF
--- a/zerocopy-derive/tests/ui-msrv/enum.stderr
+++ b/zerocopy-derive/tests/ui-msrv/enum.stderr
@@ -67,33 +67,14 @@ error: TryFromBytes not supported on enum types
    | |_^
 
 error: must have a non-align #[repr(...)] attribute in order to guarantee this type's memory layout
-<<<<<<< HEAD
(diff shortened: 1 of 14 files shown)


zerocopy-derive/tests/ui-msrv/enum.stderr
```diff
@@ -67,54 +67,39 @@ error: TryFromBytes not supported on enum types
...
   --> tests/ui-msrv/enum.rs:70:1
    |
 70 | / enum FromZeros1 {
 71 | |     A(u8),
 72 | | }
    | |_^
 
-<<<<<<< HEAD
-error: only C-like enums can implement FromZeros
-=======
```
joshlf, line 106: Uh oh. The fact that this slipped through suggests that these tests aren't actually getting validated...

jswrenn (approved):

PR #720, opened 2023-12-13
[ci] check all-jobs-succeed depends on all jobs

### LINKED ISSUE #444: Verify that `all-jobs-succeeded` CI job depends on all other jobs
Due to a limitation in GitHub's merge queue support, we need to manually list all jobs that are required to pass before a PR in the merge queue can be considered to have succeeded. In order to avoid having to list all jobs manually in the GitHub settings page for this repository, we instead provide the following job which depends on all other jobs:

https://github.com/google/zerocopy/blob/68953ed8c440dabcebb0bfa93b23329f671b96aa/.github/workflows/ci.yml#L445-L453

This works, but is still error-prone: there's nothing that ensures that this job won't fall out of sync if we add new CI jobs in the future.

Similar to jobs like [`check_versions`](https://github.com/google/zerocopy/blob/68953ed8c440dabcebb0bfa93b23329f671b96aa/.github/workflows/ci.yml#L335), we should write a new CI job that automatically extracts all jobs which should block a PR merging, and make sure that `all-jobs-succeed` depends on each such job.

## Mentoring instructions

The job should support the following behaviors:
- [ ] Enumerate every YAML file under `.github`
  - NOTE: [Both `.yml` and `.yaml` are possible file extensions](https://en.wikipedia.org/wiki/YAML)
- [ ] For each such file:
  - [ ] Parse it as YAML with [`yq`](https://github.com/mikefarah/yq)
  - [ ] Extract any jobs which run for each PR (at a minimum, it must support [this syntax](https://github.com/google/zerocopy/blob/68953ed8c440dabcebb0bfa93b23329f671b96aa/.github/workflows/ci.yml#L3-L4), but you should also confirm that there aren't other ways of indicating to GitHub that an action should be run for each PR)
- [ ] Ensure that `all-jobs-succeeded` depends on each job found in the previous step

tommy-gilligan:
Potentially a new per-PR job could be added to CI without it being included in `all-jobs-succeed`'s dependencies.  This prevents that.

I have successfully constructed a [failure](https://github.com/tommy-gilligan/zerocopy/actions/runs/7195624547/job/19598647853) (by deleting a job from `all-jobs-succeed`'s dependencies)

Closes #444

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -482,7 +482,7 @@ jobs:
                   <(yq -r '.jobs.all-jobs-succeed.needs[]' "$i" | sort | uniq)
               fi
 
-            # The sed call here excludes all-jobs-succeed from the list of jobs that
+            # The grep call here excludes all-jobs-succeed from the list of jobs that
             # all-jobs-succeed does not depend on.  If all-jobs-succeed does
             # not depend on itself, we do not care about it.
             done | sort | uniq | grep -v '^all-jobs-succeed$' || true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -490,10 +490,8 @@ jobs:
 
           if [ -n "$jobs" ]
           then
-            echo 'There are jobs triggered by pull requests that'
-            echo 'are not listed in all-jobs-succeed.needs:'
-
-            echo $jobs
+            missing_jobs="$(echo "$jobs" | tr ' ' '\n')"
+            echo "all-jobs-succeed missing dependencies on some jobs: $missing_jobs" | tee -a $GITHUB_STEP_SUMMARY
             exit 1
           fi
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -461,6 +461,38 @@ jobs:
           cargo install --locked kani-verifier       &> /dev/null || true
           cargo kani setup                           &> /dev/null || true
 
+  check-job-dependencies:
+    runs-on: ubuntu-latest
+    name: Check all-jobs-succeeded depends on all jobs
+    steps:
+      - name: Install yq (for YAML parsing)
+        run: go install github.com/mikefarah/yq/v4@latest
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - name: Run dependency check
+        run: |
+          set -eo pipefail
+          jobs=$(for i in $(find .github -iname '*.yaml' -or -iname '*.yml')
+            do
+              # Select jobs that are triggered by pull request.
+              if yq -e '.on | has("pull_request")' "$i" 2>/dev/null >/dev/null
+              then
+                # This gets the list of jobs that are not in
+                # all-jobs-succeed.needs .
+                comm -23 \
+                  <(yq -r '.jobs | keys | .[]' "$i" | sort | uniq) \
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -461,6 +461,38 @@ jobs:
...
+                # This gets the list of jobs that are not in
+                # all-jobs-succeed.needs .
+                comm -23 \
+                  <(yq -r '.jobs | keys | .[]' "$i" | sort | uniq) \
+                  <(yq -r '.jobs.all-jobs-succeed.needs[]' "$i" | sort | uniq)
+              fi
+            done | sort | uniq
+          )
+          # Effectively, this checks that all-jobs-succeed does not depend on
+          # itself. Checking for the existence of a job that we don't want
```
joshlf, line 488: IIUC, it's slightly different than this - it checks that the set of jobs that all-jobs-succeeded does not depend on is _exactly_ itself. That distinction is important because it also ensures that there are no jobs _besides_ itself that it doesn't depend on. Do I understand that correctly? If so, could you update the comment?

joshlf (changes requested):
This is awesome, thanks for your work on this! One tiny nit on one of the comments, but otherwise this looks good!

.github/workflows/ci.yml
```diff
@@ -461,6 +461,38 @@ jobs:
...
+          # all-jobs-succeed to depend on acts as a crude test that the
+          # building of the dependency list is working as expected.
+          if [ "$jobs" != "all-jobs-succeed" ]
+          then
+            exit 1
+          fi
```
joshlf, line 494: Oh also: On failure, could you echo `$jobs` so that a developer can see from the Actions output what the offending jobs are?

```suggestion
          if [ "$jobs" != "all-jobs-succeed" ]
          then
            missing_jobs="$(echo "$jobs" | tr ' ' '\n' | grep -v '^all-jobs-succeed$')"
            echo "all-jobs-succeed missing dependencies on some jobs: $missing_jobs" | tee -a $GITHUB_STEP_SUMMARY
            exit 1
          fi
```

joshlf (changes requested):

.github/workflows/ci.yml
```diff
@@ -461,6 +461,40 @@ jobs:
...
+                # This gets the list of jobs that all-jobs-succeed does not depend on.
+                comm -23 \
+                  <(yq -r '.jobs | keys | .[]' "$i" | sort | uniq) \
+                  <(yq -r '.jobs.all-jobs-succeed.needs[]' "$i" | sort | uniq)
+              fi
+
+            # The sed call here excludes all-jobs-succeed from the list of jobs that
+            # all-jobs-succeed does not depend on.  If all-jobs-succeed does
+            # not depend on itself, we do not care about it.
+            done | sort | uniq | sed '/^all-jobs-succeed$/d'
```
joshlf, line 488: IMO more developers are familiar with `grep -v` than with `sed` with the `/d` suffix. It's also easier to look up since you can just look up the `-v` flag in the `grep` manpage (I tried looking up the `/d` suffix in the `sed` manpage and wasn't sure how to search for it).

```suggestion
            done | sort | uniq | grep -v '^all-jobs-succeed$'
```

joshlf (changes requested):

joshlf (approved):

joshlf:
Thanks again for your help with this, @tommy-gilligan!

PR #728, opened 2023-12-14
Add git pre-push hook

tommy-gilligan:
There are definitely different approaches that can be taken to sharing git hooks (some projects install git hooks automatically).  Here I've gone with something simple: just have a directory `githooks` that the developer/user is expected to point their git at by manually changing `core.hooksPath`.

Documenting installation is necessary but I've not yet done it.  Didn't want to go too far down the path I've chosen before getting feedback.

I also haven't yet done anything much to deal with stdout/stderr noise.


joshlf:
Hey @tommy-gilligan, thanks for this! Just wanted to let you know that, between other work and time off for the holidays, we likely won't get a chance to review this for at least a few weeks, but we haven't forgotten about it. We'll follow up once everyone's back at work and we've cleared out our backlog.

joshlf:
Okay, had a chance to look at this. Unfortunately you'll need to rebase changes that have happened in the interim. This looks great, though! My one concern would be how long it takes all of these scripts to execute locally; I assume it's not too much time, but it's also doing enough parsing and shell script work that it wouldn't surprise me if it was at least a noticeable increase. `git push` is a very frequent part of our workflow, so I'd be worried about even small slowdowns. If you could time the execution of the git hook and let me know what you get, that'd be great. If you could time each individual step so we know what's taking more time, that'd be even better.

Thanks again for doing this! In the long run, I think this will save us a lot of needless headache with extra CI round trips.

CONTRIBUTING.md
```diff
@@ -194,6 +194,21 @@ Commit messages should never contain references to any of:
       https://github.com/google/zerocopy/commit/789b3deb)
 1. Other entities which may not make sense to arbitrary future readers
 
+### Git hooks
+
+There is a pre-push git hook available at ./githooks/pre-push .  This hook runs
```
joshlf, line 199:
```suggestion
There is a pre-push git hook available at `./githooks/pre-push`. This hook runs
```

Single period between sentences to be consistent w/ other documentation. Can you do the same in the rest of this added section?

joshlf (changes requested):
Amazing, thank you so much for your work on this!

The timings look totally acceptable to me - sub-one-second latency addition for `git push` shouldn't be a problem, especially given the benefit. You're more than welcome to work on optimizing, but maybe let's do that in a separate PR? That way we can isolate those changes in case it breaks anything.

> The githook maybe adds a bit of noise to STDOUT/STDERR when running git from a terminal. Please let me know if anything should be done about this.

I'd recommend, in the pre-push script, just piping the stdout of each subcommand to `/dev/null`. We're good about printing errors to stderr, so this shouldn't swallow any.

githooks/pre-push
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+set -eo pipefail
+echo "Running pre-push git hook: $0"
+./ci/check_fmt.sh
```
joshlf, line 4: Can you add a comment here explaining why you aren't piping the output of this one?

joshlf (changes requested):
> > I'd recommend, in the pre-push script, just piping the stdout of each subcommand to /dev/null. We're good about printing errors to stderr, so this shouldn't swallow any.
> 
> I've ended up with everything but the formatting script redirecting to `/dev/null`. `rustfmt --check` uses stdout for displaying a diff that highlights files that need to be formatted and this will be useful for the user to see. Along the way I added `-q` to a `cargo install`, I think it makes sense to keep that.

Sounds good!

> Additionally, I found a small programming error in `check_versions.sh`. Because it was small, I went ahead and fixed it here.

Oh, what was the error?

PR #794, opened 2024-01-19
Support `derive(TryFromBytes)` on `packed` structs.

jswrenn:
Now that #715 supports unaligned `Ptr`s, we get support for `repr(packed)` 'for free' — this commit merely removes the anti-`packed` check from the derive.

Makes progress towards #5.


joshlf (approved):
🎉

PR #803, opened 2024-01-21
[derive] Support TryFromBytes on unit-only enums

joshlf:
TODO:
- Runtime tests for TryFromBytes enums
- Compile-fail tests


zerocopy-derive/src/lib.rs
```diff
@@ -401,6 +399,147 @@ const STRUCT_UNION_ALLOWED_REPR_COMBINATIONS: &[&[StructRepr]] = &[
...
 ];
 
+fn derive_try_from_bytes_enum(ast: &DeriveInput, enm: &DataEnum) -> proc_macro2::TokenStream {
+    if !enm.is_fieldless() {
+        return Error::new_spanned(ast, "only field-less enums can implement TryFromBytes")
+            .to_compile_error();
+    }
+
+    let reprs = try_or_print!(ENUM_TRY_FROM_BYTES_CFG.validate_reprs(ast));
+    let discriminant_type = match reprs.as_slice() {
```
jswrenn, line 409: Unsoundness: Any of these types can be shadowed, so you'll need to refer to them instead from `core_reexport::primitive`.

You should also add an integration test to make sure we don't regress on this.

jswrenn (changes requested):

zerocopy-derive/src/lib.rs
joshlf: Done on both counts

jswrenn (approved):

PR #804, opened 2024-01-21
[derive] Rename is_c_like -> is_fieldless

joshlf:
The implementation of this method checks for the presence of fields. While this is /similar/ to the concept of a C-like enum, it's not the same. Fieldless enums are exactly as described: none of their variants has fields. C-like enums, on the other hand, are more recently referred to as "unit-only" because their variants are not permitted to contain parentheses or curly braces. E.g., the following is a fieldless enum, but is not a C-like/unit-only enum:

```rust
  #[repr(u8)]
  enum Foo {
      A,   // This is fine
      B(), // This is not fine
      C{}  // This is not fine
  }
```

This distinction is important specifically when it comes to discriminants. The syntax `Foo::A as u8` computes the discriminant value of the `A` variant. However, `Foo::B` is a type constructor, and thus is a function. As a result, `Foo::B as u8` computes the address of the function `Foo::B`.

### DIFF
--- a/zerocopy-derive/src/lib.rs
+++ b/zerocopy-derive/src/lib.rs
@@ -413,8 +413,8 @@ fn derive_from_zeros_struct(ast: &DeriveInput, strct: &DataStruct) -> proc_macro
 // - one of the variants has a discriminant of `0`
 
 fn derive_from_zeros_enum(ast: &DeriveInput, enm: &DataEnum) -> proc_macro2::TokenStream {
-    if !enm.is_c_like() {
-        return Error::new_spanned(ast, "only C-like enums can implement FromZeros")
+    if !enm.is_fieldless() {
+        return Error::new_spanned(ast, "only field-less enums can implement FromZeros")
             .to_compile_error();
     }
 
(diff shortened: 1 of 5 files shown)


zerocopy-derive/src/lib.rs
```diff
@@ -413,7 +413,7 @@ fn derive_from_zeros_struct(ast: &DeriveInput, strct: &DataStruct) -> proc_macro
 // - one of the variants has a discriminant of `0`
 
 fn derive_from_zeros_enum(ast: &DeriveInput, enm: &DataEnum) -> proc_macro2::TokenStream {
-    if !enm.is_c_like() {
+    if !enm.is_fieldless() {
         return Error::new_spanned(ast, "only C-like enums can implement FromZeros")
```
jswrenn, line 417: Can we fix up these error messages too?
```suggestion
        return Error::new_spanned(ast, "only fieldless enums can implement FromZeros")
```

jswrenn (changes requested):

jswrenn (approved):

jswrenn (approved):

PR #808, opened 2024-01-23
Add TryFromBytes::try_read_from

joshlf:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1229,8 +1229,9 @@ pub unsafe trait TryFromBytes {
     /// reinterpreted as a `Self`.
     ///
     /// Note that Rust's bit validity rules are still being decided. As such,
-    /// there exist types whose bit validity is ambiguous. See the
-    /// `TryFromBytes` docs for a discussion of how these cases are handled.
+    /// there exist types whose bit validity is ambiguous. See
+    /// [here][TryFromBytes#what-is-a-valid-instance] for a discussion of how
+    /// these cases are handled.
     // TODO(#251): Require `Self: NoCell` and allow `TryFromBytes` types to
     // contain `UnsafeCell`s.
     #[inline]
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -1254,6 +1254,35 @@ pub unsafe trait TryFromBytes {
...
     }
+
+    /// Attempts to read a `Self` from a byte slice.
+    ///
+    /// `try_read_from` validates that `bytes` contains a valid `Self`. If it
+    /// does, those bytes are read out of `bytes` and reinterpreted as a `Self`.
+    ///
+    /// Note that Rust's bit validity rules are still being decided. As such,
+    /// there exist types whose bit validity is ambiguous. See the
+    /// `TryFromBytes` docs for a discussion of how these cases are handled.
```
jswrenn, line 1265:
```suggestion
    /// there exist types whose bit validity is ambiguous. See
    /// [here][TryFromBytes#what-is-a-valid-instance] for a discussion
    /// of how these cases are handled.
```

src/lib.rs
```diff
@@ -1254,6 +1254,35 @@ pub unsafe trait TryFromBytes {
...
+    #[inline]
+    #[doc(hidden)] // TODO(#5): Finalize name before remove this attribute.
+    fn try_read_from(bytes: &[u8]) -> Option<Self>
+    where
+        Self: Sized,
+    {
+        let candidate = MaybeUninit::<Self>::read_from(bytes)?;
+        let cptr = Ptr::from_maybe_uninit_ref(&candidate);
```
jswrenn, line 1273: The nittiest of nits, but I think this is slightly more in keeping with our snake case convention:
```suggestion
        let c_ptr = Ptr::from_maybe_uninit_ref(&candidate);
```

src/pointer/ptr.rs
```diff
@@ -447,6 +449,65 @@ mod _conversions {
...
+        ///
+        /// [`MaybeUninit<T>`]: MaybeUninit
+        #[doc(hidden)]
+        #[inline]
+        pub fn from_maybe_uninit_ref(ptr: &'a MaybeUninit<T>) -> Self {
+            let muptr = core::ptr::NonNull::from(ptr);
+            let ptr = muptr.cast::<T>();
+            // SAFETY:
+            // 0. `muptr`, by invariant on `&'a T`, is derived from some valid
+            //    Rust allocation, `A`. `ptr` is as well because `.cast()`
```
jswrenn, line 468: I assume you don't just consistently shadow `ptr` (and instead introduce `muptr`) because you want to distinguish between these bindings in our safety comments. If so, consider that we have *two* bindings here named `ptr`; perhaps we should rename one of those, too.

jswrenn (changes requested):

src/lib.rs

src/pointer/ptr.rs
joshlf: Done.

src/lib.rs

src/pointer/ptr.rs
joshlf: Done.

jswrenn (approved):

joshlf:
Makes progress on #5

Issue #671, opened 2023-12-04
Release Roadmap


joshlf:
# 0.8

## Definitely include

- #5 
- #29
- #251 
- #438 
- #554
  - Specifically: https://github.com/google/zerocopy/pull/574
- #431
  - Specifically: https://github.com/google/zerocopy/pull/430

## Include if we have time

- Refactor module structure
- https://github.com/google/zerocopy/issues/358
- Overhaul naming of various functions and methods
  - https://github.com/google/zerocopy/issues/284
  - https://github.com/google/zerocopy/issues/253
- https://github.com/google/zerocopy/issues/210
- https://github.com/google/zerocopy/issues/188
- https://github.com/google/zerocopy/issues/1

# 0.9

- Safe transmute

# 0.10

General API overhaul without changing fundamental analysis. Things like:
- Error reporting (e.g. https://github.com/google/zerocopy/discussions/528)
- How to express which properties are validated statically vs dynamically

PR #814, opened 2024-01-24
Add `FromBytes::getrandom` method

newpavlov:
The method is gated on the disabled by default `getrandom` feature.

This PR is an alternative to https://github.com/rust-random/getrandom/pull/381.

Reasons for why it could be better to add this method to `zerocopy` instead of introducing the `getrandom::value` function:
- `getrandom` has lower MSRV than `zerocopy`.
- `getrandom` does not have plans for breaking releases in the near future, while `zerocopy` [has](https://github.com/rust-random/getrandom/pull/381#issuecomment-1813521214).
- It allows to generate values using a bit nicer `u32::getrandom()`, while `getrandom::value()` may require explicit type ascription.
- `zerocopy` looks like a better place for keeping `unsafe` code for transforming (random) bytes to a value of `FromBytes` type.
- If we are to give up on returning `getrandom::Error`, `zerocopy` will not have any types from `getrandom` in its public API, while `getrandom` would always rely on `FromBytes` bound in its public API.


newpavlov:
@joshlf 
Could you give feedback on this addition?

joshlf:
Apologies for the radio silence.

Right now all of our effort is going into https://github.com/google/zerocopy/issues/671. At some point we do want to figure out a unified strategy for how to support integrations with other crates, but we're not going to have the cycles to do that work until 0.8 has been released. This PR on its own is obviously reasonable, but we want to be careful about how to expose APIs like these so we don't lock ourselves into supporting API patterns that we later want to walk back, and so we want to make sure we've thought through the whole design space before we publish even a simple API like this one.

TLDR: We're interested in something like this, but it will likely be quite some time before we are able to move forward on it.

PR #835, opened 2024-02-03
Add caching dupe change

### LINKED ISSUE #340: [CI] Deduplicate `actions/cache` directives

dorryspears:
This closes #340.  I tried putting the yaml in the workflow folder for the new cache job but it wasn't able to find it, so I made it an action.


joshlf:
Thanks for this, @dorryspears !

I'm not a GitHub Actions expert, so I may be off the mark here, but will making this a "step" rather than a `needs` dependency mean that this action is executed multiple times? Currently (on `main`), it executes once at the beginning of the job run and then the results are re-used by all steps, which helps things run faster.

dorryspears:
@joshlf You are right.  I now understand what the CI was doing.  I rearranged my code and added back the generate cache step.  That action just replaces the 

```yaml
- uses: actions/cache@13aacd865c20de90d75de3b17ebe84f7a17d57d2 # v4.0.0
      with:
        path: |
          ~/.cargo/
        key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
``` 
step in the existing jobs.

.github/workflows/ci.yml
```diff
@@ -519,7 +504,7 @@ jobs:
           cargo install cargo-readme --version 3.2.0 &> /dev/null || true
           cargo install --locked kani-verifier       &> /dev/null || true
           cargo kani setup                           &> /dev/null || true
-
+
```
joshlf, line 507:
```suggestion

```

.github/workflows/ci.yml
```diff
@@ -544,7 +529,7 @@ jobs:
             <(cargo metadata --format-version 1 | \
               jq -r ".packages[] | select(.name == \"zerocopy\").metadata.\"build-rs\" | keys | .[]" | \
               sort -u)
-
+
```
joshlf, line 532:
```suggestion

```

.github/workflows/ci.yml
```diff
@@ -485,20 +473,17 @@ jobs:
           assert-match "$zerocopy_derive_dep_ver" "$zerocopy_derive_targeted_ver" \
             "Same crate version ($zerocopy_derive_dep_ver) found for optional and targeted zerocopy-derive dependency." \
             "Different crate versions found for optional ($zerocopy_derive_dep_ver) and targeted ($zerocopy_derive_targeted_ver) dependency."
-
+
```
joshlf, line 476:
```suggestion

```

joshlf (changes requested):
Okay awesome, thanks! A few nits regarding stray whitespace, but this looks good!

joshlf (approved):

PR #843, opened 2024-02-07
Support arbitrary toolchain versions in `build.rs`

joshlf:


jswrenn:
> Figure out how to handle UI test files which differ by toolchain (do we also need N of these?)

This is almost certainly a job for the `ui-test` crate.

jswrenn (approved):

PR #848, opened 2024-02-08
Lower MSRV to 1.57

joshlf:
Makes progress on #554


jswrenn (approved):
Phenomenal! 1.56, here we come!

PR #850, opened 2024-02-08
Test `build.rs` cfgs using `rustversion` crate

joshlf:
In order to test code which is conditionally-compiled on different versions of the Rust toolchain, we previously relied on the same infrastructure as the runtime code - `--cfg` arguments emitted by our `build.rs` file. This, however, is using a feature to test itself. In this change, we switch to using the `rustversion` crate so that bugs in `build.rs` can't hide themselves by silently disabling the relevant tests.


jswrenn (approved):
Clever!

PR #854, opened 2024-02-08
Implement `TryFromBytes` for pointers

jswrenn:
Makes progress towards #5.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3368,10 +3368,28 @@ safety_comment! {
     #[cfg(feature = "alloc")]
     unsafe_impl!(
         #[cfg_attr(doc_cfg, doc(cfg(feature = "alloc")))]
+        T => TryFromBytes for Option<Box<T>>;
+        |c: Maybe<Option<Box<T>>>| pointer::is_zeroed(c)
+    );
+    #[cfg(feature = "alloc")]
+    unsafe_impl!(
+        #[cfg_attr(doc_cfg, doc(cfg(feature = "alloc")))]
         T => FromZeros for Option<Box<T>>
     );
+    unsafe_impl!(
+        T => TryFromBytes for Option<&'_ T>;
+        |c: Maybe<Option<&'_ T>>| pointer::is_zeroed(c)
+    );
     unsafe_impl!(T => FromZeros for Option<&'_ T>);
+    unsafe_impl!(
+            T => TryFromBytes for Option<&'_ mut T>;
+            |c: Maybe<Option<&'_ mut T>>| pointer::is_zeroed(c)
+    );
     unsafe_impl!(T => FromZeros for Option<&'_ mut T>);
+    unsafe_impl!(
+        T => TryFromBytes for Option<NonNull<T>>;
+        |c: Maybe<Option<NonNull<T>>>| pointer::is_zeroed(c)
+    );
     unsafe_impl!(T => FromZeros for Option<NonNull<T>>);
     unsafe_impl_for_power_set!(A, B, C, D, E, F, G, H, I, J, K, L -> M => FromZeros for opt_fn!(...));
     unsafe_impl_for_power_set!(A, B, C, D, E, F, G, H, I, J, K, L -> M => FromZeros for opt_extern_c_fn!(...));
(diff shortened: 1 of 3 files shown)


jswrenn:
@joshlf, CI is blocked on tweaking the invocation (and definition?) of `impl_try_from_bytes_testable` in ways I don't exactly understand. Can you give this a `./cargo.sh +msrv test --features alloc` sometime and provide some pointers?

src/pointer/ptr.rs
```diff
@@ -539,6 +539,19 @@ mod _transitions {
             unsafe { Ptr::from_ptr(self) }
         }
 
+        /// Recalls that `Ptr`'s referent is validly-aligned for `T`.
+        #[inline]
+        pub(crate) fn recall_aligned(
```
joshlf, line 544: To me this name reads like "to recall a product", which made me think it was saying "forget that this is aligned." Maybe `into_aligned`?

src/pointer/ptr.rs
```diff
@@ -674,6 +687,38 @@ mod _casts {
...
+        I: Invariants<Validity = invariant::Initialized>,
+    {
+        /// Casts this pointer-to-initialized into a pointer-to-bytes.
+        #[allow(clippy::wrong_self_convention)]
+        pub(crate) fn as_bytes(
+            self,
+        ) -> Ptr<'a, [u8], (I::Aliasing, invariant::Aligned, invariant::Valid)> {
+            // SAFETY: We ensure that:
+            // - `cast(p)` is implemented exactly as follows: `|p: *mut T| p as
+            //   *mut U`.
```
joshlf, line 702: This isn't true anymore. Can you justify that this is equivalent, or update the safety precondition so that this is documented as valid?

joshlf (changes requested):

src/pointer/ptr.rs
jswrenn: Some days, I truly feel like one of those people who writes `MaybeUninit::uninit().assume_init()` lol

jswrenn: Done! `cast_unsized` is now generalized to specify that `cast(p)` can be any kind of cast, where a 'cast' is a provenance-preserving transformation of a pointer-like value into another pointer-like value to the same allocation. To minimize safety-comment churn at callsites, I provide callers with a choose-your-own-adventure of proof outlines.

src/lib.rs
```diff
@@ -3621,7 +3639,13 @@ safety_comment! {
     /// documentation once this PR lands.
     unsafe_impl!(T: ?Sized => NoCell for *const T);
     unsafe_impl!(T: ?Sized => NoCell for *mut T);
+    unsafe_impl!(T => TryFromBytes for *const T; |c: Maybe<*const T>| {
+        pointer::is_zeroed(c)
+    });
     unsafe_impl!(T => FromZeros for *const T);
+    unsafe_impl!(T => TryFromBytes for *mut T; |c: Maybe<*const T>| {
+        pointer::is_zeroed(c)
+    });
```
jswrenn, line 3648: @joshlf Just added these two impls, too.

src/pointer/mod.rs
```diff
@@ -80,9 +80,14 @@ where
...
     {
-        // SAFETY: The alignment of `T` is 1 and thus is always aligned
-        // because `T: Unaligned`.
-        let ptr = unsafe { self.assume_alignment::<invariant::Aligned>() };
-        ptr.as_ref()
+        self.recall_aligned().as_ref()
     }
 }
+
+/// Check if the referent is zeroed.
```
joshlf, line 87:
```suggestion
/// Checks if the referent is zeroed.
```

src/lib.rs
```diff
@@ -3621,7 +3639,13 @@ safety_comment! {
     /// documentation once this PR lands.
     unsafe_impl!(T: ?Sized => NoCell for *const T);
     unsafe_impl!(T: ?Sized => NoCell for *mut T);
+    unsafe_impl!(T => TryFromBytes for *const T; |c: Maybe<*const T>| {
```
joshlf, line 3642: Can you update the preceding safety comment to mention `TryFromBytes` in addition to `FromZeros`?

src/pointer/ptr.rs
joshlf: Still need to address this.

joshlf (changes requested):

src/lib.rs
jswrenn: Done!

src/pointer/mod.rs
jswrenn: Done!

src/pointer/ptr.rs
jswrenn: Done!

joshlf (approved):

PR #860, opened 2024-02-10
[doc] TryFromBytes, IntoBytes aren't symmetrical

joshlf:
I don't love how this is currently worded. I'd love suggestions on improving it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1181,6 +1181,16 @@ pub unsafe trait NoCell {
 /// If you are negatively affected by lack of support for a particular type,
 /// we encourage you to let us know by [filing an issue][github-repo].
 ///
+/// # `TryFromBytes` is not symmetrical with [`IntoBytes`]
+///
+/// There are some types which implement both `TryFromBytes` and [`IntoBytes`],
+/// but for which `TryFromBytes` is not guaranteed to accept all byte sequences
+/// produced by `IntoBytes`. In other words, for some `T: TryFromBytes +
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1181,6 +1181,16 @@ pub unsafe trait NoCell {
...
+/// # `TryFromBytes` is not symmetrical with [`IntoBytes`]
+///
+/// There are some types which implement both `TryFromBytes` and [`IntoBytes`],
+/// but for which `TryFromBytes` is not guaranteed to accept all byte sequences
+/// produced by `IntoBytes`. In other words, for some `T: TryFromBytes +
+/// IntoBytes`, there exist values of `t: T` such that
+/// `TryFromBytes::try_from_ref(t.as_bytes()) == None`. Code should not
+/// generally assume that values produced by `IntoBytes` will necessarily be
+/// accepted as valid by `TryFromBytes`.
+///
```
jswrenn, line 1193: Perhaps we could include an example here?

jswrenn (changes requested):

src/lib.rs
joshlf: We're in a catch-22 here. We intend to eventually add `IntoBytes` impls for reference and pointer types, but we don't have them yet (https://github.com/google/zerocopy/issues/170). Once we have them, they will be described by this section. However, we can't wait to add this section because, if we do, people might write generic code that relies on the thing we're telling them not to rely on here, and that code would only break in the future once we add those impls.

jswrenn (approved):

PR #861, opened 2024-02-10
[ci] Separately track MSRV and MWRV

### LINKED ISSUE #807: Distinguish between MSRV and minimum *working* Rust version
With #791, we have the opportunity to decrease our MSRV to as low as 1.58. For 0.8's initial release, we should not target an MSRV this low. Our [current policy](https://github.com/google/zerocopy/blob/100ba98de68f577d7d2bfa2da4fc9492272ebf00/POLICIES.md#msrv) is to treat MSRV increases as a breaking changes. Since breaking changes should be infrequent, we should be careful not to back ourselves into a corner where our MSRV blocks us from providing vital functionality in a timely fashion.

Nonetheless, we should still strive for zerocopy to work on old as a Rust version as possible. To facilitate this, I propose that we distinguish between MSRV and Minimum Working Rust Version (MWRV). Our MSRV, encoded in the `rust-version` field of our `Cargo.toml` should remain at ~1.60 and decremented cautiously. Our MWRV, encoded `package.metadata.ci` in `Cargo.toml` should be as low as technically possible at all times. We should modify our CI to test on MWRV, instead of MSRV, and additionally check that MWRV <= MSRV.

joshlf:
MWRV stands for "minimum working Rust version" (a term we invent here). It is the minimum version of the toolchain that we test with in CI. We maintain (and test) the invariant that MSRV >= MWRV so that we know at all times that our code is compatible with our published MSRV.

The reason for doing this is so that we can make progress towards lowering our MSRV without painting ourselves into a corner. We treat MSRV bumps as breaking changes, so if we were to publish a particular MSRV and then later discover a problem, we wouldn't be able to revert to a previously-published, higher MSRV. By tracking MWRV separately, we can ensure that we're making progress while still leaving ourselves wiggle room to revert changes so long as those reversions are compatible with our published MSRV. Once we have let a particular MWRV bake long enough, we can lower the published MSRV to match.

Closes #807

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ jobs:
       matrix:
         # See `INTERNAL.md` for an explanation of these pinned toolchain
         # versions.
-        toolchain: [ "msrv", "stable", "nightly",  "zerocopy-generic-bounds-in-const-fn", "zerocopy-aarch64-simd", "zerocopy-panic-in-const", ]
+        toolchain: [ "mwrv", "stable", "nightly",  "zerocopy-generic-bounds-in-const-fn", "zerocopy-aarch64-simd", "zerocopy-panic-in-const", ]
         target: [
           "i686-unknown-linux-gnu",
           "x86_64-unknown-linux-gnu",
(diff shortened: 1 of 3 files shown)


joshlf:
@jswrenn I've confirmed locally that this doesn't have the same bug as #810.

joshlf:
We're [not going to do this](https://github.com/google/zerocopy/issues/554#issuecomment-2340664377).

Issue #367, opened 2023-09-10
Test the output of zerocopy-derive


joshlf:
As zerocopy-derive gets more powerful, it becomes harder to test every edge case just through compile-success and compile-fail tests. One way we could test at a higher fidelity would be to assert the exact token stream output as a result of deriving traits on various types. This would be in some ways easier to verify for a human reviewer, and would be more stable than compiler error messages (a problem that plagues our existing trybuild tests, and requires us to keep three separate copies of the .stderr files - one for each toolchain we use in CI).

A few ways we could do this:
- Use [`cargo expand`](https://crates.io/crates/cargo-expand)
- Use synstructure's [test_derive!](https://docs.rs/synstructure/latest/synstructure/macro.test_derive.html) macro

I'm currently leaning towards `test_derive!`, as it seems to do exactly what we want.

## Mentoring

Steps:
- [ ] Add `synstructure` dev-dependency to zerocopy-derive
- [ ] Add a basic test and confirm that it works (this and the previous step could be their own PR)
- [ ] Come up with as many test cases as you can think of! Consider using our existing tests (`zerocopy-derive/tests/*.rs`) and trybuild tests (`zerocopy-derive/tests/ui-*.rs`) to help give you ideas.

PR #879, opened 2024-02-14
Support deriving traits on data-carrying enums

djkoloski:
This PR is now ready for review.

~~This PR isn't yet complete. Still to do:~~

- [x] Write a padding check for data-carrying enums
- [x] Add support for deriving IntoBytes using the padding check
- [x] Add support for deriving FromZeros with a custom type check
- [x] Un-gate support for deriving Unaligned and FromBytes
- [x] Write tests for data-carrying enums
- [x] Justify new unsafe blocks in derived impls


djkoloski:
Blocked on #1515 so we can limit explicit enum discriminants to non-MSRV.

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,491 @@
...
+    }
+}
+
+pub(crate) fn discriminant_repr(reprs: &[EnumRepr]) -> &EnumRepr {
+    if reprs.contains(&EnumRepr::C) {
+        reprs.iter().find(|&r| r != &EnumRepr::C).unwrap_or(&EnumRepr::C)
+    } else {
+        reprs.first().unwrap()
+    }
+}
```
joshlf, line 66: Add a comment explaining why we're guaranteed this won't miss anything? E.g., why `#[repr(u8, u16)]` is impossible (since we'd only return the `u8` part).

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,491 @@
...
+    generics: &Generics,
+    idents: &RawIdents,
+    with_tag: bool,
+) -> TokenStream {
+    let discriminant_ident = &idents.discriminant;
+    let variant_struct_ident = idents.variant_struct(&variant.ident);
+
+    let type_params = generics.type_params().map(|p| &p.ident);
+    let phantom_ty = quote! {
+        ::core::marker::PhantomData<(#(#type_params,)*)>
```
joshlf, line 123: Here and elsewhere: Can you use `core` via `::zerocopy::macro_util::core_reexport` for hygeine?

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,491 @@
...
+                    //     original type.
+                    // - `assume_initialized`: `cast_unsized` removes the
+                    //   initialization invariant from `p`, so we re-assert that
+                    //   all of the bytes are initialized.
+                    let candidate = unsafe {
+                        fields
+                            .cast_unsized(|p: *mut #fields_ident| p as *mut #variant_struct)
+                            .assume_initialized()
+                    };
+                    #variant_struct::is_bit_valid(candidate)
```
joshlf, line 302: Here and elsewhere: should this be fully-qualified?

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,491 @@
...
+
+            #discriminant_type
+            #discriminant_consts
+            #(#variant_structs)*
+
+            // SAFETY:
+            // - `cast` is implemented as required.
+            // - By definition, `*mut Self` and `*mut [u8; size_of::<Self>()]`
+            //   are types of the same size.
+            let discriminant = unsafe { candidate.cast_unsized(|p: *mut Self| p as *mut [core_reexport::primitive::u8; core_reexport::mem::size_of::<#discriminant_ident>()]) };
```
joshlf, line 473: For `#[repr(P)]`, could we read this as a `P` instead of a `u8` array? It would maybe allow us to generate better code since we can assume alignment.

joshlf (changes requested):

zerocopy-derive/src/enums.rs
djkoloski: Leading double colon tells the compiler to look in the external prelude (links [1](https://doc.rust-lang.org/reference/paths.html), [2](https://doc.rust-lang.org/reference/names/preludes.html#extern-prelude)). `core` is always in the external prelude

jswrenn: I believe we favor `::zerocopy::macro_util::core_reexport` so that our macros would work inside 2015 edition crates.

zerocopy-derive/src/enums.rs
djkoloski: Are the sizes of `#[repr(C)]` enums well-defined?

joshlf: Yeah, it'd need to be just for `repr(u8)`, `repr(i16)`, etc.

djkoloski: I checked the implementation in rustc and it says that `#[repr(C)]` enums always have an `isize` discriminant. That's not documented AFAICT, but I can insert a const-eval assertion that `size_of::<EnumTag>() == size_of::<isize>()` so we fail gracefully if it's not `isize`-sized. Does that sound good?

joshlf: That'd be a stability hazard - if they ever change it, older versions of zerocopy would break. I'd vote that we remain conservative for `repr(C)`, but we should also put up a PR to the reference to document that as a guarantee, and then we can do what you proposed once it lands.

djkoloski: After reading more closely, the integer primitive for `repr(C)` enum discriminants is Very Complicated. It's at least as large as `c_int::BITS` but bigger if it needs to be, and unsigned if possible but can be signed if necessary. So we can't correctly guess the type of the discriminant in most cases.

I think it would be very disappointing if we had worse codegen for `repr(C)` than for other enums. I think `repr(C)` enums will be pretty common, but I could be wrong. We could work around this with something like:

```rust
struct Int;

trait OfSize<const N: usize> {
    type Type;
}

impl OfSize<1> for Int {
    type Type = i8;
}

impl OfSize<2> for Int {
    type Type = i16;
}

impl OfSize<4> for Int {
    type Type = i32;
}

impl OfSize<8> for Int {
    type Type = i64;
}

impl OfSize<16> for Int {
    type Type = i128;
}

// in macro:

#[repr(C)]
enum Discriminant {
    A,
    // ...
}

const SIZE: usize = ::core::mem::size_of::<Discriminant>();
type DiscInt = <Int as OfSize<SIZE>>::Type;

const A_DISC: DiscInt = DiscInt::from_ne_bytes(
    unsafe { transmute::<Foo, [u8; SIZE]>(Foo::A) }
);
// ...
```
And then we can read the discriminant as `DiscInt` and match against the discriminant consts. We can get away with this because the generated discriminant enum doesn't have any generics. Does that sound more appealing?

joshlf: Didn't realize it'd be this complicated. Let's just keep it as a u8 array in this PR, but file an issue to follow up with an improvement as described here in a future PR?

zerocopy-derive/src/enums.rs
djkoloski: Added lots of comments

zerocopy-derive/src/enums.rs
djkoloski: Done

zerocopy-derive/src/enums.rs
djkoloski: For variant structs, no. These are inherent impls that have the same name as `TryFromBytes`'s methods. To keep the generated code down as much as possible, the variant structs don't actually implement `TryFromBytes`. Still, these should probably be `<#variant_struct>::is_bit_valid` so I changed that.

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,550 @@
...
+
+            #[repr(C)]
+            #[allow(non_snake_case)]
+            union #fields_ident #impl_generics #where_clause {
+                #(
+                    pub #variant_idents: core_reexport::mem::ManuallyDrop<
+                        #variant_struct_idents #ty_generics,
+                    >,
+                )*
+                __nonempty: (),
```
joshlf, line 382: I presume Rust doesn't permit field-less unions? Can you add a comment explaining that?

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,550 @@
...
+            //     subfield pointer just points to a smaller portion of the
+            //     overall struct.
+            let fields = unsafe {
+                candidate
+                    .cast_unsized(|p: *mut Self| p as *mut #raw_ident)
+                    .assume_initialized()
+                    .project(|p: *mut #raw_ident| {
+                        core_reexport::ptr::addr_of_mut!((*p).fields)
+                    })
+            };
```
joshlf, line 441: Can you split this into multiple `unsafe` blocks that have separate safety comments?

joshlf (changes requested):
*Just reviewing `zerocopy-derive/src/enums.rs` for now. I'll follow up with reviews of other files.*

Could you take a quick look at #367 and see if it'd be easy to add to this PR? If it is, it'd be great to have at least one example output that we can assert against since there are so many moving pieces here.

jswrenn:
I'm seeing compilation errors on this code:
```rust
use zerocopy::*;
use zerocopy_derive::*;

#[derive(KnownLayout, TryFromBytes, Immutable)]
#[repr(u8)]
enum MyEnum {
    UnitLike,
    StructLike {
        a: u8,
        b: u16,
    },
    TupleLike(
        bool,
        char,
    ),
}

fn main() {}
```
The error is:
```
error[E0382]: use of moved value: `this`
    --> examples/dst.rs:4:23
     |
4    | #[derive(KnownLayout, TryFromBytes, Immutable)]
     |                       ^^^^^^^^^^^^
     |                       |
     |                       `this` moved due to this method call
     |                       value used here after move
     |                       move occurs because `this` has type `Ptr<'_, RawMyEnumVariantTupleLike, (A, invariant::Any, Initialized)>`, which does not implement the `Copy` trait
```

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,550 @@
...
+    quote! {
+        #(#consts)*
+    }
+}
+
+fn generate_variant_struct(
+    variant: &Variant,
+    generics: &Generics,
+    idents: &RawIdents,
+    with_tag: bool,
```
jswrenn, line 150: Is this is only invoked with `with_tag: false`?

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,550 @@
...
+
+    // We may or may not add a `tag_field` to our variants. This is because
+    // repr(int) enum variant structs start with the discriminant tag, but
+    // repr(C) enum variant structs do not. This allows repr(int) enum variant
+    // structs to more tightly pack their variant structs in some cases.
+
+    match &variant.fields {
+        Fields::Named(fields) => {
+            let tag_field =
+                if with_tag { Some(quote! { __tag: #discriminant_ident, }) } else { None };
```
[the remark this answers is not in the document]
jswrenn: Alternatively, we could instead emit tuple-like structs. Doing so would entirely eliminate the risk of collisions and wouldn't require any fancy ident generation.

jswrenn:
This PR is hefty, in part, because it generates ad-hoc `is_bit_valid` functions, as opposed to delegating to our existing derives. If we can increase the amount of delegation, we can decrease both the PR size and the degree to which we need to reason about unsafe code.

Concretely, I believe we could expand this:
```rust
#[derive(KnownLayout, Immutable)]
#[repr(u8)]
enum MyEnum<X, Y> {
    UnitLikeVariant,
    StructLikeVariant {
        a: u8,
        b: X,
    },
    TupleLikeVariant(
        bool,
        Y,
    ),
}
```
...into this:
```rust
unsafe impl<X, Y> ::zerocopy::TryFromBytes for MyEnum<X, Y>
where
    u8: ::zerocopy::TryFromBytes,
    X: ::zerocopy::TryFromBytes,
    bool: ::zerocopy::TryFromBytes,
    Y: ::zerocopy::TryFromBytes,
{
    fn only_derive_is_allowed_to_implement_this_trait() {}

    fn is_bit_valid<A>(
        mut candidate: ::zerocopy::Maybe<'_, Self, A>,
    ) -> ::zerocopy::macro_util::core_reexport::primitive::bool
    where
        A: ::zerocopy::pointer::invariant::Aliasing
            + ::zerocopy::pointer::invariant::AtLeast<
                ::zerocopy::pointer::invariant::Shared,
            >,
    {
        mod tags {
            #[repr(u8)]
            #[allow(dead_code)]
            enum Variants {
                UnitLikeVariant /* = an explicit discr, if provided */,
                StructLikeVariant,
                TupleLikeVariant,
            }

            type Tag = ::zerocopy::macro_util::core_reexport::primitive::u8;

            /// Has the same bit-validity as the tag of
            /// `Self::UnitLikeVariant`.
            #[derive(::zerocopy_derive::TryFromBytes)] // delegate!
            #[repr(u8)]
            pub(super) enum UnitLikeVariant {
                Tag = Variants::UnitLikeVariant as Tag,
            }

            /// Has the same bit-validity as the tag of
            /// `Self::StructLikeVariant`.
            #[derive(::zerocopy_derive::TryFromBytes)] // delegate!
            #[repr(u8)]
            pub(super) enum StructLikeVariant {
                Tag = Variants::StructLikeVariant as Tag,
            }

            /// Has the same bit-validity as the tag of
            /// `Self::TupleLikeVariant`.
            #[derive(::zerocopy_derive::TryFromBytes)] // delegate!
            #[repr(u8)]
            pub(super) enum TupleLikeVariant {
                Tag = Variants::TupleLikeVariant as Tag,
            }
        }

        use ::zerocopy::macro_util::{Variant, core_reexport::marker::PhantomData};

        /// Has the same layout (including bit validity) as
        /// `Self::UnitLikeVariant`.
        #[derive(::zerocopy_derive::TryFromBytes)] // delegate!
        #[repr(C)]
        struct UnitLikeVariant<X,Y>(
            tags::UnitLikeVariant,
            PhantomData<(X, Y)>
        );

        // SAFETY: The layout (including bit validity) of
        // `UnitLikeVariant<X,Y>` is identical to that of
        // `MyEnum<X, Y>::UnitLikeVariant`.
        unsafe impl<X, Y> Variant<MyEnum<X, Y>> for UnitLikeVariant<X,Y>
        {}

        /// Has the same layout (including bit validity) as
        /// `Self::StructLikeVariant`.
        #[derive(::zerocopy_derive::TryFromBytes)] // delegate!
        #[repr(C)]
        struct StructLikeVariant<X,Y>(
            tags::StructLikeVariant,
            u8,
            X,
            PhantomData<(X, Y)>
        );

        // SAFETY: The layout (including bit validity) of
        // `StructLikeVariant<X,Y>` is identical to that of
        // `MyEnum<X, Y>::StructLikeVariant`.
        unsafe impl<X, Y> Variant<MyEnum<X, Y>> for StructLikeVariant<X,Y>
        {}

        /// Has the same layout (including bit validity) as
        /// `Self::TupleLikeVariant`.
        #[derive(::zerocopy_derive::TryFromBytes)] // delegate!
        #[repr(C)]
        struct TupleLikeVariant<X, Y>(
            tags::TupleLikeVariant,
            bool,
            Y,
            PhantomData<(X, Y)>
        );

        // SAFETY: The layout (including bit validity) of
        // `TupleLikeVariant<X,Y>` is identical to that of
        // `MyEnum<X, Y>::TupleLikeVariant`.
        unsafe impl<X, Y> Variant<MyEnum<X, Y>> for TupleLikeVariant<X,Y>
        {}

        false
            || <UnitLikeVariant<X, Y> as Variant<Self>>::is_bit_valid(candidate.reborrow())
            || <StructLikeVariant<X, Y> as Variant<Self>>::is_bit_valid(candidate.reborrow())
            || <TupleLikeVariant<X, Y> as Variant<Self>>::is_bit_valid(candidate.reborrow())
    }
}
```
...with this additional helper trait added to `macro_util`:
```rust
/// Marks that `Self` has layout (and bit validity) identical to that
/// of a variant of the enum `E`.
/// 
/// # Safety
/// 
/// `Self`'s layout (including bit validity) must be identical to that
/// of a variant of the enum `E`.
pub unsafe trait Variant<E>: Sized {
    fn cast<I>(e: Ptr<'_, E, I>) -> Ptr<'_, Self, (I::Aliasing, invariant::Any, invariant::Any)>
    where
        I: Invariants,
    {
        // SAFETY: By precondition on this trait, `Self` has layout
        // identical to that of a variant of the enum `E`, thus the
        // size of `Self` is identical to that of `E`.
        unsafe { e.cast_unsized(|p| p as *mut Self) }
    }

    fn is_bit_valid<A>(
        candidate: crate::Maybe<'_, E, A>,
    ) -> bool
    where
        A: invariant::Aliasing
            + AtLeast<
                invariant::Shared,
            >,
        Self: TryFromBytes,
    {
        let candidate = Self::cast(candidate);
        // SAFETY: The above cast does not de-initialize the
        // candidate's bytes, thus the `Ptr`'s referent remains
        // initialized.
        let candidate = unsafe { candidate.assume_initialized() };
        TryFromBytes::is_bit_valid(candidate)
    }
}
```

jswrenn:
I don't follow. In my example, there's no 're-checking data we already know to be valid'. The discriminant check happens only after the cast.

djkoloski:
Sorry, I misunderstood your code slightly. In the code you provided, we try checking each of the variants of the enum one-by-one. This means that we do multiple checks on the tag of the enum, each time checking if it matches a single variant's discriminant. This means that 1. we're generating one branch per enum variant and 2. if we match an enum variant and still fail `is_bit_valid` then we continue checking other variants.

This PR currently reads the tag out once and does a single match on it to delegate to an appropriate `is_bit_valid` function. This guarantees that the compiler can optimize it to a jump table or branch table. As a result, we need the delegated `is_bit_valid` functions to avoid re-checking the tag - we wouldn't be calling `is_bit_valid` on a specific variant if we didn't know that it was the only possible variant that could match.

jswrenn:
I think we should optimize foremost for maintainability. The Rust compiler is notoriously bad at optimizing `match`s, and, in particular, does not optimizes byte array matches into jump tables last I checked.

At any rate, I think it's not too difficult to adapt my example to that potential optimization:
```rust
unsafe impl<X, Y> ::zerocopy::TryFromBytes for MyEnum<X, Y>
where
    u8: ::zerocopy::TryFromBytes,
    X: ::zerocopy::TryFromBytes,
    bool: ::zerocopy::TryFromBytes,
    Y: ::zerocopy::TryFromBytes,
{
    fn only_derive_is_allowed_to_implement_this_trait() {}

    fn is_bit_valid<A>(
        mut candidate: ::zerocopy::Maybe<'_, Self, A>,
    ) -> ::zerocopy::macro_util::core_reexport::primitive::bool
    where
        A: ::zerocopy::pointer::invariant::Aliasing
            + ::zerocopy::pointer::invariant::AtLeast<::zerocopy::pointer::invariant::Shared>,
    {
        use ::zerocopy::macro_util::{
            core_reexport::{marker::PhantomData, primitive},
            Variant,
        };

        pub type Tag = primitive::u8;

        #[repr(u8)]
        pub enum Variants {
            UnitLikeVariant, /* = an explicit discr, if provided */
            StructLikeVariant,
            TupleLikeVariant,
        }

        /// Has the same layout (including bit validity) as
        /// `Self::UnitLikeVariant`, except for the bit
        /// validity of its leading tag field, which has the
        /// validity of `MyEnum<X, Y>`'s primitive tag type.
        #[derive(::zerocopy_derive::TryFromBytes)] // delegate!
        #[repr(C)]
        struct UnitLikeVariant<X, Y>(Tag, PhantomData<(X, Y)>);

        impl<X, Y> UnitLikeVariant<X, Y> {
            const TAG: Tag = Variants::UnitLikeVariant as Tag;
        }

        // SAFETY: The layout (including bit validity) of
        // `UnitLikeVariant<X,Y>` is identical to that of
        // `MyEnum<X, Y>::UnitLikeVariant`, except for the
        // validity of its leading tag field, which has the
        // validity of `MyEnum<X, Y>`'s primitive tag type.
        unsafe impl<X, Y> Variant<MyEnum<X, Y>> for UnitLikeVariant<X, Y> {}

        /// Has the same layout (including bit validity) as
        /// `Self::StructLikeVariant`, except for the
        /// validity of its leading tag field, which has the
        /// validity of `MyEnum<X, Y>`'s primitive tag type.
        #[derive(::zerocopy_derive::TryFromBytes)] // delegate!
        #[repr(C)]
        struct StructLikeVariant<X, Y>(Tag, u8, X, PhantomData<(X, Y)>);

        impl<X, Y> StructLikeVariant<X, Y> {
            const TAG: Tag = Variants::StructLikeVariant as Tag;
        }

        // SAFETY: The layout (including bit validity) of
        // `StructLikeVariant<X,Y>` is identical to that of
        // `MyEnum<X, Y>::StructLikeVariant`, except for the
        // validity of its leading tag field, which has the
        // validity of `MyEnum<X, Y>`'s primitive tag type.
        unsafe impl<X, Y> Variant<MyEnum<X, Y>> for StructLikeVariant<X, Y> {}

        /// Has the same layout (including bit validity) as
        /// `Self::TupleLikeVariant`, except for the
        // validity of its leading tag field, which has the
        // validity of `MyEnum<X, Y>`'s primitive tag type.
        #[derive(::zerocopy_derive::TryFromBytes)] // delegate!
        #[repr(C)]
        struct TupleLikeVariant<X, Y>(Tag, bool, Y, PhantomData<(X, Y)>);

        impl<X, Y> TupleLikeVariant<X, Y> {
            const TAG: Tag = Variants::TupleLikeVariant as Tag;
        }

        // SAFETY: The layout (including bit validity) of
        // `TupleLikeVariant<X,Y>` is identical to that of
        // `MyEnum<X, Y>::TupleLikeVariant`, except for the
        // validity of its leading tag field, which has the
        // validity of `MyEnum<X, Y>`'s primitive tag type.
        unsafe impl<X, Y> Variant<MyEnum<X, Y>> for TupleLikeVariant<X, Y> {}

        let tag = unsafe {
            candidate.reborrow().cast_unsized(|p| p as *mut Tag).assume_valid().read_unaligned()
        };

        match tag {
            UnitLikeVariant::<X, Y>::TAG => {
                <UnitLikeVariant<X, Y> as Variant<Self>>::is_bit_valid(candidate.reborrow())
            }
            StructLikeVariant::<X, Y>::TAG => {
                <StructLikeVariant<X, Y> as Variant<Self>>::is_bit_valid(candidate.reborrow())
            }
            TupleLikeVariant::<X, Y>::TAG => {
                <TupleLikeVariant<X, Y> as Variant<Self>>::is_bit_valid(candidate.reborrow())
            }
            _ => false,
        }
    }
}
```

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,550 @@
...
+// repr(int) enums. The "desugared" layouts for these two enums are taken from
+// the reference.
+//
+// repr(C) enums reference:
+// https://doc.rust-lang.org/reference/type-layout.html#reprc-enums-with-fields
+//
+// repr(int) enums reference:
+// https://doc.rust-lang.org/reference/type-layout.html#primitive-representation-of-enums-with-fields
+
+struct RawIdents {
```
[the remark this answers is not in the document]
djkoloski: Done

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,550 @@
...
+
+impl RawIdents {
+    const DISCRIMINANT_VALUE_PREFIX: &'static str = "RAW_DISCRIMINANT_";
+
+    fn for_enum(ident: &Ident) -> Self {
+        Self {
+            r#enum: Ident::new(&format!("Raw{}Enum", ident), ident.span()),
+            discriminant: Ident::new(&format!("Raw{}Discriminant", ident), ident.span()),
+            fields: Ident::new(&format!("Raw{}Fields", ident), ident.span()),
+            variant_prefix: format!("Raw{}Variant", ident),
```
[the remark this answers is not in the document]
djkoloski: No particular reason, I included the name for ease of grepping around `cargo expand`ed source

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,550 @@
...
+pub(crate) fn derive_is_bit_valid(
+    reprs: &[EnumRepr],
+    ident: &Ident,
+    generics: &Generics,
+    data: &DataEnum,
+) -> TokenStream {
+    let idents = RawIdents::for_enum(ident);
+
+    if reprs.contains(&EnumRepr::C) {
+        derive_is_bit_valid_c(data, generics, &idents)
```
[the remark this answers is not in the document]
djkoloski: Done

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,550 @@
...
+    } else {
+        derive_is_bit_valid_primitive(data, generics, discriminant_repr(reprs), &idents)
+    }
+}
+
+pub(crate) fn discriminant_repr(reprs: &[EnumRepr]) -> &EnumRepr {
+    // This returns the first non-C, non-align repr. We know that `reprs` will
+    // contain at least `C` if not also in integer because we check the valid
+    // reprs in the derive which calls this function. If we don't find a non-C,
+    // non-align repr then we know that it was `repr(C)` all along.
```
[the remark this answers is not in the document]
djkoloski: I think this is a good idea but probably better suited for a follow-up PR which cleans up our repr handling for enums more broadly

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,550 @@
...
+    reprs.iter().find(|&r| !matches!(r, EnumRepr::C | EnumRepr::Align(_))).unwrap_or(&EnumRepr::C)
+}
+
+pub(crate) fn generate_discriminant_type(
```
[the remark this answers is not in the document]
djkoloski: I've just been referencing the examples in the reference, so I linked those

zerocopy-derive/src/enums.rs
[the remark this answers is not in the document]
djkoloski: Switched to tuple structs

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,550 @@
...
+    match &variant.fields {
+        Fields::Named(fields) => {
+            let tag_field =
+                if with_tag { Some(quote! { __tag: #discriminant_ident, }) } else { None };
+
+            let struct_fields = fields.named.iter().map(|f| {
+                let ident = &f.ident;
+                let ty = &f.ty;
+                quote! {
+                    pub #ident: #ty
```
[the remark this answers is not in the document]
djkoloski: Nope, removed

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,550 @@
...
+    let (impl_generics, _, where_clause) = generics.split_for_impl();
+
+    let trait_path = Trait::TryFromBytes.path();
+
+    // We may or may not add a `tag_field` to our variants. This is because
+    // repr(int) enum variant structs start with the discriminant tag, but
+    // repr(C) enum variant structs do not. This allows repr(int) enum variant
+    // structs to more tightly pack their variant structs in some cases.
+
+    match &variant.fields {
```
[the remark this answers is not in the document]
djkoloski: Yes

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,550 @@
...
+                ) #where_clause;
+
+                impl #variant_struct_ident #impl_generics #where_clause {
+                    fn is_bit_valid<A>(this: ::zerocopy::Maybe<'_, Self, A>) -> bool
+                    where
+                        A: ::zerocopy::pointer::invariant::Aliasing
+                            + ::zerocopy::pointer::invariant::AtLeast<::zerocopy::pointer::invariant::Shared>,
+                    { true }
+                }
+            }
```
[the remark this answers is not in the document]
djkoloski: Simplified this logic

zerocopy-derive/src/enums.rs
djkoloski: Done

zerocopy-derive/src/enums.rs
djkoloski: Done

src/macro_util.rs
```diff
@@ -294,11 +294,19 @@ macro_rules! struct_has_padding {
 #[doc(hidden)] // `#[macro_export]` bypasses this module's `#[doc(hidden)]`.
 #[macro_export]
 macro_rules! union_has_padding {
-    ($t:ty, $($ts:ty),*) => {
+    ($t:ty, [$($ts:ty),*]) => {
         false $(|| ::zerocopy::macro_util::core_reexport::mem::size_of::<$t>() != ::zerocopy::macro_util::core_reexport::mem::size_of::<$ts>())*
     };
 }
 
+#[doc(hidden)]
```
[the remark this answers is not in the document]
djkoloski: Done

src/macro_util.rs
```diff
@@ -294,11 +294,19 @@ macro_rules! struct_has_padding {
...
+#[macro_export]
+macro_rules! enum_has_padding {
```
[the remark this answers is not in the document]
djkoloski: Done

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,550 @@
...
+        // NOTE: We're using `[u8; N]` as the type of the discriminant contants
+        // here because the sizes of `#[repr(C)]` fieldless enums is not well-
+        // specified. There _might_ be an opportunity to generate better code
+        // for enums with the primitive representation, but it also may not make
+        // any difference.
+        quote! {
+            // SAFETY: The raw discriminant enum is a C-style enum, and all
+            // C-style enums are represented as simple integers. All integers
+            // can be transmuted to their constituent bytes.
+            const #const_ident: [u8; core_reexport::mem::size_of::<#discriminant>()] = unsafe {
```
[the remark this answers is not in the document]
djkoloski: Done

zerocopy-derive/src/enums.rs
djkoloski: Copy-paste error, fixed now

zerocopy-derive/src/enums.rs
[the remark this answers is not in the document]
joshlf: I should clarify: it'd also be good to a top-level example of the full expansion that we could refer to.

zerocopy-derive/src/enums.rs
joshlf: Follow-up: Can we change this to an enum so it's self-documenting at the call sites? E.g., `generate_variant_struct(..., TagLocation::Outer)` or something.

djkoloski:
Documenting feedback:
1. Switch to recursively calling `TryFromBytes` derive on variant structs (make sure that this doesn't incur an additional validation cost)
2. Go ahead with primitive lookup trait to simplify match and discriminant reading
3. Try to deduplicate some more code between `repr(C)` and `repr(int)`

jswrenn:
Example that collapses `repr(C)` and `repr(int)` into a single codepath, differing only by the RHS of two type aliases:
```rust
unsafe impl<X, Y> ::zerocopy::TryFromBytes for MyEnum<X, Y>
where
    u8: ::zerocopy::TryFromBytes,
    X: ::zerocopy::TryFromBytes,
    bool: ::zerocopy::TryFromBytes,
    Y: ::zerocopy::TryFromBytes,
{
    fn only_derive_is_allowed_to_implement_this_trait() {}

    fn is_bit_valid<A>(
        mut candidate: ::zerocopy::Maybe<'_, Self, A>,
    ) -> ::zerocopy::macro_util::core_reexport::primitive::bool
    where
        A: ::zerocopy::pointer::invariant::Aliasing
            + ::zerocopy::pointer::invariant::AtLeast<::zerocopy::pointer::invariant::Shared>,
    {
        use meta::*;

        mod meta {
            use super::*;
            pub(super) use ::zerocopy::macro_util::{
                core_reexport::{marker::PhantomData, mem::size_of, primitive},
                SizeToTag, Variant,
            };

            #[repr(u8)]
            pub enum Variants {
                UnitLikeVariant, /* = an explicit discr, if provided */
                StructLikeVariant,
                TupleLikeVariant,
            }

            pub type Tag = SizeToTag<{ size_of::<Variants>() }>;

            // reverse the RHS of these, depending if repr(int)
            pub type OuterTag = Tag;
            pub type InnerTag = ();

            pub mod tags {
                /* generate a const for each variant */

                pub const UnitLikeVariant: super::Tag =
                    super::Variants::UnitLikeVariant as super::Tag;
                pub const StructLikeVariant: super::Tag =
                    super::Variants::StructLikeVariant as super::Tag;
                pub const TupleLikeVariant: super::Tag =
                    super::Variants::TupleLikeVariant as super::Tag;
            }

            pub(super) mod variants {
                use super::*;

                /* generate a struct with a leading `InnerTag` field for each fieldful variant */

                /// Has the same layout (including bit validity) as
                /// `Self::StructLikeVariant`, except for the
                /// validity of its leading tag field, which has the
                /// validity of `MyEnum<X, Y>`'s primitive tag type.
                #[derive(::zerocopy_derive::TryFromBytes)] // delegate!
                #[repr(C)]
                pub struct StructLikeVariant<X, Y>(super::InnerTag, u8, X, PhantomData<(X, Y)>);

                // SAFETY: The layout (including bit validity) of
                // `StructLikeVariant<X,Y>` is identical to that of
                // `MyEnum<X, Y>::StructLikeVariant`, except for the
                // validity of its leading tag field, which has the
                // validity of `MyEnum<X, Y>`'s primitive tag type.
                unsafe impl<X, Y> Variant<self::MyEnum<X, Y>> for StructLikeVariant<X, Y> {}

                /// Has the same layout (including bit validity) as
                /// `Self::TupleLikeVariant`, except for the
                // validity of its leading tag field, which has the
                // validity of `MyEnum<X, Y>`'s primitive tag type.
                #[derive(::zerocopy_derive::TryFromBytes)] // delegate!
                #[repr(C)]
                pub struct TupleLikeVariant<X, Y>(super::InnerTag, bool, Y, PhantomData<(X, Y)>);

                // SAFETY: The layout (including bit validity) of
                // `TupleLikeVariant<X,Y>` is identical to that of
                // `MyEnum<X, Y>::TupleLikeVariant`, except for the
                // validity of its leading tag field, which has the
                // validity of `MyEnum<X, Y>`'s primitive tag type.
                unsafe impl<X, Y> Variant<MyEnum<X, Y>> for TupleLikeVariant<X, Y> {}
            }
        }

        let tag = unsafe {
            candidate.reborrow().cast_unsized(|p| p as *mut Tag).assume_valid().read_unaligned()
        };

        match tag {
            tags::UnitLikeVariant => true,
            tags::StructLikeVariant => {
                <variants::StructLikeVariant<X, Y> as Variant<Self>>::is_bit_valid(
                    candidate.reborrow(),
                )
            }
            tags::TupleLikeVariant => {
                <variants::TupleLikeVariant<X, Y> as Variant<Self>>::is_bit_valid(
                    candidate.reborrow(),
                )
            }
            _ => false,
        }
    }
}
```
...with supporting `macro_util`:
```rust
pub unsafe trait Variant<E>: Sized {
    fn cast<I>(e: Ptr<'_, E, I>) -> Ptr<'_, Self, (I::Aliasing, invariant::Any, invariant::Any)>
    where
        I: Invariants,
    {
        // SAFETY: By precondition on this trait, `Self` has layout
        // identical to that of a variant of the enum `E`, thus the
        // size of `Self` is identical to that of `E`.
        unsafe { e.cast_unsized(|p| p as *mut Self) }
    }

    fn is_bit_valid<A>(candidate: crate::Maybe<'_, E, A>) -> bool
    where
        A: invariant::Aliasing + AtLeast<invariant::Shared>,
        Self: TryFromBytes,
    {
        let candidate = Self::cast(candidate);
        // SAFETY: The above cast does not de-initialize the
        // candidate's bytes, thus the `Ptr`'s referent remains
        // initialized.
        let candidate = unsafe { candidate.assume_initialized() };
        TryFromBytes::is_bit_valid(candidate)
    }
}

mod size_to_tag {
    pub trait SizeToTag<const SIZE: usize> {
        type Tag;
    }

    impl SizeToTag<1> for () {
        type Tag = u8;
    }
    impl SizeToTag<2> for () {
        type Tag = u16;
    }
    impl SizeToTag<4> for () {
        type Tag = u32;
    }
    impl SizeToTag<8> for () {
        type Tag = u64;
    }
    impl SizeToTag<16> for () {
        type Tag = u128;
    }
}

pub type SizeToTag<const SIZE: usize> = <() as size_to_tag::SizeToTag<SIZE>>::Tag;
```

jswrenn:
> Documenting feedback:
> 
> 1. Switch to recursively calling `TryFromBytes` derive on variant structs (make sure that this doesn't incur an additional validation cost)
> 2. Go ahead with primitive lookup trait to simplify match and discriminant reading
> 3. Try to deduplicate some more code between `repr(C)` and `repr(int)`

@djkoloski Also reducing ident interpolation, wherever possible. :-)

djkoloski:
It looks like modules aren't going to work out unfortunately:

```rust
const THREE: ::core::primitive::i8 = 3;

#[derive(Eq, PartialEq, Debug, imp::KnownLayout, imp::Immutable, imp::TryFromBytes)]
#[repr(i8)]
enum Blah {
    A = 1,
    B = 0,
    C = 1 + 2,
    D = 3 + THREE,
}
```
```
error[E0425]: cannot find value `THREE` in this scope
  --> zerocopy-derive\tests\enum_try_from_bytes.rs:86:13
   |
86 |     D = 3 + THREE,
   |             ^^^^^ not found in this scope
   |
   = help: consider importing this constant:
           crate::THREE
```

We can't just `use super::*` either because the enum being derived on could be inside a function, and that `use` would only pull in names defined in the module and not in the function itself. i.e. This wouldn't work:

```rust
fn scope() {
    const THREE: ::core::primitive::i8 = 3;

    #[derive(Eq, PartialEq, Debug, imp::KnownLayout, imp::Immutable, imp::TryFromBytes)]
    #[repr(i8)]
    enum Blah {
        A = 1,
        B = 0,
        C = 1 + 2,
        D = 3 + THREE,
    }
}
```

So it seems like a general rule that we can't use `mod`s in macros if we may emit the names of any items.

joshlf:
If we have to put everything at the top-level, presumably we have to worry about name collisions? In other words, e.g. if the innermost scope defines `const FOO`, then if our generated code also defines `const FOO`, [that's a compilation error](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=dd5618cf51831121be6d39926bddad89).

Assuming that's right, do we have any thoughts on how to avoid name collisions? Since they could be collisions with names that we have no visibility into, our best bet might just be to generate a random prefix or something.

djkoloski:
I'm currently going through and prefixing all the items we generate with triple underscores. I've seen a general convention of:
- One underscore: unused item
- Two underscores: private implementation detail (usually behind `#[doc(hidden)]`)
- Three underscores: macro output (avoid name collisions)

jswrenn:
That sounds reasonable.

joshlf:
Yeah as long as we include "zerocopy" in the prefix, that strikes me as sufficiently unlikely to collide. Worst comes to worst we can implement randomization later if need be.

jswrenn:
Is there a risk of collision here, or does hygiene save us? And if there is a type name collision, can that lead to unsoundness? If so, we should document this publicly as a known soundness hole.

joshlf:
Name collisions between named consts in the same scope [are a compile error](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=dd5618cf51831121be6d39926bddad89). Not a soundness error, though.

djkoloski:
The best hygiene we can get in a proc-macro is mixed hygiene, which only applies to local variables, labels, and `$crate`. Everything else uses call-site hygiene (i.e. is unhygenic). I poked around a bit and the worst I was able to do was get a name collision by naming a variant `___zerocopy_nonempty` (which caused a compile error). I fixed that by prefixing field names, so that's no longer an issue. I can't see any other opportunities to cause problems with name collisions.

jswrenn:
I'm seeing compiler errors when I try to `cargo check` this example:
```rust
use zerocopy::*;
use zerocopy_derive::*;

#[derive(KnownLayout, TryFromBytes, Immutable)]
#[repr(C)]
enum MyEnum<X, Y> {
    UnitLikeVariant,
    StructLikeVariant { a: u8, b: X },
    TupleLikeVariant(bool, Y),
}

fn main() {}
```

It looks like the derive isn't emitting `#[derive(TryFromBytes)]` on `___ZerocopyVariantStruct_StructLikeVariant` and `___ZerocopyVariantStruct_TupleLikeVariant`.

djkoloski:
The derive was correctly adding the derive attribute. The problem was that the attribute was deriving `::zerocopy::TryFromBytes`, but your dependencies did not enable the `derive` feature on zerocopy. Assuming that this is a configuration that we want to support (i.e. zerocopy and zerocopy-derive, but without the `derive` feature enabled in zerocopy), I have changed the macro to derive `::zerocopy_derive::TryFromBytes` instead.

jswrenn:
Ah, gotcha. Yes, its actually the recommend configuration, since it allows the crates to be compiled in parallel.

zerocopy-derive/src/enums.rs
djkoloski: Switching to outer tag / inner tag to unify repr(C) and repr(int) code paths make this obsolete

zerocopy-derive/src/enums.rs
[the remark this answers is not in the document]
djkoloski: These values are no longer used

zerocopy-derive/src/enums.rs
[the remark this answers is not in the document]
djkoloski: Filed #1581

jswrenn:
Seeing a compilation error on this:
```
use std::marker::PhantomData;

use zerocopy::*;
use zerocopy_derive::*;

#[derive(KnownLayout, TryFromBytes, Immutable)]
#[repr(C)]
enum MyEnum<'a, 'b, X, Y> {    
    Foo(PhantomData<&'a X>),
    Bar(PhantomData<&'b Y>),
}

fn main() {}
```
The error is caused by the generated `PhantomData<...>`s in the variant structs not including lifetimes. I think there's an elegant fix to this. Rather than generating this for `Foo`:
```
#[repr(C)]
#[allow(non_snake_case)]
struct ___ZerocopyVariantStruct_Foo<'a, 'b, X, Y>(
    core_reexport::mem::MaybeUninit<___ZerocopyInnerTag>,
    PhantomData<&'a X>,
    core_reexport::marker::PhantomData<(X, Y)>,
);
```
...generate this instead:
```rust
#[repr(C)]
#[allow(non_snake_case)]
struct ___ZerocopyVariantStruct_Foo<'a, 'b, X, Y>(
    core_reexport::mem::MaybeUninit<___ZerocopyInnerTag>,
    PhantomData<&'a X>,
    core_reexport::marker::PhantomData<MyEnum<'a, 'b, X, Y>>, // <- ensures all lifetimes, generics, consts are captured with correct variance
);
```

djkoloski:
Fixed

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,359 @@
...
+            // - The returned pointer is cast from `p`, and so has the same
+            //   provenance as `p`.
+            // - The raw enum has the same types at the same locations as the
+            //   original enum, and so preserves the locations of any
+            //   `UnsafeCell`s.
+            let raw_enum = unsafe {
+                candidate.cast_unsized(|p: *mut Self| {
+                    p as *mut ___ZerocopyRawEnum #ty_generics
+                })
+            };
```
jswrenn, line 333: This `forget_exclusive` call can cause issues if the enum contains `UnsafeCell`s. I think you can avoid it by re-shuffling where you do your `.reborrow()`s, like so:
```suggestion
            // SAFETY:
            // - `cast` is implemented as required.
            // - By definition, `*mut Self` and `*mut __ZerocopyTagPrimitive`
            //   are types of the same size.
            let tag = unsafe {
                candidate.reborrow().cast_unsized(|p: *mut Self| { // <-- REBORROW
                    p as *mut ___ZerocopyTagPrimitive
                })
            };
            // SAFETY: Since `candidate` has the invariant `Initialized`, we
            // know that `candidate`'s referent (and thus `tag`'s referent) is
            // as-initialized as `Self`. Since all of the possible tag types
            // always have all of their bytes initialized, we know that
            // `tag`'s referent has all of its bytes initialized. Since the
            // validity invariant of `__ZerocopyTagPrimitive` is just that all
            // of its bytes are initialized, we know that `tag`'s referent is
            // bit-valid.
            let tag = unsafe { tag.assume_valid() };
            let tag = tag.read_unaligned();

            // SAFETY:
            // - The raw enum has the same size as the input enum.
            // - The returned pointer is cast from `p`, and so has the same
            //   provenance as `p`.
            // - The raw enum has the same types at the same locations as the
            //   original enum, and so preserves the locations of any
            //   `UnsafeCell`s.
            let raw_enum = unsafe {
                candidate.reborrow().cast_unsized(|p: *mut Self| { // <-- REBORROW
                    p as *mut ___ZerocopyRawEnum #ty_generics
                })
            };
```

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,359 @@
...
+            // SAFETY: Since `candidate` has the invariant `Initialized`, we
+            // know that `candidate`'s referent (and thus `tag`'s referent) is
+            // as-initialized as `Self`. Since all of the possible tag types
+            // always have all of their bytes initialized, we know that
+            // `tag`'s referent has all of its bytes initialized. Since the
+            // validity invariant of `__ZerocopyTagPrimitive` is just that all
+            // of its bytes are initialized, we know that `tag`'s referent is
+            // bit-valid.
+            let tag = unsafe { tag.assume_valid() };
+            let tag = tag.read_unaligned();
```
jswrenn, line 320: This is a nit, but the tag extraction is a little easier to read in the generated output if it's grouped into a single binding:

```rust
            let tag = {
                // SAFETY:
                // - `cast` is implemented as required.
                // - By definition, `*mut Self` and `*mut __ZerocopyTagPrimitive`
                //   are types of the same size.
                let tag = unsafe {
                    candidate.cast_unsized(|p: *mut Self| {
                        p as *mut ___ZerocopyTagPrimitive
                    })
                };
                // SAFETY: Since `candidate` has the invariant `Initialized`, we
                // know that `candidate`'s referent (and thus `tag`'s referent) is
                // as-initialized as `Self`. Since all of the possible tag types
                // always have all of their bytes initialized, we know that
                // `tag`'s referent has all of its bytes initialized. Since the
                // validity invariant of `__ZerocopyTagPrimitive` is just that all
                // of its bytes are initialized, we know that `tag`'s referent is
                // bit-valid.
                let tag = unsafe { tag.assume_valid() };
                tag.read_unaligned()
            };
```

jswrenn, line 320: We can simplify this proof a bit by calling the unsafe `tag.assume_initialized()` — the justification for this is just that `candidate` is `Initialized` and casting it doesn't clear invariant. Then, call `tag.bikeshed_recall_valid()`, which is safe and handles the proof that initialized implies valid for the tag type.

zerocopy-derive/src/lib.rs
```diff
@@ -908,15 +846,28 @@ const STRUCT_UNION_INTO_BYTES_CFG: Config<StructRepr> = Config {
 // An enum is `IntoBytes` if it is field-less and has a defined repr.
 
 fn derive_into_bytes_enum(ast: &DeriveInput, enm: &DataEnum) -> proc_macro2::TokenStream {
```
jswrenn, line 848: Presently, this derive:
```rust
#[derive(IntoBytes)]
#[repr(u8)]
enum Foo {
    A(u8),
    B(u8),
}
```
...expands to:
```rust
const _: () = {
    #[repr(u8)]
    #[allow(dead_code)]
    enum ___ZerocopyTag {
        A,
        B,
    }
    #[allow(deprecated)]
    unsafe impl ::zerocopy::IntoBytes for Foo
    where
        u8: ::zerocopy::IntoBytes,
        u8: ::zerocopy::IntoBytes,
        ::zerocopy::macro_util::HasPadding<
            Foo,
            {
                false
                    || ::zerocopy::macro_util::core_reexport::mem::size_of::<Foo>()
                        != (::zerocopy::macro_util::core_reexport::mem::size_of::<
                            ___ZerocopyTag,
                        >()
                            + ::zerocopy::macro_util::core_reexport::mem::size_of::<
                                u8,
                            >())
                    || ::zerocopy::macro_util::core_reexport::mem::size_of::<Foo>()
                        != (::zerocopy::macro_util::core_reexport::mem::size_of::<
                            ___ZerocopyTag,
                        >()
                            + ::zerocopy::macro_util::core_reexport::mem::size_of::<
                                u8,
                            >())
            },
        >: ::zerocopy::macro_util::ShouldBe<false>,
    {
        fn only_derive_is_allowed_to_implement_this_trait() {}
    }
};
```

I think this is the only place we're generating items outside of the trait impl. Can we move that generated `___ZerocopyTag` type into the body of that const expr? If we can do so, then we don't need to generate the wrapping `const _: () = {...}` blocks.

jswrenn (changes requested):

zerocopy-derive/src/enums.rs
djkoloski: I agree, but `bikeshed_recall_valid` is `pub(crate)`.

jswrenn: Add `#[inline]`, `#[doc(hidden)]`, and make it `pub`! :-)

djkoloski: Sure. I assumed that since it was named `bikeshed` that you weren't interested in making it public yet, so hiding it seems sensible. In the interest of bikeshedding, I'd call this function `always_valid` since it's using the fact that the type is unconditionally valid to add the validity invariant to the pointer.

jswrenn: The `bikeshed` is just there for until we remove the `doc(hidden)`. Previous contenders have included `recall_valid` (as a counterpart to `forget_valid`). Josh had another idea that I can't remember.

For now, just keep the `bikeshed`.

djkoloski: Yeah, to clarify I have no intention of renaming it in this PR.

zerocopy-derive/src/enums.rs
djkoloski: Done

zerocopy-derive/src/enums.rs
djkoloski: Done

zerocopy-derive/src/lib.rs
djkoloski: Done. Note that the stderr output for `enum.rs` has changed as a result.

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,362 @@
...
+
+            let tag = {
+                // SAFETY:
+                // - The provided cast addresses a subset of the bytes addressed
+                //   by `candidate` because it addresses the starting tag of the
+                //   enum. Because the pointer is cast from `candidate`, it has
+                //   the same provenance as it. There are no `UnsafeCell`s in
+                //   the tag because it is a primitive integer.
+                // - By definition, `Self` and `__ZerocopyTagPrimitive` are
+                //   types of the same size.
```
jswrenn, line 312:
```suggestion
                // - The size of `__ZerocopyTagPrimitive` is less-than-or-equal
                // to that of `Self`.
```
`__ZerocopyTagPrimitive` will be smaller if `Self` is a dataful enum, right?

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,362 @@
...
+                let tag_ptr = unsafe {
+                    candidate.reborrow().cast_unsized(|p: *mut Self| {
+                        p as *mut ___ZerocopyTagPrimitive
+                    })
+                };
+                // SAFETY: `candidate` is initialized, and `tag_ptr` points to
+                // a region completely contained within `candidate`. Therefore,
+                // `tag_ptr` is also initialized.
```
jswrenn, line 320:
```suggestion
                // SAFETY: `tag_ptr` is casted from `candidate`, whose referent
                // is `Initialized`. Since we have not written uninitialized
                // bytes to the referent, `tag_ptr` is also `Initialized`.
```

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,362 @@
...
+                let tag_ptr = unsafe { tag_ptr.assume_initialized() };
+                tag_ptr.bikeshed_recall_valid().read_unaligned()
+            };
+
+            // SAFETY:
+            // - The raw enum has the same size as the input enum.
```
jswrenn, line 326: The derive needs to forward `align` annotations where appropriate. For example, deriving `TryFromBytes` on this:
```rust
#[repr(C, align(64))]
enum MyEnum {
    Foo(u8),
}
```
violates this safety condition. `___ZerocopyRawEnum` has a size of `8` on x86_64, whereas `MyEnum` has a size of `64`.

jswrenn (dismissed):

jswrenn:
Something I've been doing as I review this is replacing the `// SAFETY` comments in the derive with `" SAFETY: ...";` exprs. These get preserved in the expanded output, so I can evaluate the safety comments in situ.

zerocopy-derive/src/enums.rs
djkoloski: I think that bullet point is vestigial, I reworded the first bullet point to cover all of the points and didn't delete the second. Removing that in the next push.

joshlf:
> Something I've been doing as I review this is replacing the `// SAFETY` comments in the derive with `" SAFETY: ...";` exprs. These get preserved in the expanded output, so I can evaluate the safety comments in situ.

Another option would be:
- Expose `safety_comment!` as a `#[doc(hidden)] #[macro_export]` (and maybe move it to live in `macro_util`)
- In `safety_comment!`...
   - Make it optional to actually provide a macro to call (ie, it's acceptable to only provide a doc comment and nothing else)
   - In the macro expansion, emit that doc comment on a `const _: () = ();` or similar so that it has an item to live on
- Inside of `quote!(...)`, do:
  ```rust
  safety_comment! {
      /// SAFETY:
      /// Blah blah blah
  }
  ```

jswrenn:
@joshlf I initially did roughly that, but the extra `const _: () = ();`s are pretty distracting in the expanded output.

zerocopy-derive/src/enums.rs
djkoloski: That isn't be required for safety. The alignment and overall size of the original enum is higher, but we don't use the raw enum for anything except projecting into the fields. I have reworded this safety comment to note that the raw enum may have lower alignment and size.

zerocopy-derive/src/enums.rs
djkoloski: Done.

zerocopy-derive/src/lib.rs
```diff
@@ -510,105 +512,21 @@ const STRUCT_UNION_ALLOWED_REPR_COMBINATIONS: &[&[StructRepr]] = &[
...
-        // If the enum could implement `FromBytes`, we can avoid emitting a
-        // match statement. This is faster to compile, and generates code which
-        // performs better.
-        quote!({
-            // Prevent an "unused" warning.
-            let _ = candidate;
-            // SAFETY: If the enum could implement `FromBytes`, then all bit
-            // patterns are valid. Thus, this is a sound implementation.
-            true
-        })
```
joshlf, line 542: Was it intentional to remove this optimization?

zerocopy-derive/src/lib.rs
```diff
@@ -510,105 +512,21 @@ const STRUCT_UNION_ALLOWED_REPR_COMBINATIONS: &[&[StructRepr]] = &[
...
-                let d: [core_reexport::primitive::u8; core_reexport::mem::size_of::<Self>()] = unsafe { core_reexport::mem::transmute(v) };
-                // SAFETY: Here we check that the bits of the argument
-                // `candidate` are equal to the bits of a `Self` constructed
-                // using safe code. If this condition passes, then we know that
-                // `candidate` refers to a bit-valid `Self`.
-                discriminant == d
-            })*
-        )
-    };
+    // Generate the equivalent enum in terms of just structs and unions
```
joshlf, line 517: This comment feels weird here since it's describing an implementation detail of *how* `derive_is_bit_valid` works.

zerocopy-derive/src/lib.rs
```diff
@@ -1000,11 +956,6 @@ const STRUCT_UNION_UNALIGNED_CFG: Config<StructRepr> = Config {
 // - `repr(u8)` or `repr(i8)`
 
 fn derive_unaligned_enum(ast: &DeriveInput, enm: &DataEnum) -> proc_macro2::TokenStream {
-    if !enm.is_fieldless() {
-        return Error::new_spanned(ast, "only field-less enums can implement Unaligned")
-            .to_compile_error();
-    }
-
```
joshlf, line 1007: Since we're removing this restriction, remove or update the comment at the bottom of this function, which currently assumes field-less enums?

joshlf (changes requested):
I've reviewed most of this PR, but still need to review the following files:
Still need to review:
- `zerocopy-derive/src/ext.rs`
- `zerocopy-derive/src/repr.rs`
- `zerocopy-derive/src/lib.rs`
- `zerocopy-derive/tests/*`

It may be later today or tomorrow before I get a chance to review those files, so I figured it'd be better for you to at least have these comments now.

zerocopy-derive/src/enums.rs
joshlf, line 1: [Uber nit; feel free to disregard] Maybe call this `enum.rs` singular to be consistent with `repr.rs`?

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,365 @@
+// Copyright 2023 The Fuchsia Authors
```
joshlf, line 1:
```suggestion
// Copyright 2024 The Fuchsia Authors
```

zerocopy-derive/src/lib.rs
```diff
@@ -1075,10 +1041,18 @@ impl PaddingCheck {
...
 
         Ident::new(s, Span::call_site())
     }
+
+    fn validator_macro_context(&self) -> Option<&TokenStream> {
+        match self {
+            PaddingCheck::Struct | PaddingCheck::Union => None,
+            PaddingCheck::Enum(context) => Some(context),
+        }
+    }
```
joshlf, line 1055: Doc comment? It's not obvious from the definition what this is for.

zerocopy-derive/src/lib.rs
```diff
@@ -1093,16 +1067,26 @@ enum Trait {
...
 impl Trait {
-    fn path(&self) -> Path {
-        let span = Span::call_site();
-        let root = if *self == Self::Sized {
-            quote_spanned!(span=> ::zerocopy::macro_util::core_reexport::marker)
-        } else {
-            quote_spanned!(span=> ::zerocopy)
-        };
-        let ident = Ident::new(&format!("{:?}", self), span);
-        parse_quote_spanned! {span=> #root::#ident}
```
joshlf, line 1105: Does the new impl discard span information, or am I misreading? If so, is it possible to preserve span information, or is there a reason we should discard it?

zerocopy-derive/src/repr.rs
```diff
@@ -171,6 +174,33 @@ define_kind_specific_repr!(
...
+            EnumRepr::U16 => "u16",
+            EnumRepr::U32 => "u32",
+            EnumRepr::U64 => "u64",
+            EnumRepr::Usize => "usize",
+            EnumRepr::I8 => "i8",
+            EnumRepr::I16 => "i16",
+            EnumRepr::I32 => "i32",
+            EnumRepr::I64 => "i64",
+            EnumRepr::Isize => "isize",
+            EnumRepr::Align(_) => return None,
```
joshlf, line 191: Maybe put an `unimplemented!()` here so we know if we accidentally try to serialize an align repr via `ToTokens`?

zerocopy-derive/tests/enum_from_zeros.rs
```diff
@@ -44,9 +44,58 @@ enum ImplicitNonFirstVariantIsZero {
...
+#[repr(i8)]
+enum ImplicitNonFirstVariantIsZeroable {
+    A(::core::num::NonZeroU8) = 1,
+    B = -1,
+    C(bool),
+}
+
+// MSRV does not support data-carrying enum variants with explicit discriminants
+#[cfg(not(__ZEROCOPY_TOOLCHAIN = "msrv"))]
+util_assert_impl_all!(ImplicitNonFirstVariantIsZeroable: imp::FromZeros);
```
joshlf, line 101: Maybe put these in a module so you can factor out the `#[cfg(not(__ZEROCOPY_TOOLCHAIN = "msrv"))]`?

joshlf (changes requested):
High-level

zerocopy-derive/tests/enum_from_zeros.rs
djkoloski: Done

zerocopy-derive/src/repr.rs
djkoloski: Done, and changed the return type to non-optional

zerocopy-derive/src/lib.rs
djkoloski: Yes, we're going to replace this with #1303

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,365 @@
...
+            // Unit variants don't need any further validation beyond checking
+            // the tag.
+            quote! {
+                #tag_ident => true
+            }
+        } else {
+            quote! {
+                #tag_ident => {
+                    // SAFETY:
+                    // - The variant struct has the same size as the enum
```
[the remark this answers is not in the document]
djkoloski: Reworded safety comments

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,365 @@
...
+    let (_, ty_generics, _) = generics.split_for_impl();
+
+    let fields = data.variants.iter().filter_map(|variant| {
+        if matches!(variant.fields, Fields::Unit) {
+            return None;
+        }
+
+        // Field names are prefixed with `field_` to prevent name collision with
+        // the `__nonempty` field.
+        let field_name = Ident::new(&format!("field_{}", &variant.ident), variant.ident.span());
```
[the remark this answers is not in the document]
djkoloski: Done

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,365 @@
...
+    }
+}
+
+fn generate_variants_union(generics: &Generics, data: &DataEnum) -> TokenStream {
+    let (_, ty_generics, _) = generics.split_for_impl();
+
+    let fields = data.variants.iter().filter_map(|variant| {
+        if matches!(variant.fields, Fields::Unit) {
+            return None;
+        }
```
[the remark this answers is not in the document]
djkoloski: Done

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,365 @@
...
+
+fn variant_struct_ident(variant_ident: &Ident) -> Ident {
+    Ident::new(&format!("___ZerocopyVariantStruct_{}", variant_ident), variant_ident.span())
+}
+
+/// Generates variant structs for the given enum variant.
+///
+/// These are structs associated with each variant of an enum. They are
+/// `repr(C)` tuple struct with the same fields as the variant after an
+/// `InnerTag`.
```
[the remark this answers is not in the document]
djkoloski: Done.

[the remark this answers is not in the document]
djkoloski: Added a comment

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,365 @@
...
+}
+
+fn variant_struct_ident(variant_ident: &Ident) -> Ident {
+    Ident::new(&format!("___ZerocopyVariantStruct_{}", variant_ident), variant_ident.span())
+}
+
+/// Generates variant structs for the given enum variant.
+///
+/// These are structs associated with each variant of an enum. They are
+/// `repr(C)` tuple struct with the same fields as the variant after an
```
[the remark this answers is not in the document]
djkoloski: Done

zerocopy-derive/src/enums.rs
djkoloski: Done

zerocopy-derive/src/lib.rs
djkoloski: `quote` will handle the span information correctly in general, I think this should keep the hygiene information correct.

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,365 @@
...
+/// - Patterns do not currently support const expressions
+fn generate_tag_consts(data: &DataEnum) -> TokenStream {
+    let tags = data.variants.iter().map(|v| {
+        let variant_ident = &v.ident;
+        let tag_ident = tag_ident(variant_ident);
+
+        quote! {
+            #[allow(non_upper_case_globals)]
+            const #tag_ident: ___ZerocopyTagPrimitive =
+                ___ZerocopyTag::#variant_ident as ___ZerocopyTagPrimitive;
```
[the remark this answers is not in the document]
djkoloski: Done

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,365 @@
...
+
+/// Generates a constant for each variant of the enum. When we match on the
+/// enum's tag, each arm matches one of these constants. We have to use
+/// constants here because:
+///
+/// - The type that we're matching on is not the type of the tag, it's just a
+///   bunch of bytes.
+/// - We can't read the enum tag as an enum because the bytes may not represent
+///   a valid variant.
+/// - Patterns do not currently support const expressions
```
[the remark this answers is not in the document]
djkoloski: Done

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,365 @@
...
+fn tag_ident(variant_ident: &Ident) -> Ident {
+    Ident::new(&format!("___ZEROCOPY_TAG_{}", variant_ident), variant_ident.span())
+}
+
+/// Generates a constant for each variant of the enum. When we match on the
+/// enum's tag, each arm matches one of these constants. We have to use
+/// constants here because:
+///
+/// - The type that we're matching on is not the type of the tag, it's just a
+///   bunch of bytes.
```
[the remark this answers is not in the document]
djkoloski: Updated this comment

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,365 @@
...
+            #(#variants),*
+        }
+    }
+}
+
+fn tag_ident(variant_ident: &Ident) -> Ident {
+    Ident::new(&format!("___ZEROCOPY_TAG_{}", variant_ident), variant_ident.span())
+}
+
+/// Generates a constant for each variant of the enum. When we match on the
```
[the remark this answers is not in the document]
djkoloski: Done

zerocopy-derive/src/enums.rs
```diff
@@ -0,0 +1,365 @@
...
+// those terms.
+
+use ::proc_macro2::TokenStream;
+use ::quote::quote;
+use ::syn::{DataEnum, Fields, Generics, Ident};
+
+use crate::{EnumRepr, Trait};
+
+/// Returns the repr for the tag enum, given the collection of reprs on the
+/// enum. This function returns:
```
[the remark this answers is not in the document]
djkoloski: Done

src/macro_util.rs
```diff
@@ -291,11 +317,38 @@ macro_rules! struct_has_padding {
...
         false $(|| ::zerocopy::macro_util::core_reexport::mem::size_of::<$t>() != ::zerocopy::macro_util::core_reexport::mem::size_of::<$ts>())*
     };
 }
 
+/// Does the enum type `$t` have padding?
+///
+/// `$disc` is the type of the discriminant for the enum, and `$ts` is a list of
+/// fields in each square-bracket-deiminated variant. `$t` must be an enum, or
+/// else `enum_has_padding!`'s result may be meaningless. An enum has padding if
+/// any of its variant structs contain padding, and so all of the variants of an
```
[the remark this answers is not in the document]
djkoloski: Linked some sections in the reference

src/macro_util.rs
```diff
@@ -291,11 +317,38 @@ macro_rules! struct_has_padding {
...
-    ($t:ty, $($ts:ty),*) => {
+    ($t:ty, [$($ts:ty),*]) => {
         false $(|| ::zerocopy::macro_util::core_reexport::mem::size_of::<$t>() != ::zerocopy::macro_util::core_reexport::mem::size_of::<$ts>())*
     };
 }
 
+/// Does the enum type `$t` have padding?
+///
+/// `$disc` is the type of the discriminant for the enum, and `$ts` is a list of
+/// fields in each square-bracket-deiminated variant. `$t` must be an enum, or
```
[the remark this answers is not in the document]
djkoloski: Done

djkoloski (commented):
Addressed feedback

src/impls.rs
```diff
@@ -1731,7 +1731,15 @@ mod tests {
...
-        assert_impls!([NotZerocopy]: KnownLayout, !Immutable, !TryFromBytes, !FromZeros, !FromBytes, !IntoBytes, !Unaligned);
+        assert_impls!(
+            [NotZerocopy]: KnownLayout,
+            !Immutable,
+            !TryFromBytes,
+            !FromZeros,
+            !FromBytes,
+            !IntoBytes,
+            !Unaligned
+        );
```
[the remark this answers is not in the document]
joshlf: Done.

zerocopy-derive/src/enum.rs
joshlf, line 1: TODO: Confirm that `UnsafeCell` issues are sound

joshlf (changes requested):

zerocopy-derive/tests/ui-nightly/enum.rs
```diff
@@ -72,9 +72,17 @@ enum TryFromBytes1 {
...
+    B(u8),
+}
+
+struct NotTryFromBytes;
+
+#[derive(TryFromBytes)]
+#[repr(u8)]
+enum TryFromBytes3 {
+    A(NotTryFromBytes),
 }
```
joshlf, line 86: It looks like this isn't generating any error, presumably because the entire compilation is failing during an earlier compiler pass. You will likely need to put this in a separate file.

zerocopy-derive/tests/ui-nightly/enum.rs
```diff
@@ -114,6 +122,21 @@ enum FromZeros5 {
     B,
 }
 
+struct NotFromZeros;
+
+#[derive(FromZeros)]
+#[repr(u8)]
+enum FromZeros6 {
+    A(NotFromZeros),
+}
```
joshlf, line 131: Same as above - needs to be in a separate file.

zerocopy-derive/tests/ui-nightly/enum.rs
```diff
@@ -114,6 +122,21 @@ enum FromZeros5 {
...
+
+#[derive(FromZeros)]
+#[repr(u8)]
+enum FromZeros7 {
+    A = 1,
+    B(NotFromZeros),
```
joshlf, line 137: Make this a type that does implement `FromZeros` to make it clear that we're testing the fact that there's no 0 discriminant?

zerocopy-derive/tests/ui-nightly/enum.rs
```diff
@@ -160,6 +183,267 @@ enum FromBytes7 {
     A,
 }
 
+#[derive(FromBytes)]
+#[repr(u8)]
+enum FooU8 {
```
joshlf, line 188: Same as above.

zerocopy-derive/tests/ui-nightly/enum.rs
```diff
@@ -241,3 +525,31 @@ enum Unaligned12 {
...
+//
+// IntoBytes errors
+//
+
+#[derive(IntoBytes)]
+#[repr(u8)]
+enum IntoBytes1 {
+    A,
+    B(u8),
+}
```
joshlf, line 538: Same above

zerocopy-derive/tests/ui-nightly/enum.rs
```diff
@@ -241,3 +525,31 @@ enum Unaligned12 {
...
+
+#[derive(IntoBytes)]
+#[repr(C, align(4))]
+struct Align4IntoBytes(u32);
+
+#[derive(IntoBytes)]
+#[repr(u8)]
+enum IntoBytes2 {
+    A(Align4IntoBytes),
+}
```
joshlf, line 548: Same as above

zerocopy-derive/tests/ui-nightly/enum.rs
```diff
@@ -241,3 +525,31 @@ enum Unaligned12 {
...
+
+#[derive(IntoBytes)]
+#[repr(u32)]
+enum IntoBytes3 {
+    A(u32),
+    B(u16),
+}
```
joshlf, line 555: Same as above

joshlf (changes requested):

zerocopy-derive/tests/ui-nightly/enum.rs
djkoloski: Done

zerocopy-derive/tests/ui-nightly/enum.rs
djkoloski: Done

zerocopy-derive/tests/ui-nightly/enum.rs
djkoloski: Done

zerocopy-derive/tests/ui-nightly/enum.rs
djkoloski: Done

zerocopy-derive/tests/ui-nightly/enum.rs
djkoloski: Done

zerocopy-derive/tests/ui-nightly/enum.rs
djkoloski: Done

joshlf (approved):
We're going to follow up with #1634, which will block releasing 0.8.

PR #883, opened 2024-02-15
derive: generalize `require_self_sized`

jswrenn:
Generalize `require_self_sized` to require other `Trait`s.

### DIFF
--- a/zerocopy-derive/src/lib.rs
+++ b/zerocopy-derive/src/lib.rs
@@ -31,12 +31,14 @@
 mod ext;
 mod repr;
 
+use quote::quote_spanned;
+
 use {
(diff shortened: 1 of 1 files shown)


tests/ui-msrv/transmute-mut-dst-not-intobytes.stderr
```diff
@@ -1,8 +1,8 @@
-error[E0277]: the trait bound `Dst: IntoBytes` is not satisfied
+error[E0277]: the trait bound `Dst: AsBytes` is not satisfied
```
jswrenn, line 1: I'm very, very confused why this happened.

joshlf (approved):

Issue #882, opened 2024-02-15
Ensure zerocopy-derive doesn't emit code that assumes a prelude


joshlf:
Add `#![no_implicit_prelude]` to every zerocopy-derive test in the `zerocopy-derive/tests` directory (other than `trybuild.rs`), and  fix whatever breaks. If possible, come up with a more robust strategy to ensure we can't emit code that has this hygiene issue.

joshlf:
Closed in https://github.com/google/zerocopy/pull/886

Issue #11, opened 2022-09-19
Make derive macros hygienic


joshlf:
*Migrated from https://fxbug.dev/84476*

*Originally reported by @Kestrer*

Zerocopy's derive macros rely on `Sized` and `zerocopy` being in scope for them to function, but that is not always the case, for example when `zerocopy`'s derive macros are being re-exported by another crate, or in `#![no_implicit_prelude]` environments.

This is not too limiting, I can easily write:

```rust
use ::core::marker::Sized;
use ::some_other_crate::zerocopy;
```

Before the derive macro, but ideally it would just work.

To fix this, Zerocopy should use `::core::marker::Sized` instead of just `Sized`, and it should support an attribute that gives that path of the zerocopy crate [like Serde does](https://serde.rs/container-attrs.html#crate):

```rust
#[derive(::some_other_crate::zerocopy::AsBytes)]
#[zerocopy = ::some_other_crate::zerocopy]
pub struct S;
```

*Comment by @tamird*

This is needed to remove unused crate dependencies from FIDL crates which *sometimes* emit zerocopy derives, but *always* include a dependency edge to zerocopy.

Something like serde's [crate attribute](https://serde.rs/container-attrs.html#crate) would allow us to avoid this.

*Comment by @tamird*

Actually, this is not directly needed; the workaround described here works, and is implemented in [https://fuchsia-review.googlesource.com/c/fuchsia/+/702383](https://fuchsia-review.googlesource.com/c/fuchsia/+/702383).

joshlf:
@djkoloski Flagging this in case you'd be interested in it - specifically the `#[zerocopy(crate = "...")]` annotation.

Kestrer:
> do you know if it's sufficient to instead emit `::zerocopy`?

This should be the default option, since in most cases it works. However, it fails in the following cases:
1. The user renames `zerocopy` to something else, e.g. if they do `foobar = { package = "zerocopy", version = "0.6.3" }` in their `Cargo.toml`
2. The user does not depend on `zerocopy` itself, but rather imports it through some other crate — e.g. `foobar` has `pub use zerocopy;` in its crate root, and the user’s crate tries `#[derive(foobar::zerocopy::AsBytes)]`.

Supporting these cases is the motivation for adding a `#[zerocopy(crate = path::to::zerocopy)]` annotation.

> Another approach that occurred to me - which I suspect also won't work for some reason - is to emit `extern crate zerocopy` in the generated code.

[...]

PR #886, opened 2024-02-16
[derive] Disable prelude when testing; fix bugs

joshlf:
This has the effect of ensuring that derive-emitted code will fail to compile if it spuriously relies on certain identifiers being in scope - namely, identifiers which are part of the prelude.

Disabling the prelude surfaced a few bugs which are also fixed in this commit.

Makes progress on #11


jswrenn (approved):

joshlf:
Closes https://github.com/google/zerocopy/issues/882

Issue #873, opened 2024-02-13
`Ptr` type should reason about `UnsafeCell` overlap


joshlf:
Two requirements to add to the list of internal invariants:
- During the lifetime `'a`, no code will access* this memory region treating `UnsafeCell`s as existing at different ranges than they exist in `T`
- During the lifetime `'a`, no reference will exist to this memory which treats `UnsafeCell`s as existing at different ranges than they exist in `T`

* What does "access" mean? Should this just say "loads and stores"?

joshlf:
Closed in https://github.com/google/zerocopy/pull/890

PR #890, opened 2024-02-16
[`TryFromBytes`] Permit `UnsafeCells`

jswrenn:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1210,12 +1210,6 @@ pub unsafe trait NoCell {
 /// including changes that only affect soundness, which may cause code which
 /// uses those items to silently become unsound.
 ///
-/// ## `UnsafeCell`
-///
-/// `TryFromBytes` may not be implemented for any type which contains an
-/// [`UnsafeCell`]. This is a temporary limitation and will be lifted once
-/// [`try_from_ref`] has a [`NoCell`] bound.
-///
 /// [undefined behavior]: https://raphlinus.github.io/programming/rust/2018/08/17/undefined-behavior.html
 /// [github-repo]: https://github.com/google/zerocopy
 /// [`try_from_ref`]: TryFromBytes::try_from_ref
@@ -1264,7 +1258,7 @@ pub unsafe trait TryFromBytes {
     #[doc(hidden)] // TODO(#5): Finalize name before remove this attribute.
     fn try_from_ref(bytes: &[u8]) -> Option<&Self>
     where
-        Self: KnownLayout,
+        Self: KnownLayout + NoCell,
     {
         let candidate = Ptr::from_ref(bytes).try_cast_into_no_leftover::<Self>()?;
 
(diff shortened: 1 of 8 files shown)


joshlf (changes requested):
Need to make sure that all call sites have new safety proofs that mention `UnsafeCell` for:
- `Ptr::project`
- `Ptr::cast_unsized`

src/lib.rs
```diff
@@ -1293,7 +1287,7 @@ pub unsafe trait TryFromBytes {
     #[doc(hidden)] // TODO(#5): Finalize name before remove this attribute.
     fn try_read_from(bytes: &[u8]) -> Option<Self>
     where
-        Self: Sized,
+        Self: Sized + NoCell,
```
jswrenn, line 1290: We should have a TODO here, since we cannot claim to support `UnsafeCell`s in any meaningful sense until this bound is removed.

src/lib.rs
```diff
@@ -3111,7 +3105,8 @@ safety_comment! {
     ///   closure:
     ///   - Given `t: *mut bool` and `let r = *mut u8`, `r` refers to an object
     ///     of the same size as that referred to by `t`. This is true because
-    ///     `bool` and `u8` have the same size (1 byte) [1].
+    ///     `bool` and `u8` have the same size (1 byte) [1]. Neither `r` nor `t`
+    ///     contain `UnsafeCell`s because neither `bool` nor `u8` do.
```
jswrenn, line 3109: Not to nit-pick, but if we're nitpicking, [citation needed].

src/lib.rs
```diff
@@ -3162,7 +3157,8 @@ safety_comment! {
     ///   closure:
     ///   - Given `t: *mut char` and `let r = *mut u32`, `r` refers to an object
     ///     of the same size as that referred to by `t`. This is true because
-    ///     `char` and `u32` have the same size [1].
+    ///     `char` and `u32` have the same size [1]. Neither `r` nor `t` contain
+    ///     `UnsafeCell`s because neither `char` nor `u32` do.
```
jswrenn, line 3161: [citation needed]

src/lib.rs
```diff
@@ -3358,6 +3361,22 @@ safety_comment! {
...
+    /// [1] https://github.com/rust-lang/unsafe-code-guidelines/issues/492
+    ///
+    /// TODO(#896): Write a more complete safety proof before the next stable
+    /// release.
+    #[cfg(feature = "alloc")]
+    unsafe_impl!(
+        #[cfg_attr(doc_cfg, doc(cfg(feature = "alloc")))]
+        T: Sized => NoCell for Box<T>
+    );
+}
```
jswrenn, line 3378: Generally speaking, we need to justify that *none* of the primitive types are `!Freeze`, not just `Box<T>`.

joshlf: I agree, but it's not clear that `Box<T>` *is* actually a primitive type. Obviously it is in practice, but it's not clear that it's guaranteed as such. Also, I'm not sure this is the right location for a TODO comment about documenting other primitive types.

src/lib.rs
joshlf: done.

src/lib.rs

src/lib.rs

joshlf (approved):

joshlf:
Closes #873

PR #891, opened 2024-02-16
Convert shell scripts to Rust

djkoloski:
I primarily dev on Windows, and I'd prefer to not have to use WSL to work on zerocopy.


.github/workflows/ci.yml
```diff
@@ -109,8 +109,8 @@ jobs:
         #
         # ...to:
         #
-        #   toolchain: $ {{ ./cargo.sh --version matrix.toolchain }} # hypothetical syntax
-        ZC_TOOLCHAIN="$(./cargo.sh --version ${{ matrix.toolchain }})"
+        #   toolchain: $ {{ cargo run -p cargos --version matrix.toolchain }} # hypothetical syntax
```
joshlf, line 112: I'm not familiar with Windows enough to know: Is there a way that we could keep a script that would be compatible with both Windows and POSIX, and would just be a thin wrapper around `cargo run -p cargos`? The latter is more verbose, which isn't great for the command we run the most frequently during development.

tools/cargos/src/main.rs
```diff
@@ -0,0 +1,253 @@
...
+    //
+    // In CI, make sure to use the default stable toolchain. If we're testing on
+    // our MSRV, then we also have our MSRV toolchain installed. As of this
+    // writing, our MSRV is low enough that the correspoding Rust toolchain's
+    // Cargo doesn't know about the `rust-version` field, and so if we were to
+    // use Cargo with that toolchain, `pkg-meta` would return `null` when asked
+    // to retrieve the `rust-version` field. This also requires `RUSTFLAGS=''`
+    // to override any unstable `RUSTFLAGS` set by the caller.
+    let output = Command::new("rustup")
+        .args(["run", "stable", "cargo", "metadata", "--format-version", "1"])
```
joshlf, line 97: Why not `cargo +stable` like we did before? Is there an advantage to calling via `rustup`?

tools/cargos/src/main.rs
```diff
@@ -0,0 +1,253 @@
...
+// include any toolchain which is listed in the `package.metadata.build-rs`
+// Cargo.toml section.
+//
+// A common task that is especially annoying to perform by hand is to update
+// trybuild's stderr files. Using this script:
+//
+//   TRYBUILD=overwrite ./cargos +all test --workspace
+
+use std::{
+    env::{args, vars},
```
joshlf, line 28: Uber nit: Can you just `use env` and then invoke these as `env::args()` and `env::vars()` to be consistent with our existing style?

tools/generate-readme/src/main.rs
```diff
@@ -0,0 +1,52 @@
...
+
+fn main() {
+    let output = Command::new("cargo").args(["readme", "--no-license"]).output().unwrap();
+
+    let readme = String::from_utf8(output.stdout).unwrap();
+
+    // This uses the `cargo readme` tool, which you can install via `cargo install
+    // cargo-readme --version 3.2.0`.
+    //
+    // The `sed` command is used to strip code links like:
```
joshlf, line 41:
```suggestion
    // The regex is used to strip code links like:
```

joshlf (changes requested):

.github/workflows/ci.yml
djkoloski: We could just publish this as a separate package and `cargo install` it. Another option would be to leave the `cargo.sh` file in place and make it a thin wrapper over `cargo run -p cargos`.

tools/cargos/src/main.rs
djkoloski: I think there's some shell trickery going on here. When I run `cargo +stable metadata` in powershell, it acts as expected. However, if I `Command::new("cargo").args(["+stable", "metadata"])` in Rust, I get:

```
stdout: 
stderr: error: no such command: `+stable`

        Cargo does not handle `+toolchain` directives.
        Did you mean to invoke `cargo` through `rustup` instead?
```

I suspect there's some kind of aliasing going on here where "cargo" as a command resolves to a specific toolchain's `cargo.exe` binary, but from the shell it resolves to rustup's "cargo" dispatcher binary.

tools/generate-readme/src/main.rs
djkoloski: Done

tools/cargos/src/main.rs
djkoloski: Done

tools/cargos/src/main.rs
joshlf: Gotcha, that makes sense.

.github/workflows/ci.yml
```diff
@@ -109,8 +109,8 @@ jobs:
         #
         # ...to:
         #
-        #   toolchain: $ {{ ./cargo.sh --version matrix.toolchain }} # hypothetical syntax
-        ZC_TOOLCHAIN="$(./cargo.sh --version ${{ matrix.toolchain }})"
+        #   toolchain: $ {{ ./cargos.sh --version matrix.toolchain }} # hypothetical syntax
```
joshlf, line 112: Is there a reason not to leave this named `cargo.sh`? (I'll be honest I'm not sure what the etymology of `cargos` is as opposed to `cargo`)

joshlf (changes requested):

joshlf:
It looks like the invocation of `cargo` itself (e.g. the one inside `cargos.sh`) is using the default toolchain installed in the GitHub Action. You probably want to do `cargo +stable` like we do in `cargo.sh` on `main`.

tools/cargos/src/main.rs
```diff
@@ -0,0 +1,254 @@
...
+            'n' | 'N' => process::exit(1),
+            _ => (),
+        }
+    }
+
+    let version = versions.get(name)?;
+    Command::new("rustup")
+        .args(["toolchain", "install", &version, "-c", "rust-src"])
+        .status()
+        .unwrap();
```
joshlf, line 160: Could be worth factoring this out, but it can't be `fn rustup` (at least as it's written now) because that doesn't wire up stdin/stdout/stderr to the parent process's stdin/stdout/stderr.

tools/cargos/src/main.rs
```diff
@@ -0,0 +1,254 @@
...
+                    .filter_map(|(k, v)| if k == "RUSTFLAGS" { Some(v) } else { None })
+                    .next()
+                    .unwrap_or_default();
+
+                Command::new("rustup")
+                    .args(["run", &version, "cargo"])
+                    .args(args)
+                    .env("RUSTFLAGS", format!("{}{}", get_rustflags(name), rustflags))
+                    .status()
+                    .unwrap();
```
joshlf, line 229: Could be worth factoring this out, but it can't be `fn rustup` (at least as it's written now) because that doesn't wire up stdin/stdout/stderr to the parent process's stdin/stdout/stderr.

joshlf (changes requested):
I pushed a new commit that should hopefully fix (some of) the CI errors. Feel free to delete it, squash it, etc.

Since these are in their own workspace now, can you update `ci/check_fmt.sh` to format that workspace as well?

tools/cargos/src/main.rs
djkoloski: Done

tools/cargos/src/main.rs
djkoloski: Done

djkoloski:
`ci/check_fmt.sh` picks up the .rs files in the tools/ workspace, so no changes required there.

.github/workflows/ci.yml
djkoloski: It is an unfortunately mundane one. I originally kept it named `cargo.sh` and made `cargo.bat` to match. As it turns out though, when `cargos` runs a `cargo` command, Windows will run `cargo.bat` and not the installed `cargo` binary. So it effectively causes `cargos` to recursively call itself when it intends to invoke `cargo`. The easy fix was to use a different name, and I figured if I had to call it `cargos` on Windows then I should probably just make it uniform.

joshlf: Ah, that makes sense. My preference would be to keep `cargo.sh` named as-is, and rename the `.bat` file. Maybe `win-cargo.bat`? I'll leave it up to you what you want to name it.

cargo.sh
```diff
@@ -1,135 +1,12 @@
...
-        "rustup toolchain install $TOOLCHAIN -c rust-src"
-    } || exit 1
-
-    RUSTFLAGS="$(get-rustflags ${1:1}) $RUSTFLAGS" cargo "+$TOOLCHAIN" ${@:2}
-    ;;
-  *)
-    print-usage-and-exit
-    ;;
-esac
+# Build `cargos` without any RUSTFLAGS set in the environment
```
joshlf, line 9: Can the tool itself now be called `cargo` instead of `cargos`? Or is there a name conflict in doing that?

joshlf (changes requested):
Last comment that I'm aware of - once this is changed, we should be ready to merge.

Can you also squash all of these commits together? Otherwise, the entire set of commits shows up in the commit message that GitHub generates.

cargo.sh
djkoloski: Even if there's not a name conflict, it sounds like naming it `cargo` is asking for trouble. Would something like `cargo-multiplex`/`cargo-multi`/`multicargo`/`mc` be acceptable?

joshlf: Hmm yeah good point. How about something like:
- `cargo-zerocopy`
- `cargo-zerocopy-shim`

Given that this is always called from a script, I'm not too worried about a verbose name.

djkoloski: I went with `cargo-zerocopy`

joshlf (approved):

PR #892, opened 2024-02-17
Add TryFromBytes::try_from_mut

joshlf:
Makes progress on #5

### DIFF
--- a/src/pointer/ptr.rs
+++ b/src/pointer/ptr.rs
@@ -419,6 +419,38 @@ mod _conversions {
         }
     }
 
+    /// `&'a mut T` → `Ptr<'a, T>`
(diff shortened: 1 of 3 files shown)


src/pointer/ptr.rs
```diff
@@ -454,6 +487,59 @@ mod _conversions {
         }
     }
 
+    impl<'a, T, I> Ptr<'a, T, I>
+    where
+        T: 'a + ?Sized,
+        I: Invariants,
+        I::Aliasing: invariant::at_least::Shared,
+    {
+        pub(crate) fn reborrow(&'a self) -> Self {
```
joshlf, line 496: This is probably unsound. In particular, if `I::Aliasing = Exclusive`, then `Self` is an exclusive reference that can be passed by-value to methods such as `as_mut`.

jswrenn (approved):

jswrenn (approved):

PR #898, opened 2024-02-17
Remove unsound `Ptr::forget_valid`

jswrenn:
With the addition of `Initialized`, validity is no longer totally ordered.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8059,7 +8059,13 @@ mod tests {
         macro_rules! assert_impls {
             ($ty:ty: TryFromBytes) => {
                 <$ty as TryFromBytesTestable>::with_passing_test_cases(|val| {
-                    let c = Ptr::from_ref(val).forget_valid().forget_aligned();
+                    let c = Ptr::from_ref(val).forget_aligned();
+                    let c = c.forget_aligned();
+                    // SAFETY:
+                    // TODO(#899): This is unsound. `$ty` is not necessarily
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -8059,7 +8059,12 @@ mod tests {
         macro_rules! assert_impls {
             ($ty:ty: TryFromBytes) => {
                 <$ty as TryFromBytesTestable>::with_passing_test_cases(|val| {
-                    let c = Ptr::from_ref(val).forget_valid().forget_aligned();
+                    let c = Ptr::from_ref(val).forget_aligned();
+                    let c = c.forget_aligned();
+                    // SAFETY: This is unsound. `$ty` is not necessarily
+                    // `IntoBytes`, but that's the corner we've backed ourselves
+                    // into by using `Ptr::from_ref`.
+                    let c = unsafe { c.assume_initialized() };
```
jswrenn, line 8067: @joshlf, can you work your macro magic to maneuver us from this meddlesome mire?

joshlf: Since this doesn't actually change the behavior, I think the best thing to do is to merge this now - it just moves us in the good direction of at least more clearly calling out the soundness hole - and follow up to fix it afterwards: https://github.com/google/zerocopy/issues/899

joshlf (dismissed):

joshlf (approved):

src/lib.rs

joshlf (approved):

Issue #192, opened 2023-06-16
Validate `cargo package` contents in CI


joshlf:
Add a CI test that runs `cargo package` or `cargo publish --dry-run` or similar, and validates certain aspects of the resulting artifacts. This should include:
- Validate that the `LICENSE` file was included, and that its contents are correct (especially relevant since zerocopy-derive's license file is included implicitly as of https://github.com/google/zerocopy/pull/191)

joshlf:
*Consolidating #105 into this*

---

Historically, I've run into issues while doing `cargo package` or `cargo publish` that I didn't run into during normal development. We should think through this more and figure out if there's anything we can do to exercise these potential pitfalls in CI.

PR #906, opened 2024-02-18
Add packing checks in CI

dorryspears:
Makes Progress On #192

I am not sure how to check the Licenses other than checking the content of the license itself by hard coding it in the CI.
This will check if they are at least present in the package.


.github/workflows/ci.yml
```diff
@@ -431,6 +431,25 @@ jobs:
       - name: Run dependency check
         run: ./ci/check_job_dependencies.sh
 
+  test-packing:
+    runs-on: ubuntu-latest
+    name: Checks Packing
```
jswrenn, line 436: See above.
```suggestion
    name: Check packaging 
```

.github/workflows/ci.yml
```diff
@@ -431,6 +431,25 @@ jobs:
       - name: Run dependency check
         run: ./ci/check_job_dependencies.sh
 
+  test-packing:
```
jswrenn, line 434: When we use the word "packing" in the case of zerocopy, we're usually talking about uninitialized bytes added to layouts to satisfy alignment requirements. I think "packaging" is less prone to be confused with that. We also start our job names with `check-`.
```suggestion
  check-packaging:
```

.github/workflows/ci.yml
```diff
@@ -441,7 +460,7 @@ jobs:
       # https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks
       if: failure()
       runs-on: ubuntu-latest
-      needs: [build_test, kani, check_fmt, check_readme, check_msrv, check_versions, generate_cache, check-all-toolchains-tested, check-job-dependencies]
+      needs: [build_test, kani, check_fmt, check_readme, check_msrv, check_versions, generate_cache, check-all-toolchains-tested, check-job-dependencies, test-packing]
```
jswrenn, line 463:
```suggestion
      needs: [build_test, kani, check_fmt, check_readme, check_msrv, check_versions, generate_cache, check-all-toolchains-tested, check-job-dependencies, check-packaging]
```

.github/workflows/ci.yml
```diff
@@ -431,6 +431,25 @@ jobs:
       - name: Run dependency check
         run: ./ci/check_job_dependencies.sh
 
+  test-packing:
+    runs-on: ubuntu-latest
+    name: Checks Packing 
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - name: Testing Packing
```
jswrenn, line 439: A more descriptive name:
```suggestion
      - name: Check that LICENSE files are packaged
```

.github/workflows/ci.yml
```diff
@@ -431,6 +431,25 @@ jobs:
...
+        run: |
+          set -eo pipefail
+          cargo package --target-dir test-pack
+
+          # Check that Licenses exist
+          cd test-pack/package
+          cd "$(ls -d */ | head -n 1)"
```
jswrenn, line 446: What's happening here?

jswrenn (changes requested):

.github/workflows/ci.yml
dorryspears: This goes to the package folder on line 445

The pack command generates the source and the actual crate file in the test-pack/package folder with the version 
name appended at the end.  Since the version name will change, it needs another way to reference the package source file. This is where line 446 comes in which grabs the first directory in the current working directory and cd's
into it.

From there we check if the Licenses exist.

dorryspears: Will make those suggested changes now

PR #912, opened 2024-02-20
Implement TryFromBytes for UnsafeCell

joshlf:
TODO:
- Commit message body
- Get tests working
- Write tests that exercise `<Wrapper<UnsafeCell<T>> as TryFromBytes>::is_bit_valid` for all `Wrapper` types. This ensures that `is_bit_valid` never switches from an exclusive to a shared pointer, which would panic once it got to `UnsafeCell::is_bit_valid`.
- Try to see if we can statically prevent calling `is_bit_valid` with a shared pointer for `UnsafeCell`.

Makes progress on #5


src/lib.rs
```diff
@@ -3688,6 +3706,82 @@ safety_comment! {
...
+    #[allow(clippy::missing_inline_in_public_items)]
+    fn only_derive_is_allowed_to_implement_this_trait()
+    where
+        Self: Sized,
+    {
+    }
+
+    #[inline]
+    fn is_bit_valid<A: invariant::at_least::Shared>(candidate: Maybe<'_, Self, A>) -> bool {
+        // TODO: Write a comment here.
```
joshlf, line 3723: TODO

src/lib.rs
```diff
@@ -3688,6 +3706,82 @@ safety_comment! {
     unsafe_impl!(T: ?Sized + Unaligned => Unaligned for UnsafeCell<T>);
     assert_unaligned!(UnsafeCell<()>, UnsafeCell<u8>);
 }
+
+// SAFETY: See safety comment in `is_bit_valid` impl.
+//
+// TODO: Try to add `T: ?Sized` bound.
```
joshlf, line 3712: TODO: This should be doable, but we'll need to replace `UnsafeCell::get_mut` with a manual implementation.

joshlf: On second thought, we can just do this in a follow-up. It's not a 0.8 blocker.

src/lib.rs
```diff
@@ -1269,6 +1274,10 @@ pub unsafe trait TryFromBytes {
         // This call may panic. If that happens, it doesn't cause any soundness
         // issues, as we have not generated any invalid state which we need to
         // fix before returning.
+        //
+        // Note that one panic condition is calling `try_into_valid` (and thus
+        // `is_bit_valid`) with a shared pointer when `Self: !NoCell`. Since
+        // `Self: NoCell`, this panic condition will not happen.
```
jswrenn, line 1280: This is a post-monomorphization error, not a panic, right?

jswrenn (approved):

Issue #148, opened 2022-12-03
Ord instances for integer wrappers e.g. U64


jjl:
It would be nice if they could be compared directly rather than having to `.get()` them before comparison. That would allow deriving Ord over structs using them that are in disk format.

joshlf:
If memory serves, I tried this when I initially introduced those wrappers, and it caused a lot of type inference issues. I'll take a stab at it again when I get a chance to confirm that that was the reason.

joshlf:
Closed in https://github.com/google/zerocopy/issues/1120.

PR #915, opened 2024-02-20
Implement `PartialOrd` and `Ord` where possible

00xc:
All numeric types can implement `PartialOrd` and `Ord`, except floats, which can only implement `PartialOrd`.

We cannot simply implement `PartialOrd` via `@all_types` because then clippy will complain about `PartialOrd` not being implemented in terms of `Ord` (see [`clippy::non_canonical_partial_ord_impl`](https://rust-lang.github.io/rust-clippy/master/index.html#/non_canonical_partial_ord_impl) lint) for non-floats.

### DIFF
--- a/src/byteorder.rs
+++ b/src/byteorder.rs
@@ -192,6 +192,13 @@ macro_rules! impl_ops_traits {
     ($name:ident, $native:ident, "floating point number") => {
         impl_ops_traits!($name, $native, @all_types);
         impl_ops_traits!($name, $native, @signed_integer_floating_point);
+
+        impl<O: ByteOrder> PartialOrd for $name<O> {
+            #[inline(always)]
+            fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+                self.get().partial_cmp(&other.get())
+            }
+        }
     };
     ($name:ident, $native:ident, "unsigned integer") => {
         impl_ops_traits!($name, $native, @signed_unsigned_integer);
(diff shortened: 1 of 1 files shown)


joshlf:
Thanks for this! Before we merge, I'm going to need to take a bit of time to investigate https://github.com/google/zerocopy/issues/148#issuecomment-1336219104, so apologies if this sits idle for a bit before I get a chance to do that.

Out of curiosity, what use case prompted you to put up this PR? We don't have much visibility into how zerocopy is used beyond what we can see publicly in places like GitHub.

00xc:
> Thanks for this! Before we merge, I'm going to need to take a bit of time to investigate [#148 (comment)](https://github.com/google/zerocopy/issues/148#issuecomment-1336219104), so apologies if this sits idle for a bit before I get a chance to do that.

That is fine, but out of curiosity, what kind of issues were you encountering? This seems to build cleanly.

> Out of curiosity, what use case prompted you to put up this PR? We don't have much visibility into how zerocopy is used beyond what we can see publicly in places like GitHub.

My main use of the crate has been through [packit](/coconut-svsm/packit), but this particular change was motivated by a toy program I wrote to parse some information out of a vmlinux file.

joshlf:
> > Thanks for this! Before we merge, I'm going to need to take a bit of time to investigate [#148 (comment)](https://github.com/google/zerocopy/issues/148#issuecomment-1336219104), so apologies if this sits idle for a bit before I get a chance to do that.
> 
> 
> 
> That is fine, but out of curiosity, what kind of issues were you encountering? This seems to build cleanly.

The thing that makes it tricky is that it doesn't cause problems in zerocopy itself, but rather in downstream crates. At the time, I was working on a networking stack (Fuchsia's Netstack3), which is a ~100k line codebase with a lot of use of these byteorder primitives. I added impls similar to these, and a ton of type inference started failing. It wasn't obvious failures either - it was that sort of "what on earth is the compiler talking about??" type of failures. It's been years since I tried it, so I don't remember the details, but I'm wary of making life hard for downstream crates for that reason.

jswrenn:
@joshlf You've been importing zerocopy pre-releases into Fuchsia, right? After the next time you do so, what do you think about filing a draft CR that adds these impls to the vendored copy, and seeing if anything breaks in CI? If we don't get major breakage in Fuchsia, it's probably safe to merge this.

joshlf:
> @joshlf You've been importing zerocopy pre-releases into Fuchsia, right? After the next time you do so, what do you think about filing a draft CR that adds these impls to the vendored copy, and seeing if anything breaks in CI? If we don't get major breakage in Fuchsia, it's probably safe to merge this.

I ran this test, and nothing broke. That seems as good a test as any, especially given that we have customers who want this. Let's merge!

joshlf (approved):
I'm going to wait to merge until #948 has merged. Until that happens, our CI is unreliable.

Issue #929, opened 2024-02-23
Fix Kani failures


joshlf:
Kani recently released 0.47, which is causing CI failures [such as this one](https://github.com/google/zerocopy/actions/runs/8019289511/job/21906873024?pr=927).

## Steps

- [x] Pin Kani to 0.46 to unblock CI
  - #930
- [x] Fix the bug so that we can go back to the latest version
  - #939
- [x] Teach our auto-roller to roll the Kani version and keep it pinned
  - #931

joshlf:
Closed in #939

PR #931, opened 2024-02-23
[ci] Automatically roll pinned Kani version

joshlf:
Makes progress on #929


joshlf:
I'm manually testing this here: https://github.com/google/zerocopy/actions/runs/8020262121

PR #936, opened 2024-02-23
[byteorder] Test test_ops_impls under Kani

joshlf:
Makes progress on #929


joshlf:
Kani tests are taking forever, so this isn't viable. Closing in favor of another option like #939.

PR #943, opened 2024-02-23
[ci] Run `cargo fix` when rolling toolchains

### LINKED ISSUE #918: Auto-roller should run `cargo fix`
Recently, the auto-roller produced https://github.com/google/zerocopy/pull/914, which failed CI. I had to manually create https://github.com/google/zerocopy/pull/917, which fixed some new warnings introduced in the most recent nightly.

In this case, I confirmed that `./cargo.sh +nightly fix` didn't do the trick. However, there may be other cases in which it would fix errors automatically so that we wouldn't have to do it manually. There is little downside. We should have the roller run `./cargo.sh +nightly fix` as part of the roll.

joshlf:
Closes #918

### DIFF
--- a/.github/workflows/roll-pinned-toolchain-versions.yml
+++ b/.github/workflows/roll-pinned-toolchain-versions.yml
@@ -79,6 +79,13 @@ jobs:
             # Confirm that the update didn't bork `Cargo.toml`.
             validate-file "$REGEX" Cargo.toml
 
+            # Run `cargo fix` in case there are any warnings or errors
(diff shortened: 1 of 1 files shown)


joshlf:
Testing this manually: https://github.com/google/zerocopy/actions/runs/8023132238

joshlf:
Okay, looks like the roller is successfully getting past the `cargo fix` step. It's failing at a later step, but that's not relevant to this PR.

Issue #947, opened 2024-02-26
Stable and nightly toolchain roller failing


joshlf:
https://github.com/google/zerocopy/actions/runs/8030574585

jswrenn:
This is occurring because the Rust replacement to `./cargo.sh` is not setting its exit code appropriately depending on the outcome of its inner invocation. Its exit code is always `0`, and so Github Actions considers all of our CI steps to have unconditionally passed.

joshlf:
Fixed in a bunch of PRs culminating in #948.

PR #948, opened 2024-02-26
[cargo-zerocopy] Pass failure return codes through

djkoloski:
Addresses #947

### DIFF
--- a/cargo.sh
+++ b/cargo.sh
@@ -6,6 +6,8 @@
 # This file may not be copied, modified, or distributed except according to
 # those terms.
 
+set -eo pipefail
+
 # Build `cargo-zerocopy` without any RUSTFLAGS set in the environment
 env -u RUSTFLAGS cargo +stable build --manifest-path tools/Cargo.toml -p cargo-zerocopy -q
 # Thin wrapper around the `cargo-zerocopy` binary in `tools/cargo-zerocopy`
--- a/tools/cargo-zerocopy/src/main.rs
+++ b/tools/cargo-zerocopy/src/main.rs
@@ -27,7 +27,7 @@
 use std::{
     env, fmt,
     io::{self, Read as _},
-    process::{self, Command},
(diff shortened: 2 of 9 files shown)


cargo.sh
```diff
@@ -6,6 +6,8 @@
 # This file may not be copied, modified, or distributed except according to
 # those terms.
 
+set -e
```
joshlf, line 9: Just to be future-proof against using pipes in the future:

```suggestion
set -eo pipefail
```

tools/cargo-zerocopy/src/main.rs
```diff
@@ -53,6 +53,45 @@ impl fmt::Display for Error {
...
+        if let Ok(output) = self.output() {
+            if !output.status.success() {
+                eprintln!(
+                    "[cargo-zerocopy] failed while capturing output from command: {:?}",
+                    self
+                );
+                let stdout = std::str::from_utf8(&output.stdout).unwrap();
+                let stderr = std::str::from_utf8(&output.stderr).unwrap();
+                eprintln!("[cargo-zerocopy] stdout: {stdout}");
+                eprintln!("[cargo-zerocopy] stderr: {stderr}");
```
joshlf, line 72: Would it be possible to prefix `[cargo-zerocopy]` on each line (if stdout/stderr has newlines)?

joshlf:
Sorry for the pushes; those were accidental. I've force-pushed so that the state of the PR should be returned to what it was before.

tools/cargo-zerocopy/src/main.rs
joshlf: Nvm, I realized that this is only called for tiny utilities like `cargo msrv`, and it's not that important. Not worth the engineering effort IMO, especially given needing to juggle different platforms' end-of-line characters.

joshlf (approved):

PR #967, opened 2024-02-29
Add `KnownLayout`, remove `Sized` from some APIs

joshlf:
Makes progress on #29

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -263,12 +263,8 @@ pub use crate::wrappers::*;
 #[cfg_attr(doc_cfg, doc(cfg(feature = "derive")))]
 pub use zerocopy_derive::{TryFromBytes, Unaligned};
 
-// `pub use` separately here so that we can mark it `#[doc(hidden)]`.
-//
-// TODO(#29): Remove this or add a doc comment.
 #[cfg(any(feature = "derive", test))]
 #[cfg_attr(doc_cfg, doc(cfg(feature = "derive")))]
-#[doc(hidden)]
 pub use zerocopy_derive::KnownLayout;
 
 // `pub use` separately here so that we can mark it `#[doc(hidden)]`.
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1936,22 +1941,30 @@ pub unsafe trait FromBytes: FromZeros {
...
+    /// let packet = Packet::ref_from_prefix(bytes).unwrap();
+    ///
+    /// assert_eq!(packet.header.src_port, [0, 1]);
+    /// assert_eq!(packet.header.dst_port, [2, 3]);
+    /// assert_eq!(packet.header.length, [4, 5]);
+    /// assert_eq!(packet.header.checksum, [6, 7]);
+    /// assert_eq!(packet.body, [[8, 9], [10, 11], [12, 13]]);
     /// ```
     #[inline]
     fn ref_from_prefix(bytes: &[u8]) -> Option<&Self>
```
joshlf, line 1963: Need to figure out how to reword this paragraph from the doc comment:

```rust
    /// `ref_from_prefix` returns a reference to the first `size_of::<Self>()`
    /// bytes of `bytes`. If `bytes.len() < size_of::<Self>()` or `bytes` is not
    /// aligned to `align_of::<Self>()`, this returns `None`.
```

src/lib.rs
```diff
@@ -1985,9 +1998,9 @@ pub unsafe trait FromBytes: FromZeros {
     #[inline]
     fn ref_from_suffix(bytes: &[u8]) -> Option<&Self>
```
joshlf, line 1999: Need to figure out how to reword this paragraph from the doc comment:

```rust
    /// `ref_from_suffix` returns a reference to the last `size_of::<Self>()`
    /// bytes of `bytes`. If `bytes.len() < size_of::<Self>()` or the suffix of
    /// `bytes` is not aligned to `align_of::<Self>()`, this returns `None`.
```

src/lib.rs
joshlf: Added a TODO in #29 to block release on updating this doc comment, so we don't have to resolve this now.

src/lib.rs
joshlf: Added a TODO in #29 to block release on updating this doc comment, so we don't have to resolve this now.

jswrenn (approved):

PR #974, opened 2024-02-29
Document `TryFromBytes`

jswrenn:
Makes progress towards #5.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -25,6 +25,8 @@
 //!
 //! Zerocopy provides four core marker traits, each of which can be derived
 //! (e.g., `#[derive(FromZeros)]`):
+//! - [`TryFromBytes`] indicates that a type may safely be converted from certain
+//!   byte sequences (conditional on runtime checks)
 //! - [`FromZeros`] indicates that a sequence of zero bytes represents a valid
 //!   instance of a type
 //! - [`FromBytes`] indicates that a type may safely be converted from an
@@ -261,7 +263,7 @@ pub use crate::wrappers::*;
 
 #[cfg(any(feature = "derive", test))]
 #[cfg_attr(doc_cfg, doc(cfg(feature = "derive")))]
-pub use zerocopy_derive::{TryFromBytes, Unaligned};
+pub use zerocopy_derive::Unaligned;
 
 // `pub use` separately here so that we can mark it `#[doc(hidden)]`.
 //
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1146,18 +1148,87 @@ pub unsafe trait NoCell {
         Self: Sized;
 }
 
-/// Types whose validity can be checked at runtime, allowing them to be
-/// conditionally converted from byte slices.
+/// Implements [`TryFromBytes`].
 ///
-/// WARNING: Do not implement this trait yourself! Instead, use
-/// `#[derive(TryFromBytes)]`.
+/// This derive synthesizes the runtime checks required to check if a sequence
```
joshlf, line 1153:
```suggestion
/// This derive synthesizes the runtime checks required to check whether a sequence
```

src/lib.rs
```diff
@@ -1146,18 +1148,87 @@ pub unsafe trait NoCell {
...
+///
+/// [safety conditions]: trait@TryFromBytes#safety
+#[cfg(any(feature = "derive", test))]
+#[cfg_attr(doc_cfg, doc(cfg(feature = "derive")))]
+pub use zerocopy_derive::TryFromBytes;
+
+/// Types for which some bit patterns are valid.
+///
+/// A memory region of the appropriate length which contains initialized bytes
+/// can be viewed as a `TryFromBytes` type so long as the runtime value of bytes
```
joshlf, line 1192:
```suggestion
/// can be viewed as a `TryFromBytes` type so long as the runtime value of those bytes
```

src/lib.rs
```diff
@@ -1146,18 +1148,87 @@ pub unsafe trait NoCell {
...
+/// correspond to a [*valid instance*] of that type. For example, [`bool`] is
```
joshlf, line 1193:
```suggestion
/// corresponds to a [*valid instance*] of that type. For example, [`bool`] is
```

src/lib.rs
```diff
@@ -1146,18 +1148,87 @@ pub unsafe trait NoCell {
...
+/// `TryFromBytes`; zerocopy can transmute a [`bool`] into a [`u8`] so long as
```
joshlf, line 1194:
```suggestion
/// `TryFromBytes`; zerocopy can transmute a [`u8`] into a  [`bool`] so long as
```

Also, this isn't technically true in our API (right now). We don't have `try_transmute!`, so we can't actually do this. Maybe say something like "zerocopy can transmute a `[u8]` into a `bool`..."?

joshlf (changes requested):

src/lib.rs

src/lib.rs

src/lib.rs

joshlf (approved):

PR #983, opened 2024-03-01
Restructure crate-level 'Overview' documentation

jswrenn:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -21,28 +21,54 @@
 //! Zerocopy makes zero-cost memory manipulation effortless. We write `unsafe`
 //! so you don't have to.
 //!
+//! [customer-request-issue]: https://github.com/google/zerocopy/issues/new/choose
+//!
 //! # Overview
 //!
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -21,29 +21,55 @@
...
 //!
+//! [customer-request-issue]: https://github.com/google/zerocopy/issues/new/choose
+//!
 //! # Overview
 //!
-//! Zerocopy provides four core marker traits, each of which can be derived
-//! (e.g., `#[derive(FromZeros)]`):
-//! - [`TryFromBytes`] indicates that a type may safely be converted from certain
-//!   byte sequences (conditional on runtime checks)
+//! ###### Conversion Traits
```
joshlf, line 28: Why so many pound signs?

joshlf (approved):

src/lib.rs
jswrenn: H6 is the lowest level of heading tag, used here because I want paragraph-level headings, not section-level headings.

PR #988, opened 2024-03-01
Document `KnownLayout`

jswrenn:
Documents the derive, and aligns the trait documentation with that of other traits.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -287,10 +287,6 @@ mod wrappers;
 pub use crate::byteorder::*;
 pub use crate::wrappers::*;
 
-#[cfg(any(feature = "derive", test))]
-#[cfg_attr(doc_cfg, doc(cfg(feature = "derive")))]
-pub use zerocopy_derive::KnownLayout;
-
 use core::{
     cell::{self, RefMut, UnsafeCell},
     cmp::Ordering,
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -934,19 +930,102 @@ impl DstLayout {
...
+/// # /*
+///     ...
+/// # */
+/// }
+/// ```
+///
+/// # Limitations
+///
+/// This derive cannot currently be applied to unsized structs with the
+/// default-`repr`.
```
joshlf, line 968: Not sure that "default-`repr`" is a standard term. Maybe "without an explicit `repr` attribute" or something?

src/lib.rs
```diff
@@ -934,19 +930,102 @@ impl DstLayout {
     }
 }
 
-/// A trait which carries information about a type's layout that is used by the
-/// internals of this crate.
+/// Implements [`KnownLayout`].
+///
+/// This derive analyzes various qualities of a type's layout that are needed
```
joshlf, line 935:
```suggestion
/// This derive analyzes various properties of a type's layout that are needed
```

Alternatively, "aspects"?

Same below in the other doc comment.

src/lib.rs
```diff
@@ -934,19 +930,102 @@ impl DstLayout {
...
+/// for some of zerocopy's APIs. It should be used whenver possible, and can be
```
joshlf, line 936: Should we be suggesting "use this whenever possible", or just leave it up to users to discover (via bounds) when they need to derive it? It seems redundant given that the bounds will guide them to when they need to derive `KnownLayout`. It adds extra prose that, to me, sounds a tad mystical.

Same elsewhere in this PR.

joshlf (changes requested):

src/lib.rs

src/lib.rs

src/lib.rs
jswrenn: Nixed the "use this whenever possible"; I agree it sounds a bit mystical here.

jswrenn: That said, I'd like to keep it in the crate-level overview.

joshlf (changes requested):
It might be worth explicitly calling out that this supports unsized slice DST types, and providing examples in the documentation. We'll have to be a bit careful because:
- We also require `KnownLayout` on `Sized` types, so we shouldn't imply that that's the *only* time when it's needed.
- We may make `KnownLayout` smarter in the future and use it encode other information that lets us do things besides slice DST stuff.

That said, I could imagine someone thinking, "why do you need this if you have `mem::size_of` and `mem::align_of`?", and it could be good to respond to that thought.

joshlf (approved):
> It might be worth explicitly calling out that this supports unsized slice DST types, and providing examples in the documentation. We'll have to be a bit careful because:
> 
> * We also require `KnownLayout` on `Sized` types, so we shouldn't imply that that's the _only_ time when it's needed.
> * We may make `KnownLayout` smarter in the future and use it encode other information that lets us do things besides slice DST stuff.
> 
> That said, I could imagine someone thinking, "why do you need this if you have `mem::size_of` and `mem::align_of`?", and it could be good to respond to that thought.

Discussed offline; will follow up in #993

PR #995, opened 2024-03-01
KnownLayout tracks pointer metadata type

joshlf:
Makes progress on https://github.com/google/zerocopy/issues/29

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1065,14 +1065,73 @@ pub unsafe trait KnownLayout {
     where
         Self: Sized;
 
+    /// `()` for sized types and `usize` for slice DSTs.
+    #[doc(hidden)]
+    type PointerMetadata: PointerMetadata;
+
     #[doc(hidden)]
     const LAYOUT: DstLayout;
 
     /// SAFETY: The returned pointer has the same address and provenance as
     /// `bytes`. If `Self` is a DST, the returned pointer's referent has `elems`
-    /// elements in its trailing slice. If `Self` is sized, `elems` is ignored.
+    /// elements in its trailing slice.
     #[doc(hidden)]
-    fn raw_from_ptr_len(bytes: NonNull<u8>, elems: usize) -> NonNull<Self>;
+    fn raw_from_ptr_len(bytes: NonNull<u8>, meta: Self::PointerMetadata) -> NonNull<Self>;
+}
+
+/// The metadata associated with a [`KnownLayout`] type.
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -1065,14 +1065,71 @@ pub unsafe trait KnownLayout {
     where
         Self: Sized;
 
+    /// `()` for sized types and `usize` for slice DSTs.
+    #[doc(hidden)]
+    type PointerMetadata: PointerMetadata;
```
jswrenn, line 1070: Might want to remove the `doc(hidden)` here once this appears in `where` bounds.

src/lib.rs
```diff
@@ -1065,14 +1065,71 @@ pub unsafe trait KnownLayout {
...
+#[doc(hidden)]
+pub trait PointerMetadata {
+    /// Constructs a `Self` from an element count.
+    ///
+    /// If `Self = ()`, this returns `()`. If `Self = usize`, this returns
+    /// `elems`. No other types are currently supported.
+    fn from_elem_count(elems: usize) -> Self;
+
+    /// What is the size of the object with the given layout and pointer
+    /// metadata?
```
jswrenn, line 1092:
```suggestion
    /// Computes the size of the object with the given layout and pointer
    /// metadata.
```

src/lib.rs
```diff
@@ -1065,14 +1065,71 @@ pub unsafe trait KnownLayout {
...
+impl PointerMetadata for () {
+    #[inline]
+    #[allow(clippy::unused_unit)]
+    fn from_elem_count(_elems: usize) -> () {}
+
+    #[inline]
+    fn size_for_metadata(&self, layout: DstLayout) -> Option<usize> {
+        match layout.size_info {
+            SizeInfo::Sized { size } => Some(size),
+            SizeInfo::SliceDst(_) => unreachable!(),
```
jswrenn, line 1111: In the interest of minimizing our panic paths, perhaps we should do something besides `unreachable!()`, like return `None`.

(We can't use `unreachable_unchecked()`, because this method is safe and this line is obviously not actually unreachable when evaluated in isolation.)

jswrenn (changes requested):

src/lib.rs
joshlf: Done in the next PR.

src/lib.rs

src/lib.rs

jswrenn (approved):

PR #1018, opened 2024-03-04
Implement `try_transmute!`

### LINKED ISSUE #1013: Implement `try_transmute!`
We should implement `try_transmute!`, analogous to `transmute!`, that performs compile-time size and alignment checks, but also a run-time validity check.

## Considerations

Due to the run-time validity check, this macro will not be usable in const contexts for the foreseeable future. (But we could perhaps work around this by also providing `const fn is_bit_valid` inherent methods in our derive; see https://github.com/google/zerocopy/issues/115.)

The implementation should avoid performing dynamic size or alignment checks.

## Steps
Once #999 is merged, we can define a typed alternative to `is_bit_valid` on `TryFromBytes`:
```rust
#[doc(hidden)]
#[inline]
fn is_src_valid<Src>(src: &mut Src) -> bool
where
    Src: IntoBytes,
    Self: Sized,
{
    let c_ptr = crate::Ptr::from_mut(src);

    if mem::size_of::<Src>() > mem::size_of::<Self>() {
        return false;
    }

    // SAFETY:
    // - `size_of::<Src>()` <= `size_of::<Self>()`
    // - `c_ptr` is exclusively aliased, so we do not need to reason about
    //   `UnsafeCell`
    let c_ptr = unsafe { c_ptr.cast_unsized(|p| p as *mut Self) };

    // SAFETY: `c_ptr` is derived from `src` which is `IntoBytes`. By
    // invariant on `IntoByte`s, `c_ptr`'s referent consists entirely of
    // initialized bytes.
    let c_ptr = unsafe { c_ptr.assume_validity::<crate::invariant::Initialized>() };

    Self::is_bit_valid(c_ptr)
}
```

The dynamic components of `try_transmute!` can then be factored into a helper function:
```rust
/// A helper for `try_transmute!`.
#[inline(always)]
pub fn try_transmute<Src, Dst>(mut src: Src) -> Option<Dst>
where
    Src: IntoBytes,
    Dst: TryFromBytes,
{
    if !Dst::is_src_valid(&mut src) {
        return None;
    }

    let src = ManuallyDrop::new(src);

    // SAFETY:
    // - `src` is a bit-valid instance of `Dst`
    // - `src` is `ManuallyDrop`
    Some(unsafe { core::mem::transmute_copy(&*src) })
}
```

jswrenn:
Closes #1013.

Makes progress towards #5.


joshlf (changes requested):

src/lib.rs
```diff
@@ -1521,6 +1521,45 @@ pub unsafe trait TryFromBytes {
...
+        // - `size_of::<Src>()` <= `size_of::<Self>()`
+        // - `c_ptr` is exclusively aliased, so we do not need to reason about
+        //   `UnsafeCell`
+        #[allow(clippy::as_conversions)]
+        let c_ptr = unsafe { c_ptr.cast_unsized(|p| p as *mut Self) };
+
+        // SAFETY: `c_ptr` is derived from `src` which is `IntoBytes`. By
+        // invariant on `IntoByte`s, `c_ptr`'s referent consists entirely of
+        // initialized bytes.
+        let c_ptr = unsafe { c_ptr.assume_validity::<crate::invariant::Initialized>() };
```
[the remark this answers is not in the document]
jswrenn: We can't do this, because `T` (`Self`) isn't necessarily `IntoBytes`; it's the `Src` that's `IntoBytes`, but we've casted that type away.

src/lib.rs
```diff
@@ -4732,6 +4771,63 @@ macro_rules! transmute_mut {
...
+/// # Examples
+///
+/// ```
+/// # use zerocopy::try_transmute;
+/// assert_eq!(try_transmute!(0u8), Some(false));
+/// assert_eq!(try_transmute!(1u8), Some(true));
+/// assert_eq!(try_transmute!(255u8), Option::<bool>::None);
+/// ```
+#[macro_export]
+macro_rules! try_transmute {
```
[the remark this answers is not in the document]
jswrenn: Done

src/macro_util.rs
```diff
@@ -404,6 +404,25 @@ pub unsafe fn transmute_mut<'dst, 'src: 'dst, Src: 'src, Dst: 'dst>(
     unsafe { &mut *dst }
 }
 
+/// A helper for `try_transmute!`.
+#[inline(always)]
+pub fn try_transmute<Src, Dst>(mut src: Src) -> Option<Dst>
+where
+    Src: core::fmt::Debug + crate::IntoBytes,
+    Dst: core::fmt::Debug + crate::TryFromBytes,
```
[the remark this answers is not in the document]
jswrenn: Done

src/macro_util.rs
```diff
@@ -404,6 +404,25 @@ pub unsafe fn transmute_mut<'dst, 'src: 'dst, Src: 'src, Dst: 'dst>(
...
+{
+    if !Dst::is_src_valid(&mut src) {
+        return None;
+    }
+
+    let src = ManuallyDrop::new(src);
+
+    // SAFETY:
+    // - `src` is a bit-valid instance of `Dst`
```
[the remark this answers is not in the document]
jswrenn: Done

src/lib.rs
```diff
@@ -1521,6 +1521,45 @@ pub unsafe trait TryFromBytes {
...
+        Self: Sized,
+    {
+        let c_ptr = crate::Ptr::from_mut(src);
+
+        if mem::size_of::<Src>() > mem::size_of::<Self>() {
+            return false;
+        }
+
+        // SAFETY:
+        // - `size_of::<Src>()` <= `size_of::<Self>()`
```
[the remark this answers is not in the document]

src/lib.rs
```diff
@@ -1521,6 +1521,45 @@ pub unsafe trait TryFromBytes {
...
+    ///
+    /// # Safety
+    ///
+    /// Unsafe code may assume that, if `is_bit_valid(src)` returns true, `*src`
+    /// is valid instance of `Self`.
+    ///
+    /// # Panics
+    ///
+    /// `is_src_valid` panics under exactly the same circumstances as
+    /// [`TryFromBytes::is_bit_valid`].
```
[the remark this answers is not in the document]

src/lib.rs
```diff
@@ -1521,6 +1521,45 @@ pub unsafe trait TryFromBytes {
     #[doc(hidden)]
     fn is_bit_valid<A: invariant::at_least::Shared>(candidate: Maybe<'_, Self, A>) -> bool;
 
+    /// Is a given source value a valid instance of `Self`?
+    ///
+    /// # Safety
+    ///
+    /// Unsafe code may assume that, if `is_bit_valid(src)` returns true, `*src`
```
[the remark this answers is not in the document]

src/lib.rs
```diff
@@ -4732,6 +4773,65 @@ macro_rules! transmute_mut {
     }}
 }
 
+/// Conditionally transmutes a value of one type to a value of another type of
+/// the same size.
+///
+/// This macro an expression of type  `Src` and produces a `Result<Dst, Src>`.
```
joshlf, line 4779:
```suggestion
/// This macro consumes an expression of type `Src` and produces a `Result<Dst, Src>`.
```

src/lib.rs
```diff
@@ -4732,6 +4773,65 @@ macro_rules! transmute_mut {
...
+/// original `T` will be forgotten, and the value of type `U` will be returned.
+///
+/// # Examples
+///
+/// ```
+/// # use zerocopy::try_transmute;
+/// assert_eq!(try_transmute!(0u8), Ok(false));
+/// assert_eq!(try_transmute!(1u8), Ok(true));
+/// assert_eq!(try_transmute!(255u8), Result::<bool, _>::Err(255));
+/// ```
```
joshlf, line 4797: We should update or add language to discuss what validation is performed, and to discuss what happens on failure.

src/lib.rs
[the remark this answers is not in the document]
joshlf: Ack.

joshlf (changes requested):
Also: Tests.

src/lib.rs

src/lib.rs
```diff
@@ -1107,6 +1107,49 @@ pub unsafe trait TryFromBytes {
...
+    /// # Safety
+    ///
+    /// Unsafe code may assume that, if `is_mut_src_valid(src)` returns true,
+    /// `*src` is a bit-valid instance of `Self`, and that the size of `Src` is
+    /// greater than or equal to the size of `Self`.
+    ///
+    /// # Panics
+    ///
+    /// `is_src_valid` panics under exactly the same circumstances as
+    /// [`is_bit_valid`].
```
joshlf, line 1121: Maybe also mention that size mismatch can cause a panic or a compilation failure? (just like we do for `is_bit_valid`)

src/macro_util.rs
```diff
@@ -409,6 +411,31 @@ pub unsafe fn transmute_mut<'dst, 'src: 'dst, Src: 'src, Dst: 'dst>(
...
+/// Attempts to transmute `Src` into `Dst`.
+///
+/// A helper for `try_transmute!`.
+#[inline(always)]
+pub fn try_transmute<Src, Dst>(mut src: Src) -> Result<Dst, ValidityError<Src, Dst>>
+where
+    Src: crate::IntoBytes,
+    Dst: crate::TryFromBytes,
+{
+    if !Dst::is_mut_src_valid(&mut src) {
```
joshlf, line 423: If we can get `Unalign<Dst>: KnownLayout` without `Dst: KnownLayout`, then we could use `<Unalign<Dst> as TryFromBytes>::try_mut_from` here instead of needing to add `is_mut_src_valid`, right?

joshlf (changes requested):

src/macro_util.rs
joshlf: Using #1252, we might be able to use `Unalign<SizedKnownLayout<Dst>>`.

src/lib.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/80bdd61e29970f8ba17c2ce5e47f2c69b1d9dffa..dca55262906a36d52de53f88a608eb34d56c1c27#diff-bbffbe0cf99afd2b815401c126c4790999e882fb64e3cb367aea899c64c95966R425-R429)

src/macro_util.rs
jswrenn: Maybe, though doing so would entail needless runtime checks and ultimately end with throwing away the return value of `try_mut_from`.

I think `is_src_valid` (which I've now generalized to take a `Ptr<Src>`) will also be useful for `try_transmute_mut` and `try_transmute_ref`.

src/macro_util.rs
```diff
@@ -409,6 +414,83 @@ pub unsafe fn transmute_mut<'dst, 'src: 'dst, Src: 'src, Dst: 'dst>(
...
+    // - `p as *mut Self` is a provenance-preserving cast
+    // - The aliasing of `c_ptr` is neither `Any` nor `Shared`, so we do not
+    //   need to reason about the ranges of `UnsafeCell`s.
+    #[allow(clippy::as_conversions)]
+    let c_ptr = unsafe { src.cast_unsized(|p| p as *mut Dst) };
+
+    // SAFETY: `c_ptr` is derived from `src` which is `IntoBytes`. By
+    // invariant on `IntoByte`s, `c_ptr`'s referent consists entirely of
+    // initialized bytes.
+    let c_ptr = unsafe { c_ptr.assume_initialized() };
```
joshlf, line 456: Why don't we add a safe `bikeshed_recall_initialized` method on `Ptr` with `T: IntoBytes` and `I: Invariants<Validity = Valid>` bounds?

src/macro_util.rs
```diff
@@ -409,6 +414,83 @@ pub unsafe fn transmute_mut<'dst, 'src: 'dst, Src: 'src, Dst: 'dst>(
...
+{
+    crate::util::assert_dst_not_bigger_than_src::<Src, Dst>();
+
+    // SAFETY: This is a pointer cast, satisfying the following properties:
+    // - `p as *mut Self` addresses a subset of the `bytes` addressed by
+    //   `c_ptr`, because we assert above that the size of `Self` is less
+    //   than or equal to the size of `Src`.
+    // - `p as *mut Self` is a provenance-preserving cast
+    // - The aliasing of `c_ptr` is neither `Any` nor `Shared`, so we do not
+    //   need to reason about the ranges of `UnsafeCell`s.
```
joshlf, line 449: Looks like these were copy-pasted without being updated?

src/lib.rs
```diff
@@ -5607,6 +5683,50 @@ mod tests {
         assert_eq!(*y, 0);
     }
 
+    #[test]
+    fn test_try_transmute() {
+        // Test that memory is transmuted with transmuted as expected.
```
joshlf, line 5688: Typo?

src/lib.rs
```diff
@@ -5607,6 +5683,50 @@ mod tests {
...
+        let array_of_bools = [false, true, false, true, false, true, false, true];
+        let array_of_arrays = [[0, 1], [0, 1], [0, 1], [0, 1]];
+        let x: Result<[[u8; 2]; 4], _> = try_transmute!(array_of_bools);
+        assert_eq!(x, Ok(array_of_arrays));
+        let x: Result<[bool; 8], _> = try_transmute!(array_of_arrays);
+        assert_eq!(x, Ok(array_of_bools));
+
+        // Test that `transmute!` works with `!NoCell` types.
```
joshlf, line 5696:
```suggestion
        // Test that `transmute!` works with `!Immutable` types.
```

joshlf (changes requested):
UI tests? There are a lot of `transmute-*.rs` UI tests that we could duplicate, and possible some we could add (e.g. for `Dst: !TryFromBytes`).

src/lib.rs

src/lib.rs

src/macro_util.rs
jswrenn: Not just that, but a soundness issue too! Fixed, thanks to our spiffy `AliasingSafe` trait.

src/macro_util.rs
jswrenn: I don't think that would help here without much more machinery. The `cast_unsized` blows our validity away to `Any`.

jswrenn:
> UI tests? There are a lot of transmute-*.rs UI tests that we could duplicate, and possible some we could add (e.g. for Dst: !TryFromBytes).

[Done](https://github.com/google/zerocopy/compare/d3b7f6e3adf0302247fe584169a3739e361f083e..004b5b452f5b7e481e7fb3c581c09bb564d8ddd5)

src/macro_util.rs
joshlf: That's convincing.

joshlf (approved):

PR #1024, opened 2024-03-06
TransparentWrapper supports UnsafeCell variance

joshlf:
This will allow us to implement `TransparentWrapper` for `T` where `T` and `T::Inner` do not have `UnsafeCell`s at the same byte ranges. This, in turn, will let us automatically implement traits for these types so long as those traits don't require reasoning about `UnsafeCell` ranges (in other words, all traits other than `NoCell`).

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -26,13 +26,12 @@ use crate::{
 ///
 /// # Safety
 ///
-/// `T: TransparentWrapper` implies that:
-/// - `T` has the same size as [`T::Inner`]
-/// - `T` has `UnsafeCell`s covering the same byte ranges as `T::Inner`
-/// - `T` has zero-sized `UnsafeCell`s (e.g., `UnsafeCell<()>`,
-///   `[UnsafeCell<u8>; 0]`, etc) at the same byte offsets as `T::Inner`
-///
+/// `T: TransparentWrapper` implies that `T` has the same size as [`T::Inner`].
 /// Further, `T: TransparentWrapper<I>` implies that:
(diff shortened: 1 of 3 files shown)


src/util.rs
```diff
@@ -106,14 +106,17 @@ impl<I: invariant::Validity> ValidityVariance<I> for Invariant {
...
-// - Per [1], `MaybeUninit<T>` has `UnsafeCell`s at the same byte ranges as
-//   `Inner = T`, and `UnsafeCell`s at the same byte offsets as `T`.
 // - See inline comments for other safety justifications.
 //
 // [1] TODO(#896): Write a safety proof before the next stable release.
 unsafe impl<T, I: Invariants> TransparentWrapper<I> for MaybeUninit<T> {
     type Inner = T;
 
+    // SAFETY: Per [1], `MaybeUninit<T>` has `UnsafeCell`s at the same byte
+    // ranges as `Inner = T`, and `UnsafeCell`s at the same byte offsets as `T`.
```
jswrenn, line 116: What's the distinction you're drawing here between ranges and offsets?

joshlf: I should perhaps say "covering" the same byte ranges. E.g., the type `#[repr(C)] struct Foo(UnsafeCell<()>, usize)` doesn't have any byte ranges which are covered by `UnsafeCell`s, but there is a ZST `UnsafeCell` at byte offset 0. Until https://github.com/rust-lang/unsafe-code-guidelines/issues/495 is resolved (and until we remove the restriction from `NoCell` that even ZST `UnsafeCell`s are banned), this distinction is required for soundness.

Issue #1009, opened 2024-03-03
Support atomic types


joshlf:
Various `core::sync::atomic::AtomicXxx` types could implement some of our traits.

joshlf:
Blocked on https://github.com/google/zerocopy/issues/1818

PR #1028, opened 2024-03-07
Implement traits for atomic types

joshlf:
Makes progress on #1009. This commit doesn't implement all traits for `AtomicPtr<T>`, so there's more work to do.


jswrenn (approved):

josephlr:
@joshlf it seems like this PR causes `AtomicBool` to implement `FromBytes` which is unsound. This is fixed in #1091

Issue #1046, opened 2024-03-14
`transmute_mut!` documents looser requirements than what is implemented


kupiakos:
[`transmute_mut!`](https://docs.rs/zerocopy/0.8.0-alpha.7/zerocopy/macro.transmute_mut.html) documents the requirements that:

- `T: Sized + IntoBytes`
- `U: Sized + FromBytes`
- `align_of::<T>() >= align_of::<U>()`

However, this is less strict that what is actually required:

 - `T: FromBytes + IntoBytes + NoCell`
 - `U: FromBytes + IntoBytes + NoCell`
 - `size_of::<T>() == size_of::<U>()`
 - `align_of::<T>() >= align_of::<U>()`

The size and `NoCell` requirements are also similarly missing from `transmute_ref!`.

joshlf:
Good catch! We'll block [releasing 0.8](https://github.com/google/zerocopy/issues/671) on this.

jswrenn:
This was completed in #1050 and #1058. We now use the following pseudocode to document the bounds of our transmute macros:
```rust
const fn transmute<Src, Dst>(src: Src) -> Dst
where
    Src: IntoBytes,
    Dst: FromBytes,
    size_of::<Src>() == size_of::<Dst>();

const fn transmute_ref<'src, 'dst, Src, Dst>(src: &'src Src) -> &'dst Dst
where
    'src: 'dst,
    Src: IntoBytes + NoCell,
    Dst: FromBytes + NoCell,
    size_of::<Src>() == size_of::<Dst>(),
    align_of::<Src>() >= align_of::<Dst>();

const fn transmute_mut<'src, 'dst, Src, Dst>(src: &'src mut Src) -> &'dst mut Dst
where
    'src: 'dst,
    Src: FromBytes + IntoBytes + NoCell,
    Dst: FromBytes + IntoBytes + NoCell,
    size_of::<Src>() == size_of::<Dst>(),
    align_of::<Src>() >= align_of::<Dst>();
```

Issue #686, opened 2023-12-07
Allow `UnsafeCell`s when transmuting mutable references?


joshlf:
Once we land #251, we'll have the ability to granularly require or permit `UnsafeCell`s in different APIs. Currently the plan is to require `UnsafeCell`s whenever references are being transmuted, but maybe that's not necessary when the references are mutable, as that disallows interior mutation anyway? This may depend on which aliasing model we choose (stacked borrows, tree borrows, etc), as some may disallow this.

PR #1049, opened 2024-03-15
Relax `NoCell` bound on `transmute_mut!`

jswrenn:
And document that both `T` and `U` must be `FromBytes + IntoBytes`.

Makes progress towards #686.

Partially addresses #1046.


joshlf:
IMO we should wait until (the `&mut` part of) #495 is resolved to land the `NoCell` part of this PR.

jswrenn:
Superseded by #1225.

PR #1050, opened 2024-03-15
Use pseudocode to document `transmute_ref`

jswrenn:
To my eyes, this is much more easily comprehensible at a glance than the prosaic alternative.

If this looks good to you, @joshlf, I'll apply the pattern to `transmute!` and `transmute_mut!` too. 

Partially addresses #1046.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4418,14 +4418,27 @@ macro_rules! transmute {
 /// Safely transmutes a mutable or immutable reference of one type to an
 /// immutable reference of another type of the same size.
 ///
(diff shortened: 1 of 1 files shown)


joshlf (changes requested):
Maybe also mention that the argument and return types are inferred from the calling context? That aspect of the behavior differs from the given function since, with a function, you can specify the types if you want. A caller might expect our macro to support some sort of type syntax akin to `transmute_ref::<S, D>`.

joshlf (approved):

Issue #1051, opened 2024-03-15
Request: Add FromBytes::read_from_prefix_split and read_from_suffix_split


smalis-msft:
It would be handy to have a FromBytes::read_from_(prefix/suffix)_split that returns a tuple of the T and the unused portion of the byte slice. Ref::new_from_prefix isn't quite the same in case the byte slice is unaligned, but we don't want to derive Unaligned on the type.

jswrenn:
Something like this ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=859eb8eea011a0dcd733e2c5fc9fc995))?
```rust
fn read_from_prefix_split<T>(bytes: &[u8]) -> Option<(T, &[u8])>
where
    T: Sized + FromBytes,
{
    Ref::<_, Unalign<T>>::new_unaligned_from_prefix(bytes)
        .map(|(r, s)| (r.read().into_inner(), s))
}
```

smalis-msft:
Yeah that looks about right. I mean there's a number of ways to do it, you can just call FromBytes::read_from_prefix and then index into your byte buffer by sizeof<T>, but having it provided and able to be used in a small oneliner would be nice.

jstarks:
In our code, we often parse structures with fixed headers and then some dynamic portion, from a potentially unaligned buffer. To support this, it has been nice to have both options readily available (both preserving the suffix bytes and not), without having to reach for `Ref`.

`Ref` is powerful, but it's certainly more tedious to use. And it's arguably more dangerous (from a correctness perspective, not a safety one): since reading from an unaligned buffer requires even more typing than reading from/aliasing an aligned one, developers get lazy and just assume their buffer is aligned. And in their testing, it is. Then something subtle changes in the code, possibly far away from the call site, and runtime checks start failing.

Or developers write something like `let foo = Foo::read_from_prefix(buf)?` followed by `let buf = &buf[..size_of::<Foo>()]`, which is error prone from the type name duplication. (I guess `size_of_val(&foo)` would also fix this, but it's still tedious to write.)

To solve this, we have a private crate that provides these kind of extra methods for use with `zerocopy`, but we'd love to just standardize on a single crate. Since these methods aren't already in `zerocopy`, I'm curious if we're "holding it wrong"--maybe there's a better way to do this kind of parsing? Do most users have aligned buffers? Or have unaligned struct definitions?

joshlf:
[...]

Issue #884, opened 2024-02-15
`FromBytes::(mut_)?slice_from_[prefix|suffix]` also returns the prefix/suffix instead of just the read slice


kupiakos:
- `slice_from_prefix` returns `Option<(&[Self], &[u8])>`
- `slice_from_suffix` returns `Option<(&[u8], &[Self])>`
- `mut_slice_from_prefix` returns `Option<(&mut [Self], &mut [u8])>`
- `mut_slice_from_suffix` returns `Option<(&mut [u8], &mut [Self])>`

However, all of the other methods drop the prefix/suffix:

- `FromBytes::read_from_prefix` returns `Option<Self>`
- `FromBytes::read_from_suffix` returns `Option<Self>`
- `FromBytes::ref_from_prefix` returns `Option<&Self>`
- `FromBytes::ref_from_suffix` returns `Option<&Self>`
- `FromBytes::mut_from_prefix` returns `Option<&mut Self>`
- `FromBytes::mut_from_suffix` returns `Option<&mut Self>`

This is desirable: the majority of the time I drop the prefix anyways and these are convenience methods for methods in `Ref`.
`[ref|mut]_from_[prefix|suffix]` in their docs both point to using `Ref` to preserve the prefix/suffix.

joshlf:
This will be a breaking change, so we'll need to make sure we either address it as part of 0.8 or punt it to a future release. I've added it to #671 to make sure we make a decision one way or another before we release 0.8.

joshlf:
Heads up about https://github.com/google/zerocopy/issues/1051, which relates to this, and about https://github.com/google/zerocopy/pull/1059, which resolves #1051 in the opposite direction from this request.

Unless we want even more method duplication (which we'd really like to avoid), it seems that we'll need to just pick one of the following two approaches:
- The `FromBytes` methods drop the suffix/prefix, and users who want them need to use `Ref`
- The `FromBytes` methods return the suffix/prefix, and users who don't need them need to explicitly discard them

#1059 takes the latter approach. From a survey of uses in the ecosystem, it seems that needing the suffix/prefix is more common than not needing it. We also feel like discarding the suffix/prefix is more ergonomic than using the `Ref` API equivalents.

@kupiakos want to make sure you're aware that this is something we're considering and give you a chance to chime in.

kupiakos:
> From a survey of uses in the ecosystem

[...]

PR #1059, opened 2024-03-19
Also return remaining bytes in `FromBytes` conversion methods

jswrenn:
This change makes the slice and non-slice conversion methods consistent, and provides a lightweight way (i.e., without `Ref`) to access the leftover bytes after conversions. For instance, this:
```rust
trait FromBytes {
    fn ref_from_prefix(bytes: &[u8]) -> Option<&Self>
    where
        Self: KnownLayout + NoCell;
}
```
...becomes this, in this PR:
```rust
trait FromBytes {
    fn ref_from_prefix(bytes: &[u8]) -> Option<(&Self, &[u8])>
    where
        Self: KnownLayout + NoCell;
}
```
...where the second element of the tuple are the remaining bytes. Providing the remainder eases the pattern where a `T` is parsed from a buffer, and then the buffer is advanced to exclude the parsed `T`.

This PR tests out the 'feel' of tuple-returning conversion methods. Alternatively, we could provide a mirror set of `_split`  methods.

Addresses #1051, but is the antithesis of #884. Inspired by https://github.com/google/zerocopy/issues/1051#issuecomment-2007575409. It's syntactically cheap for users to discard remaining bytes at call sites if they wish, but the alternative — discarding the excess and requiring users to go through `Ref` — is syntactically and cognitively expensive.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2203,12 +2203,10 @@ pub unsafe trait FromBytes: FromZeros {
 
     /// Interprets the prefix of the given `bytes` as a `&Self` without copying.
     ///
-    /// `ref_from_prefix` returns a reference to the first `size_of::<Self>()`
-    /// bytes of `bytes`. If `bytes.len() < size_of::<Self>()` or `bytes` is not
-    /// aligned to `align_of::<Self>()`, this returns `None`.
-    ///
-    /// To also access the prefix bytes, use [`Ref::new_from_prefix`]. Then, use
-    /// [`Ref::into_ref`] to get a `&Self` with the same lifetime.
+    /// This method returns both a reference to the first `size_of::<Self>()`
+    /// bytes of `bytes` interpreted as `Self`, and a reference to the remaining
+    /// bytes. If `bytes.len() < size_of::<Self>()` or `bytes` is not aligned to
+    /// `align_of::<Self>()`, this returns `None`.
     ///
     /// # Examples
     ///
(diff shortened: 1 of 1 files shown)


smalis-msft:
While I'm obviously in favor of doing this in general, I feel like modifying the existing methods would be too large of a breaking change to be worthwhile, which was why I suggested new methods.

jswrenn:
Hm, we'll need to balance the costs of subjecting current users to breaking changes, against the costs of subjecting current and future users to a more bloated API.  Adding `_split` methods consistently would require adding eight additional methods to `FromBytes` (which has 14 methods in our latest 0.8 pre-release):
```diff
  mut_from
  mut_from_prefix
+ mut_from_prefix_split
  mut_from_suffix
+ mut_from_suffix_split
  mut_slice_from
  mut_slice_from_prefix
+ mut_slice_from_prefix_split
  mut_slice_from_suffix
+ mut_slice_from_suffix_split
  read_from
  read_from_prefix
+ read_from_prefix_split
  read_from_suffix
+ read_from_suffix_split
  ref_from
  ref_from_prefix
+ ref_from_prefix_split
  ref_from_suffix
+ ref_from_suffix_split
  slice_from_prefix
  slice_from_suffix
```
That's quite a hefty addition, and whatever course of action we take here will also likely be taken for our upcoming [`TryFromBytes`](https://docs.rs/zerocopy/0.8.0-alpha.7/zerocopy/trait.TryFromBytes.html) trait, too.

Since we're preparing for a major breaking release anyways, we have a bit more license than usual to consider breaking changes.

kupiakos:
As I discuss in https://github.com/google/zerocopy/issues/884#issuecomment-2027828809, I'm strongly in support of _not_ removing the plain-casting methods that don't return the suffix. **The code should do the minimum work necessary for what's needed.**

I personally find the zerocopy APIs lacking in useful transitive operations like this, so I would prefer the "bloat" of adding these methods and keeping the plain-casting ones.

What do you think of these methods starting with `split_` instead of ending in `_split`? `split_mut_from_prefix` reads clearly in the imperative mood and highlights that it's doing more than just a bounds-check and pointer cast: it's splitting as well. This **split is an extra operation**, both cognitively and for the compiler.

I don't care about the `*_from_suffix` API and I don't know who does - all I ask is that there's consistency where it makes sense. Perhaps as a middle ground, this PR can introduce only splitting versions for the `prefix` APIs, which adds 5 methods instead of 10:

```diff
  mut_from
  mut_from_prefix
+ split_mut_from_prefix
  mut_from_suffix
  mut_slice_from
  mut_slice_from_prefix
+ split_mut_slice_from_prefix
  mut_slice_from_suffix
  read_from
  read_from_prefix
+ split_read_from_prefix
  read_from_suffix
  ref_from
  ref_from_prefix
+ split_ref_from_prefix
  ref_from_suffix
  slice_from_prefix
+ split_slice_from_prefix
  slice_from_suffix
```

smalis-msft:
> I don't care about the *_from_suffix API and I don't know who does
We use it for parsing certain things as well as prefix, and would strongly prefer to have split versions of it as well if you decide to go this route.

jswrenn:
I've made a discussion post exploring some of the naming and return type design space: https://github.com/google/zerocopy/discussions/1095

joshlf:
I'm inclined to merge this based on the following reasoning:
- The ergonomics of discarding an unneeded prefix/suffix are better than the ergonomics of reconstructing a prefix/suffix if one isn't returned by the API
- From what little visibility we have into use cases in the ecosystem, it seems that wanting the prefix/suffix is the more common situation
- For users who care about not relying on the compiler to perform optimizations, techniques like the one described in [this comment](https://github.com/google/zerocopy/issues/884#issuecomment-2043538316) are available
- More generally, we are aware that our current internal software organization is based on the assumption that the compiler will aggressively optimize, and we are actively considering how to provide better support for users who aren't comfortable with this situation: https://github.com/google/zerocopy/issues/1125

src/lib.rs
```diff
@@ -2235,31 +2233,31 @@ pub unsafe trait FromBytes: FromZeros {
...
     ///
-    /// `ref_from_suffix` returns a reference to the last `size_of::<Self>()`
-    /// bytes of `bytes`. If `bytes.len() < size_of::<Self>()` or the suffix of
-    /// `bytes` is not aligned to `align_of::<Self>()`, this returns `None`.
-    ///
-    /// To also access the suffix bytes, use [`Ref::new_from_suffix`]. Then, use
-    /// [`Ref::into_ref`] to get a `&Self` with the same lifetime.
+    /// This method returns both a reference to the last `size_of::<Self>()`
+    /// bytes of `bytes` interpreted as `Self`, and a reference to the remaining
+    /// bytes. If `bytes.len() < size_of::<Self>()` or the suffix of `bytes` is
```
joshlf, line 2259: To make it more clear that it's the prefix, not the suffix, that is returned:

```suggestion
    /// bytes of `bytes` interpreted as `Self`, and a reference to the preceding
    /// bytes. If `bytes.len() < size_of::<Self>()` or the suffix of `bytes` is
```

src/lib.rs
```diff
@@ -2360,35 +2358,36 @@ pub unsafe trait FromBytes: FromZeros {
...
-    /// bytes of `bytes`. If `bytes.len() < size_of::<Self>()` or the suffix of
-    /// `bytes` is not aligned to `align_of::<Self>()`, this returns `None`.
+    /// Interprets the suffix of the given `bytes` as a `&mut Self` without
+    /// copying.
     ///
-    /// To also access the suffix bytes, use [`Ref::new_from_suffix`]. Then,
-    /// use [`Ref::into_mut`] to get a `&mut Self` with the same lifetime.
+    /// This method returns both a reference to the last `size_of::<Self>()`
+    /// bytes of `bytes` interpreted as `Self`, and a reference to the remaining
+    /// bytes. If `bytes.len() < size_of::<Self>()` or the suffix of `bytes` is
```
joshlf, line 2389:
```suggestion
    /// This method returns both a reference to the last `size_of::<Self>()`
    /// bytes of `bytes` interpreted as `Self`, and a reference to the preceding
    /// bytes. If `bytes.len() < size_of::<Self>()` or the suffix of `bytes` is
```

joshlf (changes requested):

joshlf (approved):

PR #1072, opened 2024-03-25
Add `TryFromBytes` prefix/suffix conversion methods

jswrenn:


joshlf (changes requested):
Rebase on top of: #1161

src/lib.rs
```diff
@@ -1132,40 +1130,237 @@ pub unsafe trait TryFromBytes {
...
+                // calling `try_into_valid` (and thus `is_bit_valid`) with a shared
+                // pointer when `Self: !Immutable`. Since `Self: Immutable`, this panic
+                // condition will not happen.
+                match candidate.try_into_valid() {
+                    Ok(valid) => Ok(valid.as_ref()),
+                    Err(e) => Err(e.map_src(|src| src.as_bytes().as_ref()).into()),
+                }
+            }
+            Err(e) => Err(e.map_src(Ptr::as_ref).into()),
+        }
```
jswrenn, line 1194: Rewritten in this manner for consistency with our other `from_ref` methods, and because doing so allows us to name `try_ref_from`'s argument "candidate" without thinking of new, weird identifier names.

jswrenn:
@joshlf Need to write a commit message, but otherwise this is ready for review. This PR intentionally omits prefix/suffix-returning versions of `try_read_from`, since I suspect those will get considerably less verbose once we finish our `UnsafeCell`-and-`Ptr` refactor.

src/lib.rs
```diff
@@ -1102,16 +1102,14 @@ pub unsafe trait TryFromBytes {
...
-    /// reinterpreted as a `Self`.
+    /// Attempts to interpret the given `candidate` as a `&Self` without
+    /// copying.
     ///
-    /// Note that Rust's bit validity rules are still being decided. As such,
-    /// there exist types whose bit validity is ambiguous. See
-    /// [here][TryFromBytes#what-is-a-valid-instance] for a discussion of how
-    /// these cases are handled.
+    /// If the bytes of `candidate` are a valid instance of `Self`, this method
+    /// returns a reference to those bytes interpreted as `Self`. If
```
joshlf, line 1109:
```suggestion
    /// returns a reference to those bytes reinterpreted as a `Self`. If
```

src/lib.rs
```diff
@@ -1102,16 +1102,14 @@ pub unsafe trait TryFromBytes {
...
+    /// `candidate.len() < size_of::<Self>()` or `candidate` is not aligned to
```
joshlf, line 1110: This supports unsized types, so `candidate.len() < size_of::<Self>()` is too restrictive.

src/lib.rs
```diff
@@ -1132,40 +1130,236 @@ pub unsafe trait TryFromBytes {
...
+    /// ```
+    /// use zerocopy::TryFromBytes;
+    /// # use zerocopy_derive::*;
+    ///
+    /// // The only valid value of this type is the byte `0xC0`
+    /// #[derive(TryFromBytes, KnownLayout, Immutable)]
+    /// #[repr(u8)]
+    /// enum C0 { xC0 = 0xC0 }
+    ///
+    /// // The only valid value of this type is the bytes `0xC0C0`.
```
joshlf, line 1145:
```suggestion
    /// // The only valid value of this type is the byte sequence `0xC0C0`.
```

src/lib.rs
```diff
@@ -1132,40 +1130,236 @@ pub unsafe trait TryFromBytes {
...
+    /// #[derive(TryFromBytes, KnownLayout, Immutable)]
+    /// #[repr(C, packed)]
```
joshlf, line 1147:
```suggestion
    /// #[repr(C)]
```

In this case, we don't need `packed`, and it might be better not to introduce more concepts for readers who aren't familiar.

src/lib.rs
```diff
@@ -1132,40 +1130,236 @@ pub unsafe trait TryFromBytes {
...
+        }
+    }
 
-        candidate.map(MaybeAligned::as_ref)
+    /// Attempts to interpret the prefix of the given `candidate` as a `&Self`
+    /// without copying.
+    ///
+    /// If the first `size_of::<Self>()` bytes of `candidate` are a valid
+    /// instance of `Self`, this method returns both a reference to those bytes
+    /// interpreted as `Self`, and a reference to the remaining bytes. If
```
joshlf, line 1202:
```suggestion
    /// reinterpreted as a `Self`, and a reference to the remaining bytes. If
```

src/lib.rs
```diff
@@ -1132,40 +1130,236 @@ pub unsafe trait TryFromBytes {
...
+    /// `candidate.len() < size_of::<Self>()` or `candidate` is not aligned to
```
joshlf, line 1203: Same here as above wrt unsized types

src/lib.rs
```diff
@@ -1132,40 +1130,236 @@ pub unsafe trait TryFromBytes {
...
+    /// # use zerocopy_derive::*;
+    ///
+    /// // The only valid value of this type is the byte `0xC0`
+    /// #[derive(TryFromBytes, KnownLayout, Immutable)]
+    /// #[repr(u8)]
+    /// enum C0 { xC0 = 0xC0 }
+    ///
+    /// // The only valid value of this type is the bytes `0xC0C0`.
+    /// #[derive(TryFromBytes, KnownLayout, Immutable)]
+    /// #[repr(C, packed)]
```
joshlf, line 1240:
```suggestion
    /// #[repr(C)]
```

src/lib.rs
```diff
@@ -1132,40 +1130,236 @@ pub unsafe trait TryFromBytes {
...
+        try_ref_from_prefix_suffix(candidate, CastType::Prefix)
     }
 
-    /// Attempts to interpret a mutable byte slice as a `Self`.
+    /// Attempts to interpret the suffix of the given `candidate` as a `&Self`
+    /// without copying.
+    ///
+    /// If the last `size_of::<Self>()` bytes of `candidate` are a valid
+    /// instance of `Self`, this method returns both a reference to those bytes
+    /// interpreted as `Self`, and a reference to the preceding bytes. If
```
joshlf, line 1281:
```suggestion
    /// reinterpreted as a `Self`, and a reference to the preceding bytes. If
```

src/lib.rs
```diff
@@ -1132,40 +1130,236 @@ pub unsafe trait TryFromBytes {
...
+    /// `candidate.len() < size_of::<Self>()` or the suffix of `candidate` is
```
joshlf, line 1282: Same here as above wrt unsized types

src/lib.rs
```diff
@@ -1132,40 +1130,236 @@ pub unsafe trait TryFromBytes {
...
+    /// # use zerocopy_derive::*;
+    ///
+    /// // The only valid value of this type is the byte `0xC0`
+    /// #[derive(TryFromBytes, KnownLayout, Immutable)]
+    /// #[repr(u8)]
+    /// enum C0 { xC0 = 0xC0 }
+    ///
+    /// // The only valid value of this type is the bytes `0xC0C0`.
+    /// #[derive(TryFromBytes, KnownLayout, Immutable)]
+    /// #[repr(C, packed)]
```
joshlf, line 1319:
```suggestion
    /// #[repr(C)]
```

src/lib.rs
```diff
@@ -1132,40 +1130,236 @@ pub unsafe trait TryFromBytes {
...
+
+    /// Attempts to interpret the given `candidate` as a `&mut Self` without
+    /// copying.
     ///
-    /// Note that Rust's bit validity rules are still being decided. As such,
-    /// there exist types whose bit validity is ambiguous. See
-    /// [here][TryFromBytes#what-is-a-valid-instance] for a discussion of how
-    /// these cases are handled.
+    /// If the bytes of `candidate` are a valid instance of `Self`, this method
+    /// returns a reference to those bytes interpreted as `Self`. If
```
joshlf, line 1359:
```suggestion
    /// returns a reference to those bytes reinterpreted as a `Self`. If
```

src/lib.rs
```diff
@@ -1132,40 +1130,236 @@ pub unsafe trait TryFromBytes {
...
+    /// `candidate.len() < size_of::<Self>()` or `candidate` is not aligned to
```
joshlf, line 1360: Same here as above wrt unsized types

src/lib.rs
```diff
@@ -1187,6 +1381,44 @@ pub unsafe trait TryFromBytes {
...
+    /// # use zerocopy_derive::*;
+    ///
+    /// // The only valid value of this type is the byte `0xC0`
+    /// #[derive(TryFromBytes, KnownLayout, Immutable)]
+    /// #[repr(u8)]
+    /// enum C0 { xC0 = 0xC0 }
+    ///
+    /// // The only valid value of this type is the bytes `0xC0C0`.
+    /// #[derive(TryFromBytes, KnownLayout, Immutable)]
+    /// #[repr(C, packed)]
```
joshlf, line 1398:
```suggestion
    /// #[repr(C)]
```

src/lib.rs
```diff
@@ -1213,15 +1445,222 @@ pub unsafe trait TryFromBytes {
         }
     }
 
-    /// Attempts to read a `Self` from a byte slice.
+    /// Attempts to interpret the prefix of the given `candidate` as a `&mut
+    /// Self` without copying.
+    ///
+    /// If the first `size_of::<Self>()` bytes of `candidate` are a valid
+    /// instance of `Self`, this method returns both a reference to those bytes
+    /// interpreted as `Self`, and a reference to the remaining bytes. If
```
joshlf, line 1453:
```suggestion
    /// reinterpreted as a `Self`, and a reference to the remaining bytes. If
```

src/lib.rs
```diff
@@ -1213,15 +1445,222 @@ pub unsafe trait TryFromBytes {
...
+    /// `candidate.len() < size_of::<Self>()` or `candidate` is not aligned to
```
joshlf, line 1454: Same here as above wrt unsized types

src/lib.rs
```diff
@@ -1213,15 +1445,222 @@ pub unsafe trait TryFromBytes {
...
+    /// # use zerocopy_derive::*;
+    ///
+    /// // The only valid value of this type is the byte `0xC0`
+    /// #[derive(TryFromBytes, KnownLayout, Immutable)]
+    /// #[repr(u8)]
+    /// enum C0 { xC0 = 0xC0 }
+    ///
+    /// // The only valid value of this type is the bytes `0xC0C0`.
+    /// #[derive(TryFromBytes, KnownLayout, Immutable)]
+    /// #[repr(C, packed)]
```
joshlf, line 1492:
```suggestion
    /// #[repr(C)]
```

src/lib.rs
```diff
@@ -1213,15 +1445,222 @@ pub unsafe trait TryFromBytes {
...
+        util::assert_dst_is_not_zst::<Self>();
+        try_mut_from_prefix_suffix(candidate, CastType::Prefix)
+    }
+
+    /// Attempts to interpret the suffix of the given `candidate` as a `&mut
+    /// Self` without copying.
+    ///
+    /// If the last `size_of::<Self>()` bytes of `candidate` are a valid
+    /// instance of `Self`, this method returns both a reference to those bytes
+    /// interpreted as `Self`, and a reference to the preceding bytes. If
```
joshlf, line 1540:
```suggestion
    /// reinterpreted as a `Self`, and a reference to the preceding bytes. If
```

src/lib.rs
```diff
@@ -1213,15 +1445,222 @@ pub unsafe trait TryFromBytes {
...
+    /// `candidate.len() < size_of::<Self>()` or the suffix of `candidate` is
```
joshlf, line 1541: Same here as above wrt unsized types

src/lib.rs
```diff
@@ -1213,15 +1445,222 @@ pub unsafe trait TryFromBytes {
...
+    /// # use zerocopy_derive::*;
+    ///
+    /// // The only valid value of this type is the byte `0xC0`
+    /// #[derive(TryFromBytes, KnownLayout, Immutable)]
+    /// #[repr(u8)]
+    /// enum C0 { xC0 = 0xC0 }
+    ///
+    /// // The only valid value of this type is the bytes `0xC0C0`.
+    /// #[derive(TryFromBytes, KnownLayout, Immutable)]
+    /// #[repr(C, packed)]
```
joshlf, line 1579:
```suggestion
    /// #[repr(C)]
```

src/lib.rs
```diff
@@ -1213,15 +1445,222 @@ pub unsafe trait TryFromBytes {
...
+    {
+        util::assert_dst_is_not_zst::<Self>();
+        try_mut_from_prefix_suffix(candidate, CastType::Suffix).map(swap)
+    }
+
+    /// Attempts to read the given `candidate` as a `Self`.
+    ///
+    /// If the bytes of `candidate` are a valid instance of `Self`, reads those
+    /// bytes as `Self`. If `candidate.len() < size_of::<Self>()` or `candidate`
+    /// is not aligned to `align_of::<Self>()` or the bytes are not a valid
```
joshlf, line 1626: Alignment isn't checked, right?

src/lib.rs
```diff
@@ -1213,15 +1445,222 @@ pub unsafe trait TryFromBytes {
...
+    /// # use zerocopy_derive::*;
+    ///
+    /// // The only valid value of this type is the byte `0xC0`
+    /// #[derive(TryFromBytes, KnownLayout, Immutable)]
+    /// #[repr(u8)]
+    /// enum C0 { xC0 = 0xC0 }
+    ///
+    /// // The only valid value of this type is the bytes `0xC0C0`.
+    /// #[derive(TryFromBytes, KnownLayout, Immutable)]
+    /// #[repr(C, packed)]
```
joshlf, line 1642:
```suggestion
    /// #[repr(C)]
```

joshlf (changes requested):

src/lib.rs
jswrenn: This is the phrasing we use on `FromBytes` which, iirc, we selected because it's roughly correct and the actual property (especially for suffix casts) is really hard to explain. Do you have a suggestion for an alternative phrasing we should be using on these methods?

joshlf: I've updated #29 to block on this instead of blocking this PR.

src/lib.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/5cec64e2ad733d07413967e70cc0bc0be4749ebb..c24a1f2f667d619e060d75093a46196278b03ae2#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R1109)

joshlf (approved):

PR #1076, opened 2024-03-27
Make `Unalign`'s field `pub`

### LINKED ISSUE #836: Make the inner type wrapped in `Unalign` public
This is not exactly a customer request since I'm not using this crate currently, but I was looking on the docs and I wondered why won't you make the inner field of [`Unalign`](https://docs.rs/zerocopy/latest/zerocopy/struct.Unalign.html) public?

Usually when prefer to make newtype's field private, but in this case making it public has a big advantage: we cannot make a reference to it and so we cannot expose it in a non-copying way, but if it is public the compiler allows reads/writes to fields of it as long as we never take a reference and only read/write directly. This can make working with this type more comfortable.

jswrenn:
This allows for directly writing to the field safely. Since Rust ensures that unsafe uses of this field require `unsafe`, we should additionally consider deprecating many of the methods of `Unalign`.

Fixes #836

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9099,7 +9099,15 @@ mod tests {
             Unaligned,
             !FromBytes
         );
-        assert_impls!([NotZerocopy]: KnownLayout, !NoCell, !TryFromBytes, !FromZeros, !FromBytes, !IntoBytes, !Unaligned);
+        assert_impls!(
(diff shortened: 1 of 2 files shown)


joshlf:
Closing per https://github.com/google/zerocopy/issues/836#issuecomment-2023775423. We can re-open later if we decide to continue with this.

PR #1091, opened 2024-04-04
Gate implementations for Atomic* types behind #[cfg(target_has_atomic)] and Rust 1.60

### LINKED ISSUE #1086: Build failure in v0.8.0-alpha.7 on Cortex M
```
$ cargo build --target=thumbv6m-none-eabi --no-default-features
   Compiling zerocopy v0.8.0-alpha.7
error[E0412]: cannot find type `AtomicI64` in module `core::sync::atomic`
    --> /home/dhardy/.cargo/registry/src/index.crates.io-6f17d22bba15001f/zerocopy-0.8.0-alpha.7/src/util.rs:434:29
     |
434  |           core::sync::atomic::AtomicI64 [i64],
     |                               ^^^^^^^^^ help: a struct with a similar name exists: `AtomicI16`
     |
    ::: /home/dhardy-extra/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/sync/atomic.rs:3055:1
     |
3055 | / atomic_int! {
3056 | |     cfg(target_has_atomic = "16"),
3057 | |     cfg(target_has_atomic_equal_alignment = "16"),
3058 | |     stable(feature = "integer_atomics_stable", since = "1.34.0"),
...    |
3070 | |     i16 AtomicI16
3071 | | }
     | |_- similarly named struct `AtomicI16` defined here

error[E0412]: cannot find type `AtomicU64` in module `core::sync::atomic`
    --> /home/dhardy/.cargo/registry/src/index.crates.io-6f17d22bba15001f/zerocopy-0.8.0-alpha.7/src/util.rs:435:29
     |
435  |           core::sync::atomic::AtomicU64 [u64],
     |                               ^^^^^^^^^ help: a struct with a similar name exists: `AtomicU16`
     |
    ::: /home/dhardy-extra/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/sync/atomic.rs:3073:1
     |
3073 | / atomic_int! {
3074 | |     cfg(target_has_atomic = "16"),
3075 | |     cfg(target_has_atomic_equal_alignment = "16"),
3076 | |     stable(feature = "integer_atomics_stable", since = "1.34.0"),
...    |
3088 | |     u16 AtomicU16
3089 | | }
     | |_- similarly named struct `AtomicU16` defined here

For more information about this error, try `rustc --explain E0412`.
error: could not compile `zerocopy` (lib) due to 2 previous errors
```

This came up in [rand CI](https://github.com/rust-random/rand/actions/runs/8513971767/job/23350259516?pr=1425). Alpha 6 passes.

josephlr:
Fixes #1086 

Note: this makes the `Atomic*` `impl`s require Rust 1.60 or later.

The atomic `impl`s are moved to their own `atomics` module, and now traits are only implemented on types like `AtomicU32` if `#[cfg(target_has_atomic = "32")]`.

`doc_cfg` attributes are also added so that the generated documentation notes this restriction:
impl2

We also make sure to not implement `AtomicBool: FromBytes`.

This also fixes a bug in the `impl_for_transparent_wrapper!` macro where code like:
```
impl_for_transparent_wrapper!(FromBytes for AtomicBool);
```
would not emit a `UnsafeCell<bool>: FromBytes` trait bound. This resulted in #1028 being unsound. This is fixed by:
  - Simplifying the macro (no more duplicated code)
  - adding a `$ty::Inner: $trait` bound


josephlr:
@joshlf I can split out the `impl_for_transparent_wrapper!` fix into a separate PR if you want.

josephlr:
@joshlf I was eventually able to get the UI tests to pass, but it was a little annoying figuring out which command to run to regenerate them. For reference, it's:
```
TRYBUILD=overwrite ./cargo.sh +nightly test --workspace --all-features
TRYBUILD=overwrite ./cargo.sh +stable test --workspace --features=__internal_use_only_features_that_work_on_stable
```
(not `TRYBUILD=overwrite ./cargo.sh +all test --workspace` like some of the documentation suggests).

It might be worth noting the correct procedure in `CONTRIBUTING.md` or `INTERNAL.md`, or adding an easier way to run the "correct" UI tests.

josephlr:
@joshlf merge conflict has been resolved, let me know if you want me to change anything here or split the PR.

src/atomics.rs
joshlf, line 1: Can you add the copyright header (see other `.rs` files) with the year 2024?

src/atomics.rs
```diff
@@ -0,0 +1,185 @@
+//! Conditionally implement traits for atomic types
+//!
+//! Note that we only implement traits for atomics types if the corresponding
+//! [`#[cfg(target_has_atomic)]`](https://doc.rust-lang.org/reference/conditional-compilation.html#target_has_atomic)
```
joshlf, line 4: Small nit: To be consistent with comment style used elsewhere:

```suggestion
//! [`#[cfg(target_has_atomic)]`][target_has_atomic]
```

...and then, at the bottom of the comment:

```rust
/// [target_has_atomic]: https://doc.rust-lang.org/reference/conditional-compilation.html#target_has_atomic
```

src/atomics.rs
```diff
@@ -0,0 +1,185 @@
...
+//! configuration option is true.
+//!
+//! Note: On targets like `thumbv6m-none-eabi` (which has an [`AtomicU32`]
+//! type but `#[cfg(target_has_atomic = "32")]` is false), we don't implement
+//! the zerocopy traits. We will be able to handle this when Rust stabilizes
+//! [`#[cfg(target_has_atomic_load_store)]`](https://www.github.com/rust-lang/rust/issues/94039).
```
joshlf, line 10: Same here as previous comment.

joshlf (changes requested):
Apologies for taking so long to review this. I've left a few basic comments from a late-night skim, but I should have more comments once I get a chance to take a look at this with more brain cells.

Really impressed by how well you picked up on all of our idiosyncratic design patterns! This PR feels like it was written by one of the maintainers!

joshlf:
Oh one other thing: Can you squash all of your commits into one? That way you'll get to control the commit message that ends up in `main` (otherwise it's a GitHub-generated amalgam of different commit messages smushed together).

joshlf:
Taking another look at this: I think it'd be easier to review (and also to reason about after the fact if need be) if we split out the change that moves code into a new module, since that makes it so there's no clear diff for the code that moved. Would you be able to move the code back to its old location and then we can follow up with the code movement in a follow-on PR?

src/atomics.rs
josephlr: Fixed!

src/atomics.rs
josephlr: Fixed

src/atomics.rs
josephlr: No longer adding the `atomics.rs` file, so not needed.

josephlr:
> Oh one other thing: Can you squash all of your commits into one? That way you'll get to control the commit message that ends up in `main` (otherwise it's a GitHub-generated amalgam of different commit messages smushed together).

Should now be a single commit


> Would you be able to move the code back to its old location and then we can follow up with the code movement in a follow-on PR?

Code moved back to `lib.rs` and `util.rs`.

@joshlf this should be ready for another look. Note that I changed how the atomic types implement `TransparentWrapper`. Now the atomic types implement a much simpler `util::Atomic`, and then use a generic `impl<A: Atomic> TransparentWrapper`. This allows us to have simpler macros and less generated code.

Cargo.toml
```diff
@@ -35,6 +35,10 @@ exclude = [".*"]
 # From 1.61.0, Rust supports generic types with trait bounds in `const fn`.
 zerocopy-generic-bounds-in-const-fn = "1.61.0"
 
+# From 1.60.0, Rust supports detecting if a target supports atomics, so we can
```
joshlf, line 38:
```suggestion
# From 1.60.0, Rust supports detecting whether a target supports atomics, so we can
```

src/lib.rs
```diff
@@ -3799,46 +3795,96 @@ safety_comment! {
...
-    };
+/// Implements the zerocopy traits for atomic integer types.
+///
+/// Note that we only implement traits for atomics types if the corresponding
+/// [`#[cfg(target_has_atomic)]`][target_has_atomic] configuration is true.
+///
+/// Note: On targets like `thumbv6m-none-eabi` (which has an [`AtomicU32`]
+/// type but `#[cfg(target_has_atomic = "32")]` is false), we don't implement
+/// the zerocopy traits. We will be able to handle this when Rust stabilizes
+/// [`#[cfg(target_has_atomic_load_store)]`][target_has_atomic_load_store].
```
joshlf, line 3806:
```suggestion
/// TODO(#1009): On targets like `thumbv6m-none-eabi` (which has an [`AtomicU32`]
/// type but `#[cfg(target_has_atomic = "32")]` is false), we don't implement
/// the zerocopy traits. We will be able to handle this when Rust stabilizes
/// [`#[cfg(target_has_atomic_load_store)]`][target_has_atomic_load_store].
```

I've updated #1009 to track this. Feel free to suggest edits to that issue if there's more detail I should include.

src/lib.rs
```diff
@@ -3799,46 +3795,96 @@ safety_comment! {
...
+///
+/// [target_has_atomic]: https://doc.rust-lang.org/reference/conditional-compilation.html#target_has_atomic
+/// [target_has_atomic_load_store]: https://www.github.com/rust-lang/rust/issues/94039
+///
+/// # Safety
+///
+/// `$atomic` must be an atomic type with corresponding native type `$native`.
```
joshlf, line 3813: Can this be more precise? "Atomic type" here is obviously alluding to the fact that all of the atomics in `core::sync::atomic` have similar semantics, and in fact their doc comments are macro-generated under the hood. 

Maybe we could say something like:
- `$atomic` has the same size and bit validity as `$native`
- All of the bytes of `$atomic` are covered by `UnsafeCell`s

src/lib.rs
```diff
@@ -3799,46 +3795,96 @@ safety_comment! {
...
+mod atomic_8 {
+    use super::*;
+    unsafe_impl_traits_for_atomics!(AtomicBool[bool], AtomicU8[u8], AtomicI8[i8]);
+    safety_comment! {
+        /// SAFETY:
+        /// Per [1], `AtomicBool`, `AtomicU8`, and `AtomicI8` have an alignment of 1.
+        ///
+        /// [1] Per https://doc.rust-lang.org/nightly/core/sync/atomic/struct.AtomicBool.html:
+        ///
+        ///     This type has the same size, alignment, and bit validity as [the native type].
```
joshlf, line 3842:
```suggestion
        /// SAFETY:
        /// Per [1], `AtomicBool`, `AtomicU8`, and `AtomicI8` have an alignment of 1.
        ///
        /// [1] Per [2], [3], and [4]:
        ///
        ///     This type has the same size, alignment, and bit validity as [the native type].
        ///
        /// [2] https://doc.rust-lang.org/nightly/core/sync/atomic/struct.AtomicBool.html
        /// [3] https://doc.rust-lang.org/nightly/core/sync/atomic/struct.AtomicU8.html
        /// [4] https://doc.rust-lang.org/nightly/core/sync/atomic/struct.AtomicI8.html
```

I know we've been fast-and-loose with this in the past, but I'm trying to move us towards being more precise with our safety comments.

src/lib.rs
```diff
@@ -3799,46 +3795,96 @@ safety_comment! {
...
-    AtomicBool [bool],
-    AtomicI16 [i16], AtomicI32 [i32], AtomicI8 [i8], AtomicIsize [isize],
-    AtomicU16 [u16], AtomicU32 [u32], AtomicU8 [u8], AtomicUsize [usize],
-);
+#[cfg(zerocopy_atomics)]
+#[cfg(target_has_atomic = "8")]
+#[cfg_attr(doc_cfg, doc(cfg(target_has_atomic = "8")))]
+mod atomic_8 {
+    use super::*;
+    unsafe_impl_traits_for_atomics!(AtomicBool[bool], AtomicU8[u8], AtomicI8[i8]);
```
joshlf, line 3835: Safety comment, and put inside a `safety_comment! { ... }` block?

src/lib.rs
```diff
@@ -3799,46 +3795,96 @@ safety_comment! {
...
+///
+/// Note: On targets like `thumbv6m-none-eabi` (which has an [`AtomicU32`]
+/// type but `#[cfg(target_has_atomic = "32")]` is false), we don't implement
+/// the zerocopy traits. We will be able to handle this when Rust stabilizes
+/// [`#[cfg(target_has_atomic_load_store)]`][target_has_atomic_load_store].
+///
+/// [target_has_atomic]: https://doc.rust-lang.org/reference/conditional-compilation.html#target_has_atomic
+/// [target_has_atomic_load_store]: https://www.github.com/rust-lang/rust/issues/94039
+///
+/// # Safety
```
joshlf, line 3811: Can we add code which is unsafe to call so that, if you don't call this from inside `safety_comment! { ... }`, you'll get a compilation failure?

src/lib.rs
```diff
@@ -3799,46 +3795,96 @@ safety_comment! {
...
+        unsafe_impl!(AtomicI8: Unaligned);
+    }
+    assert_unaligned!(AtomicBool, AtomicU8, AtomicI8);
+}
+#[cfg(zerocopy_atomics)]
+#[cfg(target_has_atomic = "16")]
+#[cfg_attr(doc_cfg, doc(cfg(target_has_atomic = "16")))]
+mod atomic_16 {
+    use super::*;
+    unsafe_impl_traits_for_atomics!(AtomicU16[u16], AtomicI16[i16]);
```
joshlf, line 3854: Safety comment, and put inside a `safety_comment! { ... }` block?

src/lib.rs
```diff
@@ -3799,46 +3795,96 @@ safety_comment! {
...
+}
+#[cfg(zerocopy_atomics)]
+#[cfg(target_has_atomic = "32")]
+#[cfg_attr(doc_cfg, doc(cfg(target_has_atomic = "32")))]
+mod atomic_32 {
+    use super::*;
+    unsafe_impl_traits_for_atomics!(AtomicU32[u32], AtomicI32[i32]);
```
joshlf, line 3861: Safety comment, and put inside a `safety_comment! { ... }` block?

src/lib.rs
```diff
@@ -3799,46 +3795,96 @@ safety_comment! {
...
+}
+#[cfg(zerocopy_atomics)]
+#[cfg(target_has_atomic = "64")]
+#[cfg_attr(doc_cfg, doc(cfg(target_has_atomic = "64")))]
+mod atomic_64 {
+    use super::*;
+    unsafe_impl_traits_for_atomics!(AtomicU64[u64], AtomicI64[i64]);
```
joshlf, line 3868: Safety comment, and put inside a `safety_comment! { ... }` block?

src/lib.rs
```diff
@@ -3799,46 +3795,96 @@ safety_comment! {
...
+}
+#[cfg(zerocopy_atomics)]
+#[cfg(target_has_atomic = "ptr")]
+#[cfg_attr(doc_cfg, doc(cfg(target_has_atomic = "ptr")))]
+mod atomic_ptr {
+    use super::*;
+    unsafe_impl_traits_for_atomics!(AtomicUsize[usize], AtomicIsize[isize]);
```
joshlf, line 3875: Safety comment, and put inside a `safety_comment! { ... }` block?

src/lib.rs
```diff
@@ -3799,46 +3795,96 @@ safety_comment! {
...
-    ///     Alignment is measured in bytes, and must be at least 1.
-    ///
-    /// [3] Per https://doc.rust-lang.org/reference/type-layout.html#size-and-alignment:
-    ///
-    ///     The size of a value is always a multiple of its alignment.
-    unsafe_impl!(AtomicBool: Unaligned);
-    unsafe_impl!(AtomicU8: Unaligned);
-    unsafe_impl!(AtomicI8: Unaligned);
-    assert_unaligned!(AtomicBool, AtomicU8, AtomicI8);
+    /// See the comments on the `*mut T` impls for more information.
```
joshlf, line 3881: This safety comment smells to me. We shouldn't have to talk about `FromBytes`/`IntoBytes` in order to justify this impl - we should only need to reference the safety invariants on `util::Atomic`.

src/util.rs
```diff
@@ -330,110 +326,66 @@ unsafe impl<T, I: Invariants> TransparentWrapper<I> for Unalign<T> {
...
-
-        fn cast_from_inner(ptr: *mut UnsafeCell<$native>) -> *mut $atomic {
-            // SAFETY: Per [1] (from comment on impl block), `$atomic` has the
-            // same size as `$native`. Thus, this cast preserves size.
-            //
-            // This cast trivially preserves provenance.
-            ptr.cast::<$atomic>()
-        }
-    };
+/// `Self` must be an atomic type whose native equivalent is `Self::Native`.
```
joshlf, line 333: Same here as in `lib.rs` regarding the safety invariant. Ideally this safety invariant should be sufficient to prove that the following impl (`unsafe impl<A: Atomic, I: Invariants> TransparentWrapper<I> for A`) is sound without reference to the stdlib documentation.

I know that I'm asking for a higher quality bar than we have on `main` already 🙂  This refactor is making it so that the safety reasoning is more "at a distance", so I'm a bit more worried about making the safety reasoning precise and watertight.

joshlf (changes requested):
Awesome, thank you for reworking this! In general this looks awesome. I'm going to try to push hard on how we express and propagate our safety invariants, but the rest of the PR looks great. This is definitely nicer than what we have currently.

djkoloski:
In the process of rebasing and updating this PR, I discovered that some unsoundness in the implementation. The `TransparentWrapper` trait requires that the implementing type has the same layout as its `Inner` type. All atomics have the same size and bit validity as their non-atomic counterparts, but the alignments of atomics are always equal to their sizes which may differ from their non-atomic counterparts. As a concrete example, on 32-bit platforms which support 64-bit atomics, the align of `u64` is likely 4 but the align of `AtomicU64` must be 8.

joshlf:
I've carved the soundness fix for `impl_for_transparent_wrapper!` into #1585 so we can hopefully land that separately. My intention is to incrementally carve out different parts of this so we don't have to review everything at once.

joshlf:
I've carved out the `target_has_atomic` fix into #1586.

joshlf:
> In the process of rebasing and updating this PR, I discovered that some unsoundness in the implementation. The `TransparentWrapper` trait requires that the implementing type has the same layout as its `Inner` type. All atomics have the same size and bit validity as their non-atomic counterparts, but the alignments of atomics are always equal to their sizes which may differ from their non-atomic counterparts. As a concrete example, on 32-bit platforms which support 64-bit atomics, the align of `u64` is likely 4 but the align of `AtomicU64` must be 8.

IIUC, this isn't quite true - `T: TransparentWrapper` requires that `T` and `T::Inner` share some aspects of layout, but which aspects in particular are controlled by the associated `T::XxxVariance` types. In this case, [`AlignmentVariance` is set to `Invariant`](https://github.com/google/zerocopy/pull/1091/files#diff-4007187965c1fea4fd252a76f5f1007b2ea59f6d86e2f454edee547c5036be24R363), which means that no guarantees are made regarding the alignment of `T` relative to the alignment of `T::Inner`.

joshlf:
Closing this now that the important changes have landed in #1585 and #1586. We may still carve out the addition of the `Atomic` trait - this is tracked by #1592.

Thanks again for all of your work on this, and for finding and fixing the unsoundness in `impl_for_transparent_wrapper!` along the way!

PR #1119, opened 2024-04-18
Rename `TryFromBytes` methods to be consistent with `FromBytes`

jswrenn:
For example, `try_from_ref` is now `try_ref_from`.

See: https://github.com/google/zerocopy/discussions/1095

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1479,7 +1479,7 @@ pub use zerocopy_derive::TryFromBytes;
 /// but for which `TryFromBytes` is not guaranteed to accept all byte sequences
 /// produced by `IntoBytes`. In other words, for some `T: TryFromBytes +
 /// IntoBytes`, there exist values of `t: T` such that
-/// `TryFromBytes::try_from_ref(t.as_bytes()) == None`. Code should not
+/// `TryFromBytes::try_ref_from(t.as_bytes()) == None`. Code should not
 /// generally assume that values produced by `IntoBytes` will necessarily be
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1496,14 +1496,14 @@ pub use zerocopy_derive::TryFromBytes;
...
-/// [`try_from_ref`]: TryFromBytes::try_from_ref
+/// [`try_ref_from`]: TryFromBytes::try_ref_from
 /// [*valid instance*]: #what-is-a-valid-instance
 #[cfg_attr(feature = "derive", doc = "[derive]: zerocopy_derive::TryFromBytes")]
 #[cfg_attr(
     not(feature = "derive"),
     doc = concat!("[derive]: https://docs.rs/zerocopy/", env!("CARGO_PKG_VERSION"), "/zerocopy/derive.TryFromBytes.html"),
 )]
-// TODO(#5): Update `try_from_ref` doc link once it exists
+// TODO(#5): Update `try_ref_from` doc link once it exists
```
joshlf, line 1506: Is this TODO done? Can we remove it while we're here?

joshlf (changes requested):
Can you add to the commit message:

> Makes progress on #5

src/lib.rs

joshlf:
> Can you add to the commit message:
> 
> > Makes progress on #5

Still need to do this ^

jswrenn:
Whoops, missed that. Done.

joshlf (approved):

Issue #1139, opened 2024-04-24
Error Handling in 0.8


jswrenn:
## Overview
Zerocopy's deserialization methods will return `Result` instead of `Option`. The `Err` type of the `Result` will provide:
- the cause of the deserialization error
- access to the un-deserializable bytes 

### Providing better error messages

As it stands, `.unwrap()` presently produces an uninformative error message. Using a custom error type will allow us to provide precise diagnostic text.

### Providing programatic access to the cause of the failure

The bytemuck crate provides this information with its [`PodCastError`](https://docs.rs/bytemuck/latest/bytemuck/enum.PodCastError.html) type. A [cursory search of Github](https://github.com/search?q=PodCastError%3A%3ATargetAlignmentGreaterAndInputNotAligned+language%3ARust&type=code&l=Rust) suggests that this fidelity is useful to some customers.

### Providing the original bytes

Our current API does not always provide access to the original bytes in the case of deserialization failure. For instance, attempting this:

```rust
loop {
    let Some((des, rem)) = T::mut_from_prefix(buffer) else {
        panic!("could not parse from {buffer:?}");
    };
    buffer = rem;
}
```

...produces this error message:

```text
error[E0502]: cannot borrow `buffer` as immutable because it is also borrowed as mutable
 --> src/main.rs:9:38
  |
8 |     let Some((d, remainder)) = mut_from_prefix_split::<u8>(buffer) else {
  |                                                            ------ mutable borrow occurs here
9 |         panic!("could not parse from {buffer:?}");
  |                                      ^^^^^^^^^^
  |                                      |
  |                                      immutable borrow occurs here
  |                                      mutable borrow later used here
```

## Steps
- [x] Implement Proposal
  - #1144
- [x] Make `Unalign`'s methods return `Result<_, AlignmentError<_, _>>` 
  - #1198
- [x] Resolve undecided API issues, either:
  - Make `ConvertError` public, or
  - Evolve the API towards #1150

PR #1144, opened 2024-04-26
Overhaul error reporting (enum version)

jswrenn:
This is an implementation of the ideas presented in #1139.

This PR:
1. introduces the public `error` module (whose items are also re-exported)
2. revises zerocopy's APIs to return `Result`s with informative error types, rather than `Option`s

## Fundamental Error Types

Generally speaking, zerocopy’s conversions may fail for one of up to three reasons, each reflected in a fundamental error type:
-  `AlignmentError<Src, Dst>`: the conversion source was improperly aligned
- `SizeError<Src, Dst>`: the conversion source was of incorrect size
- `ValidityError<Src, Dst>`: the conversion source contained invalid data

These fundamental error types are parameterized with the failed conversion's:
- `Src`: the input value of the conversion, stored within the error; and
- `Dst`: the output type of the conversion, retained as phantom type for error reporting purposes.

Methods that only have one failure mode return that mode's corresponding error type directly. For example, `Ref::new_unaligned`'s only failure mode is a size mismatch. This PR revises its return type from `Option<Ref<B, T>>` to `Result<Ref<B, T>, SizeError<B, T>>`. 

The input value of the conversion is retrievable from each of these fundamental types with an `into_src` method.

## Compound Error Types

Conversion methods that have either two or three possible failure modes return one of these error types:
- `CastError<Src, Dst>`: the error type of ref-conversions.
- `TryCastError<Src, Dst>`: the error type of fallible ref-conversions.
- `TryReadError<Src, Dst>`: the error type of fallible read-conversions.

The input value of the conversion is retrievable from each of these compound types with an `into_src` method.

These three types are parameterized aliases for the enum `ConvertError<A, S, V>`:
```rust
pub enum ConvertError<A, S, V> {
    Alignment(A),
    Size(S),
    Validity(V),
}
```
(This enum is never used directly in zerocopy's public API, and is only referred to via these three parameterized aliases.)

These aliases parameterize their reachable branches of `ConvertError` with the appropriate fundamental error type, and their unreachable branches with `Infallible`.

## Shared Interfaces

Both the compound and fundamental error types provide a shared set of affordances; they:

- provide access to the underlying `Src`
- provide implementations of `Debug`, `PartialEq` and `Eq`
- provide identical implementations of `Display`

## User Experience Examples

Accessing the source and treating the error as opaque:
```rust
let packet = match Packet::ref_from(source) {
    Ok((packet, suffix)) => {
        source = suffix;
        packet
    },
    Err(e) => {
        tracing::error!("parse error: {}", e);
        source = e.into_src();
    }
};
```

Accessing the source and reflecting on the error:
```rust
let packet = match Packet::try_mut_from_prefix(source) {
    Ok((packet, suffix)) => {
        source = suffix;
        packet
    },
    Err(TryCastError::Alignment(e)) => {
        panic!("bug: {}", e)
    },
    Err(TryCastError::Size(e)) => {
        tracing::trace!("waiting for additional bytes: {}", e);
        source = e.into_src();
        continue; // wait for more bytes
    },
    Err(TryCastError::Validity(e)) => {
        return MyError::InvalidPacket(e);
    },
};
```

### Future Outlook

The stabilization of `min_exhaustive_patterns` (https://github.com/rust-lang/rust/issues/119612) will allow omitting infallible variants from matches.

#### Surface Area

This PR introduces seven public types to support error reporting. By contrast, #1150 introduces eleven.

#### API Complexity

In comparison to #1150, this PR slightly pessimizes access to the underlying byte slice: it provides by-value access via a `into_src()` method rather than via `.src`. By contrast, this PR comparatively optimizes access to failure mode: it can be determined by matching directly on the returned error, as opposed to first accessing the failure mode via `.err`.

#### Conceptual Complexity

With the possible exception of `ConvertError`, this PR's types all implement a [common, consistent interface](https://github.com/google/zerocopy/pull/1144#:~:text=Shared%20Interfaces): they all provide access to the kind of failure that occurred, access to underlying byte slice, and consistently rich error messages.

By contrast, the capabilities provided by #1150's types vary depending on the type. Access to the source value is provided by `Error`, `AlignmentError`, `SizeError`, `ValidityError`, `CastError`, `TryCastError` and `TryReadError`, but *not* by `ConvertError`, `Alignment`, `Size`, or `Validity`.

The fidelity of the `Display` implementations vary by type: `Error`, `AlignmentError`, `SizeError`, `ValidityError`, `CastError`, `TryCastError` and `TryReadError` provide high-fidelity error messages, while `ConvertError`, `Alignment`, `Size`, or `Validity` provide low fidelity error messages.

## Future Work

The return type of `try_transmute!` (which has not yet been implemented) should be `Result<Src, ValidityError<Src, Dst>>`.

The type `ConvertError` is not used directly in any public APIs and is defined in a private module to remove it and its variants (which are also marked `#[doc(hidden)]` from our SemVer obligations. Before `0.8`, we should strive to make these items `pub` so that we can demonstrate matching on error kind in our documentation and release notes.

Makes progress towards #1139.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,0 +1,463 @@
+//! Types related to error reporting.
+//!
+//! ## Single failure mode errors
+//!
+//! Generally speaking, zerocopy's conversions may fail for one of up to three reasons:
+//! - [`AlignmentError`]: the conversion source was improperly aligned
+//! - [`SizeError`]: the conversion source was of incorrect size
+//! - [`ValidityError`]: the conversion source contained invalid data
+//!
+//! Methods that only have one failure mode, like [`Ref::new_unaligned`], return
+//! that mode's corresponding error type directly.
+//!
+//! ## Compound errors
+//!
+//! Conversion methods that have either two or three possible failure modes
+//! return one of these error types:
+//! - [`CastError`]: the error type of reference conversions
+//! - [`TryCastError`]: the error type of fallible reference conversions
+//! - [`TryReadError`]: the error type of fallible read conversions
+//!
+//! ## Accessing the conversion source
+//!
+//! All error types provide an `into_src` method that converts the error into
+//! the source value underlying the failed conversion.
+
+use core::{convert::Infallible, fmt, marker::PhantomData, ops::Deref};
+
+use crate::TryFromBytes;
+#[cfg(doc)]
+use crate::{FromBytes, Ref};
+
+// This is private to remove `ConvertError` from our SemVer obligations for the
+// time being.
+// TODO(#1139): Remove this wrapping `private` module.
+mod private {
+    #[cfg(doc)]
+    use super::*;
+
+    /// Zerocopy's generic error type.
+    ///
+    /// Generally speaking, zerocopy's conversions may fail for one of up to three reasons:
+    /// - [`AlignmentError`]: the conversion source was improperly aligned
+    /// - [`SizeError`]: the conversion source was of incorrect size
+    /// - [`ValidityError`]: the conversion source contained invalid data
+    ///
+    /// However, not all conversions produce all errors. For instance,
+    /// [`FromBytes::ref_from`] may fail due to alignment or size issues, but not
+    /// validity issues. This generic error type captures these (im)possibilities
+    /// via parameterization: `A` is parameterized with [`AlignmentError`], `S` is
+    /// parameterized with [`SizeError`], and `V` is parameterized with
+    /// [`Infallible`].
+    ///
+    /// Zerocopy never uses this type directly in its API. Rather, we provide three
+    /// pre-parameterized aliases:
+    /// - [`CastError`]: the error type of reference conversions
+    /// - [`TryCastError`]: the error type of fallible reference conversions
+    /// - [`TryReadError`]: the error type of fallible read conversions
+    #[derive(PartialEq, Eq)]
+    pub enum ConvertError<A, S, V> {
+        /// The conversion source was improperly aligned.
+        #[doc(hidden)]
+        Alignment(A),
+        /// The conversion source was of incorrect size.
+        #[doc(hidden)]
+        Size(S),
+        /// The conversion source contained invalid data.
+        #[doc(hidden)]
+        Validity(V),
+    }
+}
+
+use private::ConvertError;
+
+impl<A: fmt::Debug, S: fmt::Debug, V: fmt::Debug> fmt::Debug for ConvertError<A, S, V> {
+    #[inline]
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Alignment(e) => f.debug_tuple("Alignment").field(e).finish(),
+            Self::Size(e) => f.debug_tuple("Size").field(e).finish(),
+            Self::Validity(e) => f.debug_tuple("Validity").field(e).finish(),
+        }
+    }
+}
+
+/// Produces a human-readable error message.
+impl<A: fmt::Display, S: fmt::Display, V: fmt::Display> fmt::Display for ConvertError<A, S, V> {
+    #[inline]
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Alignment(e) => e.fmt(f),
+            Self::Size(e) => e.fmt(f),
+            Self::Validity(e) => e.fmt(f),
+        }
+    }
+}
+
+/// The error emitted if the conversion source is improperly aligned.
+#[derive(PartialEq, Eq)]
+pub struct AlignmentError<Src, Dst: ?Sized> {
+    /// The source value involved in the conversion.
+    src: Src,
+    /// The inner destination type inolved in the conversion.
+    dst: PhantomData<Dst>,
+}
+
+impl<Src, Dst: ?Sized> AlignmentError<Src, Dst> {
+    pub(crate) fn new(src: Src) -> Self {
+        Self { src, dst: PhantomData }
+    }
+
+    /// Produces the source underlying the failed conversion.
+    #[inline]
+    pub fn into_src(self) -> Src {
+        self.src
+    }
+
+    pub(crate) fn with_src<NewSrc>(self, new_src: NewSrc) -> AlignmentError<NewSrc, Dst> {
+        AlignmentError { src: new_src, dst: PhantomData }
+    }
+
+    pub(crate) fn map_src<NewSrc>(self, f: impl Fn(Src) -> NewSrc) -> AlignmentError<NewSrc, Dst> {
+        AlignmentError { src: f(self.src), dst: PhantomData }
+    }
+
+    pub(crate) fn into<S, V>(self) -> ConvertError<Self, S, V> {
+        ConvertError::Alignment(self)
+    }
+}
+
+impl<Src, Dst: ?Sized> fmt::Debug for AlignmentError<Src, Dst> {
+    #[inline]
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("AlignmentError").finish()
+    }
+}
+
+/// Produces a human-readable error message.
+// The bounds on this impl are intentionally conservative, and can be relaxed
+// either once a `?Sized` alignment accessor is stabilized, or by storing the
+// alignment as a runtime value.
+impl<Src, Dst> fmt::Display for AlignmentError<Src, Dst>
+where
+    Src: Deref,
+{
+    #[inline]
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        #[cfg_attr(__INTERNAL_USE_ONLY_NIGHTLY_FEATURES_IN_TESTS, allow(lossy_provenance_casts))]
+        #[allow(clippy::as_conversions)]
+        let addr = self.src.deref() as *const _ as *const () as usize;
+        let addr_align = 2usize.pow(addr.trailing_zeros());
+        f.write_str("the conversion failed because the address of the source (a multiple of ")?;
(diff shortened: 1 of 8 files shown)


src/lib.rs
```diff
@@ -1642,23 +1651,29 @@ pub unsafe trait TryFromBytes {
...
+                // fix before returning.
+                //
+                // Note that one panic or post-monomorphization error condition is
+                // calling `try_into_valid` (and thus `is_bit_valid`) with a shared
+                // pointer when `Self: !Immutable`. Since `Self: Immutable`, this panic
+                // condition will not happen.
+                match candidate.try_into_valid() {
+                    Ok(candidate) => Ok(candidate.as_mut()),
+                    Err(e) => {
+                        Err(e.map_src(|src| src.as_bytes().as_mut()).into()) // TODO: FIXME
```
jswrenn, line 1671: In many places of this PR, returning the remaining bytes required restructuring the function body so the borrow checker can reason better about control flow.

This is the one place where I've yet to figure out how to make the borrow checker happy. The code here doesn't violate the borrow checker, but it fails to compile because `.as_bytes()` is only callable on `Sized` referents. We could make `.as_bytes()` work on `?Sized` referents with `size_of_val_raw`, but that's still yet to be stabilized.

src/error.rs
```diff
@@ -0,0 +1,395 @@
...
+// either once a `?Sized` alignment accessor is stabilized, or by storing the
+// alignment as a runtime value.
+impl<Src, Dst> fmt::Display for AlignmentError<Src, Dst>
+where
+    Src: Deref,
+{
+    #[inline]
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str("the conversion failed because the address of the source (")?;
+        f.write_fmt(format_args!("{:p}", self.src.deref()))?;
```
joshlf, line 143: Leaking the pointer might be considered security sensitive by some users. It can reveal aspects of the program's execution and help with memory corruption exploits (e.g. to help an attacker know where their attack payload will land in memory). We might want to omit this.

jswrenn: Good call!

jswrenn: Done!

jswrenn:
@joshlf, let's take another look at this and #1150 soon. I've added some examples to this PR's description illustrating what it feels like to use this API, and some discussion of why I think it compares favorably to #1150 in both API and cognitive complexity.

src/error.rs
```diff
@@ -0,0 +1,391 @@
...
+        #[doc(hidden)]
+        Alignment(A),
+        /// The conversion source was of incorrect size.
+        #[doc(hidden)]
+        Size(S),
+        /// The conversion source contained invalid data.
+        #[doc(hidden)]
+        Validity(V),
+    }
+}
```
jswrenn, line 66: @joshlf I've gone ahead and encapsulated the `ConvertError` in a private struct, so the public API consists solely of:
-  `AlignmentError<Src, Dst>`: the conversion source was improperly aligned
- `SizeError<Src, Dst>`: the conversion source was of incorrect size
- `ValidityError<Src, Dst>`: the conversion source contained invalid data
- `CastError<Src, Dst>`: the compound error type of ref-conversions
- `TryCastError<Src, Dst>`: the compound error type of fallible ref-conversions
- `TryReadError<Src, Dst>`: the compound error type of fallible read-conversions

src/error.rs
joshlf, line 1: Can you add tests for the `Display` impls (mostly so we have a place where we can write out how they're formatted - especially so we can see a diff if we change the format in the future)?

src/lib.rs
```diff
@@ -281,6 +281,11 @@ mod macros;
 
 pub mod byteorder;
 mod deprecated;
+// Thid module is `pub` so that zerocopy's error types and error handling
```
joshlf, line 284:
```suggestion
// This module is `pub` so that zerocopy's error types and error handling
```

src/pointer/ptr.rs
```diff
@@ -992,17 +997,36 @@ mod _casts {
...
+        }
+    }
+
+    impl<'a, T, I> Ptr<'a, T, I>
+    where
+        T: 'a + ?Sized,
+        I: Invariants<Validity = Initialized>,
+        T: Immutable,
+    {
+        /// Casts this pointer-to-initialized into a pointer-to-bytes.
```
joshlf, line 1012: Safety precondition?

src/lib.rs
```diff
@@ -1111,24 +1119,40 @@ pub unsafe trait TryFromBytes {
...
+                        .map_src(|src| {
+                            // SAFETY: The provided `size`, below, is computed from
+                            // `bytes.len()`. By contract on `try_into_valid` we can
+                            // count on `src` being equal to `candidate`. Since
+                            // `candidate` was derived from the entire extent of
+                            // `bytes`, the size of `bytes` is exactly equal to the
+                            // size of `src`.
+                            let bytes = unsafe { src.as_bytes_with_size(size) };
+                            bytes.as_mut()
+                        })
```
joshlf, line 1150: It feels sketchy to me that we have to make this implementation so much messier (esp with introducing new `unsafe`) than it was before just to support this change to our API. I wonder if there's something we can do to refactor this so it stays reasonable?

src/pointer/ptr.rs
```diff
@@ -845,10 +845,15 @@ mod _transitions {
         ///
         /// This method will panic if
         /// [`T::is_bit_valid`][TryFromBytes::is_bit_valid] panics.
+        ///
+        /// # Safety
+        ///
+        /// Unsafe code may rely on this method's returned `ValidityError`
+        /// containing `self`.
```
joshlf, line 852:
```suggestion
        /// On error, unsafe code may rely on this method's returned `ValidityError`
        /// containing `self`.
```

src/pointer/ptr.rs
```diff
@@ -1081,7 +1105,7 @@ mod _casts {
     /// alignment of `[u8]` is 1.
     impl<'a, I> Ptr<'a, [u8], I>
     where
-        I: Invariants<Validity = Valid>,
+        I: Invariants<Alignment = Aligned, Validity = Valid>,
```
joshlf, line 1108: Why?

joshlf (changes requested):

src/lib.rs
jswrenn: Perhaps; I'll poke at this a bit more. This is all done to work around our `as_bytes` method having a `Sized` bound. We'll be able to remove that bound once `std::mem::size_of_val_raw` is stabilized, but until then, the workaround is to stash the size while we have it. I don't think it's likely we'll be able to avoid the new `unsafe` here.

src/error.rs
```diff
@@ -0,0 +1,391 @@
...
+//! Methods that only have one failure mode, like [`Ref::new_unaligned`], return
+//! that mode's corresponding error type directly.
+//!
+//! ## Compound errors
+//!
+//! Conversion methods that have either two or three possible failure modes
+//! return one of these error types:
+//! - [`CastError`]: the error type of ref-conversions.
+//! - [`TryCastError`]: the error type of fallible ref-conversions.
+//! - [`TryReadError`]: the error type of fallible read-conversions.
```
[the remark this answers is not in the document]
jswrenn: [Done](https://github.com/google/zerocopy/compare/fcaf8242f32f4efcfd5a8aaddb048a287d0ff485..d45389beb7f929e4cfa2770a5f33b18145bcc508#diff-97e25e2a0e41c578875856e97b659be2719a65227c104b992e3144efa000c35eR17)

src/error.rs
```diff
@@ -0,0 +1,391 @@
...
+//!
+//! All error types provide an `into_src` method that converts the error into
+//! the source value underlying the failed conversion.
+
+use core::{convert::Infallible, fmt, marker::PhantomData, ops::Deref};
+
+#[cfg(doc)]
+use crate::{FromBytes, Ref, TryFromBytes};
+
+mod private {
```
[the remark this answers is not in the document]
jswrenn: [Done](https://github.com/google/zerocopy/compare/fcaf8242f32f4efcfd5a8aaddb048a287d0ff485..d45389beb7f929e4cfa2770a5f33b18145bcc508#diff-97e25e2a0e41c578875856e97b659be2719a65227c104b992e3144efa000c35eR32)

src/pointer/ptr.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/fcaf8242f32f4efcfd5a8aaddb048a287d0ff485..d45389beb7f929e4cfa2770a5f33b18145bcc508#diff-86b9160ad9b43e7e08657e37fd78b8666d627607c29d8a24f8cc549da6e54b57R1014)

src/error.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/fcaf8242f32f4efcfd5a8aaddb048a287d0ff485..d45389beb7f929e4cfa2770a5f33b18145bcc508#diff-97e25e2a0e41c578875856e97b659be2719a65227c104b992e3144efa000c35eR414)

src/lib.rs
```diff
@@ -4946,9 +5045,14 @@ where
...
-    pub fn new_unaligned(bytes: B) -> Option<Ref<B, T>> {
+    pub fn new_unaligned(bytes: B) -> Result<Ref<B, T>, SizeError<B, T>> {
         util::assert_dst_is_not_zst::<T>();
-        Ref::new(bytes)
+        match Ref::new(bytes) {
+            Ok(dst) => Ok(dst),
+            Err(CastError::Size(e)) => Err(e),
+            Err(CastError::Alignment(_)) => unreachable!(),
+            Err(CastError::Validity(i)) => match i {},
+        }
```
joshlf, line 5055: We're doing this a lot; any way we can factor this out somehow?

joshlf (changes requested):

joshlf (approved):

PR #1150, opened 2024-04-29
Overhaul error reporting (struct version)

joshlf:
Alternative to https://github.com/google/zerocopy/pull/1144 with a struct as the top-level error type, and the source and destination fields in this struct instead of in enum variants.


joshlf:
Closing in favor of https://github.com/google/zerocopy/pull/1144

Issue #1149, opened 2024-04-29
Consider compiler-erroring instead of panicking in DST APIs


jswrenn:
*This issue goes partway towards addressing #325.*

For example, `Ref::from_prefix_with_trailing_elements` documents a panic if `Self` is a ZST. However, the ZST-iness (pronounced "zestiness") of a type relevant here is always a static property and can be reported via postmonomorphization error.

The advantages of this are several-fold:
- it allows us to maintain the consistent policy that our APIs do not panic (with the exception of panics in *user*-supplied `TryFromBytes` validators)
- it is forwards compatible with us relaxing the compile error to a panic in future non-breaking releases
- it may have optimizer advantages 

Postmonomorphization errors currently have a bit of a 'weirdness' factor (and an unwieldy name), but I suspect that will change. The recent [stabilization notes for inline const exprs](https://github.com/rust-lang/rust/pull/104087) cite ZST-checking as a motivating case of what the feature will make possible.

joshlf:
Note that this partially addresses #325. I think we should leave #325 open and continue the main discussion there, but it's good to have this as a separate issue so we can block #671 on it.

PR #1154, opened 2024-04-30
Use compile-time assertions to deny ZSTy DSTs

jswrenn:
Presently, we deny ZSTy DSTs in our APIs via panicking at runtime. However, the ZSTiness of a DST is statically detectable and can be denied instead at compile time. This PR replaces our ZSTy DST panics with compile-time assertions. Doing gives us the freedom later provide meaningful runtime semantics in such cases.

Partially addresses https://github.com/google/zerocopy/issues/325
Closes https://github.com/google/zerocopy/issues/1149

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1609,12 +1609,33 @@ pub unsafe trait TryFromBytes {
     /// there exist types whose bit validity is ambiguous. See
     /// [here][TryFromBytes#what-is-a-valid-instance] for a discussion of how
     /// these cases are handled.
+    ///
+    /// # Compile-Time Assertions
+    ///
+    /// This method cannot yet be used on unsized types whose dynamically-sized
+    /// component is zero-sized. Attempting to use this method on such types
+    /// results in a compile-time assertion error; e.g.:
+    ///
+    /// ```compile_fail,E0080
+    /// use zerocopy::*;
+    /// # use zerocopy_derive::*;
+    ///
+    /// #[derive(Immutable, KnownLayout, TryFromBytes)]
+    /// #[repr(C)]
+    /// struct ZSTy {
+    ///     leading_sized: u16,
+    ///     trailing_dst: [()],
+    /// }
+    ///
+    /// let _ = ZSTy::try_ref_from(0u16.as_bytes()); // ⚠ Compile Error!
+    /// ```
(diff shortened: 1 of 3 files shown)


src/util.rs
```diff
@@ -589,6 +589,22 @@ pub(crate) const fn min(a: NonZeroUsize, b: NonZeroUsize) -> NonZeroUsize {
...
 
+pub(crate) const fn assert_dst_is_not_zst<T>()
+where
+    T: crate::KnownLayout + ?Sized
+{
+    trait ConstAssert: crate::KnownLayout {
+        const DST_IS_NOT_ZST: bool = {
+            assert!(!Self::DST_IS_ZST);
+            !Self::DST_IS_ZST
+        };
```
joshlf, line 600: Couldn't we just do this based on the existing `KnownLayout::LAYOUT` const rather than introducing new machinery in `KnownLayout`?

joshlf (changes requested):

src/util.rs
jswrenn: Done

src/lib.rs
```diff
@@ -1609,12 +1609,33 @@ pub unsafe trait TryFromBytes {
     /// there exist types whose bit validity is ambiguous. See
     /// [here][TryFromBytes#what-is-a-valid-instance] for a discussion of how
     /// these cases are handled.
+    ///
+    /// # Compile-Time Assertions
+    ///
+    /// This method cannot yet be used on unsized types whose dynamically-sized
```
joshlf, line 1615:
```suggestion
    /// This method cannot be used on unsized types whose dynamically-sized
```

IMO we shouldn't imply that we'll lift this restriction in the future. We might, but it's not guaranteed, and it's also pointless cognitive burden for users to think about our long-term plans when there's no specific timeline and so there's no meaningful difference between "you can't yet" and "you can't."

Same elsewhere in this PR.

joshlf (changes requested):

src/lib.rs
jswrenn: We don't want folks relying on this function as a compile-time assert on DST ZSTiness so the "yet" is load-bearing.

joshlf (approved):

PR #1160, opened 2024-05-01
Ptr::try_cast_into takes explicit element count 

joshlf:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -545,7 +545,7 @@ pub unsafe trait KnownLayout {
 
 /// The metadata associated with a [`KnownLayout`] type.
 #[doc(hidden)]
-pub trait PointerMetadata {
+pub trait PointerMetadata: Copy + Eq + Debug {
     /// Constructs a `Self` from an element count.
     ///
     /// If `Self = ()`, this returns `()`. If `Self = usize`, this returns
@@ -598,7 +598,9 @@ impl PointerMetadata for usize {
                 let without_padding = offset.checked_add(slice_len)?;
                 without_padding.checked_add(util::padding_needed_for(without_padding, layout.align))
             }
-            SizeInfo::Sized { .. } => unreachable!(),
+            // NOTE: This branch is unreachable, but we return `None` rather
+            // than `unreachable!()` to avoid generating panic paths.
+            SizeInfo::Sized { .. } => None,
         }
     }
 }
@@ -1177,7 +1179,7 @@ pub unsafe trait TryFromBytes {
         Self: KnownLayout + Immutable,
     {
         util::assert_dst_is_not_zst::<Self>();
-        match Ptr::from_ref(candidate).try_cast_into_no_leftover::<Self, BecauseImmutable>() {
+        match Ptr::from_ref(candidate).try_cast_into_no_leftover::<Self, BecauseImmutable>(None) {
             Ok(candidate) => {
                 // This call may panic. If that happens, it doesn't cause any soundness
                 // issues, as we have not generated any invalid state which we need to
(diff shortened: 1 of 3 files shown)


src/pointer/ptr.rs
```diff
@@ -1128,7 +1142,17 @@ mod _casts {
...
+
             let elems = <U as KnownLayout>::PointerMetadata::from_elem_count(elems);
+            // For a slice DST type, if `meta` is `Some(elems)`, then we
+            // synthesize `layout` to describe a sized type whose size is equal
+            // to the size of the instance that we are asked to cast. For sized
+            // types, `validate_cast_and_convert_metadata` returns `elems == 0`.
+            // Thus, in this case, we need to use the `elems` passed by the
+            // caller, not the one returned by
+            // `validate_cast_and_convert_metadata`.
+            let elems = meta.unwrap_or(elems);
```
joshlf, line 1154: This feels super sketchy. Is there a better way to structure this?

src/lib.rs
```diff
@@ -2697,6 +2592,34 @@ pub unsafe trait FromBytes: FromZeros {
...
+    util::assert_dst_is_not_zst::<T>();
+    let (slf, prefix_suffix) = Ptr::from_mut(bytes)
+        .try_cast_into(cast_type, meta)
+        .map_err(|err| err.map_src(|s| s.as_mut()))?;
+    Ok((slf.bikeshed_recall_valid().as_mut(), prefix_suffix.as_mut()))
+}
+
+fn swap<T, U>((t, u): (T, U)) -> (U, T) {
+    (u, t)
+}
```
jswrenn, line 2621: Slap some `#[inline(always)]` on these?

src/lib.rs
```diff
@@ -1912,11 +1914,7 @@ pub unsafe trait FromBytes: FromZeros {
     where
         Self: KnownLayout + Immutable,
     {
-        util::assert_dst_is_not_zst::<Self>();
-        let (slf, suffix) = Ptr::from_ref(bytes)
-            .try_cast_into(CastType::Prefix)
-            .map_err(|err| err.map_src(|s| s.as_ref()))?;
-        Ok((slf.bikeshed_recall_valid().as_ref(), suffix.as_ref()))
+        ref_from_prefix_suffix(bytes, None, CastType::Prefix)
```
jswrenn, line 1917: This is so slick.

src/lib.rs
```diff
@@ -1912,11 +1914,7 @@ pub unsafe trait FromBytes: FromZeros {
     where
         Self: KnownLayout + Immutable,
     {
-        util::assert_dst_is_not_zst::<Self>();
```
jswrenn, line 1915: These assertions should be retained here. It really helps with the error messages to have const asserts as high up on the call stack as possible.

joshlf: Done. In fact, this was incorrect anyway since it'd error in the case where an explicit count is provided.

src/lib.rs

src/pointer/ptr.rs
```diff
@@ -1660,7 +1692,7 @@ mod tests {
 
                     for cast_type in [CastType::Prefix, CastType::Suffix] {
                         if let Ok((slf, remaining)) =
-                            Ptr::from_ref(bytes).try_cast_into::<T>(cast_type)
+                            Ptr::from_ref(bytes).try_cast_into::<T>(cast_type, None)
```
joshlf, line 1695: TODO: Test `Some(meta)` for different values of `meta`

src/pointer/ptr.rs
```diff
@@ -1677,7 +1709,7 @@ mod tests {
                         }
                     }
 
-                    if let Ok(slf) = Ptr::from_ref(bytes).try_cast_into_no_leftover::<T>() {
+                    if let Ok(slf) = Ptr::from_ref(bytes).try_cast_into_no_leftover::<T>(None) {
```
joshlf, line 1712: TODO: Test `Some(meta)` for different values of `meta`

jswrenn (approved):

PR #1161, opened 2024-05-02
[pointer] Ptr::try_cast_into returns two Ptrs

joshlf:
The first return value is unchanged - it is the target of the cast. The second return value replaces the previous `split_at: usize`, and replaces it with a pointer to the prefix or suffix bytes as a `Ptr<[u8]>`.

This permits us to replace uses of `Ref` with direct uses of `Ptr` in some places.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -344,7 +344,7 @@ pub use crate::layout::*;
 //
 // See the documentation on `util::polyfills` for more information.
 #[allow(unused_imports)]
-use crate::util::polyfills::{self, NonNullExt as _};
+use crate::util::polyfills::{self, NonNullExt as _, NumExt as _};
 
 #[rustversion::nightly]
 #[cfg(all(test, not(__INTERNAL_USE_ONLY_NIGHTLY_FEATURES_IN_TESTS)))]
@@ -1913,7 +1913,10 @@ pub unsafe trait FromBytes: FromZeros {
         Self: KnownLayout + Immutable,
     {
         util::assert_dst_is_not_zst::<Self>();
-        Ref::<&[u8], Self>::new_from_prefix(bytes).map(|(r, s)| (r.into_ref(), s))
+        let (slf, suffix) = Ptr::from_ref(bytes)
+            .try_cast_into(CastType::Prefix)
+            .map_err(|err| err.map_src(|s| s.as_ref()))?;
+        Ok((slf.bikeshed_recall_valid().as_ref(), suffix.as_ref()))
     }
 
     /// Interprets the suffix of the given `bytes` as a `&Self` without copying.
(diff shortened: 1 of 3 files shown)


joshlf (commented):
TODO: Make @jswrenn a co-author; this is refactored from his work in #1072.

src/pointer/ptr.rs
```diff
@@ -1390,6 +1380,109 @@ mod _project {
...
+            // referent does not wrap around the address space. Since `l_ptr` is
+            // non-null, if `l_ptr.add(l_len)` does not wrap around the address
+            // space, then it is also non-null.
+            let r_slice = unsafe { NonNull::new_unchecked(r_slice) };
+
+            // TODO: Update this safety comment to prove that `l_slice` and
+            // `r_slice` satisfy the allocation, lifetime, isize, and
+            // wrap-around invariants of `Ptr`. Update it to justify that
+            // `l_slice` and `r_slice` are non-overlapping (required in case
+            // aliasing is exclusive).
```
joshlf, line 1426: TODO

src/pointer/ptr.rs
```diff
@@ -1390,6 +1380,109 @@ mod _project {
...
+
+            // TODO: The numbering on these invariants is off.
```
joshlf, line 1428: TODO

src/pointer/ptr.rs
```diff
@@ -1414,10 +1403,79 @@ mod _project {
...
+
+            let ptr = core::ptr::slice_from_raw_parts_mut(base, len);
+
+            // SAFETY: By invariant, `self`'s address is non-null and its range
+            // does not wrap around the address space. Since, by the preceding
+            // lemma, `ptr` addresses a range within that addressed by `self`,
+            // `ptr` is non-null.
+            let ptr = unsafe { NonNull::new_unchecked(ptr) };
+
+            // SAFETY: TODO(#896)
```
joshlf, line 1440: NOTE: Leaving this undone, but it will block 0.8.

src/lib.rs
```diff
@@ -4892,13 +4904,14 @@ where
...
-        let split_at = match Ptr::from_ref(bytes.deref()).try_cast_into::<T>(CastType::Prefix) {
-            Ok((_, split_at)) => split_at,
+        let remainder = match Ptr::from_ref(bytes.deref()).try_cast_into::<T>(CastType::Prefix) {
+            Ok((_, remainder)) => remainder,
             Err(e) => {
                 return Err(e.with_src(()).with_src(bytes));
             }
         };
 
+        let split_at = bytes.len().checked_sub(remainder.len()).unwrap();
```
joshlf: Heads up about this `.expect()`. If we go the "always return an error instead of panicking" route, even for errors we know are unreachable, then it's not clear to me what error to construct here (it was easier when we were returning `Option`s). I could also be convinced to do `unchecked_sub`.

src/lib.rs
```diff
@@ -4892,13 +4904,17 @@ where
...
+            Ok((_, remainder)) => remainder,
             Err(e) => {
                 return Err(e.with_src(()).with_src(bytes));
             }
         };
 
+        // PANICS: This is guaranteed not to underflow (and thus panic) since
+        // `remainder` addresses  subset of `bytes`, and thus can't be longer
+        // than `bytes`.
+        let split_at = bytes.len().checked_sub(remainder.len()).expect("zerocopy internal error: `bytes.len().checked_sub(remainder.len())` should not underflow");
```
joshlf, line 4917: Heads up about this `.expect()`. If we go the "always return an error instead of panicking" route, even for errors we know are unreachable, then it's not clear to me what error to construct here (it was easier when we were returning `Option`s). I could also be convinced to do `unchecked_sub`.

jswrenn: Since it's unreachable, it doesn't matter what error we'd return (though I'd argue that a `SizeError` is most appropriate). But let's just do an `unchecked_sub` here!

jswrenn (approved):
Giving this an approval, but there's one suggestion in my review.

src/lib.rs
joshlf: Done!

Issue #1155, opened 2024-04-30
Name TBD


joshlf:
## Progress

- [ ] Restrict `Immutable` documentation so that it requires recursive freedom from interior mutability, not just shallow
- [ ] Relax `Immutable` documentation so that it refers only to interior mutability
- [ ] Relax `Immutable` documentation so that it promises *nothing* to callers outside of zerocopy
- [ ] Get formal semantics of "interior mutability" documented in the Rust reference or stdlib docs
- [ ] Document in the reference or stdlib docs that certain operations are guaranteed sound depending *only* on whether interior mutation happens at runtime, and not on whether the type system believes that `UnsafeCell`s exist at certain offsets or covering certain ranges
- [ ] Update `Immutable` documentation to refer to this formal notion of interior mutability
- [ ] Update use sites to rely only on this soundness guarantee
- [ ] Permit `UnsafeCell`s in some places as appropriate (e.g., introduce a wrapper type which "disables" interior mutability)

### Intuitive definition of interior mutability + compiler optimizations

TODO:
- Provide an intuitive description of interior mutability and discuss how it relates to the compiler's right to make certain assumptions or optimizations
- Mention that there is currently no formal definition of interior mutability

### Interior mutability vs `UnsafeCell`s

TODO: Describe how interior mutability is implemented via `UnsafeCell`s today, and justify our intention to relax `Immutable`'s requirements in the future to *only* mention interior mutability, not `UnsafeCell`s in particular.

### Stacked Borrows, `UnsafeCell` overlap, and Ralf's future plans

TODO:
- Describe how Stacked Borrows reasons about `UnsafeCell` overlap
- Describe Ralf's intended formal semantics and how it diverges from Stacked Borrows

### `Immutable` vs `Freeze`

TODO:
- Explain why, in order to support general Safe Transmute, `Immutable` must be recursive, while the stdlib `Freeze` will not be recursive
- Conclude that we can never replace `Immutable` with `Freeze`

josephlr:
@joshlf I'm a little confused why we need the different semantics from `Freeze`. If we have a type like:
```
struct Foo<'a> {
  a: u32,
  b: &'a UnsafeCell<u64>,
}
```
It will be `Immutable`, but not `FromBytes` or `IntoBytes`, so it doesn't seem like we would run into problems with the various transmute methods? Or am I missing something?

[...]

PR #1165, opened 2024-05-02
Restate `Immutable`'s guarantees

jswrenn:
This PR adjust's `Immutable`'s informal public contract. Previously, we publicly documented that `Immutable` denoted that a type does not directly contain `UnsafeCell`. Now, we only document that `Immutable` types do not permit interrior mutation.

This change is made in recognition that:
 - Containing `UnsafeCell` may be fine, so long as that the special properties of that `UnsafeCell` cannot be exercised in safe code.
 - In contrast to what's guaranteed by the compiler's `Freeze` trait, we almost certainly need *recursive* immutability for safe transmutation.

Makes progress towards #1155.


README.md
```diff
@@ -44,7 +44,8 @@ functionality themselves, but are required to call certain methods provided
 by the conversion traits:
 - `KnownLayout` indicates that zerocopy can reason about certain layout
   qualities of a type
-- `Immutable` indicates that a type does not contain any `UnsafeCell`s
+- `Immutable` indicates that a type is free from interrior mutation,
```
joshlf, line 47:
```suggestion
- `Immutable` indicates that a type is free from interior mutation,
```

And elsewhere in this PR.

joshlf, line 47:
```suggestion
- `Immutable` indicates that a type is free from interrior mutability,
```

Let's standardize on "mutability" rather than "mutation"? The PR currently uses both, but IMO "mutability" is the more standard term.

src/lib.rs
```diff
@@ -776,11 +776,10 @@ pub use zerocopy_derive::FromZeros;
...
 pub use zerocopy_derive::Immutable;
 
-/// Types which do not directly contain any [`UnsafeCell`]s.
+/// Types which are free from interior mutability.
 ///
-/// `T: Immutable` indicates that `T` does not contain any `UnsafeCell`s in its
-/// fields. `T` may still refer to types which contain `UnsafeCell`s: for
-/// example, `&UnsafeCell<T>` implements `Immutable`.
+/// `T: Immutable` indicates that `T` does not permit interior mutation, except
+/// by ownership or an exclusive borrow.
```
joshlf, line 782:
```suggestion
/// by ownership or an exclusive (`&mut`) borrow.
```

In my experience, most casual Rust programmers aren't familiar with the "mutable references are actually exclusive references" perspective.

joshlf (changes requested):

README.md
jswrenn: Done

jswrenn: Done

src/lib.rs

joshlf (approved):

PR #1172, opened 2024-05-03
Add "release new crate version" workflow

joshlf:
This can be triggered manually via the GitHub web UI, and will generate a PR which updates the crate versions in both `Cargo.toml` files.

### DIFF
--- a/.github/workflows/release-crate-version.yml
+++ b/.github/workflows/release-crate-version.yml
@@ -1,0 +1,38 @@
+name: Release new crate versions
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Crate version'
+        required: true
+        default: "0.8.COOL"
+
+permissions: read-all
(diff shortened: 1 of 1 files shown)


joshlf:
I tested this manually by temporarily adding a `pull_request` trigger to this workflow, which generated https://github.com/google/zerocopy/pull/1173.

Note that the GitHub Actions meta-variable `${{ github.event.inputs.version }}` is not set when triggered by `pull_request`, so the fact that the version string is empty in that PR is expected given the testing methodology.

jswrenn (approved):
It'd also be nice to parameterize this by branch so we could use the same process for releases to older version trains, but this looks like a great start to me!

joshlf:
Now that this is merged, here's a better example of a PR - this one generated via the manual trigger UI: https://github.com/google/zerocopy/pull/1178

PR #1176, opened 2024-05-03
[release-crate-versions] Support arbitrary branches

joshlf:

### DIFF
--- a/.github/workflows/release-crate-version.yml
+++ b/.github/workflows/release-crate-version.yml
@@ -5,7 +5,10 @@ on:
       version:
         description: 'Crate version'
         required: true
-        default: "0.8.COOL"
+      branch:
+        description: 'Branch to update'
+        required: true
+        default: 'main'
 
 permissions: read-all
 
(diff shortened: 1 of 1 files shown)


joshlf:
Actually, looks like this might not be necessary. It looks like we can run the action "from" a particular branch in the UI (although I suspect we'll still have to update the part of the action that checks out `main` to detect what branch it's running from):

Here's the UI for `main`:
Screenshot 2024-05-03 at 8 32 01 AM

...and for another branch:
Screenshot 2024-05-03 at 8 32 13 AM

joshlf:
Alternatively, maybe this PR is better because it won't require us to backport this file to other branches (and do the same with future changes/bug fixes).

jswrenn (approved):

joshlf:
Example PR: #1179

Looks like the `sed` command that replaces `^version` in `Cargo.toml` affects dependencies that we specify using full TOML table syntax. That's a bug, but a difficult one to fix without wading into the [very nasty parts of sed](https://stackoverflow.com/questions/148451/how-to-use-sed-to-replace-only-the-first-occurrence-in-a-file). Given that this works on `main` right now, I'm inclined to not fix this bug and just continue manually cutting releases to version trains before 0.8, which will presumably be rare.

PR #1184, opened 2024-05-03
Implement `Ref::try_into_{ref,mut}`

jswrenn:
Wip. Currently a borrow-checker puzzle.


joshlf:
> Wip. Currently a borrow-checker puzzle.

I believe this is impossible to fix w/o adding bounds. `IntoByteSlice` doesn't imply `Copy`, and so `self.0.into()` irrevocably moves `self.0`. You'd need to add a `Copy` bound or some way of reversing the `into()` operation (ie, to reconstruct the value stored in `self.0` given the error).

jswrenn:
Perhaps there's a way forward doing `&*self.0` instead? We only need to call `is_bit_valid`, here.

joshlf:
Unfortunately that wouldn't live long enough. It would have the lifetime of the stack frame, but we need to return from the stack frame

jswrenn:
> Unfortunately that wouldn't live long enough. It would have the lifetime of the stack frame, but we need to return from the stack frame

Would you like me to provide that hat you'll eat, or will you be providing your own? ;-)

src/lib.rs
```diff
@@ -5187,13 +5187,64 @@ where
...
+                // valid size for `T`. Since properties are checked upon
+                // constructing `Ref`, these failures are unreachable.
+                unsafe { core::hint::unreachable_unchecked() }
+            }
+        }
+    }
+
+    /// Attempts to convert this `Ref<_, T>` into a `&T`.
+    ///
+    /// `try_into_ref` consumes the `Ref`, and returns a reference to `T`.
```
joshlf, line 5205: Document why this could fail?

src/lib.rs
```diff
@@ -5187,13 +5187,64 @@ where
...
+                // is aligned to `T`'s alignment and its size corresponds to a
+                // valid size for `T`. Since properties are checked upon
+                // constructing `Ref`, these failures are unreachable.
+                unsafe { core::hint::unreachable_unchecked() }
+            }
+        };
+        if !is_bit_valid {
+            Err(ValidityError::new(self))
+        } else {
+            match Ptr::from_ref(self.0.into()).try_cast_into_no_leftover::<T>() {
```
joshlf, line 5227: Is there any way we can keep the original `Ptr` and avoid calling constructing it twice?

src/lib.rs
```diff
@@ -5210,13 +5261,65 @@ where
...
+                // valid size for `T`. Since properties are checked upon
+                // constructing `Ref`, these failures are unreachable.
+                unsafe { core::hint::unreachable_unchecked() }
+            }
+        }
+    }
+
+    /// Attempts to convert this `Ref<_, T>` into a `&mut T`.
+    ///
+    /// `try_into_mut` consumes the `Ref`, and returns a reference to `T`.
```
joshlf, line 5279: Document why this could fail?

src/lib.rs
```diff
@@ -5210,13 +5261,65 @@ where
...
+                // is aligned to `T`'s alignment and its size corresponds to a
+                // valid size for `T`. Since properties are checked upon
+                // constructing `Ref`, these failures are unreachable.
+                unsafe { core::hint::unreachable_unchecked() }
+            }
+        };
+        if !is_bit_valid {
+            Err(ValidityError::new(self))
+        } else {
+            match Ptr::from_mut(self.0.into()).try_cast_into_no_leftover::<T>() {
```
joshlf, line 5301: Is there any way we can keep the original `Ptr` and avoid calling constructing it twice?

joshlf (changes requested):

src/lib.rs
jswrenn: Not that I can find, unfortunately. It might be possible with more `unsafe` code.

jswrenn: We might be able to do something like this, but lifetime extension makes me nervous:
```rust
    /// Attempts to convert this `Ref<_, T>` into a `&mut T`.
    ///
    /// `try_into_mut` consumes the `Ref`, and returns a reference to `T`.
    #[must_use = "has no side effects"]
    #[inline(always)]
    pub fn try_into_mut(mut self) -> Result<&'a mut T, ValidityError<Self, T>>
    where
        T: TryFromBytes,
    {
        // Extend the lifetime of `&mut *self.0` to `'a`.
        // SAFETY: Idk, this feels sketchy.
        let bytes = unsafe { mem::transmute::<&mut [u8], &'a mut [u8]>(&mut *self.0) };
        match Ptr::from_mut(bytes).try_cast_into_no_leftover::<T>() {
            Ok(candidate) => {
                match candidate.try_into_valid() {
                    Ok(candidate) => Ok(candidate.as_mut()),
                    Err(e) => Err(e.with_src(self)),
                }
            }
            Err(CastError::Validity(i)) => match i {},
            Err(CastError::Alignment(_) | CastError::Size(_)) => {
                // SAFETY: By invariant on `Ref::0`, the referenced byte slice
                // is aligned to `T`'s alignment and its size corresponds to a
                // valid size for `T`. Since properties are checked upon
                // constructing `Ref`, these failures are unreachable.
                unsafe { core::hint::unreachable_unchecked() }
            },
        }
    }
```

jswrenn: If `IntoByteSlice` documented that `.into()` behaved identically to `.deref()`, this might be fine?

PR #1188, opened 2024-05-04
Roll toolchains, release 0.7.33

joshlf:
This may fix the test failures in rust-random/rand#1446

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@
 [package]
 edition = "2018"
 name = "zerocopy"
-version = "0.7.32"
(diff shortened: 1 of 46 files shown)


jswrenn (approved):

joshlf:
The remaining failing tests are cargo-semver-checks on nightly with `--all-features` enabled, which is attempting to build 0.7.32 (published on crates.io) with the `simd-nightly` feature, which in turn enables nightly compiler features which changed since 0.7.32 was published. 

Discussed offline with @jswrenn, and we decided it made sense to manually override those tests and merge.

PR #1198, opened 2024-05-06
Use `AlignmentError` in `Unalign`'s failure conditions

jswrenn:
This make's `Unalign`'s methods consistent with zerocopy's other methods, and, in the case of `Unalign::try_deref_mut`, allows the original `&mut Unalign<T>` to be reused in the event of failure.

Makes progress towards #1139

### DIFF
--- a/src/pointer/ptr.rs
+++ b/src/pointer/ptr.rs
@@ -613,12 +613,43 @@ mod _conversions {
             c
         }
     }
+
+    /// `Ptr<'a, T, (_, _, _)>` → `Ptr<'a, Unalign<T>, (_, Aligned, _)>`
+    impl<'a, T, I> Ptr<'a, T, I>
+    where
+        I: Invariants,
+    {
+        /// Converts a `Ptr` an unaligned `T` into a `Ptr` to an aligned
+        /// `Unalign<T>`.
+        pub(crate) fn into_unalign(
+            self,
+        ) -> Ptr<'a, crate::Unalign<T>, (I::Aliasing, Aligned, I::Validity)> {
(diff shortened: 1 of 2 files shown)


src/pointer/ptr.rs
```diff
@@ -613,12 +613,41 @@ mod _conversions {
             c
         }
     }
+
+    impl<'a, T, I> Ptr<'a, T, I>
+    where
+        I: Invariants,
+    {
+        /// Converts a `Ptr` to a `Unalign<T>` into an `Ptr` to an unaligned `T`.
+        pub(crate) fn into_unaligned(
```
joshlf, line 622:
```suggestion
        pub(crate) fn into_unalign(
```

src/pointer/ptr.rs
```diff
@@ -613,12 +613,41 @@ mod _conversions {
             c
         }
     }
+
+    impl<'a, T, I> Ptr<'a, T, I>
```
joshlf, line 617: Fancy doc comment here with the UTF-8 arrow character?

src/pointer/ptr.rs
```diff
@@ -613,12 +613,41 @@ mod _conversions {
...
+    where
+        I: Invariants,
+    {
+        /// Converts a `Ptr` to a `Unalign<T>` into an `Ptr` to an unaligned `T`.
```
joshlf, line 621:
```suggestion
        /// Converts an unaligned `Ptr<T>` to an aligned `Ptr<Unalign<T>>`.
```

joshlf (changes requested):

joshlf (approved):
Please fix the outstanding comments, but feel free to merge after that.

PR #1218, opened 2024-05-08
[ref] Move `Ref` definition into its own module

joshlf:
This allows us to more strictly enforce the requirement that, for `Ref`'s `ByteSlice` field, methods are never called other than those defined on the `ByteSlice`, `ByteSliceMut`, `IntoByteSlice`, and `IntoByteSliceMut` traits. We accomplish this by keeping the field private, and only exposing unsafe getters which return, for example, `&impl ByteSlice`.

We also introduce new traits - `CopyableByteSlice` and `CloneableByteSlice`. We tighten the `Copy` and `Clone` impls for `Ref` to require these bounds. Previously, `ByteSlice` required that its "dereference stability" requirement applied to any `Copy` or `Clone` impls, but this would have prevented us from implementing `ByteSlice` for `Vec`. By splitting the requirement into separate traits, we open the door to supporting `Vec` (#992).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4709,20 +4709,50 @@ macro_rules! include_value {
 ///
 /// Implementations of `ByteSlice` must promise that their implementations of
 /// [`Deref`] and [`DerefMut`] are "stable". In particular, given `B: ByteSlice`
-/// and `b: B`, `b` must always dereference to a byte slice with the same
-/// address and length. This is true for both `b.deref()` and `b.deref_mut()`.
-/// If `B: Copy` or `B: Clone`, then the same is also true of copies or clones
-/// of `b`. For example, `b.deref_mut()` must return a byte slice with the same
-/// address and length as `b.clone().deref()`.
+/// and `b: B`, two calls, each to either `b.deref()` or `b.deref_mut()`, must
+/// return a byte slice with the same address and length. This must hold even if
+/// the two calls are separated by an arbitrary sequence of calls to methods on
+/// `ByteSlice`, [`ByteSliceMut`], [`IntoByteSlice`], or [`IntoByteSliceMut`],
+/// or on their super-traits. This does *not* need to hold if the two calls are
+/// separated by any method calls, field accesses, or field modifications *other
+/// than* those from these traits.
+///
+/// Note that this also implies that, given `b: B`, the address and length
+/// cannot be modified via objects other than `b`, either on the same thread or
+/// on another thread.
 pub unsafe trait ByteSlice: Deref<Target = [u8]> + Sized {}
 
-#[allow(clippy::missing_safety_doc)] // TODO(fxbug.dev/99068)
 /// A mutable reference to a byte slice.
 ///
 /// `ByteSliceMut` abstracts over various ways of storing a mutable reference to
 /// a byte slice, and is implemented for various special reference types such as
 /// `RefMut<[u8]>`.
-pub unsafe trait ByteSliceMut: ByteSlice + DerefMut {}
+///
+/// `ByteSliceMut` is a shorthand for [`ByteSlice`] and [`DerefMut`].
+pub trait ByteSliceMut: ByteSlice + DerefMut {}
+impl<B: ByteSlice + DerefMut> ByteSliceMut for B {}
+
+/// A [`ByteSlice`] which can be copied without violating dereference stability.
+///
+/// # Safety
+///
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -4709,21 +4709,51 @@ macro_rules! include_value {
...
+///
+/// Note that this also implies that, given `b: B`, the address and length
+/// cannot be modified via objects other than `b`, either on the same thread or
+/// on another thread.
 pub unsafe trait ByteSlice: Deref<Target = [u8]> + Sized {}
 
-#[allow(clippy::missing_safety_doc)] // TODO(fxbug.dev/99068)
+// Technically `ByteSliceMut` could be a safe trait since its safety invariants
+// are simply those of `ByteSlice`, but we make it `unsafe` to be conservative.
+#[allow(clippy::missing_safety_doc)]
```
jswrenn, line 4727: The `allow` here wigs me out a bit. We anticipate that outsiders might want to implement this trait, right? If so, we either need a safety comment or (better yet) we need to remove the `unsafe` modifier. Otherwise, we should seal the trait.

src/lib.rs
```diff
@@ -4834,6 +4864,14 @@ pub trait IntoByteSliceMut<'a>: ByteSliceMut + Into<&'a mut [u8]> {}
 #[allow(clippy::undocumented_unsafe_blocks)]
 unsafe impl<'a> ByteSlice for &'a [u8] {}
 
+// TODO(#429): Add a "SAFETY" comment and remove this `allow`.
+#[allow(clippy::undocumented_unsafe_blocks)]
+unsafe impl<'a> CopyableByteSlice for &'a [u8] {}
+
+// TODO(#429): Add a "SAFETY" comment and remove this `allow`.
+#[allow(clippy::undocumented_unsafe_blocks)]
+unsafe impl<'a> CloneableByteSlice for &'a [u8] {}
```
jswrenn, line 4873: It might be nice to burn down these TODOs while we're here, rather than add to our safety TODO backlog, but we can also punt on this.

jswrenn (changes requested):

src/lib.rs
```diff
@@ -4709,21 +4709,51 @@ macro_rules! include_value {
...
 /// [`Deref`] and [`DerefMut`] are "stable". In particular, given `B: ByteSlice`
-/// and `b: B`, `b` must always dereference to a byte slice with the same
-/// address and length. This is true for both `b.deref()` and `b.deref_mut()`.
-/// If `B: Copy` or `B: Clone`, then the same is also true of copies or clones
-/// of `b`. For example, `b.deref_mut()` must return a byte slice with the same
-/// address and length as `b.clone().deref()`.
+/// and `b: B`, two calls, each to either `b.deref()` or `b.deref_mut()`, must
+/// return a byte slice with the same address and length. This must hold even if
+/// the two calls are separated by an arbitrary sequence of calls to methods on
+/// `ByteSlice`, [`ByteSliceMut`], [`IntoByteSlice`], or [`IntoByteSliceMut`].
```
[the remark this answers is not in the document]
joshlf: They don't, but they have super-traits that do. Do you have any suggestions on how else we could word it?

src/lib.rs
joshlf: I *believe* we can make this safe. Technically redundant, but maybe we could just have the safety invariant be that the type satisfies the stability property with respect to `DerefMut`? It's redundant in the sense that the `ByteSlice` safety docs already mention `DerefMut`.

src/lib.rs
joshlf: My reasoning is that we've just split the trait's safety obligations into multiple traits, and not added any new obligations. Happy to do this now if you want - I just figured it'd be good not to block on that.

src/lib.rs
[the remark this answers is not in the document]
jswrenn: We could say "`Deref`, `DerefMut`, `Into<&[u8]>` and `Into<&mut [u8]>`".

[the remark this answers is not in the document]
joshlf: Discussed with @jswrenn offline: we're going to add "...or on their super-traits" to the current wording.

src/lib.rs
joshlf: Discussed with @jswrenn offline: decided to just make `ByteSliceMut` a safe trait.

jswrenn (approved):

PR #1224, opened 2024-05-09
Add `AliasingSafe` framework

joshlf:
This commit adds the `pointer::aliasing_safety::AliasingSafe` trait, which is implemented for pointer conversions which do not violate aliasing. This can happen either because the aliasing is exclusive or because neither type contains `UnsafeCell`s.

This paves the way for us to remove `Immutable` bounds from some of our API, including from some derives.

Makes progress on #251

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -322,7 +322,7 @@ use core::{
     },
 };
 
-use crate::pointer::invariant;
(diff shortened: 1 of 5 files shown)


jswrenn (approved):
(pair programmed)

PR #1227, opened 2024-05-10
[ci] Publish `cargo doc` to GitHub Pages

joshlf:


joshlf:
Tested manually by setting the job to run on the `pull_request` trigger, which resulted in: https://google.github.io/zerocopy/

I've now removed that trigger.

jswrenn:
> Tested manually by setting the job to run on the pull_request trigger

Per-PR docs actually sounds awesome. Could you restore that functionality?

joshlf:
Not without more work - the impl for testing just clobbered the docs every time a new PR is uploaded. It'd be cool to have separate docs for each PR, but it'd take more engineering work. We should do it as a follow-up though.

PR #1252, opened 2024-05-13
[util] Add SizedKnownLayout

joshlf:


joshlf:
Turns out that this is just `Wrapping`.

PR #1268, opened 2024-05-15
Improve UI/UX of post-monomorphization errors

jswrenn:
By asserting in a macro, rather than a helper function, rustc's diagnostics point to the user's callsite, rather than our own.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -674,7 +674,8 @@ macro_rules! const_panic {
 
 /// Either assert (if the current Rust toolchain supports panicking in `const
 /// fn`) or evaluate the expression and, if it evaluates to `false`, call
-/// `const_panic!`.
+/// `const_panic!`. This is used in place of `assert!` in const contexts to
+/// accommodate old toolchains.
 macro_rules! const_assert {
     ($e:expr) => {{
         #[cfg(zerocopy_panic_in_const)]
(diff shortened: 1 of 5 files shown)


joshlf (changes requested):
Can we add UI tests for these? We honestly should have had them all along.

src/macros.rs
```diff
@@ -720,3 +720,53 @@ macro_rules! const_unreachable {
         loop {}
     }};
 }
+
+/// Asserts at compile time that `$condition` is true for `Self` or the given
+/// `$tyvar`s.
+macro_rules! ct_assert {
```
joshlf, line 726: Is there a way to name this (and phrase the doc comment) so that it's clear why it's different from `const_assert` (immediately above in this file)?

src/macros.rs
```diff
@@ -720,3 +720,53 @@ macro_rules! const_unreachable {
...
+
+        impl<$($tyvar $(: $(? $optbound +)* $($bound +)*)?,)*> ConstAssert for ($($tyvar,)*) {
+            const ASSERT: bool = {
+                const_assert!($condition);
+                $condition
+            };
+        }
+
+        const_assert!(<($($tyvar,)*) as ConstAssert>::ASSERT);
+    };
```
joshlf, line 754: Wrap these bodies in an extra set of curly braces? Otherwise `ConstAssert` is introduced into the caller's namespace.

src/macros.rs
```diff
@@ -720,3 +720,53 @@ macro_rules! const_unreachable {
...
+        ct_assert!($tyvar: ?Sized + KnownLayout => {
+            let dst_is_zst = match $tyvar::LAYOUT.size_info {
+                crate::SizeInfo::Sized { .. } => false,
+                crate::SizeInfo::SliceDst(TrailingSliceLayout { elem_size, .. }) => {
+                    elem_size == 0
+                }
+            };
+            !dst_is_zst
+        });
+    }
```
joshlf, line 771: Wrap this in an extra set of curly braces? Otherwise, `use crate::KnownLayout;` is introduced into the caller's namespace.

joshlf (changes requested):

jswrenn:
We cannot, because `trybuild` only does a `cargo check`, not a `cargo build`.

joshlf:
> We cannot, because `trybuild` only does a `cargo check`, not a `cargo build`.

Oh look, [not the first time we've run into this issue](https://github.com/dtolnay/trybuild/issues/241#issuecomment-1854700765). Bummer.

src/macros.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/e36690aa8bf91b56899b9ad2b70eeb5f75a49362..2a433d49b23b4eba40cf8843490b0940b24894ac#diff-315c02cd05738da173861537577d159833f70f79cfda8cd7cf1a0d7a28ace31bR727)

src/macros.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/e36690aa8bf91b56899b9ad2b70eeb5f75a49362..2a433d49b23b4eba40cf8843490b0940b24894ac#diff-315c02cd05738da173861537577d159833f70f79cfda8cd7cf1a0d7a28ace31bR760)

joshlf:
Still need a resolution on https://github.com/google/zerocopy/pull/1268/files#r1601997671.

src/macros.rs
jswrenn: Note that this name will appear in compiler errors, so we want it to be both externally intelligible and succinct.

I can definitely improve the documentation, but would welcome suggestions of alternative names.

jswrenn: Oh, I suppose `static_assert` might be fitting.

jswrenn: [Done](https://github.com/google/zerocopy/compare/b2cb6b7adeff72a5dfa2d9a273308117574a28ed..6cfde9151b0b74d4f51e6cdb14a1dbd1ef7be5ad)

src/macros.rs
```diff
@@ -720,3 +720,55 @@ macro_rules! const_unreachable {
...
+            };
+        }
+
+        const_assert!(<($($tyvar,)*) as ConstAssert>::ASSERT);
+    }};
+}
+
+/// Assert at compile time that `tyvar` does not have a zero-sized DST
+/// component.
+macro_rules! ct_assert_dst_is_not_zst {
```
joshlf, line 761: Also update this name?

src/macros.rs
```diff
@@ -720,3 +720,55 @@ macro_rules! const_unreachable {
         loop {}
     }};
 }
+
+/// Asserts at compile time that `$condition` is true for `Self` or the given
+/// `$tyvar`s. Unlike `const_assert`, this is *strictly* a compile-time check.
+/// The condition is checked after monomorphization and, upon failure, emits a
+/// compile error.
```
joshlf, line 727: `const_assert!` is also strictly a compile-time check so long as it's always called from `const` code; its fanciness has to do with its behavior on older toolchains that don't support explicitly panicking in `const` code. Maybe I should update the docs to be more clear about that.

The distinction is actually that `static_assert` uses post-monomorphization errors only, and so its errors are generated during a different compiler phase. Arguably we could just inline `static_assert` into `ct_assert_dst_is_not_zst` below.

joshlf (changes requested):

joshlf (approved):

PR #1269, opened 2024-05-15
Rephrase conversion methods' documentation to accommodate DSTs

jswrenn:
This entails removing reference to both `size_of` and `align_of`, which cannot be invoked on slice DSTs.


src/lib.rs
```diff
@@ -1201,13 +1200,14 @@ pub unsafe trait TryFromBytes {
...
-    /// interpreted as a `Self`, and a reference to the remaining bytes. If
-    /// `candidate.len() < size_of::<Self>()` or `candidate` is not aligned to
-    /// `align_of::<Self>()` or the bytes are not a valid instance of `Self`,
-    /// this returns `Err`.
+    /// 
+    /// This method attempts to deserialize the largest possible size of `Self`
+    /// from the leading bytes of `candidate`. If that prefix is a valid
+    /// instance of `Self`, this method returns a reference to those bytes
+    /// reinterpreted as `Self`, and a reference to the remaining bytes. If
+    /// there are insufficient bytes, or if `candidate` is not appropriately
```
joshlf, line 1208:
```suggestion
    /// there are insufficient bytes, if `candidate` is not appropriately
```

src/lib.rs
```diff
@@ -1281,12 +1281,13 @@ pub unsafe trait TryFromBytes {
...
-    /// instance of `Self`, this method returns both a reference to those bytes
-    /// interpreted as a `Self`, and a reference to the preceding bytes. If
-    /// `candidate.len() < size_of::<Self>()` or the suffix of `candidate` is
-    /// not aligned to `align_of::<Self>()` or the suffix of `candidate` is not
-    /// a valid instance of `Self`, this returns `Err`.
+    /// This method attempts to deserialize the largest possible size of `Self`
+    /// from the trailing bytes of `candidate`. If that suffix is a valid
+    /// instance of `Self`, this method returns a reference to those bytes
+    /// reinterpreted as `Self`, and a reference to the preceeding bytes. If
+    /// there are insufficient bytes, or if the suffix of `candidate` would not
```
joshlf, line 1288:
```suggestion
    /// there are insufficient bytes, if the suffix of `candidate` would not
```

src/lib.rs
```diff
@@ -1201,13 +1200,14 @@ pub unsafe trait TryFromBytes {
...
     /// without copying.
-    ///
-    /// If the first `size_of::<Self>()` bytes of `candidate` are a valid
-    /// instance of `Self`, this method returns both a reference to those bytes
-    /// interpreted as a `Self`, and a reference to the remaining bytes. If
-    /// `candidate.len() < size_of::<Self>()` or `candidate` is not aligned to
-    /// `align_of::<Self>()` or the bytes are not a valid instance of `Self`,
-    /// this returns `Err`.
+    /// 
+    /// This method attempts to deserialize the largest possible size of `Self`
```
joshlf, line 1204: Can we pick a different verb besides "deserialize"? IMO it has connotations of copying.

src/lib.rs
```diff
@@ -1281,12 +1281,13 @@ pub unsafe trait TryFromBytes {
     /// Attempts to interpret the suffix of the given `candidate` as a `&Self`
     /// without copying.
     ///
-    /// If the last `size_of::<Self>()` bytes of `candidate` are a valid
-    /// instance of `Self`, this method returns both a reference to those bytes
-    /// interpreted as a `Self`, and a reference to the preceding bytes. If
-    /// `candidate.len() < size_of::<Self>()` or the suffix of `candidate` is
-    /// not aligned to `align_of::<Self>()` or the suffix of `candidate` is not
-    /// a valid instance of `Self`, this returns `Err`.
+    /// This method attempts to deserialize the largest possible size of `Self`
```
joshlf, line 1284: Can we pick a different verb besides "deserialize"? IMO it has connotations of copying.

src/lib.rs
jswrenn: I think the "or" is load-bearing. We're describing a disjunction of error conditions, not a conjunction.

joshlf: "If x, y, or z" describes a disjunction, just like "if x, y, and z" describes a conjunction. Am I missing something?

jswrenn: The structure you're proposing isn't "if x, y, or z", it's "If x, if y, or z". The operator precedence isn't totally unambiguous.

And if we used "if x, y, or z", it's not clear until that very last "or" whether the reader should understand those conditions to be in a conjunction or disjunction. We shouldn't hold the reader in suspense like that.

joshlf: > And if we used "if x, y, or z", it's not clear until that very last "or" whether the reader should understand those conditions to be in a conjunction or disjunction. We shouldn't hold the reader in suspense like that.

I'm convinced by this.

src/lib.rs
```diff
@@ -2764,12 +2766,10 @@ pub unsafe trait FromBytes: FromZeros {
...
     ///
-    /// This method verifies that `bytes.len() >= size_of::<T>() * count` and
-    /// that `bytes` is aligned to `align_of::<T>()`. It reinterprets the first
-    /// `size_of::<T>() * count` bytes from `bytes` to construct a `&[Self]`,
-    /// and returns the remaining bytes to the caller. It also ensures that
-    /// `sizeof::<T>() * count` does not overflow a `usize`. If any of the
-    /// length, alignment, or overflow checks fail, it returns `Err`.
+    /// This method attempts to return a reference to the prefix of `bytes`
+    /// reinterpreted as a `Self` with `count` trailing elements, and reference
+    /// to the preceeding bytes. If there are insufficient bytes, or if `bytes`
```
joshlf, line 2771:
```suggestion
    /// to the remaining bytes. If there are insufficient bytes, or if `bytes`
```

Or something else, but "preceding" only works if this were a suffix cast.

src/lib.rs
```diff
@@ -2847,12 +2847,10 @@ pub unsafe trait FromBytes: FromZeros {
...
     ///
-    /// This method verifies that `bytes.len() >= size_of::<T>() * count` and
-    /// that `bytes` is aligned to `align_of::<T>()`. It reinterprets the last
-    /// `size_of::<T>() * count` bytes from `bytes` to construct a `&[Self]`,
-    /// and returns the preceding bytes to the caller. It also ensures that
-    /// `sizeof::<T>() * count` does not overflow a `usize`. If any of the
-    /// length, alignment, or overflow checks fail, it returns `Err`.
+    /// This method attempts to return a reference to the suffix of `bytes`
+    /// reinterpreted as a `Self` with `count` trailing elements, and reference
+    /// to the preceeding bytes. If there are insufficient bytes, or if that
```
joshlf, line 2852:
```suggestion
    /// to the preceding bytes. If there are insufficient bytes, or if that
```

joshlf (changes requested):

src/lib.rs
```diff
@@ -2275,9 +2276,9 @@ pub unsafe trait FromBytes: FromZeros {
 
     /// Interprets the given `bytes` as a `&Self` without copying.
     ///
-    /// If `bytes.len()` does not correspond to a valid length for `Self`, or if
-    /// `bytes` is not aligned to `Self`'s alignment requirement, this returns
-    /// `Err`.
+    /// This method attempts to return a reference to `bytes` interpreted as a
+    /// `Self`. If there are insufficient or excess bytes, or if `bytes` is not
+    /// aligned to `Self`'s alignment requirement, this returns `Err`.
```
joshlf, line 2281: Here I think the old wording - "valid length for `Self`" (although I'd say size instead of length) - makes more sense. "Insufficient or excess" implies that there's one correct size, but that's not true - there could be many correct sizes.

src/lib.rs
```diff
@@ -2481,8 +2484,9 @@ pub unsafe trait FromBytes: FromZeros {
 
     /// Interprets the given `bytes` as a `&mut Self` without copying.
     ///
-    /// If `bytes.len() != size_of::<Self>()` or `bytes` is not aligned to
-    /// `align_of::<Self>()`, this returns `Err`.
+    /// This method attempts to return a reference to `bytes` interpreted as a
+    /// `Self`. If there are insufficient or excess bytes, or if `bytes` is not
+    /// aligned to `Self`'s alignment requirement, this returns `Err`.
```
joshlf, line 2489: Same here as above.

joshlf (changes requested):

src/lib.rs
jswrenn: How could there be many correct sizes? Isn't this backed by a `try_cast_into_no_leftover`?

jswrenn: Ah, I see your point. There's only one correct size per DST length though.

joshlf: Yeah exactly

src/lib.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/1b87f8faee7d14408c266e59a37f359b5084daba..a1ee12c2621aca15bfb635c0a32b14bd8e9004b8)

src/lib.rs

src/lib.rs

src/lib.rs
```diff
@@ -2347,10 +2348,11 @@ pub unsafe trait FromBytes: FromZeros {
 
     /// Interprets the prefix of the given `bytes` as a `&Self` without copying.
     ///
-    /// This method returns both a reference to the first `size_of::<Self>()`
-    /// bytes of `bytes` interpreted as a `Self`, and a reference to the remaining
-    /// bytes. If `bytes.len() < size_of::<Self>()` or `bytes` is not aligned to
-    /// `align_of::<Self>()`, this returns `Err`.
+    /// This method computes the largest possible size of `Self` that can fit in
+    /// the leading bytes of `candidate`, then attempts to return both a
```
jswrenn, line 2352: Whoops, this shouldn't be "bytes of `candidate`.

"bytes of `bytes`" is awkward, so perhaps I'll rename `bytes` to `source`.

jswrenn: Done.

src/lib.rs
```diff
@@ -2682,22 +2688,22 @@ pub unsafe trait FromBytes: FromZeros {
...
+    /// Interprets the given bytes as a `&Self` with a DST length equal to
     /// `count`.
     ///
-    /// This method verifies that `bytes.len() == size_of::<T>() * count` and
-    /// that `bytes` is aligned to `align_of::<T>()`. It also ensures that
-    /// `sizeof::<T>() * count` does not overflow a `usize`. If any of the
-    /// length, alignment, or overflow checks fail, it returns `Err`.
+    /// This method attempts to return a reference to `source` interpreted as a
+    /// `Self` with `count` trailing elements. If the length of `source` is not
+    /// a valid size of `Self`, or if `source` is not appropriately aligned,
```
joshlf, line 2705: "Valid size of `Self`" isn't the right condition when an explicit count is given.

src/ref.rs
```diff
@@ -434,20 +433,24 @@ where
...
-    // TODO(#29), TODO(#871): Pick a name and make this public. Make sure to
-    // update references to this name in `#[deprecated]` attributes elsewhere.
-    #[doc(hidden)]
+    /// Constructs a `Ref` from the given bytes with DST length equal to `count`
+    /// without copying.
+    ///
+    /// This method attempts to return a `Ref` to the prefix of `source`
+    /// interpreted as a `T` with `count` trailing elements, and a reference to
+    /// the remaining bytes. If the length of `source` is not a valid size of
+    /// `T`, or if `source` is not appropriately aligned, this returns `Err`.
```
joshlf, line 442: "Valid size of `T`" isn't the right condition when an explicit count is given.

src/ref.rs
```diff
@@ -635,16 +640,20 @@ where
...
-    // TODO(#29), TODO(#871): Pick a name and make this public. Make sure to
-    // update references to this name in `#[deprecated]` attributes elsewhere.
-    #[doc(hidden)]
+    /// Constructs a `Ref` from the given bytes with DST length equal to `count`
+    /// without copying.
+    ///
+    /// This method attempts to return a `Ref` to the prefix of `source`
+    /// interpreted as a `T` with `count` trailing elements, and a reference to
+    /// the remaining bytes. If the length of `source` is not a valid size of
+    /// `T`, this returns `Err`.
```
joshlf, line 649: "Valid size of `T`" isn't the right condition when an explicit count is given.

src/ref.rs
```diff
@@ -657,38 +666,40 @@ where
...
-    // TODO(#29), TODO(#871): Pick a name and make this public. Make sure to
-    // update references to this name in `#[deprecated]` attributes elsewhere.
-    #[doc(hidden)]
+    /// Constructs a `Ref` from the suffix of the given bytes with DST length
+    /// equal to `count` without copying.
+    ///
+    /// This method attempts to return a `Ref` to the suffix of `source`
+    /// interpreted as a `T` with `count` trailing elements, and a reference to
+    /// the preceding bytes. If there are insufficient bytes, or if that suffix
+    /// of `source` is not appropriately aligned, this returns `Err`.
```
joshlf, line 695: Alignment isn't required here.

src/ref.rs
```diff
@@ -657,38 +666,40 @@ where
...
-    // TODO(#29), TODO(#871): Pick a name and make this public. Make sure to
-    // update references to this name in `#[deprecated]` attributes elsewhere.
-    #[doc(hidden)]
+    /// Constructs a `Ref` from the prefix of the given bytes with DST
+    /// length equal to `count` without copying.
+    ///
+    /// This method attempts to return a `Ref` to the prefix of `source`
+    /// interpreted as a `T` with `count` trailing elements, and a reference to
+    /// the remaining bytes. If there are insufficient bytes, or if `source` is
+    /// not appropriately aligned, this returns `Err`.
```
joshlf, line 675: Alignment isn't required here.

joshlf (changes requested):

src/lib.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/f31d47afecf6f6652a2343d33ea5bc13662b1800..0c1f3c5d10b2842e836b549d7a8fbea1d5613700#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R2705)

src/ref.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/f31d47afecf6f6652a2343d33ea5bc13662b1800..0c1f3c5d10b2842e836b549d7a8fbea1d5613700#diff-d3f14a3d45e76b3a701b00d92f105dbbde3fee7002742276173d74a5bab8d16cR441)

src/ref.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/f31d47afecf6f6652a2343d33ea5bc13662b1800..0c1f3c5d10b2842e836b549d7a8fbea1d5613700#diff-d3f14a3d45e76b3a701b00d92f105dbbde3fee7002742276173d74a5bab8d16cR649)

src/ref.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/f31d47afecf6f6652a2343d33ea5bc13662b1800..0c1f3c5d10b2842e836b549d7a8fbea1d5613700#diff-d3f14a3d45e76b3a701b00d92f105dbbde3fee7002742276173d74a5bab8d16cR675)

src/ref.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/f31d47afecf6f6652a2343d33ea5bc13662b1800..0c1f3c5d10b2842e836b549d7a8fbea1d5613700#diff-d3f14a3d45e76b3a701b00d92f105dbbde3fee7002742276173d74a5bab8d16cR695)

joshlf (approved):

PR #1273, opened 2024-05-16
Document slice DSTs and size validity

### LINKED ISSUE #1263: Document slice DSTs and size validity at the crate root
Add a section to our crate root docs roughly with the following shape:

```rust
//! # Dynamically-sized types
//! 
//! Zerocopy supports slice-based dynamically sized types ("slice DSTs", or just
//! "DSTs" for short) via the [`KnownLayout`] trait.
//! 
//! A slice DST is a type...
//! 
//! ## What is a valid size?
//! 
//! ...
```

The section should give an overview of slice DSTs at a level of detail appropriate for our users. The "What is a valid size?" sub-section should describe the concept of "has a valid size for `T`", which is used in places in our API where we support slice DSTs. Those places should link to this sub-section.

joshlf:
Closes #1263

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,10 @@ Zerocopy provides four derivable traits for zero-cost conversions:
 - `IntoBytes` indicates that a type may safely be converted *to* a byte
   sequence
 
+This traits support sized types, slices, and [slice DSTs][slice-dsts].
+
+[slice-dsts]: KnownLayout#dynamically-sized-types
+
 ###### Marker Traits
 
 Zerocopy provides three derivable marker traits that do not provide any
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -37,6 +37,10 @@
 //! - [`IntoBytes`] indicates that a type may safely be converted *to* a byte
 //!   sequence
 //!
+//! This traits support sized types, slices, and [slice DSTs][slice-dsts].
+//!
+//! [slice-dsts]: KnownLayout#dynamically-sized-types
+//!
 //! ##### Marker Traits
 //!
 //! Zerocopy provides three derivable marker traits that do not provide any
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -71,6 +71,139 @@ Zerocopy provides byte-order aware integer types that support these
...
+attribute such as `#[repr(C)]`. [Other representations are
+supported][reprs], but in this section, we'll use `#[repr(C)]` as our
+example.
+
+A `#[repr(C)]` slice DST is laid out [just like sized `#[repr(C)]`
+types][repr-c-structs], but the presenence of a variable-length field
+introduces the possibility of *dynamic padding*. In particular, it may be
+necessary to add trailing padding *after* the trailing slice field in order
+to satisfy the outer type's alignment, and the amount of padding required
+may be a function of the length of the trailing slice field. This is just a
```
jswrenn, line 117: "may"? I'm not sure this qualification is technically necessary.

joshlf: Discussed offline; sticking with "may" since there are cases such as this one in which padding is fixed regardless of trailing slice length:

```rust
#[repr(C)]
struct Foo {
    a: u32,
    b: u16,
    z: [[u8; 4]],
}
```

README.md
```diff
@@ -65,6 +65,12 @@ These macros perform *compile-time* alignment and size checks, but cannot be
 used in generic contexts. For generic conversions, use the methods defined
 by the [conversion traits](#conversion-traits).
 
+###### Dynamically-sized types
+
+Zerocopy supports sized types, slices, and [slice DSTs][slice-dsts].
+
+[slice-dsts]: KnownLayout#dynamically-sized-types
```
jswrenn, line 72: IMO, this sentence should just be moved under the "Conversion Traits" heading, rather than introducing a new heading.

jswrenn (changes requested):

README.md
joshlf: Done.

jswrenn (approved):

Issue #1297, opened 2024-05-18
Make error types usable in more contexts


joshlf:
Per https://github.com/google/zerocopy/discussions/1288#discussioncomment-9476006:

> Error types like `SizeError` do not implement [`std::error::Error`](https://doc.rust-lang.org/std/error/trait.Error.html), so the functions like `read_from_bytes()` that now return a `Result` can't be used with [`anyhow::Context`](https://docs.rs/anyhow/latest/anyhow/trait.Context.html), while the previous version that returned `Option` could. Since `Debug` and `Display` are already implemented, I think this should just be a matter of adding `impl std::error::Error for SizeError {}`, but I didn't actually try this out. It may still not be possible to call `.context()` due to other constraints (`Send + Sync`). This is easy enough to work around (assuming the caller doesn't really care about the underlying error), and we didn't use this pattern in many places, so it's not really a big deal for us.

Implementing `std::error::Error` is implemented in #1298.

Supporting `Send + Sync` may be harder, since the contained `Src` type may not be `Send` or `Sync`. We may need to provide a way to discard the `Src` type, and we will need to consider how discoverable this mechanism is for people who are encountering a trait-not-satisfied error. I could see a few options:
- Provide an `Error<Src> -> Error<()>` conversion
- Provide a new type alias `type Foo = Error<()>` so that type signatures are cleaner
- Provide an entirely new type which contains similar information to the original error, but without the source
- Provide an entirely new type which contains the *name* of the source type, but doesn't contain an instance of the source type (just like we do for `Dst` right now)

PR #1298, opened 2024-05-18
[error] Implement std::error::Error on errors

joshlf:
While we're here, also relax `Dispaly for AlignmentError<Src, Dst>` to permit `Dst: ?Sized` in exchange for `Dst: KnownLayout`. This is an important relaxation since our APIs permit performing conversions into unsized destination types with runtime alignment checking.

Makes progress on #1297

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,10 +60,11 @@ alloc = []
 derive = ["zerocopy-derive"]
 simd = []
 simd-nightly = ["simd"]
+std = ["alloc"]
 # This feature depends on all other features that work on the stable compiler.
 # We make no stability guarantees about this feature; it may be modified or
 # removed at any time.
-__internal_use_only_features_that_work_on_stable = ["alloc", "derive", "simd"]
+__internal_use_only_features_that_work_on_stable = ["alloc", "derive", "simd", "std"]
 
 [dependencies]
 zerocopy-derive = { version = "=0.8.0-alpha.14", path = "zerocopy-derive", optional = true }
--- a/src/error.rs
+++ b/src/error.rs
@@ -32,15 +32,40 @@
 //! All error types provide an `into_src` method that converts the error into
 //! the source value underlying the failed conversion.
 //!
+//! ## Display formatting
+//!
+//! All error types provide a `Display` implementation that produces a
+//! human-readable error message. When `debug_assertions` are enabled, these
+//! error messages are verbose and may include potentially sensitive
+//! information, including:
+//!
+//! - the names of the involved types
+//! - the sizes of the involved types
+//! - the addresses of the involved types
+//! - the contents of the involved types
+//!
+//! When `debug_assertions` are disabled (as is default for `release` builds),
+//! such potentially sensitive information is excluded.
+//!
+//! In the future, we may support manually configuring this behavior. If you are
+//! interested in this feature, [let us know on GitHub][issue-1457] so we know
+//! to prioritize it.
+//!
+//! [issue-1457]: https://github.com/google/zerocopy/issues/1457
+//!
 //! ## Validation order
 //!
 //! Our conversion methods typically check alignment, then size, then bit
 //! validity. However, we do not guarantee that this is always the case, and
 //! this behavior may change between releases.
 
-use core::{convert::Infallible, fmt, marker::PhantomData, ops::Deref};
+use core::{
+    convert::Infallible,
+    fmt::{self, Debug, Write},
+    ops::Deref,
+};
 
-use crate::TryFromBytes;
+use crate::{util::SendSyncPhantomData, KnownLayout, TryFromBytes};
 #[cfg(doc)]
 use crate::{FromBytes, Ref};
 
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -60,10 +60,11 @@ alloc = []
 derive = ["zerocopy-derive"]
 simd = []
 simd-nightly = ["simd"]
+std-error = ["alloc"]
```
jswrenn, line 63: I kinda think we should just call this feature `std`, for symmetry with `alloc`.

src/error.rs
```diff
@@ -126,9 +135,10 @@ impl<Src, Dst: ?Sized> fmt::Debug for AlignmentError<Src, Dst> {
 // The bounds on this impl are intentionally conservative, and can be relaxed
 // either once a `?Sized` alignment accessor is stabilized, or by storing the
 // alignment as a runtime value.
-impl<Src, Dst> fmt::Display for AlignmentError<Src, Dst>
+impl<Src, Dst: ?Sized> fmt::Display for AlignmentError<Src, Dst>
 where
     Src: Deref,
+    Dst: KnownLayout,
```
jswrenn, line 141: Perhaps we should have this bound on our `Display` impls for the other error types, too?

src/error.rs
```diff
@@ -139,14 +149,22 @@ where
         f.write_str("the conversion failed because the address of the source (a multiple of ")?;
         addr_align.fmt(f)?;
         f.write_str(") is not a multiple of the alignment (")?;
-        core::mem::align_of::<Dst>().fmt(f)?;
+        <Dst as KnownLayout>::LAYOUT.align.get().fmt(f)?;
```
jswrenn, line 152: I really like the technique of using `KnownLayout` to enrich the error messages! We could also apply this to `SizeError`, I think, to revalidate the size at runtime from `src` in `Display`, rather than add another field to `SizeError`.

jswrenn (changes requested):

src/error.rs
joshlf: Not sure I understand what you're suggesting here.

Incidentally, it turns out that some of the `Src: Deref` bounds were unnecessary (ie, not actually used), so I removed them.

src/error.rs
joshlf: To what end? We only need it here because of its use in the body.

Cargo.toml
joshlf: Done.

joshlf:
I added a `SendSyncPhantomData` type and tests to make sure that the errors are `Send + Sync` only dependent upon `Src: Send + Sync`, but regardless of `Dst`, which only exists at the type level.

src/error.rs
jswrenn: @joshlf Improved diagnostics. I went ahead and added a `KnownLayout` bound to `SizeError`: https://github.com/google/zerocopy/compare/f8bd5db3e615ba1c93451e2cce6789c73788b2ca..235cd25c1b93f85e90287ada15d5a4fb2d5770c5

I'm sure we can use this bound to provide better diagnostics in the DST case too.

src/error.rs
```diff
@@ -206,17 +224,26 @@ impl<Src, Dst: ?Sized> fmt::Debug for SizeError<Src, Dst> {
 /// Produces a human-readable error message.
 impl<Src, Dst: ?Sized> fmt::Display for SizeError<Src, Dst>
 where
-    Src: Deref,
```
jswrenn, line 209: I think we should keep this bound, too. Once we have `size_of_val`, we can use this bound to provide information about the length of the source.

jswrenn (changes requested):
I went ahead and committed an example of how having a `KnownLayout` bound can be used to improve our diagnostics. Generally speaking, I think we should consistently have `Src: Deref` and `Dst: ?Sized + KnownLayout` on our `Display` and `Error` impls. We can always relax the bounds later.

src/error.rs
joshlf: I suspect this may also be considered too much data leak for some users. Size often correlates with content, which in turn can be used to [violate confidentiality](https://en.wikipedia.org/wiki/CRIME_(security_exploit)).

src/error.rs
joshlf: What do you mean "once we have `size_of_val`"? [We already do](https://doc.rust-lang.org/std/mem/fn.size_of_val.html).

(Same concern about leaking size, btw.)

joshlf: Hmmm yeah I could see that. Maybe a "rich-error-text" feature or something? We could probably gate a lot behind that so we're super anal without it but pretty verbose with it.

joshlf:
Discussed offline with @jswrenn. Here's the plan:
- `Debug` impl is always verbose
- Conditional on `debug_assertions`:
  - On: `Display` impls are verbose (assume that this is in a non-prod context where there's no concern of data leak
  - Off: `Display` impls attempt to leak as little information as is reasonable
- File an issue to track: have some way of configuring this behavior; I can see two use cases:
  - Users enable `debug_assertions` in prod
  - Users want the output of `Display` to be stable regardless of `debug_assertions`
- Document this behavior at the crate root. Suggest that users who need a way of changing it chime in on this new issue so we know how much demand there is.

jswrenn (approved):

PR #1299, opened 2024-05-18
Move `IntoByteSlice[Mut]: Into` into methods

### LINKED ISSUE #1260: Remove `Into<&[u8]>` bound from `IntoByteSlice`?
The `Into<&[u8]>` bound on `IntoByteSlice` means that we can't implement `IntoByteSlice` for `&mut [u8]`. If instead we used a method directly on `IntoByteSlice`, we could remove this restriction.

This restriction means, for example, that `Ref::into_ref` isn't available for `Ref<&mut [u8], _>`.

joshlf:
Remove the `Into<&[u8]>` and `Into<&mut [u8]>` bounds on `IntoByteSlice` and `IntoByteSliceMut` and replace them with methods. This allows `IntoByteSlice` to be implemented for `&mut [u8]` (and, as a consequence, permits `Ref::<&mut [u8], _>::into_ref`), which was previously barred.

Closes #1260

### DIFF
--- a/src/byte_slice.rs
+++ b/src/byte_slice.rs
@@ -160,6 +160,7 @@ where
 pub trait SplitByteSliceMut: SplitByteSlice + ByteSliceMut {}
 impl<B: SplitByteSlice + ByteSliceMut> SplitByteSliceMut for B {}
 
+#[allow(clippy::missing_safety_doc)] // There's a `Safety` section on `into_byte_slice`.
 /// A [`ByteSlice`] that conveys no ownership, and so can be converted into a
 /// byte slice.
 ///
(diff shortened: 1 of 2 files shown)


src/byte_slice.rs
```diff
@@ -218,16 +233,17 @@ unsafe impl<'a> SplitByteSlice for &'a [u8] {
...
-// something semantically equivalent to it). Thus, `ByteSlice`'s stability
-// guarantees are not violated by the `Into<&[u8]>` impl.
-//
-// [1] https://doc.rust-lang.org/std/convert/trait.From.html#impl-From%3CT%3E-for-T
-unsafe impl<'a> IntoByteSlice<'a> for &'a [u8] {}
+// SAFETY: See inline.
+unsafe impl<'a> IntoByteSlice<'a> for &'a [u8] {
+    #[inline(always)]
+    fn into_byte_slice(self) -> &'a [u8] {
+        // SAFETY: It would be patently insane to implement `<Deref for
```
jswrenn, line 240: 🤣

jswrenn (approved):

PR #1301, opened 2024-05-18
Disable coverage tracking on tests

joshlf:
This ensure that our coverage tracking more accurately reflects what percentage of production (ie, non-test) code is covered, and makes warnings about uncovered code less likely to be false positives (e.g., for test code that we add just to test whether something compiles).

### DIFF
--- a/src/byteorder.rs
+++ b/src/byteorder.rs
@@ -855,6 +855,7 @@ module!(network_endian, NetworkEndian, "network-endian");
 module!(native_endian, NativeEndian, "native-endian");
 
 #[cfg(any(test, kani))]
+#[cfg_attr(coverage_nightly, coverage(off))]
 mod tests {
     use super::*;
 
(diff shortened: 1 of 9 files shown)


joshlf:
Looks like this attribute unfortunately can't be applied at the module level: https://github.com/rust-lang/rust/issues/84605#issuecomment-2118986532

briansmith:
> This ensure that our coverage tracking more accurately reflects what percentage of production (ie, non-test) code is covered, and makes warnings about uncovered code less likely to be false positives (e.g., for test code that we add just to test whether something compiles).

I don't think this is a good idea anyway. We still want code coverage measurement for tests so we can be notified about any branches in test code that aren't executed (i.e. bugs in tests). Code coverage reporting tools like codecov.io have features like [components](https://docs.codecov.com/docs/components) and [flags](https://docs.codecov.com/docs/flags) that could be used to separate test and non-test code with various levels of inconvenience.

PR #1303, opened 2024-05-19
Derive trivial is_bit_valid when possible

joshlf:
When deriving `FromBytes`, the implied `TryFromBytes` derive's `is_bit_valid` impl is generated as always returning `true`. This is faster to codegen, is faster to compile, and is friendlier on the optimizer.

TODO: Tests

Makes progress on #5


jswrenn (dismissed):

zerocopy-derive/src/lib.rs
```diff
@@ -405,40 +409,53 @@ fn derive_unaligned_inner(ast: &DeriveInput) -> proc_macro2::TokenStream {
...
-                })*
-            }
-        )
-    });
+fn derive_try_from_bytes_struct(
+    ast: &DeriveInput,
+    strct: &DataStruct,
+    top_level: Trait,
+) -> proc_macro2::TokenStream {
+    let extras = if top_level == Trait::FromBytes {
```
joshlf, line 417: Oops, this check is not sufficient! In particular, `#[derive(FromBytes)]` applied to a type with generics will compile successfully, but that doesn't mean that the type is always `FromBytes` - the generated impl is bounded by its generic field types also being `FromBytes`.

It *might* be sufficient to do `top_level == Trait::FromBytes` *and* check that the type has no generics. However, this would rely for soundness on compilation always failing for non-`FromBytes` types. E.g., what if non-`FromBytes` types just result in the `FromBytes` impl not being available, but the compilation still succeeding?

What we probably should do is to add an explicit `Self: FromBytes` check in the body of the generated `is_bit_valid`. I'm pretty sure that's foolproof.

(Same applies to derives for unions and enums)

joshlf: > It _might_ be sufficient to do `top_level == Trait::FromBytes` _and_ check that the type has no generics. However, this would rely for soundness on compilation always failing for non-`FromBytes` types. E.g., what if non-`FromBytes` types just result in the `FromBytes` impl not being available, but the compilation still succeeding?
> 
> What we probably should do is to add an explicit `Self: FromBytes` check in the body of the generated `is_bit_valid`. I'm pretty sure that's foolproof.

Done - the latest version does both of these.

joshlf:
@jswrenn Take a look at the `.stderr` files. The extra `Self: FromBytes` check added in `is_bit_valid` results in significantly more verbose error messages. It's still obvious what's happening, so it's not a *huge* deal. I could also see replacing this with one of the following:
- Convince ourselves that there's absolutely no way for compilation to succeed if `FromBytes` isn't satisfied (risky - what about future language changes a la trivial bounds?)
- In the `#[derive(FromBytes)]` case, do `impl TryFromBytes for Type where Type: FromBytes`; maybe this will generate better error messages? Significantly harder to wire through our existing codebase, though
- In all cases of `#[derive(Subtrait)]`, do `impl Trait for Type where Type: Subtrait` (so that we don't have to special-case this); now all code can more soundly rely on the top-level trait actually being satisfied. Probably still more verbose errors than currently, but maybe not as bad as with the current solution in this PR.

jswrenn:
I expected the expansion for this:
```rust
#[derive(FromBytes)]
struct Foo {
    a: u8,
    b: u16,
}
```
...to look like this:
```rust
unsafe impl ::zerocopy::TryFromBytes for Foo
where
    u8: ::zerocopy::FromBytes,
    u16: ::zerocopy::FromBytes,
{
    fn only_derive_is_allowed_to_implement_this_trait() {}
    fn is_bit_valid<A>(candidate: ::zerocopy::Maybe<Self, A>) -> bool
    where
        A: ::zerocopy::pointer::invariant::Aliasing
            + ::zerocopy::pointer::invariant::AtLeast<
                ::zerocopy::pointer::invariant::Shared,
            >,
    {
        true
    }
}
```
Why doesn't it? Am I missing some subtlety here?

joshlf:
Okay, rebasing this on our recent work makes the error messages *much* less verbose. See the `.stderr` files in https://github.com/google/zerocopy/commit/b8eeb756bba6e211ee25213ccc7390274d934474 (old) vs https://github.com/google/zerocopy/commit/9c82f0c71840d46395825e4c1e57d369161c7a27 (new) (unfortunately, diffing these directly is not helpful because of how many changes I rebased on). It's still not perfect, but it's a *lot* better.

In #1752, we bubble errors up as errors rather than as `TokenStream`s, and so we can bail earlier and not generate multiple errors in a single derive invocation. I suspect that what happened is that that change is responsible for the errors here becoming much less verbose.

jswrenn (approved):

zerocopy-derive/src/output_tests.rs
joshlf, line 1: TODO: Add more output tests for this optimization.

zerocopy-derive/src/lib.rs
```diff
@@ -510,15 +515,29 @@ fn derive_try_from_bytes_union(ast: &DeriveInput, unn: &DataUnion) -> TokenStrea
...
+fn derive_try_from_bytes_enum(
+    ast: &DeriveInput,
+    enm: &DataEnum,
+    top_level: Trait,
+) -> Result<TokenStream, Error> {
     let repr = EnumRepr::from_attrs(&ast.attrs)?;
 
-    // The enum derive requires some extra scaffolding
-    let extra = Some(r#enum::derive_is_bit_valid(&ast.ident, &repr, &ast.generics, enm)?);
+    // TODO: Safety comments here and below.
```
joshlf, line 529: TODO

PR #1311, opened 2024-05-19
[ci] Run Miri test alias models in parallel

joshlf:
In CI, we run Miri tests using both the Stacked Borrows and Tree Borrows aliasing models. This commit runs these in parallel to improve utilization of available parallelism.


joshlf:
[This workflow run](https://github.com/google/zerocopy/actions/runs/9149390574?pr=1311) took 18m50s, which is within error bars of the previous times. I'm going to close this as not buying us enough to justify the added complexity.

PR #1319, opened 2024-05-20
Move contributing guide to GitHub discussion

joshlf:
This will make it easier to make edits going forward, and permits discussion.


INTERNAL.md
jswrenn, line 1: Should we be deleting this? Any details which are specific to moments-of-time in our repo should probably be tracked by the repo so that the documentation remains valid when we need to go backport a feature.

joshlf: I've also moved it to be a discussion, but I take your point about history. My concern is that, in practice, we never keep it up to date because it's a pain to update. Even when we do update it, it's rarely in the same commit that we introduce a change, which means it doesn't have the benefit of tracking the state of the repo for the purposes of reversions and such. Hopefully as a discussion that we can just edit, we'll be more likely to actually do that.

jswrenn (approved):

Issue #553, opened 2023-10-27
Don't generate warning when deriving on deprecated type


joshlf:
While working on [this PR](https://github.com/rust-lang/libc/pull/3407), I wrote the following code:

```rust
    #[cfg_attr(feature = "zerocopy", derive(FromZeroes, FromBytes, AsBytes))]    
    #[deprecated(                                                                
        since = "0.2.55",                                                        
        note = "Use the `mach2` crate instead",                                  
    )]                                                                           
    pub struct mach_timebase_info {                                              
        pub numer: u32,                                                          
        pub denom: u32,                                                          
    }
```

Running `cargo check --features zerocopy` on `cargo 1.74.0-nightly (b4ddf95ad 2023-09-18)`, I get the following warning:

```text
warning: use of deprecated struct `unix::bsd::apple::mach_timebase_info`: Use the `mach2` crate instead
   --> src/unix/bsd/apple/mod.rs:306:16
    |
306 |     pub struct mach_timebase_info {
    |                ^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(deprecated)]` on by default
```

## Mentoring instructions

- [ ] Figure out the conditions under which this warning is emitted
- [ ] Write a test that reproduces this behavior
- [x] Modify our derives as needed to not generate this warning
  - Done: https://github.com/google/zerocopy/pull/556

PR #1332, opened 2024-05-21
Add trybuild test to ensure deprecated types are derivable.

daniel-levin:
Addresses https://github.com/google/zerocopy/issues/553.

#553 talks about warnings, but trybuild tests operate on compile errors. I think this change still manages to meet the spirit of #553. The docstring at the top of `absence_of_deprecated_warning.rs` explains the test. In short, there is "only" one compile error in the test.

If you comment out the `#[allow(deprecated)]` attribute in `zerocopy-derive/src/lib.rs`, then this test will fail.

### DIFF
--- a/zerocopy-derive/tests/ui-nightly/absence_of_deprecated_warning.rs
+++ b/zerocopy-derive/tests/ui-nightly/absence_of_deprecated_warning.rs
@@ -1,0 +1,33 @@
+// Copyright 2024 The Fuchsia Authors
+//
+// Licensed under a BSD-style license <LICENSE-BSD>, Apache License, Version 2.0
+// <LICENSE-APACHE or https://www.apache.org/licenses/LICENSE-2.0>, or the MIT
+// license <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your option.
+// This file may not be copied, modified, or distributed except according to
+// those terms.
+
+//! See: https://github.com/google/zerocopy/issues/553
+//! zerocopy must still allow derives of deprecated types.
+//! This test has a hand-written impl of a deprecated type, and should result in a compilation
+//! error. If zerocopy does not tack an allow(deprecated) annotation onto its impls, then this
+//! test will fail because more than one compile error will be generated.
+#![deny(deprecated)]
+
+extern crate zerocopy;
+
+use zerocopy::IntoBytes;
+
+#[deprecated = "Do not use"]
+#[derive(IntoBytes)]
+#[repr(C)]
+struct OldHeader {
+    field_a: usize,
+    collection: [u8; 8],
+}
+
+trait T {}
+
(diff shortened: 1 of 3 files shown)


zerocopy-derive/tests/ui-nightly/absence_of_deprecated_warning.rs
```diff
@@ -0,0 +1,24 @@
+//! See: https://github.com/google/zerocopy/issues/553
```
jswrenn, line 1:
```suggestion
// Copyright 2024 The Fuchsia Authors
//
// Licensed under a BSD-style license <LICENSE-BSD>, Apache License, Version 2.0
// <LICENSE-APACHE or https://www.apache.org/licenses/LICENSE-2.0>, or the MIT
// license <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your option.
// This file may not be copied, modified, or distributed except according to
// those terms.

//! See: https://github.com/google/zerocopy/issues/553
```

zerocopy-derive/tests/ui-nightly/absence_of_deprecated_warning.rs
```diff
@@ -0,0 +1,24 @@
...
+#[derive(IntoBytes)]
+#[repr(C)]
+struct OldHeader {
+    field_a: usize,
+    collection: [u8; 8],
+}
+
+trait T {}
+
+impl T for OldHeader {}
```
jswrenn, line 22: What purpose does this serve?

jswrenn (changes requested):
Thank you for this! Just a minor nit.

zerocopy-derive/tests/ui-nightly/absence_of_deprecated_warning.rs
daniel-levin: Ah yes, I was wondering about that. The contributing guide didn't mention it but probably should.

zerocopy-derive/tests/ui-nightly/absence_of_deprecated_warning.rs
daniel-levin: The useless impl triggers the specific compiler error we want. That is, zerocopy is implementing traits for the derived type as well, but only this impl should become an error.

jswrenn: That's clever! Could you add a comment to that effect; e.g.:
```suggestion
trait T {}

// Intentionally trigger a deprecation error
impl T for OldHeader {}
```

daniel-levin: Sure!

daniel-levin:
I needed to update the line numbers in the trybuild test. I squashed the modifications and force pushed.

PR #1339, opened 2024-05-22
Make macros robust against never type fallback

joshlf:


joshlf:
Credit to @WaffleLapkin for explaining why the lint is still triggered: https://github.com/rust-lang/rust/issues/123748#issuecomment-2124984745

Based on this explanation, I think it makes sense to merge as-is.

joshlf:
Decided to just merge #1338 instead and follow up in #1383

PR #1352, opened 2024-05-27
Add byteorder::U24

JackWolfard:
There are binary file formats which utilize 24-bit unsigned integers, so this will make it easier to work with those files.

### DIFF
--- a/src/byteorder.rs
+++ b/src/byteorder.rs
@@ -189,9 +189,9 @@ macro_rules! impl_fmt_traits {
 }
 
 macro_rules! impl_ops_traits {
-    ($name:ident, $native:ident, "floating point number") => {
-        impl_ops_traits!($name, $native, @all_types);
-        impl_ops_traits!($name, $native, @signed_integer_floating_point);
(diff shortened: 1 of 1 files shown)


joshlf:
Thanks for the contribution! We've currently got a pretty big backlog, so I expect it'll be a few weeks until we'll get a chance to look at this - likely late June. Apologies for the delay, but I promise we haven't forgotten about it 🙂

joshlf:
Sorry for letting this sit for so long!

Can you say a bit about your use case for this feature?

JackWolfard:
> Sorry for letting this sit for so long!
> 
> Can you say a bit more about your use case for this feature? What specific file formats are you hoping to parse, and in what context? Do you have other use cases besides that?

No worries! I ended up going with a different library, https://github.com/jam1garner/binrw. The specific file format I was processing was BYML (https://zeldamods.org/wiki/BYML) which uses a `U24` for number of entries in a few node types.

joshlf:
Sounds good!

PR #1375, opened 2024-05-31
Specify directories/files when caching $CARGO_HOME

### LINKED ISSUE #1312: CI cache not including installed Cargo tools?
Our `generate_cache` CI job, which is a dependency of the main `build_test` job, runs [`cargo install cargo-nextest`](https://github.com/google/zerocopy/blob/889ac1bc2f38b1ccf338a1bf12f222102b338bc0/.github/workflows/ci.yml#L476). However, in [this run](https://github.com/google/zerocopy/actions/runs/9149280094/job/25152797173?pr=1311), we can see that it's getting re-installed in `build_test`. This likely means that binaries installed via `cargo install` are not getting cached.

sl4m:
This fixes #1312.

I'm making the presumption that the issue concerns the re-installation of `cargo-nextest` in the step `Run tests under Miri`.

```
 Running Miri tests with 8 threads
    Updating crates.io index
 Downloading crates ...
  Downloaded cargo-nextest v0.9.70
  Installing cargo-nextest v0.9.70

 ....
```

The fix expands the Cargo cache path from `~/.cargo/` to the specific directories and files listed in the [Cargo book](https://doc.rust-lang.org/cargo/guide/cargo-home.html#caching-the-cargo-home-in-ci). I initially just had the directories from `action/cache` Rust [example](https://github.com/actions/cache/blob/main/examples.md#rust---cargo), but also encountered the error: `binary cargo-readme already exists in destination` that Josh experienced.

Here is a [test run](https://github.com/sl4m/zerocopy/actions/runs/9322316917/job/25664391430) on the fix of my fork. It is the same Linux target that was referenced in #1312.

```
Running Miri tests with 8 threads
    Updating crates.io index
     Ignored package `cargo-nextest v0.9.72` is already installed, use --force to override
Preparing a sysroot for Miri (target: aarch64-unknown-linux-gnu)... done

 ...
```

### DIFF
--- a/.github/actions/cache/action.yml
+++ b/.github/actions/cache/action.yml
@@ -11,9 +11,14 @@ description: 'Caches cargo dependencies'
 runs:
   using: composite
   steps:
-    - uses: actions/cache@v3 
-      with: 
-        path: | 
-          ~/.cargo/ 
+    - uses: actions/cache@v3
+      with:
+        path: |
+          ~/.cargo/.crates.toml
+          ~/.cargo/.crates2.json
+          ~/.cargo/bin/
+          ~/.cargo/git/db/
+          ~/.cargo/registry/cache/
+          ~/.cargo/registry/index/
         key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
-        
+


.github/actions/cache/action.yml
```diff
@@ -11,9 +11,14 @@ description: 'Caches cargo dependencies'
...
-          ~/.cargo/ 
+    - uses: actions/cache@v3
+      with:
+        path: |
+          ~/.cargo/.crates.toml
+          ~/.cargo/.crates2.json
+          ~/.cargo/bin/
+          ~/.cargo/git/db/
+          ~/.cargo/registry/cache/
+          ~/.cargo/registry/index/
```
joshlf, line 22: Can you add a comment here that explains why we do this instead of just doing `~/.cargo/` (and mentions #1312)? I know I'd personally be confused if I saw this with no explanation.

joshlf (changes requested):
Amazing, thank you for fixing this! I'm not sure I understand why this fix works, though?

.github/actions/cache/action.yml
joshlf: Also, including the link you mentioned above would be a great breadcrumb: https://doc.rust-lang.org/cargo/guide/cargo-home.html#caching-the-cargo-home-in-ci

sl4m:
> Amazing, thank you for fixing this! I'm not sure I understand why this fix works, though?

Ok, I think I know what's really going on. My change worked in my fork because I did not have existing caches.

I believe `cargo-nextest` was recently introduced in #1307? If so, it may be a result of an old cache. A new cache is created only if `Cargo.toml` changes (see [here](https://github.com/google/zerocopy/blob/4b18d5b490b159abb72ded9fc334ea1a3025eae2/.github/actions/cache/action.yml#L18)) and since GitHub Actions cache are [immutable](https://github.com/actions/cache/blob/main/tips-and-workarounds.md#update-a-cache), it did not preserve the `cargo-nextest` installation.

`Cargo.toml` has since been updated via google-pr-creation bot, so the caches should now have `cargo-nextest`.

I checked a [recent build](https://github.com/google/zerocopy/actions/runs/9323668480/job/25667424858) and it looks like it's ignoring the `cargo-nextest` installation as expected:

``` 
Running Miri tests with 8 threads
    Updating crates.io index
     Ignored package `cargo-nextest v0.9.72` is already installed, use --force to override
Preparing a sysroot for Miri (target: aarch64-unknown-linux-gnu)... done
```

I think I will close this PR if you agree and you can probably close #1312.

joshlf:
Oh damn, great forensics work. Definitely sounds like that's what happened. Well I really appreciate your time on this! That actually uncovers more work we can do - namely, tweaking the cache key so it's caching more closely to what we actually need.

PR #1405, opened 2024-06-09
Implement Inline Size Assertion Annotations

### LINKED ISSUE #1329: Inline size assertion annotations
In some use-cases, it's valuable to define a format that allows for future expansion; e.g., reserves some number of bytes as "padding", that may then be used by later additions to the format so long as the total size of the record does not change; e.g.:
```rust
#[repr(C)]
#[derive(IntoBytes, FromBytes)]
struct Abc {
    field_a: usize,
    field_b: usize,
    padding: [u8; 240],
}
```

Right now, customers may use `static_assertions::assert_eq_size` to defend against accidental size changes. In the interest of sparing users from pulling in another dependency, and in the interest of making a best-practice readily available, we might want to consider supporting *inline* size assertions on non-generic types; e.g.:

```rust
#[repr(C)]
#[derive(IntoBytes, FromBytes)]
#[zerocopy::assert_size(256)]
struct Abc {
    field_a: usize,
    field_b: usize,
    padding: [u8; 240],
}
```

zoo868e:
Implem the procedural macro `inline_assert_size_eq` to check the size of a type at compile time. Additionally, add the test file `inline_assert_size_eq_failed.rs` to ensure that the macro tirggers a compile-time error when the size is not as expected Fixes #1329


zerocopy-derive/src/lib.rs
```diff
@@ -55,6 +55,45 @@ macro_rules! try_or_print {
...
+    input: proc_macro::TokenStream,
+) -> proc_macro::TokenStream {
+    let ast: Result<DeriveInput, _> = syn::parse(input.clone());
+    let expected_size: Result<usize, _> = args.to_string().trim().parse();
+    let mut ret: proc_macro::TokenStream = input;
+    if let (Ok(ast), Ok(size)) = (ast, expected_size) {
+        let name = &ast.ident;
+        ret = quote! {
+            #ast
+            static_assertions::const_assert!(core::mem::size_of::<#name>() == #size);
```
jswrenn, line 90: We can't defer to the `static_assertions` crate, since we don't want that to be a public dependency of zerocopy, nor do we want to be an explicit dependency of our consumers.

You should instead add a hygeinic helper macro to [`macro_util.rs`](https://github.com/google/zerocopy/blob/main/src/macro_util.rs) and invoke that in the proc macro.

zerocopy-derive/src/lib.rs
```diff
@@ -55,6 +55,45 @@ macro_rules! try_or_print {
...
+/// ```compile_fail
+/// # use zero_derive::inline_assert_size_eq;
+/// # use static_assertions;
+/// #[inline_assert_size_eq(1)]
+/// struct MyStruct {
+///     val: i32
+/// }
+/// ```
+#[proc_macro_attribute]
+pub fn inline_assert_size_eq(
```
jswrenn, line 79: I would prefer to call this `assert_size_eq`, and drop the `inline_` prefix.

jswrenn (changes requested):
Thanks for the contribution! This will need a few changes before we can consider merging it.

@joshlf, I think this functionality would be nicely complemented by a `#[assert_offset(N)]` field attribute macro (though I'd prefer us tackle that in a separate PR).

zerocopy-derive/src/lib.rs
zoo868e: There is `assert_size_eq` in [`macro_util.rs`](https://github.com/google/zerocopy/blob/main/src/macro_util.rs) which check if the size are equal between two types. I think we need another name. How about `assert_size_eq_val`?

jswrenn: The items in `macro_util.rs` are internal implementation details of public macros. We can rename these internal implementation details.

zerocopy-derive/src/lib.rs
jswrenn, line 1: I wonder if we could get this to work for generic types by instead making this a helper annotation on `KnownLayout`.

We would add a hidden, associated const to `KnownLayout` that is checked on any operation that reflects the layout:
```rust
trait KnownLayout {
    #[doc(hidden)]
    const HAS_EXPECTED_LAYOUT: bool = true;
    
    /* other items */
}
```
...and then this:
```rust
#[derive(KnownLayout)]
#[repr(C)]
#[zerocopy::assert(
     align = core::mem::size_of::<T>(),
     size = core::mem::size_of::<T>())
]
struct Foo<T> {
    #[zerocopy::assert(offset = 0, size = 2 * core::mem::size_of::<T>())]
    a: [T; 2],
}
```
...could expand to this:
```rust
impl<T> KnownLayout for Foo<T> {
    const HAS_EXPECTED_LAYOUT: bool = {
        assert!(::zerocopy::core_reexport::mem::size_of::<Self>() == core::mem::size_of::<T>());
        assert!(::zerocopy::core_reexport::mem::align_of::<Self>() == core::mem::align_of::<T>());
        assert!(::zerocopy::core_reexport::mem::offset_of!(Self, a) == 0);
        assert!(::zerocopy::core_reexport::mem::size_of::<[T; 2]>() == 2 * core::mem::size_of::<T>());
        true
    };
}
```

This approach is much more powerful, but it results in compile errors at use-sites, not at definition sites. My inclination is that this tradeoff is worth it. And, perhaps for non-generic types, we can automatically emit definition-site assertions. Thoughts, @joshlf?

zerocopy-derive/src/lib.rs
```diff
@@ -55,6 +55,45 @@ macro_rules! try_or_print {
...
+    args: proc_macro::TokenStream,
+    input: proc_macro::TokenStream,
+) -> proc_macro::TokenStream {
+    let ast: Result<DeriveInput, _> = syn::parse(input.clone());
+    let expected_size: Result<usize, _> = args.to_string().trim().parse();
+    let mut ret: proc_macro::TokenStream = input;
+    if let (Ok(ast), Ok(size)) = (ast, expected_size) {
```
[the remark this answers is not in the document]
zoo868e: Since parsing of `args` and `input` may fail, I have left the return value without using the macro. The size assertion will be performed only if these parameters are parsed correctly.

zerocopy-derive/src/lib.rs
```diff
@@ -55,6 +55,45 @@ macro_rules! try_or_print {
...
+///     val: i32
+/// }
+/// ```
+#[proc_macro_attribute]
+pub fn inline_assert_size_eq(
+    args: proc_macro::TokenStream,
+    input: proc_macro::TokenStream,
+) -> proc_macro::TokenStream {
+    let ast: Result<DeriveInput, _> = syn::parse(input.clone());
+    let expected_size: Result<usize, _> = args.to_string().trim().parse();
```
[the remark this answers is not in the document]
zoo868e: Since `proc_macro::TokenStream` cannot be directly appended to `quote` and can only be parsed as a string, I first parse it into a string and then convert it to a `usize`.
Another way is parse it into `proc_macro2::TokenStream`.

zerocopy-derive/src/lib.rs
zoo868e: Done

zerocopy-derive/src/lib.rs
zoo868e: Done

jswrenn:
Thanks for being so responsive to feedback! Just a heads up: It might be awhile before we're ready to merge this. We'll need to be absolutely certain about both the syntax and its interaction with generic types before we merge, and we are currently prioritizing the backlog of issues that are blocking the 0.8 release.

zoo868e:
Hi @jswrenn, I'm currently learning about derive macros and am trying to figure out how to make them work with generic types.

I'm struggling with how to capture and parse attributes such as `offset = 0, size = 2 * core::mem::size_of::<T>()`. I'm not entirely sure what to refer to these as or how to manage them, so I've been searching for some tutorials to help me out.


From the video [A Practical Introduction to Derive Macros with Attributes](https://www.youtube.com/watch?v=GFijwucFJqw) suggests that to parse attributes, one should use the `deluxe` crate and create a struct that derives from `deluxe::ExtractAttributes`. This struct would then declare the attributes as fields. However, I'm stumped by expressions like `2 * core::mem::size_of::<T>()` because I don't know their type, making it challenging to declare them in the struct.

Am I on the right track here? Could you offer any advice or guidance?

jswrenn:
I think we'd like to avoid taking additional dependencies for this. To do it with `syn`, you'd leverage its [`parse`](https://docs.rs/syn/latest/syn/parse/index.html) module, like so:
```rust
use syn::{
    Attribute,
    Expr,
    parse::{Parse, ParseStream},
    parse_quote,
    Token,
    punctuated::Punctuated,
};

/// Declarations the qualities of a layout.
///
/// These function both as checked assertions, and as hints to `KnownLayout`.
struct Hints {
    hints: Punctuated<Hint, Token![,]>,
}

impl Parse for Hints {
    fn parse(input: ParseStream) -> syn::Result<Self> {
        Ok(Self {
            hints: input.parse_terminated(Hint::parse, Token![,])?,
        })
    }
}

/// A declaration about type/field layout qualities.
struct Hint {
    kind: HintKind,
    colon: Token![:],
    expr: Expr,
}

impl Parse for Hint {
    fn parse(input: ParseStream) -> syn::Result<Self> {
        Ok(Self {
            kind: input.parse()?,
            colon: input.parse()?,
            expr: input.parse()?,
        })
    }
}

mod keywords {
    syn::custom_keyword!(align);
    syn::custom_keyword!(offset);
    syn::custom_keyword!(size);
}

/// The layout quality a hint is about.
///
/// A hint is either about alignment, offset or size.
enum HintKind {
    /// The alignment of a type or field.
    Align(keywords::align),
    /// The offset of a field.
    Offset(keywords::offset),
    /// The size of a type or field.
    Size(keywords::size),
}

impl Parse for HintKind {
    fn parse(input: ParseStream) -> syn::Result<Self> {
        let lookahead = input.lookahead1();
        if lookahead.peek(keywords::align) {
            input.parse().map(HintKind::Align)
        } else if lookahead.peek(keywords::offset) {
            input.parse().map(HintKind::Offset)
        } else if lookahead.peek(keywords::size) {
            input.parse().map(HintKind::Size)
        } else {
            Err(lookahead.error())
        }
    }
}

fn main() -> syn::Result<()> {
    let attr: Attribute = parse_quote! {
        #[hint(
            align: 4,
            offset: 8,
            size: 0,
        )]
    };
    
    if attr.path().is_ident("hint") {
        let hints: Hints = attr.parse_args()?;
        // do something with `hints`
    }

    Ok(())
}
```

jswrenn:
Then, to siphon the hints out of the AST, you'll want to leverage the [`visit`](https://docs.rs/syn/latest/syn/visit/index.html) module, like so ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f3b8709447f0284caa2bffe551133ff2)):

```rust
use syn::{
    parse,
    visit::{self, Visit},
    DeriveInput, Field, Generics, Ident,
};

/// Visits the AST and collects layout hints.
#[derive(Debug, Default)]
struct HintCollector<'ast> {
    hints: Vec<TargetedHint<'ast>>,
    errors: Vec<parse::Error>,
}

/// A hint and its target (i.e., type or field).
#[derive(Debug)]
struct TargetedHint<'ast> {
    target: HintTarget<'ast>,
    hints: Hints,
}

/// The target of a hint.
#[derive(Copy, Clone, Debug)]
enum HintTarget<'ast> {
    Type(&'ast DeriveInput),
    Field(&'ast Field),
}

impl<'ast> HintCollector<'ast> {
    /// Collect the `Hints` for the given target from its given attribute list.
    fn visit_attributes(&mut self, target: HintTarget<'ast>, attrs: &'ast Vec<Attribute>) {
        let hints = attrs.iter().filter(|attr| attr.path().is_ident("hint"));
        for attr in hints {
            match attr.parse_args::<Hints>() {
                Ok(hints) => {
                    self.hints.push(TargetedHint { target, hints });
                }
                Err(err) => {
                    self.errors.push(err);
                }
            }
        }
    }
}

impl<'ast> Visit<'ast> for HintCollector<'ast> {
    fn visit_derive_input(&mut self, i: &'ast DeriveInput) {
        self.visit_attributes(HintTarget::Type(i), &i.attrs);
        visit::visit_derive_input(self, i);
    }

    fn visit_field(&mut self, i: &'ast Field) {
        self.visit_attributes(HintTarget::Field(i), &i.attrs);
        visit::visit_field(self, i);
    }
}

fn main() -> syn::Result<()> {
    use syn::DeriveInput;

    let input: DeriveInput = parse_quote! {
        struct Foo<T, const N: usize> {
            #[hint(
                align: 2,
                offset: 0,
                size: 2,
            )]
            a: u16,
            #[hint(
                align: core::mem::align_of::<T>(),
                offset: 4,
                size: core::mem::size_of::<T>() * N,
            )]
            b: [T; N]
        }
    };

    let mut collector = HintCollector::default();
    collector.visit_derive_input(&input);

    println!("{:#?}", collector);

    Ok(())
}
```

PR #1413, opened 2024-06-11
Roll nightly toolchain

joshlf:


joshlf:
As-is, this won't work until 1.81 is stable. Instead, we could remove the newly-added version from the toolchain matrix, and add a hard-coded exception to the test-pinned-toolchains-match job so that that job does not complain about the mismatch. It would still be tested since we test nightly, and we could set a reminder to remove the exception once 1.81 is stabilized.

joshlf:
Superseded by #1441

PR #1441, opened 2024-06-20
Roll nightly toolchain

joshlf:
This supercedes #1413, taking the simpler approach of removing the `unused_qualifications` lint from our list of denied lints.


joshlf:
@jswrenn Note [this particular aspect of the diff](https://github.com/google/zerocopy/pull/1441/files#diff-5825b1b5e3ba149bbe3e3d6a11475afc26cddfdb9a2bff655562a6e71705b871). I think this is fine, but it'd be good to get a second pair of eyes on it.

jswrenn (approved):

jswrenn:
The regression on our error messages is unfortunate, but I don't believe there's a soundness issue here.

Issue #200, opened 2023-07-27
`write_to_prefix` generates panic path


jettr:
The `write_to_prefix` function is generating a panic path in our code, and I am pretty sure it is the `bytes[..size].copy_from_slice(self.as_bytes());` line. The compiler doesn't have enough information to elide the slice comparison check.

If the code was written like the following, the compiler should not generate a panic path

```rust
    fn write_to_prefix<B: ByteSliceMut>(&self, mut bytes: B) -> Option<()> {
        let source = self.as_bytes();
        if bytes.len() < source.len() {
            return None;
        }

        bytes[..size].copy_from_slice(source);
        Some(())
    }

```

joshlf:
What tool(s) are you using to inspect the output? Is it something I could run myself?

If you're able to, could you see whether this version of the code generates a panic path? It's similar to what you wrote, but keeps the same structure of the existing code:

```rust
    fn write_to_prefix<B: ByteSliceMut>(&self, mut bytes: B) -> Option<()> {
        let self_bytes = self.as_bytes();
        bytes.get_mut(..self_bytes.len())?.copy_from_slice(self_bytes);
        Some(())
    }
```

djkoloski:
There are also some clippy lints that we can turn on to help root out these kinds of silent panic paths:
- `clippy::indexing_slicing` can detect uses of indexing and slicing which can panic at runtime
- `clippy::arithmetic_side_effects` forces careful arithmetic operations to avoid panicking
These are extremely pedantic lints, so it's not something we can address in the near term. We can root out panics on demand when they're affecting users.

joshlf:
I minimized this example and [ran it through Godbolt](https://godbolt.org/z/189vWjEqv), and found that the panic is optimized out. My guess is that your issue was that the compiler ran out of optimization budget since you're presumably compiling more code than this toy example.

I filed #202 so that we can hopefully handle these cases in a more principled way than to just experiment to find out whether, in practice, panics are being optimized out.

jettr:
> What tool(s) are you using to inspect the output? Is it something I could run myself?

I was manually looking at the LST file for our binary trying to track down all of our panic paths when I discovered this. I think godbolt is a very good alternative to test out potential panic paths though. The example from the second comment should not contain any panic paths either.

jettr:
[...]

PR #1448, opened 2024-06-21
Use `copy_nonoverlapping` instead of `slice::copy_from`

jswrenn:
In doing so, we eliminate a potential panic path. Although I was not able to observe panic paths emitted in toy examples, they might be emitted in complex examples in which the optimizer is low on gas. Regardless, I expect this change will ease our future adoption of call-graph analysis techniques of potential panic paths.

Ref https://github.com/google/zerocopy/issues/200#issuecomment-2181544061

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3929,9 +3929,9 @@ pub unsafe trait IntoBytes {
         unsafe { slice::from_raw_parts_mut(slf.cast::<u8>(), len) }
     }
 
-    /// Writes a copy of `self` to `bytes`.
+    /// Writes a copy of `self` to `dst`.
     ///
-    /// If `bytes.len() != size_of_val(self)`, `write_to` returns `Err`.
+    /// If `dst.len() != size_of_val(self)`, `write_to` returns `Err`.
     ///
     /// # Examples
     ///
@@ -3977,21 +3977,27 @@ pub unsafe trait IntoBytes {
     /// ```
     #[must_use = "callers should check the return value to see if the operation succeeded"]
     #[inline]
-    fn write_to(&self, bytes: &mut [u8]) -> Result<(), SizeError<&Self, &mut [u8]>>
+    fn write_to(&self, dst: &mut [u8]) -> Result<(), SizeError<&Self, &mut [u8]>>
     where
         Self: Immutable,
     {
-        if bytes.len() != mem::size_of_val(self) {
-            return Err(SizeError::new(self));
+        let src = self.as_bytes();
+        if dst.len() == src.len() {
+            // SAFETY: Within this branch of the conditional, we have ensured
+            // that `dst.len()` is equal to `src.len()`. Neither the size of the
+            // source nor the size of the destination change between the above
+            // size check and the invocation of `copy_unchecked`.
+            unsafe { util::copy_unchecked(src, dst) }
+            Ok(())
+        } else {
+            Err(SizeError::new(self))
         }
-        bytes.copy_from_slice(self.as_bytes());
-        Ok(())
     }
 
-    /// Writes a copy of `self` to the prefix of `bytes`.
+    /// Writes a copy of `self` to the prefix of `dst`.
     ///
     /// `write_to_prefix` writes `self` to the first `size_of_val(self)` bytes
-    /// of `bytes`. If `bytes.len() < size_of_val(self)`, it returns `Err`.
+    /// of `dst`. If `dst.len() < size_of_val(self)`, it returns `Err`.
     ///
     /// # Examples
     ///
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -3981,11 +3981,21 @@ pub unsafe trait IntoBytes {
...
+            // SAFETY: Within this branch of the conditional, we have ensured
+            // that `bytes.len()` (the size of the destination) is equal to
+            // `mem::size_of_val(self)` the size of the source value. Neither
+            // the size of the source nor the size of the destination change
+            // between the above size check and the invocation of
+            // `copy_unchecked`. The size of `self` (the source) cannot change,
+            // because `Self` is bounded as `Immutable`, and the size of `bytes`
+            // cannot change, because we hold an exclusive reference to it and
+            // do not change its size.
+            unsafe { util::copy_unchecked(self.as_bytes(), bytes) }
```
joshlf, line 3994:
```suggestion
        let self_bytes = self.as_bytes();
        if bytes.len() == self_bytes.len() {
            // SAFETY: Within this branch of the conditional, we have ensured
            // that `bytes.len()` (the size of the destination) is equal to
            // `self_bytes.len()` (the size of the source). Neither
            // the size of the source nor the size of the destination change
            // between the above size check and the invocation of
            // `copy_unchecked`.
            unsafe { util::copy_unchecked(self_bytes, bytes) }
```

(And reflow the safety commnet)

This makes the safety argument simpler since we don't need to reason about the relationship between `size_of_val(self)` and `self.as_bytes()`. It is probably ever so slightly tougher on the optimizer, but I doubt it's enough to make a difference.

src/lib.rs
```diff
@@ -4044,7 +4054,18 @@ pub unsafe trait IntoBytes {
...
+                // equal to `mem::size_of_val(self)`.
+                //
+                // Neither the size of the source nor the size of the
+                // destination change between the above subslicing operation and
+                // the invocation of `copy_unchecked`. The size of `self` (the
+                // source) cannot change, because `Self` is bounded as
+                // `Immutable`, and the size of `bytes` cannot change, because
+                // we hold an exclusive reference to it and do not change its
+                // size.
+                unsafe { util::copy_unchecked(self.as_bytes(), bytes) }
```
joshlf, line 4068: Same here as above - do `self.as_bytes()` first and bounds check using `self_bytes.len()`.

src/lib.rs
```diff
@@ -4125,7 +4146,18 @@ pub unsafe trait IntoBytes {
...
+        // `bytes.len()` is equal to `mem::size_of_val(self)`.
+        //
+        // Neither the size of the source nor the size of the destination change
+        // between the above subslicing operation and the invocation of
+        // `copy_unchecked`. The size of `self` (the source) cannot change,
+        // because `Self` is bounded as `Immutable`, and the size of `bytes`
+        // cannot change, because we hold an exclusive reference to it and do
+        // not change its size.
+        unsafe {
+            util::copy_unchecked(self.as_bytes(), bytes);
```
joshlf, line 4159: Same here as above - do `self.as_bytes()` first and bounds check using `self_bytes.len()`.

src/util.rs
```diff
@@ -584,6 +584,29 @@ pub(crate) const fn min(a: NonZeroUsize, b: NonZeroUsize) -> NonZeroUsize {
...
+/// # Safety
+///
+/// The caller guarantees that `src.len() <= dst.len()`.
+#[inline(always)]
+pub(crate) unsafe fn copy_unchecked(src: &[u8], dst: &mut [u8]) {
+    debug_assert!(src.len() <= dst.len());
+    // SAFETY: This invocation satisfies the safety contract of
+    // copy_nonoverlapping [1]:
+    // - `src.as_ptr()` is trivially valid for reads of `src.len()` bytes
+    // - `dst.as_ptr()` is valid for writes of `dst.len.len()` bytes, because the
```
joshlf, line 598:
```suggestion
    // - `dst.as_ptr()` is valid for writes of `src.len()` bytes, because the
```

Typo?

joshlf (changes requested):

src/lib.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/6e1ee44e19fac8be1e2122528c3d88ebbbd00cd0..78d9f495abc6235b27f4d76b5f3c7e299990ef3e#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R3984)

src/lib.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/6e1ee44e19fac8be1e2122528c3d88ebbbd00cd0..78d9f495abc6235b27f4d76b5f3c7e299990ef3e#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R4050)

src/util.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/6e1ee44e19fac8be1e2122528c3d88ebbbd00cd0..78d9f495abc6235b27f4d76b5f3c7e299990ef3e#diff-4007187965c1fea4fd252a76f5f1007b2ea59f6d86e2f454edee547c5036be24R598)

joshlf (approved):

Issue #1310, opened 2024-05-19
Optimize CI


joshlf:
As of this writing, our CI tests (specified in `.github/workflows/ci.yml`) take [~5.5m to run end-to-end](https://github.com/google/zerocopy/actions/runs/9149665633) during PR development and [~19.5m to run end-to-end](https://github.com/google/zerocopy/actions/runs/9149612263) in the merge queue. This significantly affects developer velocity, especially when developing a sequence of features which stack (ie, one PR needs to land before the next PR can be seriously considered).

This task tracks optimizing our end-to-end CI latency. Anything is on the table!

Note that both the PR latency and the merge queue latency are on the table. The PR latency is obviously the more important metric, since PR tests may run multiple times during PR development. However, given that GitHub has no automated way to merge a stack of PRs, we often have to actively keep an eye on the merge queue in order to know when we can kick off the next PR's merge. For this reason, merge queue latency is important as well.

## Advice

As of this writing, we skip 5 out of 7 build targets and all Miri tests during PR development. Thus, the merge queue CI tests have somewhat different performance characteristics than PR CI tests.

In my own investigations, I've discovered the following:
- In the merge queue, the bottleneck seems to be the `build_test` job, which encompasses the primary test matrix (there are other ancillary jobs such as `kani`, `check_fmt`, etc; these do not appear to be the bottleneck)
- Among individual matrix jobs, the distribution of times appears to be highly bimodal:
  - Most matrix jobs take ~1-2m to complete
  - Some matrix jobs take ~13m to complete
  - What distinguishes the two appears to be Miri tests, which are run only in the latter (~13m) group
- It also seems to take a few minutes just to spawn all of the ~200 jobs in the matrix (before they start executing)

We've already done some work to speed up Miri test execution (recently, #1307, #1308, and #1313). There is probably a lot more that could be done there.

There are probably also a lot of other optimization opportunities besides Miri; I just haven't taken the time to investigate in detail.

See also: #1312, #1314

## Failed attempts

I tried these, but found no speedup, or wasn't able to get them working:
- #1311 - no measurable speedup
- #1309 - confusing build failures

Aditya-PS-05:
@joshlf , is this issue still open. Can I work on it?

PR #1471, opened 2024-07-01
Added conditional dependancy downloading in CI

dorryspears:
Progresses #1310

This will skip downloading the cargo dependencies if the cache is present. The cache gets refreshed based on the hash of the `Cargo.toml` file, so when that file doesn't change, this could save 2 - 3 minutes per run.

### DIFF
--- a/.github/actions/cache/action.yml
+++ b/.github/actions/cache/action.yml
@@ -8,10 +8,15 @@
 
 name: Cache
 description: 'Caches cargo dependencies'
+outputs:
+  cache-hit:
+    description: 'Cache hit status'
+    value: ${{ steps.cache.outputs.cache-hit }}
 runs:
   using: composite
   steps:
     - uses: actions/cache@v4
+      id: cache
       with:
         path: |
           ~/.cargo/
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -483,9 +483,11 @@ jobs:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
 
       - name: Populate cache
+        id: populate-cache
         uses: ./.github/actions/cache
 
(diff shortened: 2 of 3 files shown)


dorryspears:
@joshlf If some of these commands need to be run regardless of the cache filling, I can split them up.

I have a feeling we still may want to run 
- `cargo check --workspace --tests`
- `cargo kani setup`


https://github.com/google/zerocopy/blob/8151bdbe5eae92b119d64c8f60d00279d75eae0b/.github/workflows/ci.yml#L507-L512

joshlf:
Hmm my thinking (and I should probably add this as a comment somewhere) was that, if there's a cache hit, then the "populate cache" step should be quick because all of the `cargo` tools we run won't need to actually download. But there might be steps that are unconditionally re-run.

Can you see if it's possible to coax GitHub into running this in such a way that one of the runs has a cache hit so we can see what the timing looks like w/ and w/o this PR?

dorryspears:
@joshlf yes of course, and thanks for the catch as I think you are right. I don't think this will save nearly as much time as I thought. It might save **20s** since it won't run that download step at all.

## With PR
- cache hit **7s** [view](https://github.com/google/zerocopy/actions/runs/9863671980/job/27236952033)
- cache not hitting **3m 9s** [view](https://github.com/google/zerocopy/actions/runs/9863624979/job/27236805542)

## Without PR
- cache hit **26s** [view](https://github.com/dorryspears/zerocopy/actions/runs/9863252798/job/27235610067)
- cache not hitting **3m 10s** [view](https://github.com/dorryspears/zerocopy/actions/runs/9863210513/job/27235474617)

joshlf:
Ah, bummer. Thanks for running the numbers, though! I don't think 20s justifies the complexity, so I'm gonna close this, but thanks for your work on it, and keep the PRs coming!

PR #1526, opened 2024-07-16
Hide large enums from rust-analyzer

djkoloski:
The large enums in zerocopy-derive's `enum_from_bytes` test make rust-analyzer hang during analysis. This change adds a VSCode config that allows us to hide that test from rust-analyzer. The test is still run as normal.

(split out from #879)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,8 @@
+{
+    "rust-analyzer.cargo.cfgs": {
+        // This adds an extra cfg to rust-analyzer so we can prevent it from
+        // analyzing tests with a lot of generated code. These can cause it to
+        // hang or run slowly in general, which is bad for workflows.
+        "rust_analyzer": null,
+    }
+}
--- a/zerocopy-derive/tests/enum_from_bytes.rs
+++ b/zerocopy-derive/tests/enum_from_bytes.rs
@@ -10,6 +10,8 @@
 #![no_implicit_prelude]
 #![allow(warnings)]
 #![deny(deprecated)]
+// The giant enums in this file cause rust-analyzer to hang during analysis.
+#![cfg(not(rust_analyzer))]
 
 include!("include.rs");
 


.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "rust-analyzer.check.extraEnv": {
+        "RUSTFLAGS": "--cfg rust_analyzer",
```
joshlf, line 3: Are we sure that this adds to `RUSTFLAGS` (as opposed to clobbers it?)

joshlf (approved):
Approved pending one comment.

.vscode/settings.json
joshlf: Discussed offline. @djkoloski and I determined that this clobbers `RUSTFLAGS`, but that that's not particularly problematic since it should be rare for rust-analyzer to be started from a context in which the `RUSTFLAGS` environment variable is already set and the user wants its behavior preserved through rust-analyzer. E.g., it will be common for rust-analyzer to be started by an editor which is in turn run not from a shell (such as VSCode).

joshlf (approved):

.vscode/settings.json
djkoloski: Found a better way to do this by reading the RA manual. This is no longer a problem.

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "rust-analyzer.cargo.cfgs": {
+        "rust_analyzer": null,
```
joshlf, line 3: Is it possible to comment in these files? It'd be good to explain what this does.

joshlf (approved):
Approved, but take a look at the comment and address it if possible. Feel free to merge without re-review.

.vscode/settings.json
djkoloski: Done

PR #1527, opened 2024-07-16
Improve zero discriminant detection for enums

djkoloski:
Enums with negative explicit discriminants can allow variants which are not first but have an implicit discriminant to be assigned a discriminant of zero.

### DIFF
--- a/zerocopy-derive/src/lib.rs
+++ b/zerocopy-derive/src/lib.rs
@@ -38,7 +38,8 @@ use {
     quote::quote,
     syn::{
         parse_quote, parse_quote_spanned, Data, DataEnum, DataStruct, DataUnion, DeriveInput,
-        Error, Expr, ExprLit, GenericParam, Ident, Lit, Path, Type, WherePredicate,
+        Error, Expr, ExprLit, ExprUnary, GenericParam, Ident, Lit, Path, Type, UnOp,
+        WherePredicate,
     },
 };
 
@@ -634,6 +635,84 @@ fn derive_from_zeros_struct(ast: &DeriveInput, strct: &DataStruct) -> proc_macro
 // - all of its variants are fieldless
 // - one of the variants has a discriminant of `0`
 
+// Returns `Ok(index)` if variant `index` of the enum has a discriminant of
+// zero. If `Err(bool)` is returned, the boolean is true if the enum has unknown
+// discriminants (e.g. discriminants set to const expressions which we can't
+// evaluate in a proc macro). If the enum has unknown discriminants, then it
+// might have a zero variant that we just can't detect.
+fn find_zero_variant(enm: &DataEnum) -> Result<usize, bool> {
+    // Discriminants can be anywhere in the range [i128::MIN, u128::MAX] because
+    // the discriminant type may be signed or unsigned. Since we only care about
+    // tracking the discriminant when it's less than or equal to zero, we can
+    // avoid u128 -> i128 conversions and bounds checking by making the "next
+    // discriminant" value implicitly negative.
+    // Technically 64 bits is enough, but 128 is better for future compatibility
+    // with https://github.com/rust-lang/rust/issues/56071
+    let mut next_negative_discriminant = Some(0);
+
+    // Sometimes we encounter explicit discriminants that we can't know the
+    // value of (e.g. a constant expression that requires evaluation). These
+    // could evaluate to zero or a negative number, but we can't assume that
+    // they do (no false positives allowed!). So we treat them like strictly-
+    // positive values that can't result in any zero variants, and track whether
+    // we've encountered any unknown discriminants.
+    let mut has_unknown_discriminants = false;
+
+    for (i, v) in enm.variants.iter().enumerate() {
+        match v.discriminant.as_ref() {
+            // Implicit discriminant
+            None => {
+                match next_negative_discriminant.as_mut() {
+                    Some(0) => return Ok(i),
+                    // n is nonzero so subtraction is always safe
+                    Some(n) => *n -= 1,
+                    None => (),
+                }
+            }
+            // Explicit positive discriminant
+            Some((_, Expr::Lit(ExprLit { lit: Lit::Int(int), .. }))) => {
+                match int.base10_parse::<u128>().ok() {
+                    Some(0) => return Ok(i),
+                    Some(_) => next_negative_discriminant = None,
+                    None => {
+                        // Numbers should never fail to parse, but just in case:
+                        has_unknown_discriminants = true;
+                        next_negative_discriminant = None;
+                    }
+                }
+            }
+            // Explicit negative discriminant
+            Some((_, Expr::Unary(ExprUnary { op: UnOp::Neg(_), expr, .. }))) => match &**expr {
+                Expr::Lit(ExprLit { lit: Lit::Int(int), .. }) => {
+                    match int.base10_parse::<u128>().ok() {
+                        Some(0) => return Ok(i),
+                        // x is nonzero so subtraction is always safe
+                        Some(x) => next_negative_discriminant = Some(x - 1),
+                        None => {
+                            // Numbers should never fail to parse, but just in
+                            // case:
+                            has_unknown_discriminants = true;
+                            next_negative_discriminant = None;
+                        }
+                    }
+                }
+                // Unknown negative discriminant (e.g. const repr)
+                _ => {
+                    has_unknown_discriminants = true;
+                    next_negative_discriminant = None;
+                }
+            },
+            // Unknown discriminant (e.g. const expr)
+            _ => {
+                has_unknown_discriminants = true;
+                next_negative_discriminant = None;
+            }
+        }
+    }
+
+    Err(has_unknown_discriminants)
+}
+
 fn derive_from_zeros_enum(ast: &DeriveInput, enm: &DataEnum) -> proc_macro2::TokenStream {
     if !enm.is_fieldless() {
         return Error::new_spanned(ast, "only field-less enums can implement FromZeros")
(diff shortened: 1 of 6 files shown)


zerocopy-derive/src/lib.rs
```diff
@@ -634,6 +635,47 @@ fn derive_from_zeros_struct(ast: &DeriveInput, strct: &DataStruct) -> proc_macro
 // - all of its variants are fieldless
 // - one of the variants has a discriminant of `0`
 
+fn find_zero_variant<'a>(variants: impl IntoIterator<Item = &'a Variant>) -> Option<usize> {
+    let mut current_implicit_discriminant = Some(0isize);
```
joshlf, line 639: Maybe we should call this `next_implicit_discriminant` since it describes the descriminant of the next variant we're about to process?

zerocopy-derive/src/lib.rs
```diff
@@ -634,6 +635,47 @@ fn derive_from_zeros_struct(ast: &DeriveInput, strct: &DataStruct) -> proc_macro
...
+            None => {
+                if current_implicit_discriminant == Some(0) {
+                    return Some(i);
+                }
+            }
+            // Explicit discriminant
+            Some((_, Expr::Lit(ExprLit { lit: Lit::Int(int), .. }))) => {
+                match int.base10_parse::<usize>().ok() {
+                    Some(0) => return Some(i),
+                    Some(x) => current_implicit_discriminant = Some(x as isize),
```
joshlf, line 655: Either use `isize::from` or comment justifying that this `as` is okay?

zerocopy-derive/src/lib.rs
```diff
@@ -634,6 +635,47 @@ fn derive_from_zeros_struct(ast: &DeriveInput, strct: &DataStruct) -> proc_macro
...
+                    None => current_implicit_discriminant = None,
+                }
+            }
+            // Explicit negative discriminant
+            Some((_, Expr::Unary(ExprUnary { op: UnOp::Neg(_), expr, .. }))) => match &**expr {
+                Expr::Lit(ExprLit { lit: Lit::Int(int), .. }) => {
+                    match int.base10_parse::<usize>().ok() {
+                        Some(0) => return Some(i),
+                        Some(x) => current_implicit_discriminant = Some(-(x as isize)),
```
joshlf, line 664: Either use `isize::from` or comment justifying that this `as` is okay?

zerocopy-derive/src/lib.rs
```diff
@@ -634,6 +635,47 @@ fn derive_from_zeros_struct(ast: &DeriveInput, strct: &DataStruct) -> proc_macro
...
+                        None => current_implicit_discriminant = None,
+                    }
+                }
+                _ => current_implicit_discriminant = None,
+            },
+            _ => current_implicit_discriminant = None,
+        }
+        if let Some(n) = current_implicit_discriminant.as_mut() {
+            *n += 1;
```
joshlf, line 673: Can we bounds check this to make sure it won't overflow?

zerocopy-derive/src/lib.rs
```diff
@@ -634,6 +635,47 @@ fn derive_from_zeros_struct(ast: &DeriveInput, strct: &DataStruct) -> proc_macro
 // - all of its variants are fieldless
 // - one of the variants has a discriminant of `0`
 
+fn find_zero_variant<'a>(variants: impl IntoIterator<Item = &'a Variant>) -> Option<usize> {
```
joshlf, line 638: How does this interact with reprs?
- Are there any reprs this is invalid for?
- Does it handle discriminants that don't fit in the given repr? E.g., `#[repr(u8)]` combined with `A = 300`

joshlf (changes requested):

zerocopy-derive/src/lib.rs
djkoloski: Done

zerocopy-derive/src/lib.rs
djkoloski: Rewrote some of the implementation to avoid these edge cases

zerocopy-derive/src/lib.rs
djkoloski: Done, same as above.

zerocopy-derive/src/lib.rs
djkoloski: Done, same as above

zerocopy-derive/src/lib.rs
djkoloski: Reworked this to handle signed and unsigned discriminants of all sizes.

zerocopy-derive/src/lib.rs
```diff
@@ -644,25 +717,23 @@ fn derive_from_zeros_enum(ast: &DeriveInput, enm: &DataEnum) -> proc_macro2::Tok
...
-            ast,
-            "FromZeros only supported on enums with a variant that has a discriminant of `0`",
-        )
-        .to_compile_error();
+    if let Err(has_unknown_variants) = find_zero_variant(enm) {
+        if has_unknown_variants {
+            return Error::new_spanned(
+                ast,
+                "FromZeros only supported on enums with a variant that has a discriminant of `0`\n\
+                This enum has discriminants that the derive macro cannot evaluate (e.g. const \
```
joshlf, line 725:
```suggestion
                This enum has discriminants that `#[derive(FromZeros)]` cannot evaluate (e.g. const \
```

Just to be friendly to people who aren't familiar with the terminology (ie, new Rustaceans who just saw that they needed to do this derive thing and don't know that they're technically called macros). I don't feel strongly about this suggestion - feel free to ignore.

zerocopy-derive/src/lib.rs
```diff
@@ -644,25 +717,23 @@ fn derive_from_zeros_enum(ast: &DeriveInput, enm: &DataEnum) -> proc_macro2::Tok
...
+                expressions). If one of the variants has a discriminant of zero, set it to zero \
+                directly.",
```
joshlf, line 727: This isn't exactly the right advice. It might be that they have an enum like:

```
const NEGATIVE_ONE: isize = -1;

#[repr(isize)]
enum Foo {
    A = NEGATIVE_ONE,
    B,
}
```

...in which case they need to set `A = -1`, not `A = 0`.

zerocopy-derive/src/lib.rs
```diff
@@ -634,6 +635,78 @@ fn derive_from_zeros_struct(ast: &DeriveInput, strct: &DataStruct) -> proc_macro
...
+            }
+            // Explicit positive discriminant
+            Some((_, Expr::Lit(ExprLit { lit: Lit::Int(int), .. }))) => {
+                match int.base10_parse::<usize>().ok() {
+                    Some(0) => return Ok(i),
+                    Some(_) => next_negative_discriminant = None,
+                    None => {
+                        has_unknown_discriminants = true;
+                        next_negative_discriminant = None;
+                    }
```
joshlf, line 678: Under what conditions can we arrive here? Wouldn't it just be a case where Rust has permitted, e.g., `#[repr(u128)]`, and the user has provided a discriminant which can't fit in `usize`? It must at least be a valid positive number since it's inside a `Lit::Int`.

If that's right, then we should consider perhaps expanding the error messages we can emit to mentioning this case, or maybe even tracking "a number we couldn't parse" as a separate condition.

zerocopy-derive/src/lib.rs
```diff
@@ -634,6 +635,78 @@ fn derive_from_zeros_struct(ast: &DeriveInput, strct: &DataStruct) -> proc_macro
...
+            Some((_, Expr::Unary(ExprUnary { op: UnOp::Neg(_), expr, .. }))) => match &**expr {
+                Expr::Lit(ExprLit { lit: Lit::Int(int), .. }) => {
+                    match int.base10_parse::<usize>().ok() {
+                        Some(0) => return Ok(i),
+                        // x is nonzero so subtraction is always safe
+                        Some(x) => next_negative_discriminant = Some(x - 1),
+                        None => {
+                            has_unknown_discriminants = true;
+                            next_negative_discriminant = None;
+                        }
```
joshlf, line 691: Same here as above.

zerocopy-derive/src/lib.rs
```diff
@@ -634,6 +635,78 @@ fn derive_from_zeros_struct(ast: &DeriveInput, strct: &DataStruct) -> proc_macro
...
+                    }
+                }
+                _ => {
+                    has_unknown_discriminants = true;
+                    next_negative_discriminant = None;
+                }
```
joshlf, line 697:
```suggestion
                // Unknown discriminant (e.g. const expr)
                _ => {
                    has_unknown_discriminants = true;
                    next_negative_discriminant = None;
                }
```

joshlf (changes requested):

zerocopy-derive/src/lib.rs
djkoloski: `#[repr(u128)]` enums are unstable: https://github.com/rust-lang/rust/issues/56071. I modified this to use `u64` instead of `usize` as a result. I think this should cover all the cases where we could fail to parse, so this branch shouldn't be reachable in practice. That said, I think we'd prefer to fail gracefully rather than error here.

zerocopy-derive/src/lib.rs
djkoloski: Done

zerocopy-derive/src/lib.rs
djkoloski: Clarified this message

zerocopy-derive/src/lib.rs
djkoloski: I like the wording change

zerocopy-derive/src/lib.rs
joshlf: True, but if it is stabilized, then this code may be used with `#[repr(u128)]` enums. Maybe we could just use `u128`s here (and document that it's for forwards-compatibility with https://github.com/rust-lang/rust/issues/56071)? I'm not worried about `#[repr(u256)]` enums, especially considering that this is just an issue of how nice our error messages are, not a correctness issue.

joshlf: Sorry, waiting on this change or a response to this comment before approving.

djkoloski: Sure, bumped it up to 128 bits

joshlf (approved):

PR #1564, opened 2024-07-29
Parse the package manifest directly for toolchains

djkoloski:
This prevents the lockfile for the workspace from being generated by a newer version of cargo, which may choose more up-to-date dependencies than our msrv toolchain can compile.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ zerocopy-panic-in-const = "1.57.0"
 [package.metadata.ci]
 # The versions of the stable and nightly compiler toolchains to use in CI.
 pinned-stable = "1.80.0"
-pinned-nightly = "nightly-2024-07-25"
+pinned-nightly = "nightly-2024-07-28"
 
 [package.metadata.docs.rs]
 all-features = true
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -78,11 +78,7 @@ zerocopy-derive = { version = "=0.8.0-alpha.16", path = "zerocopy-derive" }
 [dev-dependencies]
 assert_matches = "1.5"
 itertools = "0.11"
-# We don't use this directly, but trybuild does. On the MSRV toolchain, the
-# version resolver fails to select any version for once_cell unless we
-# depend on it directly.
-once_cell = "=1.9"
```
joshlf, line 84: Why is this no longer needed?

joshlf (approved):

Cargo.toml
djkoloski: The version resolver failing for syn and once_cell was a bug we caused. When we ran `cargo metadata` with the stable toolchain, the lockfile wasn't generated yet so it generated a lockfile and resolved packages using the stable toolchain. Later, when we went to compile/run with the MSRV toolchain, the lockfile already existed so it reused it. However, that lockfile was incompatible with the MSRV toolchain since it allowed versions of crates with an MSRV up to stable. This change avoids ever creating that "stable lockfile" and so the workspace packages are only ever resolved with the toolchain we're testing. The MSRV toolchain knows how to select compatible versions of `once_cell` and `syn`.

PR #1574, opened 2024-08-06
Strengthen soundness proof for size_of_val_raw

joshlf:
Previously, we needed to rely on the fact that the instance of any valid Rust type with 0 elements has a size (in number of bytes) which is not greater than `isize::MAX`. Providing this as a guarantee turned out to be controversial. [1]

This was made possible by https://github.com/rust-lang/rust/pull/126152.

[1] https://github.com/rust-lang/unsafe-code-guidelines/issues/465#issuecomment-2133405687

### DIFF
--- a/src/macro_util.rs
+++ b/src/macro_util.rs
@@ -132,7 +132,8 @@ macro_rules! trailing_field_offset {
         let min_size = {
             let zero_elems: *const [()] =
                 $crate::macro_util::core_reexport::ptr::slice_from_raw_parts(
-                    #[allow(clippy::incompatible_msrv)] // Work around https://github.com/rust-lang/rust-clippy/issues/12280
+                    // Work around https://github.com/rust-lang/rust-clippy/issues/12280
+                    #[allow(clippy::incompatible_msrv)]
                     $crate::macro_util::core_reexport::ptr::NonNull::<()>::dangling()
                         .as_ptr()
                         .cast_const(),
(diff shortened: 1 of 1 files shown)


src/macro_util.rs
```diff
@@ -144,16 +144,16 @@ macro_rules! trailing_field_offset {
...
-            //     initialized to 0 above, and Rust guarantees that no type's
-            //     minimum size may overflow `isize`. [1]
+            //   - If `$ty` is a slice DST, we have constructed `zero_elems` to
+            //     have zero trailing slice elements. Per the `size_of_val_raw`
+            //     docs, "For the special case where the dynamic tail length is
+            //     0, this function is safe to call." [2]
             //
             // [1] TODO(#429),
             // TODO(https://github.com/rust-lang/unsafe-code-guidelines/issues/465#issuecomment-1782206516):
             // Citation for this?
```
jswrenn, line 154: Is [1] still referenced somewhere after your revision?

joshlf: Oops, fixed.

jswrenn (approved):

joshlf:
> Can we just use `offset_of!` now that it's stable?

A few issues:
- It doesn't support our MSRV
- It doesn't support DSTs

Note that `size_of_val_raw` isn't stable at all - this code is only tested on nightly (and not actually used anywhere) to make sure we're forwards-compatible with using it in the future. So DST support is really the issue that matters.

djkoloski (approved):

PR #1601, opened 2024-08-29
never construct value on stack in new_box_zeroed

Freax13:
On lower opt-levels the compiler might not optimize out the `layout.size() == 0` branch and emits code for the if-body. This will cause a stack allocation for `Self`. Avoid calling new_zeroed() and directly construct the Box from a dangling pointer instead.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2128,7 +2128,27 @@ pub unsafe trait FromZeros: TryFromBytes {
         // no allocation, but `Box` does require a correct dangling pointer.
         let layout = Layout::new::<Self>();
         if layout.size() == 0 {
-            return Box::new(Self::new_zeroed());
+            // Construct the `Box` from a dangling pointer to avoid calling
+            // `Self::new_zeroed`. This ensures that stack space is never
+            // allocated for `Self` even on lower opt-levels where this branch
+            // might not get optimized out.
+
+            // SAFETY: Per [1], when `T` is a ZST, `Box<T>`'s only validity
+            // requirements are that the pointer is non-null and sufficiently
+            // aligned. Per [2], `NonNull::dangling` produces a pointer which
+            // is sufficiently aligned. Since the produced pointer is a
+            // `NonNull`, it is non-null.
+            //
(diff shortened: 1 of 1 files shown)


joshlf:
Thanks for this! It looks reasonable, although I have a concern with the cited Box docs: https://github.com/rust-lang/unsafe-code-guidelines/issues/529

Once that's resolved, I'll approve.

joshlf:
Blocked on https://github.com/rust-lang/rust/pull/129748.

src/lib.rs
```diff
@@ -2128,7 +2128,12 @@ pub unsafe trait FromZeros: TryFromBytes {
         // no allocation, but `Box` does require a correct dangling pointer.
         let layout = Layout::new::<Self>();
         if layout.size() == 0 {
-            return Box::new(Self::new_zeroed());
+            // SAFETY: Contructing a Box to a ZST from a dangling pointer is
+            // explicitly allowed:
+            // https://doc.rust-lang.org/std/boxed/index.html#memory-layout
+            unsafe {
+                return Box::from_raw(NonNull::dangling().as_ptr());
+            }
```
joshlf, line 2136: Can you also add a comment that explains why we do this instead of `Box::new(Self::new_zeroed())`? The argument in your commit message makes sense, but I don't think it's obvious enough that we should expect future readers to figure it out from context.

joshlf (changes requested):
Okay, https://github.com/rust-lang/rust/pull/129748 merged, so we're good to continue here.

When you make changes, please squash everything together into the same commit and force-push instead of adding new commits.

joshlf (approved):
Great, thanks again for putting this PR up!

joshlf:
Update on this: I've published [0.8.0-alpha.18](https://crates.io/crates/zerocopy/0.8.0-alpha.18), which includes this change. It should be trivial to backport to 0.7 as well (in progress in #1604), but we're having unrelated CI issues that are blocking it from merging. We probably won't have time to burn those CI issues down for at least a week or two, and possibly longer.

Issue #1358, opened 2024-05-30
Accepted risk: `Box` et al containing `UnsafeCell`s


joshlf:
## Overview

As part of #1155, we need guarantees that certain standard library types such as `Box` and `NonZeroXxx` and even built-ins such as `usize` do not contain `UnsafeCell`s. Per https://github.com/rust-lang/unsafe-code-guidelines/issues/492, it is difficult to get this guarantee, but it's also incredibly obvious that it's true and would be absurd to change (and certainly absurd in a way that doesn't just result in these types containing `UnsafeCell`s, but in a way that permits interior mutation to be exercised, which is really what `Immutable` cares about).

This issue tracks that this is an accepted risk, and we're going to proceed with 0.8 anyway.

## Progress

- [ ] Update code TODOs to point to this issue instead
- [ ] Fill out details section below
- [ ] Come up with a better title for this issue

## Details

TODO

Issue #896, opened 2024-02-17
Tracking issue for outstanding safety proofs blocking 0.8


joshlf:
This issue tracks all safety proofs which are not complete and need to be completed before we perform a stable release. All such proofs are marked as `// TODO(#896)`.

---

Outstanding TODOs as of b96b620e4bd17ec0c0e5b371dd0180391ace8107 (2024-04-29):

https://github.com/google/zerocopy/blob/b96b620e4bd17ec0c0e5b371dd0180391ace8107/src/lib.rs#L3831-L3870

https://github.com/google/zerocopy/blob/b96b620e4bd17ec0c0e5b371dd0180391ace8107/src/lib.rs#L3908-L3920

https://github.com/google/zerocopy/blob/b96b620e4bd17ec0c0e5b371dd0180391ace8107/src/lib.rs#L3984-L3987

https://github.com/google/zerocopy/blob/b96b620e4bd17ec0c0e5b371dd0180391ace8107/src/lib.rs#L4009-L4029

https://github.com/google/zerocopy/blob/b96b620e4bd17ec0c0e5b371dd0180391ace8107/src/lib.rs#L4267-L4270

https://github.com/google/zerocopy/blob/b96b620e4bd17ec0c0e5b371dd0180391ace8107/src/util.rs#L107-L119

https://github.com/google/zerocopy/blob/b96b620e4bd17ec0c0e5b371dd0180391ace8107/src/util.rs#L149-L161

https://github.com/google/zerocopy/blob/b96b620e4bd17ec0c0e5b371dd0180391ace8107/src/util.rs#L194-L206

https://github.com/google/zerocopy/blob/b96b620e4bd17ec0c0e5b371dd0180391ace8107/src/util.rs#L349-L401

PR #1615, opened 2024-09-07
Write some safety proofs

joshlf:
Adds some informal proofs as accepted risks per #1358

Makes progress on #896

### DIFF
--- a/src/impls.rs
+++ b/src/impls.rs
@@ -307,8 +307,8 @@ safety_comment! {
     ///   `NonZeroU16` is guaranteed to have the same layout and bit validity as
     ///   `u16` with the exception that `0` is not a valid instance.
     ///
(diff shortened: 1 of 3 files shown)


src/impls.rs
```diff
@@ -435,7 +433,9 @@ safety_comment! {
 
 safety_comment! {
     /// SAFETY:
-    /// TODO(#896): Write this safety proof before the next stable release.
+    /// Function objects (`fn()` and `extern "C" fn()`) self-evidently do not
```
jswrenn, line 436:
```suggestion
    /// Function pointers (`fn()` and `extern "C" fn()`) self-evidently do not
```
Function *objects* describes the broader set that includes closures, which may contain `UnsafeCell`s.

jswrenn (approved):

PR #1622, opened 2024-09-08
Implement `try_transmute_{ref,mut}`

jswrenn:
Makes progress towards #5.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -60,10 +60,12 @@
 //!
 //! - ([`try_`][try_transmute])[`transmute`] (conditionally) converts a value of
 //!   one type to a value of another type of the same size
-//! - [`transmute_mut`] converts a mutable reference of one type to a mutable
-//!   reference of another type of the same size
-//! - [`transmute_ref`] converts a mutable or immutable reference
-//!   of one type to an immutable reference of another type of the same size
+//! - ([`try_`][try_transmute_mut])[`transmute_mut`] (conditionally) converts a
+//!   mutable reference of one type to a mutable reference of another type of
+//!   the same size
+//! - ([`try_`][try_transmute_ref])[`transmute_ref`] (conditionally) converts a
+//!   mutable or immutable reference of one type to an immutable reference of
+//!   another type of the same size
 //!
 //! These macros perform *compile-time* alignment and size checks, but cannot be
 //! used in generic contexts. For generic conversions, use the methods defined
@@ -4750,7 +4752,7 @@ macro_rules! transmute_mut {
 /// This macro behaves like an invocation of this function:
 ///
 /// ```ignore
-/// const fn try_transmute<Src, Dst>(src: Src) -> Result<Dst, ValidityError<Src, Dst>>
+/// fn try_transmute<Src, Dst>(src: Src) -> Result<Dst, ValidityError<Src, Dst>>
 /// where
 ///     Src: IntoBytes,
 ///     Dst: TryFromBytes,
@@ -4784,10 +4786,9 @@ macro_rules! transmute_mut {
 ///
 /// // 2u8 → bool = error
 /// assert!(matches!(
-///     try_transmute!(3u8),
+///     try_transmute!(2u8),
 ///     Result::<bool, _>::Err(ValidityError { .. })
 /// ));
-///
 /// ```
 #[macro_export]
 macro_rules! try_transmute {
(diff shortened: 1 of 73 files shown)


src/macro_util.rs
```diff
@@ -500,6 +500,74 @@ where
...
+    let src = ManuallyDrop::new(src);
+
+    // SAFETY: By contract on `is_src_valid`, we have confirmed both that `Dst`
+    // is no larger than `Src`, and that `src` is a bit-valid instance of `Dst`.
+    // These conditions are preserved through the `ManuallyDrop<T>` wrapper,
+    // which is documented to have identical and layout bit validity to its
+    // inner value [1].
+    //
+    // [1]: https://doc.rust-lang.org/std/mem/struct.ManuallyDrop.html
+    Ok(unsafe { mem::transmute_copy(&*src) })
```
jswrenn, line 533: The `transmute_copy` here is overly awkward (I'll fix that), but the bigger issue is the `Sized` bound. If we retain it, we can keep the static size errors. If we ditch it, we can handle DSTs (but size errors for all types will be dynamic, not static). What should we do?

jswrenn: perhaps we support both:
- `try_transmute_ref!(src)` with return type `Result<Dst, ValidityError<&Src, Dst>>` and static size/alignment checking
- `try_transmute_ref!(unsized src)` with return type `Result<Dst, TryCastError<&Src, Dst>>` (dynamic size/alignment checking), but support for DSTs.

src/lib.rs
```diff
@@ -4814,6 +4815,231 @@ macro_rules! try_transmute {
...
+///
+/// However, unlike a function, this macro can only be invoked when the types of
+/// `Src` and `Dst` are completely concrete. The types `Src` and `Dst` are
+/// inferred from the calling context; they cannot be explicitly specified in
+/// the macro invocation.
+///
+/// Note that the `Src` produced by the expression `$e` will *not* be dropped.
+/// Semantically, its bits will be copied into a new value of type `Dst`, the
+/// original `Src` will be forgotten, and the value of type `Dst` will be
+/// returned.
```
joshlf, line 4845: This isn't relevant for reference transmutations.

```suggestion
```

src/lib.rs
```diff
@@ -4814,6 +4815,231 @@ macro_rules! try_transmute {
...
+            // This branch, though never taken, ensures that `size_of::<T>() ==
+            // size_of::<U>()` and that that `align_of::<T>() >=
+            // align_of::<U>()`.
+
+            // `t` is inferred to have type `T` because it's assigned to `e` (of
+            // type `&T`) as `&t`.
+            let mut t = loop {};
+            e = &t;
+
+            // `u` is inferred to have type `U` because it's used as `&u` as the
```
joshlf, line 4915:
```suggestion
            // `u` is inferred to have type `U` because it's used as `Ok(&u)` as the
```

src/lib.rs
```diff
@@ -4814,6 +4815,231 @@ macro_rules! try_transmute {
...
+///
+/// However, unlike a function, this macro can only be invoked when the types of
+/// `Src` and `Dst` are completely concrete. The types `Src` and `Dst` are
+/// inferred from the calling context; they cannot be explicitly specified in
+/// the macro invocation.
+///
+/// Note that the `Src` produced by the expression `$e` will *not* be dropped.
+/// Semantically, its bits will be copied into a new value of type `Dst`, the
+/// original `Src` will be forgotten, and the value of type `Dst` will be
+/// returned.
```
joshlf, line 4956:
```suggestion
```

src/lib.rs
```diff
@@ -4814,6 +4815,231 @@ macro_rules! try_transmute {
...
+            // size_of::<U>()` and that that `align_of::<T>() >=
+            // align_of::<U>()`.
+
+            // `t` is inferred to have type `T` because it's assigned to `e` (of
+            // type `&mut T`) as `&mut t`.
+            let mut t = loop {};
+            e = &mut t;
+
+            // `u` is inferred to have type `U` because it's used as `&mut u` as
+            // the value returned from this branch.
```
joshlf, line 5030:
```suggestion
            // `u` is inferred to have type `U` because it's used as `Ok(&mut u)` as
            // the value returned from this branch.
```

src/macro_util.rs
```diff
@@ -479,25 +509,83 @@ where
...
+pub fn try_transmute<Src, Dst>(src: Src) -> Result<Dst, ValidityError<Src, Dst>>
 where
     Src: IntoBytes,
     Dst: TryFromBytes,
 {
-    if !is_src_valid::<Src, Dst, _, BecauseExclusive>(Ptr::from_mut(&mut src)) {
-        return Err(ValidityError::new(src));
+    let mut src = ManuallyDrop::new(src);
+    let ptr = Ptr::from_mut(&mut src);
+    match try_cast_or_pme::<_, ManuallyDrop<Unalign<Dst>>, _, BecauseExclusive>(ptr) {
```
joshlf, line 519: Can you add a comment explaining why this is guaranteed not to fail due to increased alignment?

joshlf (changes requested):

src/lib.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/322d842147fb3191d649e3a8d876e05a9732e9a4..b34e4755dbf87ed8976bf1219dc629eafb52e87a#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759L4842)

src/lib.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/322d842147fb3191d649e3a8d876e05a9732e9a4..b34e4755dbf87ed8976bf1219dc629eafb52e87a#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759L4915)

src/lib.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/322d842147fb3191d649e3a8d876e05a9732e9a4..b34e4755dbf87ed8976bf1219dc629eafb52e87a#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759L4953)

src/lib.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/322d842147fb3191d649e3a8d876e05a9732e9a4..b34e4755dbf87ed8976bf1219dc629eafb52e87a#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759L5029)

joshlf (approved):

PR #1628, opened 2024-09-10
[ci] Don't fail fast when rolling pinned versions

joshlf:


joshlf:
Testing this PR [here](https://github.com/google/zerocopy/actions/runs/10792801091).

PR #1654, opened 2024-09-15
Update document about field-less enums in lib.rs

### LINKED ISSUE #985: Audit uses of "C-like" and prefer "fieldless"
Much of our documentation refers to "C-like" enums, where *fieldless* would be a more apt description. We should audit our documentation and replace "C like" for "fieldless" wherever applicable.

A fieldless enum is one without fields; e.g.:
```rust
enum Fieldless {
  Unit,
  Struct {},
  Enum(),
}
```

A C-like enum is a fieldless enum in which every constructor is unit-like. This is usually more restrictive than we need.

BatuhanDagistanli:
"C-like" enums reference was changed to "field-like", which was the only reference.

Closes #985

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4284,7 +4284,7 @@ pub unsafe trait IntoBytes {
 ///     `repr(packed)` or `repr(packed(1))`.
 /// - If the type is an enum:
 ///   - If `repr(align(N))` is provided, `N` must equal 1.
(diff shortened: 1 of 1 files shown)


joshlf (approved):
Thanks for this!

joshlf:
If you're interested in other doc cleanup issues, I just filed https://github.com/google/zerocopy/issues/1655.

Issue #1648, opened 2024-09-12
Defend against padding instability


jswrenn:
In light of https://github.com/rust-lang/rust/pull/129778, we should do more to defend against misconceptions about the durability of padding bytes across moves. For example, `FromZeros::new_zeroed` returns its value by-move, and so padding bytes within `Self` may not be `0`.

We can do some combination of the following. First, we can aggressively sign-post this risk. As a rule, one *shouldn't* depend on the value of padding bytes. 

Second, we can wrap zeroed types in a type that ensures padding remains preserved; e.g.:
```rust
union PaddingStable<T> {
    a: core::mem::ManuallyDrop<T>,
    b: [u8; core::mem::size_of::<u8>()]
}
```

RalfJung:
```rust
union PaddingStable<T> {
    a: core::mem::ManuallyDrop<T>,
    b: [u8; core::mem::size_of::<u8>()]
}
```
The last `u8` should be `T`, right?

jswrenn:
Yeah, that's a typo. Unfortunately, using `T` there also isn't currently stable. For now, the best we can do is documentation, I think (https://github.com/google/zerocopy/pull/1660).

joshlf:
If we need to add a utility type, we can do that when the need arises.

For now, I'll consider this closed by #1660.

PR #1660, opened 2024-09-15
Document that writes to padding are not preserved

joshlf:
Document that writes to padding bytes are not preserved when an object is moved or copied.

Makes progress on #1648

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -84,7 +84,7 @@
 //!
 //! # Cargo Features
 //!
-//! - **`alloc`**   
+//! - **`alloc`**
 //!   By default, `zerocopy` is `no_std`. When the `alloc` feature is enabled,
 //!   the `alloc` crate is added as a dependency, and some allocation-related
 //!   functionality is added.
@@ -94,10 +94,10 @@
 //!   `std` crate is added as a dependency (ie, `no_std` is disabled), and
 //!   support for some `std` types is added. `std` implies `alloc`.
 //!
-//! - **`derive`**   
+//! - **`derive`**
 //!   Provides derives for the core marker traits via the `zerocopy-derive`
 //!   crate. These derives are re-exported from `zerocopy`, so it is not
-//!   necessary to depend on `zerocopy-derive` directly.   
+//!   necessary to depend on `zerocopy-derive` directly.
 //!
 //!   However, you may experience better compile times if you instead directly
 //!   depend on both `zerocopy` and `zerocopy-derive` in your `Cargo.toml`,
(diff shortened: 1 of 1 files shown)


joshlf:
@saethlin Following up on our conversation at RustConf, does the wording in this warning accurately describe the behavior wrt padding bytes?

src/lib.rs
```diff
@@ -1953,6 +1953,15 @@ fn swap<T, U>((t, u): (T, U)) -> (U, T) {
 /// overhead. This is useful whenever memory is known to be in a zeroed state,
 /// such memory returned from some allocation routines.
 ///
+/// # Warning: Padding bytes
+///
+/// Note that, when an object is moved or copied, only the non-padding bytes of
```
saethlin, line 1958: Rust does not have objects. We have allocations and typed or untyped copies. Untyped copies copy all bytes in the specified range, and currently the only way you can do this is `ptr::copy` and `ptr::copy_nonoverlapping`.

Typed copies are every other copy, which includes (some of these may be surprising to users) assignment, reads/writes through pointers, function arguments, and function return values.

Typed copies are not guaranteed to preserve bytes which are padding in all variants. That's probably what this comment means to identify.

joshlf: Yeah, "typed copy" is what I'm referring to. Do you know if that term is defined anywhere in the Reference or stdlib docs? A quick Google doesn't turn anything up.

saethlin: The docs for ptr::copy contain a probably-unhelpful mention of "untyped".

joshlf: Do you know of any internal docs/discussions that define the term? Or is this just one of those "compiler people use these terms and know what they mean when they talk to each other and they weren't expecting you to look inside the compiler" things?

saethlin: Ugh of course that doesn't work. 
@CAD97 @digama0 @JakobDegen @RalfJung do we have documentation for what a typed copy is?

RalfJung: I don't know of any proper docs for this, no. That's clearly something we should fix. Any volunteers? :)

One question is where these docs would even go. The primary syntax for a typed copy is `a = b;`, so maybe this should be [here](https://doc.rust-lang.org/nightly/reference/expressions/operator-expr.html#assignment-expressions)?

jswrenn, line 1958:
```suggestion
/// Note that, when an value is moved or copied, only the non-padding bytes of
```

jswrenn (approved):
Approved, contingent on this change.

src/lib.rs
joshlf: Done

joshlf: [s/object/value](https://github.com/google/zerocopy/pull/1660#discussion_r1761911299), but otherwise I'm gonna merge this and we can nitpick the wording later once docs have landed to Rust that we can cite.

Issue #1634, opened 2024-09-10
Improve testing of custom derive support for data-carrying enums


joshlf:
Following up on #879.

## Progress

- [x] ~Write tests which ensure that a malicious use can't introduce variables which are not shadowed by our derive-generated code and cause problems~
  - Migrated to [#1684](https://github.com/google/zerocopy/issues/1684) since this doesn't need to block 0.8
- [x] Write tests which assert the token stream output of the derives
  - Done: #1665
- [x] https://github.com/google/zerocopy/pull/1665/files#r1761803556
  - #1683

PR #1664, opened 2024-09-15
[derive] Test exact output of derives

joshlf:
Makes progress on #367, #1634

### DIFF
--- a/zerocopy-derive/src/lib.rs
+++ b/zerocopy-derive/src/lib.rs
@@ -30,6 +30,8 @@
 
 mod r#enum;
 mod ext;
+#[cfg(test)]
+mod output_tests;
 mod repr;
 
 use proc_macro2::{TokenStream, TokenTree};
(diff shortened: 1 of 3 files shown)


zerocopy-derive/src/lib.rs
```diff
@@ -419,7 +421,7 @@ fn derive_try_from_bytes_struct(ast: &DeriveInput, strct: &DataStruct) -> proc_m
             // validity of a struct is just the composition of the bit
             // validities of its fields, so this is a sound implementation of
             // `is_bit_valid`.
-            fn is_bit_valid<A: ::zerocopy::pointer::invariant::Aliasing + ::zerocopy::pointer::invariant::AtLeast<::zerocopy::pointer::invariant::Shared>>(
+            fn is_bit_valid<A: ::zerocopy::pointer::invariant::Aliasing + ::zerocopy::pointer::invariant::AtLeast<::zerocopy::pointer::invariant::Shared> >(
```
jswrenn, line 424: What's with this change?

zerocopy-derive/src/output_tests.rs
```diff
@@ -0,0 +1,193 @@
...
+    FromBytes => derive_from_bytes_inner,
+    IntoBytes => derive_into_bytes_inner,
+    Unaligned => derive_unaligned_inner,
+);
+
+#[test]
+fn test_known_layout() {
+    test_derive! {
+        KnownLayout {
+            struct Foo;
```
jswrenn, line 36: Let's exercise a bit more of the derive:
```suggestion
            struct Foo<const N: usize>
            where
                for<'a> Foo<N>: Sized;
```

jswrenn (changes requested):

zerocopy-derive/src/lib.rs
joshlf: `>>` is the right-shift token while `> >` is two separate tokens. This matters because these tests care about the exact sequence of tokens.

Issue #1296, opened 2024-05-18
Improve error message on missing `Immutable`


joshlf:
Per https://github.com/google/zerocopy/discussions/1288#discussioncomment-9476006:

> The compiler errors when a type is missing `Immutable` are somewhat misleading, although I'm not sure there's much `zerocopy` can do about this. An example where the compiler recommends adding a reference rather than implementing `Immutable` (some were even more confusing than this):

```
error[E0277]: the trait bound `virtio::wl::CtrlVfdNewDmabuf: zerocopy::Immutable` is not satisfied                                          
   --> devices/src/virtio/wl.rs:317:20                                                                                                      
    |                                                                                                                                       
317 |         .write_obj(ctrl_vfd_new_dmabuf)                                                                                               
    |          --------- ^^^^^^^^^^^^^^^^^^^ the trait `zerocopy::Immutable` is not implemented for `virtio::wl::CtrlVfdNewDmabuf`          
    |          |                                                                                                                            
    |          required by a bound introduced by this call                                                                                  
    |                                                                                                                                       
note: required by a bound in `virtio::descriptor_utils::Writer::write_obj`                                                                  
   --> devices/src/virtio/descriptor_utils.rs:536:25                                                                                        
    |                                                                                                                                       
536 |     pub fn write_obj<T: Immutable + IntoBytes>(&mut self, val: T) -> io::Result<()> {                                                 
    |                         ^^^^^^^^^ required by this bound in `Writer::write_obj`                                                       
help: consider borrowing here                                                                                                               
    |                                                                                                                                       
317 |         .write_obj(&ctrl_vfd_new_dmabuf)                                                                                              
    |                    +                                                                                                                  
317 |         .write_obj(&mut ctrl_vfd_new_dmabuf)                                                                                          
    |                    ++++                                                
```

> Taking the compiler's suggestion results in a different error with a recommendation to remove the reference (back to the original code).

PR #1682, opened 2024-09-19
Use #[diagnostic::on_unimplemented] on traits

### LINKED ISSUE #1168: Use `#[diagnostic::on_unimplemented]` on public traits
`#[diagnostic::on_unimplemented]` (documented [here](https://doc.rust-lang.org/nightly/reference/attributes/diagnostics.html#the-diagnosticon_unimplemented-attribute)) was [stabilized in 1.78](https://blog.rust-lang.org/2024/05/02/Rust-1.78.0.html#diagnostic-attributes). Even though our MSRV is lower than 1.78, we can use our toolchain version detection in `build.rs` to detect 1.78, and use this to conditionally emit this attribute on some of our public traits, which might improve our ergonomics.

Note:
- Good candidates are the `HasPadding` and `ShouldBe` traits used to detect padding when deriving `IntoBytes`
- For derived traits, it would be ideal to emit a different error message if the trait is not implemented syntactically (ie, not derived) than if the trait is derived but there are unsatisfied trait bounds

Candidates for errors to fix:
- #1296

joshlf:
Closes #1168
Makes progress on #1296

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -638,6 +638,10 @@ pub use zerocopy_derive::KnownLayout;
     not(feature = "derive"),
     doc = concat!("[derive]: https://docs.rs/zerocopy/", env!("CARGO_PKG_VERSION"), "/zerocopy/derive.KnownLayout.html"),
 )]
+#[cfg_attr(
+    zerocopy_diagnostic_on_unimplemented,
+    diagnostic::on_unimplemented(note = "Consider adding `#[derive(KnownLayout)]` to `{Self}`")
+)]
 pub unsafe trait KnownLayout {
     // The `Self: Sized` bound makes it so that `KnownLayout` can still be
     // object safe. It's not currently object safe thanks to `const LAYOUT`, and
@@ -1086,6 +1090,10 @@ pub use zerocopy_derive::Immutable;
     doc = concat!("[derive]: https://docs.rs/zerocopy/", env!("CARGO_PKG_VERSION"), "/zerocopy/derive.Immutable.html"),
     doc = concat!("[derive-analysis]: https://docs.rs/zerocopy/", env!("CARGO_PKG_VERSION"), "/zerocopy/derive.Immutable.html#analysis"),
 )]
+#[cfg_attr(
+    zerocopy_diagnostic_on_unimplemented,
+    diagnostic::on_unimplemented(note = "Consider adding `#[derive(Immutable)]` to `{Self}`")
+)]
 pub unsafe trait Immutable {
     // The `Self: Sized` bound makes it so that `Immutable` is still object
     // safe.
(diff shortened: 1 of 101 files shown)


joshlf:
@danielverkamp This is an attempt to solve the [bad diagnostic you encountered](https://github.com/google/zerocopy/discussions/1288#discussioncomment-9476006), although unfortunately it seems like it hasn't done the trick. In particular if you look at the following files in this PR:
-  `tests/ui-nightly/diagnostic-not-implemented-issue-1296.rs`
- `tests/ui-nightly/diagnostic-not-implemented-issue-1296.stderr

...you'll see that the error message we request via `#[diagnostic::on_unimplemented]` does not show up for `Immutable`, while it does show up for `IntoBytes`. My speculation is that this is caused by us emitting different blanket impls for `Immutable` than for other traits, possibly combined with bad heuristics in rustc. I'll need to do more digging to figure out what's going on, but definitely let me know if you have any insight!

src/lib.rs
```diff
@@ -638,6 +638,12 @@ pub use zerocopy_derive::KnownLayout;
     not(feature = "derive"),
     doc = concat!("[derive]: https://docs.rs/zerocopy/", env!("CARGO_PKG_VERSION"), "/zerocopy/derive.KnownLayout.html"),
 )]
+#[cfg_attr(
+    zerocopy_diagnostic_on_unimplemented,
+    diagnostic::on_unimplemented(
+        note = "Consider deriving `KnownLayout` for `{Self}` (if it is defined in this crate)"
```
jswrenn, line 644: In the interest of keeping things short and sweet, cut the parenthetical. Also, since `{Self}` can be arbitrarily long, it's best to end the message with the type.
```suggestion
        note = "Consider adding `#[derive(KnownLayout)]` for `{Self}`"
```

And finally, even if it's _not_ defined in the current crate: They should totally still go and file a PR on whatever crate defines `Self`. ;-)

joshlf: My concern is that this message can trigger on built-in types like `fn()` or `&T`, where that message is misleading. Or is your feeling that it'd be obvious that it shouldn't be taken literally in all circustances?

joshlf:
> @danielverkamp This is an attempt to solve the [bad diagnostic you encountered](https://github.com/google/zerocopy/discussions/1288#discussioncomment-9476006), although unfortunately it seems like it hasn't done the trick. In particular if you look at the following files in this PR:
> 
> * `tests/ui-nightly/diagnostic-not-implemented-issue-1296.rs`
> * `tests/ui-nightly/diagnostic-not-implemented-issue-1296.stderr
> 
> ...you'll see that the error message we request via `#[diagnostic::on_unimplemented]` does not show up for `Immutable`, while it does show up for `IntoBytes`.
> 
> EDIT: It appears that this is caused by our impls of `Immutable` for `&T` and `&mut T`. When I add similar impls for `IntoBytes`, the same problem happens for `IntoBytes`. I'll file a Rust issue.

Okay, I minimized and filed an issue: https://github.com/rust-lang/rust/issues/130563

src/lib.rs
jswrenn: I think it's obvious it should be taken literally, and we already soften the language by leading with "Consider".

joshlf: Done.

jswrenn (approved):

PR #1683, opened 2024-09-19
[derive] More complex TryFromBytes enum output tests

joshlf:
Makes progress on #1634

### DIFF
--- a/zerocopy-derive/src/output_tests.rs
+++ b/zerocopy-derive/src/output_tests.rs
@@ -197,19 +197,28 @@ fn test_try_from_bytes_enum() {
     test_derive! {
         TryFromBytes {
             #[repr(u8)]
-            enum ComplexWithGenerics<X, Y> {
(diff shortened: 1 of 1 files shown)


zerocopy-derive/src/output_tests.rs
```diff
@@ -197,19 +197,28 @@ fn test_try_from_bytes_enum() {
     test_derive! {
         TryFromBytes {
             #[repr(u8)]
-            enum ComplexWithGenerics<X, Y> {
+            enum ComplexWithGenerics<'a, const N: usize, X, Y: Deref>
```
jswrenn, line 200: Just to also make sure that inline lifetime bounds are forwarded:
```suggestion
            enum ComplexWithGenerics<'a: 'static, const N: usize, X, Y: Deref>
```

jswrenn (changes requested):

jswrenn (approved):

PR #1705, opened 2024-09-20
Rename generic param to avoid name collision

djkoloski:
Also makes the signatures of is_bit_valid generated for structs, unions, and enums the same.

### DIFF
--- a/zerocopy-derive/tests/enum_try_from_bytes.rs
+++ b/zerocopy-derive/tests/enum_try_from_bytes.rs
@@ -268,3 +268,13 @@ enum GenericWithLifetimes<'a, 'b, X: 'a, Y: 'b> {
     Foo(::core::marker::PhantomData<&'a X>),
     Bar(::core::marker::PhantomData<&'b Y>),
 }
+
+#[derive(Clone, Copy, imp::TryFromBytes)]
+struct A;
+
+#[derive(imp::TryFromBytes)]
+#[repr(C)]
+enum B {
+    A(A),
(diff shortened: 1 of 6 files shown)


zerocopy-derive/tests/enum_try_from_bytes.rs
```diff
@@ -268,3 +268,13 @@ enum GenericWithLifetimes<'a, 'b, X: 'a, Y: 'b> {
...
+
+#[derive(Clone, Copy, imp::TryFromBytes)]
+struct A;
+
+#[derive(imp::TryFromBytes)]
+#[repr(C)]
+enum B {
+    A(A),
+    A2 { a: A },
+}
```
jswrenn, line 280: Could you add a UI test for the case where a field type is named `___ZerocopyAliasing`? We want to be sure this doesn't regress (i.e., start compiling successfully).

jswrenn (changes requested):

zerocopy-derive/tests/enum_try_from_bytes.rs
joshlf: So that we can unblock Fuchsia users who are depending on this, I'm going to merge this w/o extra tests, but I added the following to the release roadmap to track (note the last bullet, which links to this comment thread and to #1725):

Screenshot 2024-09-23 at 4 24 38 PM

joshlf (approved):

jswrenn (approved):

PR #1709, opened 2024-09-20
Inline TryFromBytes derive for variant structs

djkoloski:
I didn't update the goldens because it was too hard

### DIFF
--- a/zerocopy-derive/src/enum.rs
+++ b/zerocopy-derive/src/enum.rs
@@ -9,8 +9,9 @@
 use ::proc_macro2::TokenStream;
 use ::quote::quote;
 use ::syn::{DataEnum, Fields, Generics, Ident};
+use syn::parse_quote;
 
-use crate::{EnumRepr, Trait};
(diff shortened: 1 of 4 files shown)


jswrenn:
The goldens?

djkoloski:
The derive macro output tests (I call them goldens because they're effectively golden files)

jswrenn:
> I didn't update the goldens because it was too hard

@joshlf Told ya so. 😛 We *really* need to make updating these as easy as `TRYBUILD=overwrite`.

joshlf:
> > I didn't update the goldens because it was too hard
> 
> 
> 
> @joshlf Told ya so. 😛 We *really* need to make updating these as easy as `TRYBUILD=overwrite`.

My hope is that https://github.com/google/zerocopy/pull/1711 will make this much less painful.

joshlf (approved):

joshlf:
> I didn't update the goldens because it was too hard

Updated them and pushed.

PR #1711, opened 2024-09-20
[derive] Make output tests insensitive to syntax

joshlf:
Our previous output tests asserted on the exact token stream, which is sensitive to subtle differences like `> >` vs `>>` in type signatures. With this change, we instead test for string equality *after* performing a pretty-printing pass, which has the effect of normalizing most such differences, thus making our tests more robust and less sensitive to these inconsequential syntax details.

### DIFF
--- a/zerocopy-derive/src/output_tests.rs
+++ b/zerocopy-derive/src/output_tests.rs
@@ -6,20 +6,18 @@
 // This file may not be copied, modified, or distributed except according to
 // those terms.
 
-use synstructure::test_derive;
+use dissimilar::Chunk;
+use proc_macro2::TokenStream;
 
-macro_rules! impl_synstructure_derive {
-    ($($outer:ident => $inner:ident),* $(,)?) => {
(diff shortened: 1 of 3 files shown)


jswrenn (approved):

zerocopy-derive/src/output_tests.rs
```diff
@@ -29,14 +27,76 @@ impl_synstructure_derive!(
...
+fn assert_eq_streams(expect: TokenStream, res: TokenStream) {
+    let pretty =
+        |ts: TokenStream| prettyplease::unparse(&syn::parse_file(&ts.to_string()).unwrap());
+
+    let expect = pretty(expect.clone());
+    let res = pretty(res.clone());
+    if expect != res {
+        let diff = dissimilar::diff(&expect, &res)
+            .into_iter()
+            .map(|chunk| {
```
djkoloski, line 69: Per CI failures, use `flat_map` here

PR #1712, opened 2024-09-20
[derive] Improve IntoBytes error messages

joshlf:
On Rust toolchains 1.78.0 or later, use
`#[diagnostic::on_unimplemented]` to improve the error message emitted when `IntoBytes` is derived on a type with inter-field padding.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3837,25 +3837,26 @@ fn mut_from_prefix_suffix<T: FromBytes + KnownLayout + ?Sized>(
 ///
 /// # Error Messages
 ///
-/// Due to the way that the custom derive for `IntoBytes` is implemented, you
-/// may get an error like this:
+/// On Rust toolchains prior to 1.78.0, due to the way that the custom derive
+/// for `IntoBytes` is implemented, you may get an error like this:
 ///
 /// ```text
-/// error[E0277]: the trait bound `HasPadding<Foo, true>: ShouldBe<false>` is not satisfied
+/// error[E0277]: the trait bound `(): PaddingFree<Foo, true>` is not satisfied
 ///   --> lib.rs:23:10
 ///    |
 ///  1 | #[derive(IntoBytes)]
-///    |          ^^^^^^^ the trait `ShouldBe<false>` is not implemented for `HasPadding<Foo, true>`
+///    |          ^^^^^^^^^ the trait `PaddingFree<Foo, true>` is not implemented for `()`
 ///    |
-///    = help: the trait `ShouldBe<VALUE>` is implemented for `HasPadding<T, VALUE>`
+///    = help: the following implementations were found:
+///                   <() as PaddingFree<T, false>>
 /// ```
 ///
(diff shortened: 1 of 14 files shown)


zerocopy-derive/Cargo.toml
```diff
@@ -28,7 +28,7 @@ proc-macro = true
 [dependencies]
 proc-macro2 = "1.0.1"
 quote = "1.0.10"
-syn = "2.0.46"
+syn = { version = "2.0.46", features = ["full"] }
```
joshlf, line 31: @djkoloski Do you have a sense of how we could implement using the same trick that you pulled to avoid the `full` feature?

I tested locally on a MacBook Pro on low battery mode, and found that the `full` feature increased `zerocopy-derive` `cargo build` time from 8.9s to 13.3s. That's a large percentage, but a small absolute difference. So I'm not opposed to just sticking with this impl, but it'd be cool if we could avoid it.

src/lib.rs
```diff
@@ -3782,25 +3782,26 @@ fn mut_from_prefix_suffix<T: FromBytes + KnownLayout + ?Sized>(
...
 /// # Error Messages
 ///
-/// Due to the way that the custom derive for `IntoBytes` is implemented, you
-/// may get an error like this:
+/// On Rust toolchains prior to 1.78.0, due to the way that the custom derive
+/// for `IntoBytes` is implemented, you may get an error like this:
 ///
 /// ```text
-/// error[E0277]: the trait bound `HasPadding<Foo, true>: ShouldBe<false>` is not satisfied
+/// error[E0277]: the trait bound `(): PaddingFree<Foo, true>` is not satisfied
```
jswrenn, line 3789: This fallback is arguably an improvement over the previous error!

zerocopy-derive/tests/ui-nightly/enum.stderr
```diff
@@ -413,41 +413,53 @@ help: add `#![feature(trivial_bounds)]` to the crate attributes to enable
 9   + #![feature(trivial_bounds)]
     |
 
-error[E0277]: the trait bound `HasPadding<IntoBytes1, true>: ShouldBe<false>` is not satisfied
+error[E0277]: `IntoBytes1` has inter-field padding
    --> tests/ui-nightly/enum.rs:533:10
     |
 533 | #[derive(IntoBytes)]
-    |          ^^^^^^^^^ the trait `ShouldBe<false>` is not implemented for `HasPadding<IntoBytes1, true>`
+    |          ^^^^^^^^^ types with padding cannot implement `IntoBytes`
```
jswrenn, line 420: This is ridiculously good. I wonder if we could even say how many bytes of inter-field padding there with const generics, but that's just yak shaving.

jswrenn (approved):

zerocopy-derive/tests/ui-nightly/enum.stderr
joshlf: Filed https://github.com/google/zerocopy/issues/1717 to track

PR #1710, opened 2024-09-20
Add test to capture bad `IntoBytes` derive error msg

joshlf:
Makes progress on #1708

PR #1713, opened 2024-09-20
Improve IntoBytes derive error message

### LINKED ISSUE #1708: Improve ergonomics of `IntoBytes` on unsized types
This source code:

```rust
#[derive(IntoBytes)]
#[repr(C)]
struct IntoBytes4 {
    a: u8,
    b: [u8],
}
```

...results in this error:

```text
error[E0277]: the size for values of type `[u8]` cannot be known at compilation time
   --> tests/ui-nightly/struct.rs:125:8
    |
125 | struct IntoBytes4 {
    |        ^^^^^^^^^^ doesn't have a size known at compile-time
    |
    = help: within `IntoBytes4`, the trait `Sized` is not implemented for `[u8]`, which is required by `IntoBytes4: Sized`
note: required because it appears within the type `IntoBytes4`
   --> tests/ui-nightly/struct.rs:125:8
    |
125 | struct IntoBytes4 {
    |        ^^^^^^^^^^
note: required by an implicit `Sized` bound in `std::mem::size_of`
   --> $RUST/core/src/mem/mod.rs
    |
    | pub const fn size_of<T>() -> usize {
    |                      ^ required by the implicit `Sized` requirement on this type parameter in `size_of`

error[E0277]: the size for values of type `[u8]` cannot be known at compilation time
   --> tests/ui-nightly/struct.rs:127:8
    |
127 |     b: [u8],
    |        ^^^^ doesn't have a size known at compile-time
    |
    = help: the trait `Sized` is not implemented for `[u8]`
note: required by an implicit `Sized` bound in `std::mem::size_of`
   --> $RUST/core/src/mem/mod.rs
    |
    | pub const fn size_of<T>() -> usize {
    |                      ^ required by the implicit `Sized` requirement on this type parameter in `size_of`

```

What's happening here is that the derive can't tell that this type can't have padding, and so it tries to guarantee that by emitting a padding check bound (that the size of the type is equal to the sizes of its fields). However, since this type is unsized, this check can't compile because `size_of` requires `T: Sized`. Currently, the only way to make this code compile is to add `#[repr(packed)]`.

joshlf:
Closes #1708

### DIFF
--- a/src/util/macro_util.rs
+++ b/src/util/macro_util.rs
@@ -285,6 +285,32 @@ mod size_to_tag {
 #[doc(hidden)]
 pub type SizeToTag<const SIZE: usize> = <() as size_to_tag::SizeToTag<SIZE>>::Tag;
 
+// We put `Sized` in its own module so it can have the same name as the standard
+// library `Sized` without shadowing it in the parent module.
+#[cfg(zerocopy_diagnostic_on_unimplemented)]
+mod __size_of {
+    #[diagnostic::on_unimplemented(
+        message = "`{Self}` is unsized",
+        label = "`IntoBytes` needs all field types to be `Sized` in order to determine whether there is inter-field padding",
+        note = "consider using `#[repr(packed)]` to remove inter-field padding",
+        note = "`IntoBytes` does not require the fields of `#[repr(packed)]` types to be `Sized`"
+    )]
+    pub trait Sized: core::marker::Sized {}
+    impl<T: core::marker::Sized> Sized for T {}
+
+    #[inline(always)]
+    #[must_use]
+    #[allow(clippy::needless_maybe_sized)]
+    pub const fn size_of<T: Sized + ?core::marker::Sized>() -> usize {
+        core::mem::size_of::<T>()
+    }
+}
+
+#[cfg(zerocopy_diagnostic_on_unimplemented)]
+pub use __size_of::size_of;
+#[cfg(not(zerocopy_diagnostic_on_unimplemented))]
+pub use core::mem::size_of;
+
 /// Does the struct type `$t` have padding?
 ///
 /// `$ts` is the list of the type of every field in `$t`. `$t` must be a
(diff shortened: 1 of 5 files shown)


joshlf:
@jswrenn if you're on board with this, I think this can supersede #1710.

src/util/macro_util.rs
```diff
@@ -285,6 +285,32 @@ mod size_to_tag {
...
+        note = "consider using `#[repr(packed)]` to remove inter-field padding",
+        note = "`IntoBytes` does not require the fields of `#[repr(packed)]` types to be `Sized`"
+    )]
+    pub trait Sized: core::marker::Sized {}
+    impl<T: core::marker::Sized> Sized for T {}
+
+    #[inline(always)]
+    #[must_use]
+    #[allow(clippy::needless_maybe_sized)]
+    pub const fn size_of<T: Sized + ?core::marker::Sized>() -> usize {
```
jswrenn, line 304: This is such a cursed bound.

zerocopy-derive/tests/ui-nightly/struct.stderr
```diff
@@ -269,8 +269,43 @@ help: add `#![feature(trivial_bounds)]` to the crate attributes to enable
...
+   --> $WORKSPACE/src/util/macro_util.rs
+    |
+    |     pub const fn size_of<T: Sized + ?core::marker::Sized>() -> usize {
+    |                             ^^^^^ required by this bound in `size_of`
+
+error[E0277]: `[u8]` is unsized
+   --> tests/ui-nightly/struct.rs:129:8
+    |
+129 |     b: [u8],
+    |        ^^^^ `IntoBytes` needs all field types to be `Sized` in order to determine whether there is inter-field padding
```
jswrenn, line 295: NICE!

jswrenn (approved):

src/util/macro_util.rs
mainrs: I just stumbled across this randomly: why is that bound there? Wouldn't any `T` automatically also be `core::mem::Sized` due to the trait bound of `__size_of::Sized`? Thanks!

joshlf: From a correctness perspective, you're right. However, without `?core::marker::Sized`, when Rust encounters a type that is unsized that is passed to this `size_of`, it emits a `Sized` error first and bails, so it never gets to the point of discovering that the `Sized` bound (our bespoke `Sized` trait) isn't satisfied, and so it never emits our custom diagnostic. Adding `?core::marker::Sized` "uncovers" our unsatisfied trait bound so that the desired diagnostic actually gets emitted.

It is, as all things in proc macro land, mildly cursed 😛

mainrs: @joshlf thanks for the explanation!

joshlf: np!

PR #1718, opened 2024-09-21
Encode infallible alignment errors in types

joshlf:
Permit callers to prove at compile time that alignment errors are unreachable for unaligned destination types. This permits them to infallibly ignore this error condition.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -101,7 +101,7 @@ use core::error::Error;
 #[cfg(all(not(zerocopy_core_error), any(feature = "std", test)))]
 use std::error::Error;
 
-use crate::{util::SendSyncPhantomData, KnownLayout, TryFromBytes};
+use crate::{util::SendSyncPhantomData, KnownLayout, TryFromBytes, Unaligned};
 #[cfg(doc)]
 use crate::{FromBytes, Ref};
 
@@ -133,6 +133,51 @@ pub enum ConvertError<A, S, V> {
     Size(S),
     /// The conversion source contained invalid data.
     Validity(V),
+}
+
+impl<Src, Dst: ?Sized + Unaligned, S, V> From<ConvertError<AlignmentError<Src, Dst>, S, V>>
+    for ConvertError<Infallible, S, V>
+{
+    /// Infallibly discards the alignment error from this `ConvertError` since
+    /// `Dst` is unaligned.
+    ///
+    /// Since [`Dst: Unaligned`], it is impossible to encounter an alignment
+    /// error. This method permits discarding that alignment error infallibly
+    /// and replacing it with [`Infallible`].
+    ///
+    /// [`Dst: Unaligned`]: crate::Unaligned
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use core::convert::Infallible;
+    /// use zerocopy::*;
+    /// # use zerocopy_derive::*;
+    ///
+    /// #[derive(TryFromBytes, KnownLayout, Unaligned, Immutable)]
+    /// #[repr(C, packed)]
+    /// struct Bools {
+    ///     one: bool,
+    ///     two: bool,
+    ///     many: [bool],
+    /// }
+    ///
+    /// impl Bools {
+    ///     fn parse(bytes: &[u8]) -> Result<&Bools, UnalignedTryCastError<&[u8], Bools>> {
+    ///         // Since `Bools: Unaligned`, we can infallibly discard
+    ///         // the alignment error.
+    ///         Bools::try_ref_from_bytes(bytes).map_err(Into::into)
+    ///     }
+    /// }
+    /// ```
+    #[inline]
+    fn from(err: ConvertError<AlignmentError<Src, Dst>, S, V>) -> ConvertError<Infallible, S, V> {
+        match err {
+            ConvertError::Alignment(e) => ConvertError::Alignment(Infallible::from(e)),
+            ConvertError::Size(e) => ConvertError::Size(e),
+            ConvertError::Validity(e) => ConvertError::Validity(e),
+        }
+    }
 }
 
 impl<A: fmt::Debug, S: fmt::Debug, V: fmt::Debug> fmt::Debug for ConvertError<A, S, V> {
@@ -177,11 +222,20 @@ pub struct AlignmentError<Src, Dst: ?Sized> {
     /// The source value involved in the conversion.
     src: Src,
     /// The inner destination type inolved in the conversion.
+    ///
+    /// INVARIANT: An `AlignmentError` may only be constructed if `Dst`'s
+    /// alignment requirement is greater than one.
     dst: SendSyncPhantomData<Dst>,
 }
 
 impl<Src, Dst: ?Sized> AlignmentError<Src, Dst> {
-    pub(crate) fn new(src: Src) -> Self {
+    /// # Safety
+    ///
+    /// The caller must ensure that `Dst`'s alignment requirement is greater
+    /// than one.
+    pub(crate) unsafe fn new_unchecked(src: Src) -> Self {
+        // INVARIANT: The caller guarantees that `Dst`'s alignment requirement
+        // is greater than one.
         Self { src, dst: SendSyncPhantomData::default() }
     }
 
@@ -192,6 +246,9 @@ impl<Src, Dst: ?Sized> AlignmentError<Src, Dst> {
     }
 
     pub(crate) fn with_src<NewSrc>(self, new_src: NewSrc) -> AlignmentError<NewSrc, Dst> {
+        // INVARIANT: `with_src` doesn't change the type of `Dst`, so the
+        // invariant that `Dst`'s alignment requirement is greater than one is
+        // preserved.
         AlignmentError { src: new_src, dst: SendSyncPhantomData::default() }
     }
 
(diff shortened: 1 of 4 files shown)


joshlf:
Still need to bikeshed naming and how to write the doc comment, but otherwise it's done.

src/util/mod.rs
```diff
@@ -547,14 +548,20 @@ impl<T: ?Sized> AsAddress for *mut T {
     }
 }
 
-/// Is `t` aligned to `align_of::<U>()`?
+/// Validates that `t` is aligned to `align_of::<U>()`.
 #[inline(always)]
-pub(crate) fn aligned_to<T: AsAddress, U>(t: T) -> bool {
+pub(crate) fn validate_aligned_to<T: AsAddress, U>(t: T) -> Result<(), AlignmentError<(), U>> {
```
jswrenn, line 553: Take it or leave it:
```suggestion
pub(crate) fn validate_aligned_to<T: AsAddress, U>(t: T) -> Result<(), AlignmentError<T, U>> {
```

src/error.rs
```diff
@@ -135,6 +135,17 @@ pub enum ConvertError<A, S, V> {
     Validity(V),
 }
 
+impl<Src, Dst: ?Sized + Unaligned, S, V> ConvertError<AlignmentError<Src, Dst>, S, V> {
+    // TODO: Bikeshed this name
+    pub fn recall_aligned(self) -> ConvertError<Infallible, S, V> {
```
jswrenn, line 140: I continue to stand by my extreme fondness for `recall_*` as a naming pattern. :-)

jswrenn (changes requested):
This is really elegant! I particularly like the error-ization of `validate_aligned_to`.

src/util/mod.rs
joshlf: I tried that, but it ends up being worse because no callers actually pass a byte slice directly - they pass things like `bytes.deref()` - and so all of them would end up needing to do `.with_src` anyway.

Also, this would require changing `AsAddress::addr` to take a `&self` receiver - otherwise, calling `t.addr()` consumes the `t` and so it can't be returned.

src/error.rs
joshlf: I just updated and added `Into` impls that I think supersede this - I think we can just remove it entirely.

src/error.rs
joshlf, line 1: TODO: Add module-level and type-level docs that highlight this feature and how to use it.

src/error.rs
```diff
@@ -626,6 +683,17 @@ impl<Src, Dst: ?Sized> CastError<Src, Dst> {
     }
 }
 
+impl<Src, Dst: ?Sized + Unaligned> Into<SizeError<Src, Dst>> for CastError<Src, Dst> {
+    fn into(self) -> SizeError<Src, Dst> {
```
jswrenn, line 687: Slap an `#[inline(always)]` on this — we want the effects of `unreachable_unchecked` to propagate up the call chain.

src/error.rs
```diff
@@ -255,6 +290,28 @@ impl<Src, Dst: ?Sized> AlignmentError<Src, Dst> {
     }
 }
 
+impl<Src, Dst: ?Sized + Unaligned> AlignmentError<Src, Dst> {
+    fn into_infallible(self) -> Infallible {
```
jswrenn, line 294: Slap an `#[inline(always)]` on this — we want the effects of `unreachable_unchecked` to propagate up the call chain.

jswrenn (changes requested):
A few `#[inline(always)]` nits, but then I'm happy to merge this.

I wonder if this is sufficiently elegant that we should remove the `unaligned` methods from `Ref`!

src/error.rs
```diff
@@ -135,6 +135,29 @@ pub enum ConvertError<A, S, V> {
...
+    pub fn recall_aligned(self) -> ConvertError<Infallible, S, V> {
+        match self {
+            Self::Alignment(e) => ConvertError::Alignment(e.into_infallible()),
+            Self::Size(e) => ConvertError::Size(e),
+            Self::Validity(e) => ConvertError::Validity(e),
+        }
+    }
+}
+
+impl<Src, Dst: ?Sized + Unaligned, S, V> Into<ConvertError<Infallible, S, V>>
```
jswrenn, line 149: Why are we implementing `Into` here instead of `From`?

src/error.rs
jswrenn, line 294: For consistency, should this be an implementation of `From`? Then we could have a consistent messaging that calling `.into()` at all levels makes impossible error branches `Infallible`.

src/error.rs
joshlf: Done

src/error.rs
joshlf: Done

src/error.rs
joshlf: Done

src/error.rs
joshlf: Done

src/error.rs
joshlf: Discussed offline; removed.

src/error.rs
```diff
@@ -135,6 +135,19 @@ pub enum ConvertError<A, S, V> {
     Validity(V),
 }
 
+impl<Src, Dst: ?Sized + Unaligned, S, V> From<ConvertError<AlignmentError<Src, Dst>, S, V>>
+    for ConvertError<Infallible, S, V>
+{
+    #[inline]
+    fn from(err: ConvertError<AlignmentError<Src, Dst>, S, V>) -> ConvertError<Infallible, S, V> {
```
joshlf, line 142: TODO: Doc comment with example

src/error.rs
```diff
@@ -255,6 +280,29 @@ impl<Src, Dst: ?Sized> AlignmentError<Src, Dst> {
     }
 }
 
+impl<Src, Dst: ?Sized + Unaligned> From<AlignmentError<Src, Dst>> for Infallible {
+    #[inline(always)]
+    fn from(_: AlignmentError<Src, Dst>) -> Infallible {
```
joshlf, line 285: Do we care about documenting this? It's unlikely that callers will ever interact with this impl directly.

src/error.rs
```diff
@@ -135,6 +135,62 @@ pub enum ConvertError<A, S, V> {
...
+    /// #[repr(C)]
+    /// struct Packet {
+    ///     magic_number: Unalign<C0C0>,
+    ///     mug_size: u8,
+    ///     temperature: u8,
+    ///     marshmallows: [[u8; 2]],
+    /// }
+    ///
+    /// impl Packet {
+    ///     pub fn parse(bytes: &[u8]) -> Result<&Packet, ConvertError<Infallible, SizeError<&[u8], Packet>, ValidityError<&[u8], Packet>>> {
```
joshlf, line 177: This type signature is awful. Should we figure out a way to express this using type aliases instead? Perhaps add `UnalignedConvertError` and `UnalignedCastError` type aliases?

src/error.rs
joshlf: Will follow up later: #1721

src/error.rs
jswrenn: We have an alias for this: https://docs.rs/zerocopy/0.8.0-alpha.22/zerocopy/error/type.TryReadError.html

joshlf:
> I wonder if this is sufficiently elegant that we should remove the `unaligned` methods from `Ref`!

Probably, but let's [test it in Fuchsia first](https://github.com/google/zerocopy/issues/1720).

src/error.rs
joshlf: Discussed offline; added an `UnalignedTryCastError`.

src/error.rs
jswrenn: Agreed.

src/error.rs
```diff
@@ -135,6 +135,51 @@ pub enum ConvertError<A, S, V> {
...
+    /// }
+    ///
+    /// impl Bools {
+    ///     fn parse(bytes: &[u8]) -> Result<&Bools, UnalignedTryCastError<&[u8], Bools>> {
+    ///         // Since `Bools: Unaligned`, we can infallibly discard
+    ///         // the alignment error.
+    ///         Bools::try_ref_from_bytes(bytes).map_err(Into::into)
+    ///     }
+    /// }
+    /// ```
```
jswrenn, line 172: I don't think users are primed to expect bespoke documentation on trait impls. Let's move this documentation to the `error` module docs.

src/error.rs
```diff
@@ -135,6 +135,51 @@ pub enum ConvertError<A, S, V> {
     Validity(V),
 }
 
+impl<Src, Dst: ?Sized + Unaligned, S, V> From<ConvertError<AlignmentError<Src, Dst>, S, V>>
+    for ConvertError<Infallible, S, V>
```
jswrenn, line 139: I think we should make this a general `From` conversion of `A`, `S`, and `V`:
```suggestion
impl<Src, Dst: ?Sized + Unaligned, A, S, V, NewA, NewS, NewV>
    From<ConvertError<NewA, NewS, NewV>>
for
    ConvertError<Infallible, S, V>
where
    NewA: From<A>,
    NewS: From<S>,
    NewV: From<V>,
```

jswrenn (changes requested):

src/error.rs
jswrenn: Hm, this is going to conflict. Let me see if I can make it work...

jswrenn: I cannot.

src/error.rs
jswrenn: Eh, without the generalization recommendation (which is impossible), I'm okay having these docs here.

jswrenn (approved):

PR #1724, opened 2024-09-21
Add debug check for #[inline] attribute in public derive macro functions

mepatrick73:
This commit introduces a debug-level check to ensure all public functions 
generated by derive macros include the #[inline] attribute. The check 
triggers an assertion if any function lacks the attribute, implemented 
manually since Clippy does not currently check for missing inline lints 
in proc macro-generated code. This may change if Clippy adds support 
for this lint in expanded code.

Additionally, this commit adds #[inline] attributes to public functions 
in the derived code.

### DIFF
--- a/zerocopy-derive/src/lib.rs
+++ b/zerocopy-derive/src/lib.rs
@@ -87,9 +87,40 @@ macro_rules! derive {
         #[proc_macro_derive($trait)]
         pub fn $outer(ts: proc_macro::TokenStream) -> proc_macro::TokenStream {
             let ast = syn::parse_macro_input!(ts as DeriveInput);
-            $inner(&ast).into()
+            let expanded = $inner(&ast);
+
+            // We manually check if public functions are inlined
+            // We do this because clippy doesn't check for the inline lint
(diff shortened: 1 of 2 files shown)


zerocopy-derive/src/lib.rs
```diff
@@ -87,11 +87,42 @@ macro_rules! derive {
...
+                    debug_assert!(
+                        has_inline,
+                        "Method {} does not have #[inline] attribute",
+                        method.sig.ident
+                    );
+                }
+            }
+        }
+    }
+}
```
joshlf, line 124: Can you move this to the end of the file, just above `mod tests`?

joshlf (changes requested):
You'll also need to update the output in `zerocopy-derive/src/output_tests.rs` to include the new `#[inline(always)]` attributes. You can see the required diff [here](https://github.com/google/zerocopy/actions/runs/10976480064/job/30477289266?pr=1724#step:13:334).

joshlf:
Thanks for doing this, @mepatrick73! Just one small nit and a failing test and we should be good to go!

Issue #1292, opened 2024-05-17
`KnownLayout` violates private-in-public rules in some cases


joshlf:
In some cases, `#[derive(KnownLayout)]` generates [code like this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6eab52187622f8ff35e0f641aea6eb47):

```rust
trait KnownLayout {
    type PointerMetadata;
}

mod foo {
    use super::*;
    
    pub struct Foo {
        bar: Bar,
    }
    
    impl KnownLayout for Foo {
        type PointerMetadata = <Bar as KnownLayout>::PointerMetadata;
    }
    
    struct Bar;
    
    impl KnownLayout for Bar {
        type PointerMetadata = ();
    }
}
```

This causes a private-in-public error:

```text
error[E0446]: private type `Bar` in public interface
  --> src/lib.rs:13:9
   |
13 |         type PointerMetadata = <Bar as KnownLayout>::PointerMetadata;
   |         ^^^^^^^^^^^^^^^^^^^^ can't leak private type
...
16 |     struct Bar;
   |     ---------- `Bar` declared as private
```

This is probably a rustc bug.

jswrenn:
Discussion in the rust-lang zulip here: https://rust-lang.zulipchat.com/#narrow/stream/213817-t-lang/topic/Relaxing.20priv-in-path.20error/near/439582018

jswrenn:
This is a compiler bug, and is tracked by https://github.com/rust-lang/rust/issues/45713

PR #1734, opened 2024-09-23
[doc] Document rustc type privacy bug

jswrenn:
See #1292

Makes progress towards #671

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -446,6 +446,69 @@ use {FromZeros as FromZeroes, IntoBytes as AsBytes, Ref as LayoutVerified};
 ///
 /// This derive cannot currently be applied to unsized structs without an
 /// explicit `repr` attribute.
+///
+/// Some invocations of this derive run afoul of a [known bug] in Rust's type
+/// privacy checker. For example, this code:
+///
+/// ```compile_fail,E0446
+/// use zerocopy::*;
+/// # use zerocopy_derive::*;
+///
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -446,6 +446,40 @@ use {FromZeros as FromZeroes, IntoBytes as AsBytes, Ref as LayoutVerified};
...
+/// 3 | #[derive(KnownLayout)]
+///   |          ^^^^^^^^^^^ can't leak private type
+/// ...
+/// 8 | struct PrivateType;
+///   | ------------------- `PrivateType` declared as private
+///   |
+///   = note: this error originates in the derive macro `KnownLayout` (in Nightly builds, run with -Z macro-backtrace for more info)
+/// ```
+///
+/// To work around this bug, mark private field types as `pub` and annotate them with `#[doc(hidden)]`.
```
joshlf, line 480: This is only required under some circumstances. I'd double check the KnownLayout derive to get the exact list, but off the top of my head:
- It only applies to the trailing field, so I'd update the example to have two fields
- There is no separate trailing field type if it's generic (ie, if its type is `T` where `T` is a type parameter - this doesn't apply for, eg `Foo<T>`, since `Foo` could still be a private type)
- There might be reprs for which this isn't a problem? Idr.

joshlf (changes requested):

joshlf (approved):

Issue #1735, opened 2024-09-23
Support container transmutation via macro


joshlf:
*This is one possible avenue for supporting #114.*

Most container transmutations require three components:
1. `Src: FromBytes + IntoBytes + Immutable` and `Dst: FromBytes + IntoBytes + Immutable`
2. `size_of::<Src>() == size_of::<Dst>()` and `align_of::<Src>() == align_of::<Dst>()`
3. Some way of invoking the right `into_raw` and `from_raw` associated functions on the container type

Currently, macros like `transmute_ref!` have the ability to enforce conditions (1) and (2). It should be possible in principle to support condition (3) using a trait:

```rust
#[doc(hidden)]
pub unsafe trait Container {
    type Element;
    type Raw;

    fn into_raw(self) -> Self::Raw;
    unsafe fn from_raw(Self::Raw) -> Self;
}
```

The associated `Raw` type is needed because not all containers use the same raw representations. For example, [`Arc` uses `*const T`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.into_raw) while [`Vec` uses `(*mut T, usize, usize)`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.into_raw_parts).

A few open questions:
- In order to perform the size and alignment checks, we need Rust to be able to infer the types of certain variables. This is easy with `transmute_ref!`, but how would we do it when those types are inside containers? Off the top of my head, I *think* it should be possible via a function like `fn unwrap<C: Container>(c: C) -> C::Element`.
- Can we relax the trait bounds only in certain cases? E.g., `Arc` only requires `Src: IntoBytes + Immutable` and `Dst: FromBytes + Immutable`, but does not require `Src: FromBytes` or `Dst: IntoBytes`

PR #1736, opened 2024-09-23
Add transmute_container! macro

joshlf:
This is a draft implementation of #1735

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,9 @@ zerocopy-core-error = "1.81.0"
 # From 1.78.0, Rust supports the `#[diagnostic::on_unimplemented]` attribute.
 zerocopy-diagnostic-on-unimplemented = "1.78.0"
 
+# From 1.65.0, Rust supports generic associated types (GATs).
+zerocopy-gat = "1.65.0"
+
 # From 1.61.0, Rust supports generic types with trait bounds in `const fn`.
 zerocopy-generic-bounds-in-const-fn = "1.61.0"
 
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -38,6 +38,9 @@ zerocopy-core-error = "1.81.0"
 # From 1.78.0, Rust supports the `#[diagnostic::on_unimplemented]` attribute.
 zerocopy-diagnostic-on-unimplemented = "1.78.0"
 
+# From 1.65.0, Rust supports generic associated types (GATs).
+zerocopy-gat = "1.65.0"
```
jswrenn, line 42: In keeping with our other version-dependent functionality, we should attempt to make this a CFG instead.

src/lib.rs
```diff
@@ -303,6 +303,7 @@
     __ZEROCOPY_INTERNAL_USE_ONLY_NIGHTLY_FEATURES_IN_TESTS,
     feature(layout_for_ptr, strict_provenance, coverage_attribute)
 )]
+#![cfg_attr(feature = "unstable-transmute-vec", feature(vec_into_raw_parts))]
```
jswrenn, line 306: We should attempt to move the trait impl for `Vec` into the macro, so that `feature(vec_into_raw_parts)` is something our customers write instead of us. This makes it harder for customers to accidentally and unkowingly depend on a nightly feature (e.g., because another dependency enabled this cargo feature).

jswrenn (commented):
I like the general approach!

src/lib.rs
joshlf: Any ideas for how to do that given that the macro delegates to a trait impl and doesn't have access to the name of the container type being used?

jswrenn: Ugly but bear with me: Move the trait and impl to inside the macro.

joshlf: Discussed offline. Going to just remove `Vec` support for now.

PR #1737, opened 2024-09-23
[docs] Make infallible align errors discoverable

### LINKED ISSUE #1721: Make infallible alignment errors more discoverable
Following up on #1718, do the following:
- In the `error` module docs, reference infallible alignment errors
- In the doc comment for every method that returns an alignment error (or composite error that might be an alignment error), reference infallible alignment errors

joshlf:
Wherever alignment errors are possible, mention that it is possible to discard alignment errors when `Dst: Unaligned`.

Closes #1721

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -11,7 +11,8 @@
 //!
 //! ## Single failure mode errors
 //!
-//! Generally speaking, zerocopy's conversions may fail for one of up to three reasons:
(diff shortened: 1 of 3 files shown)


joshlf:
@jswrenn What do you think of this approach? If it looks good, I'll propagate it through the codebase.

jswrenn (approved):

PR #1738, opened 2024-09-23
Add TryFromBytes::try_read_from_{prefix,suffix}

joshlf:


src/lib.rs
```diff
@@ -1841,16 +1841,16 @@ pub unsafe trait TryFromBytes {
     /// # use zerocopy_derive::*;
     ///
     /// // The only valid value of this type is the byte `0xC0`
-    /// #[derive(TryFromBytes, KnownLayout, Immutable)]
```
jswrenn, line 1844: Don't need the `Immutable` bound either, right?

jswrenn (changes requested):

src/lib.rs
joshlf: Done

jswrenn (approved):

PR #1752, opened 2024-09-25
[derive] Overhaul repr parsing

### LINKED ISSUE #1748: Soundness hole in `#[derive(IntoBytes)]` on types with `#[repr(align)]`
Consider the following type:

```rust
#[derive(IntoBytes)]
#[repr(C, align(8))]
struct Foo<T> {
    t: T,
}
```

`#[derive(IntoBytes)]` emits an `IntoBytes` impl for `Foo` with a `T: Unaligned` bound. The reasoning is based on the `repr(C)` layout algorithm, but this reasoning is unsound in the presence of `#[repr(align(8))]`, which `#[derive(IntoBytes)]` spuriously ignores.

In particular, `Foo<u8>` satisfies `u8: Unaligned`, but has size 8 (7 bytes of padding) in order to satisfy its alignment requirement.

We need to either ban `#[repr(align(...))]` in `#[derive(IntoBytes)]` or at least ban it when generics are present.

joshlf:
Represent the result of parsing all `#[repr(...)]` attributes on a type as a high-level type which is only capable of representing valid combinations of `#[repr(...)]` attributes and processes them into a concise representation that's easier for high-level code to work with.

This prepares us to more easily fix https://github.com/google/zerocopy/issues/1748.

While we're here, we make a number of other improvements.

1) Errors are now converted to `TokenStream`s as late as possible rather than as early as possible, which was the previous behavior. This allows us to bail early when deriving an implied trait fails (e.g., deriving `TryFromBytes` when the user wrote `#[derive(FromZeros)]`).
2) Avoid re-computing some repr information in `TryFromBytes` enum support.

### DIFF
--- a/zerocopy-derive/src/enum.rs
+++ b/zerocopy-derive/src/enum.rs
@@ -6,42 +6,11 @@
 // This file may not be copied, modified, or distributed except according to
 // those terms.
 
-use ::proc_macro2::TokenStream;
-use ::quote::quote;
-use ::syn::{DataEnum, Fields, Generics, Ident};
-use syn::parse_quote;
+use proc_macro2::{Span, TokenStream};
+use quote::quote;
+use syn::{parse_quote, DataEnum, Error, Fields, Generics, Ident};
 
-use crate::{derive_try_from_bytes_inner, EnumRepr, Trait};
-
-/// Returns the repr for the tag enum, given the collection of reprs on the
-/// enum.
-///
-/// This function returns:
-/// - `Some(C)` for `repr(C)`
-/// - `Some(int)` for `repr(int)` and `repr(C, int)`
-/// - `None` for all other reprs
-pub(crate) fn tag_repr(reprs: &[EnumRepr]) -> Option<&EnumRepr> {
-    let mut result = None;
-    for repr in reprs {
-        match repr {
-            EnumRepr::C => result = Some(repr),
-            EnumRepr::U8
-            | EnumRepr::U16
-            | EnumRepr::U32
-            | EnumRepr::U64
-            | EnumRepr::Usize
-            | EnumRepr::I8
-            | EnumRepr::I16
-            | EnumRepr::I32
-            | EnumRepr::I64
-            | EnumRepr::Isize => {
-                return Some(repr);
-            }
-            _ => (),
-        }
-    }
-    result
-}
+use crate::{derive_try_from_bytes_inner, repr::EnumRepr, Trait};
 
 /// Generates a tag enum for the given enum. This generates an enum with the
 /// same `repr`s, variants, and corresponding discriminants, but none of the
@@ -57,7 +26,7 @@ pub(crate) fn generate_tag_enum(repr: &EnumRepr, data: &DataEnum) -> TokenStream
     });
 
     quote! {
-        #[repr(#repr)]
+        #repr
         #[allow(dead_code)]
         enum ___ZerocopyTag {
             #(#variants,)*
@@ -157,7 +126,8 @@ fn generate_variant_structs(
 
         // We do this rather than emitting `#[derive(::zerocopy::TryFromBytes)]`
         // because that is not hygienic, and this is also more performant.
-        let try_from_bytes_impl = derive_try_from_bytes_inner(&variant_struct);
+        let try_from_bytes_impl = derive_try_from_bytes_inner(&variant_struct)
+            .expect("derive_try_from_bytes_inner should not fail on synthesized type");
 
         Some(quote! {
             #variant_struct
@@ -231,21 +201,23 @@ fn generate_variants_union(generics: &Generics, data: &DataEnum) -> TokenStream 
 /// - `repr(C, int)`: <https://doc.rust-lang.org/reference/type-layout.html#combining-primitive-representations-of-enums-with-fields-and-reprc>
 pub(crate) fn derive_is_bit_valid(
     enum_ident: &Ident,
-    reprs: &[EnumRepr],
+    repr: &EnumRepr,
     generics: &Generics,
     data: &DataEnum,
-) -> TokenStream {
-    let repr =
-        tag_repr(reprs).expect("cannot derive is_bit_valid for enum without a well-defined repr");
-
+) -> Result<TokenStream, Error> {
     let trait_path = Trait::TryFromBytes.crate_path();
     let tag_enum = generate_tag_enum(repr, data);
     let tag_consts = generate_tag_consts(data);
 
-    let (outer_tag_type, inner_tag_type) = if matches!(repr, EnumRepr::C) {
+    let (outer_tag_type, inner_tag_type) = if repr.is_c() {
         (quote! { ___ZerocopyTag }, quote! { () })
-    } else {
+    } else if repr.is_primitive() {
         (quote! { () }, quote! { ___ZerocopyTag })
+    } else {
+        return Err(Error::new(
+            Span::call_site(),
+            "must have #[repr(C)] or #[repr(Int)] attribute in order to guarantee this type's memory layout",
+        ));
     };
 
     let variant_structs = generate_variant_structs(enum_ident, generics, data);
@@ -297,7 +269,7 @@ pub(crate) fn derive_is_bit_valid(
         }
     });
 
-    quote! {
+    Ok(quote! {
         // SAFETY: We use `is_bit_valid` to validate that the bit pattern of the
         // enum's tag corresponds to one of the enum's discriminants. Then, we
         // check the bit validity of each field of the corresponding variant.
@@ -392,5 +364,5 @@ pub(crate) fn derive_is_bit_valid(
                 _ => false,
             }
         }
-    }
+    })
 }
--- a/zerocopy-derive/src/lib.rs
+++ b/zerocopy-derive/src/lib.rs
@@ -48,17 +48,6 @@ use {
 
 use {crate::ext::*, crate::repr::*};
 
-/// Unwraps a `Result<_, Vec<Error>>`, converting any `Err` value into a
-/// `TokenStream` and returning it.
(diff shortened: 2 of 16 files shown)


zerocopy-derive/tests/ui-msrv/enum.stderr
```diff
@@ -101,125 +103,143 @@ error: FromZeros only supported on enums with a variant that has a discriminant
 138 | | }
     | |_^
 
-error: FromBytes requires repr of "u8", "u16", "i8", or "i16"
+error:
```
joshlf, line 106: TODO: Figure out why this is happening and fix it.

zerocopy-derive/src/enum.rs
```diff
@@ -57,7 +26,7 @@ pub(crate) fn generate_tag_enum(repr: &EnumRepr, data: &DataEnum) -> TokenStream
     });
 
     quote! {
-        #[repr(#repr)]
+        #repr
```
joshlf, line 29: This new behavior implicates #1758. In particular, this now delegates to the `ToTokens` impl for `EnumRepr`, which includes `align` reprs.

cc @djkoloski

zerocopy-derive/src/repr.rs
```diff
@@ -6,202 +6,276 @@
...
+    Transparent(S),
+    /// A compound representation: `repr(C)`, `repr(Rust)`, or `repr(Int)`
+    /// optionally combined with `repr(packed(...))` or `repr(align(...))`
+    Compound(Spanned<CompoundRepr<Prim>, S>, Option<Spanned<AlignRepr<Packed>, S>>),
+}
+
+/// A compound representation: `repr(C)`, `repr(Rust)`, or `repr(Int)`.
+#[cfg_attr(test, derive(Copy, Clone, Debug, Eq, PartialEq))]
+pub(crate) enum CompoundRepr<Prim> {
+    C,
```
djkoloski, line 38: Data-carrying enums can be `repr(C, int)`: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=555030b1832a3da32661af452c7d5809

That repr combination can only be used on data-carrying enums, otherwise the compiler emits an error. When `C` and `int` reprs are used together, the enum follows C-style enum layout (outer tag) and explicitly sets the tag primitive.

Also, are we missing a test with this repr combination?

zerocopy-derive/src/repr.rs
```diff
@@ -6,202 +6,276 @@
...
-impl<R: KindRepr> Config<R> {
-    /// Validate that `input`'s representation attributes conform to the
-    /// requirements specified by this `Config`.
+/// The computed representation of a type.
+///
+/// This is the result of processing all `#[repr(...)]` attributes on a type, if
+/// any. A `Repr` is only capable of representing legal combinations of
+/// `#[repr(...)]` attributes.
+#[cfg_attr(test, derive(Copy, Clone, Debug, Eq, PartialEq))]
+pub(crate) enum Repr<Prim, Packed, S> {
```
djkoloski, line 27: Why are these repr types generic? This seems like something that has a pretty concrete definition. I didn't see any code which obviously made use of the generic parameters.

zerocopy-derive/src/repr.rs
djkoloski, line 1: This file now contains a lot of type machinery for supporting valid reprs. I suspect that the complexity could be cut down by converting at least some of that machinery from typing into logic. Here's the repr parsing code I wrote for rkyv for comparison: https://github.com/rkyv/rkyv/blob/main/rkyv_derive/src/repr.rs

djkoloski (changes requested):

zerocopy-derive/src/lib.rs
```diff
@@ -87,11 +76,30 @@ macro_rules! derive {
         #[proc_macro_derive($trait)]
         pub fn $outer(ts: proc_macro::TokenStream) -> proc_macro::TokenStream {
             let ast = syn::parse_macro_input!(ts as DeriveInput);
-            $inner(&ast).into()
+            $inner(&ast).into_ts().into()
         }
     };
 }
 
+trait IntoTokenStream {
```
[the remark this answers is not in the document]
joshlf: Maybe. It's just added boilerplate in a number of places. This allows the support to be more local rather than force all of the functions to return `Result`s and wrap things in `Ok`, but either's fine - it's not a huge difference either way.

zerocopy-derive/src/lib.rs
```diff
@@ -131,35 +139,26 @@ fn derive_known_layout_inner(ast: &DeriveInput) -> proc_macro2::TokenStream {
...
-                },
-            )
-            .map(|repr_align| quote!(#core_path::num::NonZeroUsize::new(#repr_align as usize)))
+        let repr_align = repr
+            .get_align()
+            .map(|align| {
+                let align = align.t.get();
+                quote!(#core_path::num::NonZeroUsize::new(#align as usize))
+            })
             .unwrap_or(quote!(#core_path::option::Option::None));
```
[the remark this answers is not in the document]
joshlf: Done.

zerocopy-derive/src/lib.rs
```diff
@@ -131,35 +139,26 @@ fn derive_known_layout_inner(ast: &DeriveInput) -> proc_macro2::TokenStream {
...
-                Repr::PackedN(repr_packed) => Some(*repr_packed),
-                _ => None,
+        let repr_packed = repr
+            .get_packed()
+            .map(|packed| {
+                let packed = packed.get();
+                quote!(#core_path::num::NonZeroUsize::new(#packed as usize))
             })
-            .map(|repr_packed| quote!(#core_path::num::NonZeroUsize::new(#repr_packed as usize)))
             .unwrap_or(quote!(#core_path::option::Option::None));
```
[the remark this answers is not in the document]
joshlf: Done.

zerocopy-derive/src/repr.rs
joshlf: That's probably true, and I can work to cut it down somewhat. However, I'd like to keep the property that only valid reprs are representable, since that permits user code to exhaustively match, which I believe is important in some cases.

There's also a good amount of machinery to track span information, which is useful for error reporting. One thing I'd *love* to figure out - maybe you have thoughts here - is how to make it easier to match on a given pattern without having to write `Spanned { t: <pattern>, span: _ }`, which makes a lot of the match statements much more verbose than they need to be in theory.

Part of the extra machinery here (specifically regarding generics) exists to support testing (namely, abstracting over `Span` vs `()` for span information). I'm not sure how to get rid of that in a way that doesn't introduce runtime panic opportunities (e.g. make every `Span` an `Option<Span>`).

zerocopy-derive/src/repr.rs
joshlf: Yeah, my worry is that the logic around those is super subtle. My intention (which I should have written down) was to just not support them for now and then follow up later to add support if we need to. I'm not aware of any of our users using them.

> Also, are we missing a test with this repr combination?

No, we currently test that they're rejected:

```rust
test!(@error #[repr(C, u8)] => EnumRepr => FromRawReprs(Conflict).into());
```

zerocopy-derive/src/repr.rs
joshlf: - Structs and unions cannot be `repr(Int)`, so the `Prim` type is set to `Immutable` for them (in `type StructUnionRepr`)
- Enums cannot be `repr(packed)`, so the `Packed` type is set to `Immutable` for enums (in `type EnumRepr`)
- `S` is for span information - it's `Span` normally, but stubbed as `()` in tests since `Span`s aren't easy to synthesize in testing

I could be convinced that there's a better approach for stubbing out span information - e.g. maybe I can just write the tests so that they manually ignore span information when computing equality.

zerocopy-derive/src/repr.rs
joshlf: > Part of the extra machinery here (specifically regarding generics) exists to support testing (namely, abstracting over `Span` vs `()` for span information). I'm not sure how to get rid of that in a way that doesn't introduce runtime panic opportunities (e.g. make every `Span` an `Option<Span>`).

You inspired me to remove this. Turned out to be fairly straightforward.

joshlf: Okay, did another round of cleanups. Turns out that the `With` machinery is powerful enough to get rid of a bunch of the `TryFrom` trait bounds.

zerocopy-derive/src/repr.rs
joshlf: > I could be convinced that there's a better approach for stubbing out span information - e.g. maybe I can just write the tests so that they manually ignore span information when computing equality.

Did this.

zerocopy-derive/src/enum.rs
joshlf: IMO the right thing to do here is to strip out any align attribute and only keep transparent, c, or int. That will ensure that this PR doesn't change behavior.

Then, we should shortly afterwards follow up with https://github.com/google/zerocopy/issues/1758 and resolve any bugs.

zerocopy-derive/src/repr.rs
```diff
@@ -212,128 +266,561 @@ pub(crate) enum Repr {
...
-impl KindRepr for Repr {
-    fn is_align(&self) -> bool {
-        false
+pub(crate) use util::*;
+mod util {
+    use super::*;
+    /// A value with an associated span.
+    #[derive(Copy, Clone)]
+    #[cfg_attr(test, derive(Debug))]
+    pub(crate) struct Spanned<T> {
```
joshlf, line 590: TODO: Make this a tuple struct to make pattern matching more ergonomic

joshlf: Moved to #1780

jswrenn (approved):
Reviewed over video call.

PR #1785, opened 2024-09-30
[byteorder] Impl traits for byte arrays

### LINKED ISSUE #1749: [byteorder] Support trait impls on byte arrays
We currently support `PartialEq<[u8; N> for T` and `PartialEq<T> for [u8; N]` for byteorder type `T`. We should do the same for other traits.

joshlf:
Closes #1749


joshlf (commented):
TODO:
- Audit impls for correctness (I banged them out pretty mechanically, so there could be endianness errors)
- Tests

PR #1786, opened 2024-10-01
aarch64 fix for big endian types

lucarlig:
NEON intrinsics are not supported on aarch64 big endian targets. Change to use fall back types for big endian aarch64.
tested on aarch64_be-unknown-linux-gnu_ilp32 and aarch64_be-unknown-linux-gnu

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -575,6 +575,40 @@ jobs:
           # `roll-pinned-toolchain-versions.yml`.
           kani-version: 0.55.0
 
+  # NEON intrinsics are currently broken on big-endian platforms. [1] This test ensures
+  # that we don't accidentally attempt to compile these intrinsics on such platforms. We
+  # can't use this as part of the build matrix because rustup doesn't support the
+  # `aarch64_be-unknown-linux-gnu` target.
+  #
+  # [1] https://github.com/rust-lang/stdarch/issues/1484
+  check_be_aarch64:
+    runs-on: ubuntu-latest
+    name: Build (zerocopy / nightly / --simd / aarch64_be-unknown-linux-gnu)
+    steps:
+      - uses: actions/checkout@d632683dd7b4114ad314bca15554477dd762a938 # v4.2.0
+      - name: Configure environment variables
+        run: |
+          set -eo pipefail
+          ZC_TOOLCHAIN="$(./cargo.sh --version nightly)"
+          RUSTFLAGS="$RUSTFLAGS $ZC_NIGHTLY_RUSTFLAGS"
+          echo "RUSTFLAGS=$RUSTFLAGS" >> $GITHUB_ENV
+          echo "ZC_TOOLCHAIN=$ZC_TOOLCHAIN" >> $GITHUB_ENV
+      - name: Rust Cache
+        uses: Swatinem/rust-cache@23bce251a8cd2ffc3c1075eaa2367cf899916d84 # v2.7.3
+        with:
+          key: aarch64_be-unknown-linux-gnu
+      - name: Install stable Rust for use in 'cargo.sh'
+        uses: dtolnay/rust-toolchain@00b49be78f40fba4e87296b2ead62868750bdd83 # stable
+        with:
+          toolchain: stable
+      - name: Install Rust with nightly toolchain (${{ env.ZC_TOOLCHAIN }}) and target aarch64_be-unknown-linux-gnu
+        uses: dtolnay/rust-toolchain@00b49be78f40fba4e87296b2ead62868750bdd83 # stable
+        with:
+            toolchain: ${{ env.ZC_TOOLCHAIN }}
+            components: clippy, rust-src
+      - name: Check big endian for aarch64_be-unknown-linux-gnu target
+        run: ./cargo.sh +nightly build --target=aarch64_be-unknown-linux-gnu -Zbuild-std --features simd
+
   check_fmt:
     runs-on: ubuntu-latest
     name: Check Rust formatting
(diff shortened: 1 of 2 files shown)


joshlf:
Thanks for this! A few small things:
- Can you rebase on `main` and force-push this as a single commit?
- Can you update your commit message to follow our [commit message format](https://github.com/google/zerocopy/discussions/1318)?
- Is there a citation for the fact that NEON intrinsics aren't supported on big-endian targets? If you could add a comment with a citation in the source code, that'd be good.
- Can you add an `include` section [here](https://github.com/google/zerocopy/blob/8abf7dacd56339e7a232bf7e2ec4906edc3d4a26/.github/workflows/ci.yml#L242) and use [matrix includes](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#expanding-or-adding-matrix-configurations) to add the following two build combinations. Please add a comment that references this PR and explains why we test on these combinations.
  - `toolchain: nightly` / `target: aarch64_be-unknown-linux-gnu_ilp32` / `features: --all-features` / `crate: zerocopy`
  - `toolchain: nightly` / `target: aarch64_be-unknown-linux-gnu` / `features: --all-features` / `crate: zerocopy`

lucarlig:
@joshlf  fixed the stuff that you requested,
this is the issue you are looking for as evidence https://github.com/rust-lang/stdarch/issues/1484
I don't think the ilp32 version tests anything different so I skipped including that

lucarlig:
it's failing because 'aarch64_be-unknown-linux-gnu' is not provided by rustup, only rustc
to test should do something like:
```
cargo build --target=aarch64_be-unknown-linux-gnu -Zbuild-std
```

src/impls.rs
```diff
@@ -934,7 +934,7 @@ mod simd {
         );
         #[cfg(zerocopy_aarch64_simd)]
         simd_arch_mod!(
-            #[cfg(target_arch = "aarch64")]
+            #[cfg(all(target_arch = "aarch64", target_endian = "little"))]
```
joshlf, line 937:
```suggestion
            // NOTE(https://github.com/rust-lang/stdarch/issues/1484): NEON intrinsics are currently
            // broken on big-endian platforms.
            #[cfg(all(target_arch = "aarch64", target_endian = "little"))]
```

joshlf:
> it's failing because 'aarch64_be-unknown-linux-gnu' is not provided by rustup, only rustc to test should do something like:
> 
> ```
> cargo build --target=aarch64_be-unknown-linux-gnu -Zbuild-std
> ```

Ah gotcha. I confirmed locally that this works:

```shell
$ ./cargo.sh +nightly build --target=aarch64_be-unknown-linux-gnu -Zbuild-std --features simd
```

Can you add a new build target (under `kani` and before `check_fmt`) that runs that command? That can replace the other changes to `.github/workflows/ci.yml`.

joshlf (changes requested):

joshlf (approved):
Okay amazing, thank you for your work on this! Once this merges, I can publish a new release that includes it.

joshlf:
Unfortunately I think that the build failure you're currently seeing is going to require you to compute the correct nightly to install based on the contents of `Cargo.toml`. You can extract that via `./cargo.sh --version nightly` and then tell the GitHub action to install that particular nightly.

Here's how we do that in the build matrix - here it should be much simpler because you only need to support a single toolchain and none of the Miri stuff is relevant:

https://github.com/google/zerocopy/blob/a9f09d7afc526a3a54a7a37b991eca0f22358950/.github/workflows/ci.yml#L264-L321

lucarlig:
@joshlf not sure what dependency is missing

.github/workflows/ci.yml
```diff
@@ -575,6 +575,42 @@ jobs:
...
+  check_be_aarch64:
+    runs-on: ubuntu-latest
+    name: Build (zerocopy / nightly / --simd / aarch64_be-unknown-linux-gnu)
+    steps:
+      - uses: actions/checkout@d632683dd7b4114ad314bca15554477dd762a938 # v4.2.0
+      - name: Configure environment variables
+        run: |
+          set -eo pipefail
+          ZC_TOOLCHAIN="$(./cargo.sh --version nightly)"
+          MIRIFLAGS="$MIRIFLAGS $ZC_NIGHTLY_MIRIFLAGS"
```
joshlf, line 593: I don't believe you need this since we don't use Miri in this job.

```suggestion
```

.github/workflows/ci.yml
```diff
@@ -575,6 +575,42 @@ jobs:
...
+          RUSTFLAGS="$RUSTFLAGS $ZC_NIGHTLY_RUSTFLAGS"
+          echo "RUSTFLAGS=$RUSTFLAGS" >> $GITHUB_ENV
+          echo "ZC_TOOLCHAIN=$ZC_TOOLCHAIN" >> $GITHUB_ENV
+          echo "MIRIFLAGS=$MIRIFLAGS" >> $GITHUB_ENV
```
joshlf, line 597: Same here.

```suggestion
```

.github/workflows/ci.yml
```diff
@@ -575,6 +575,42 @@ jobs:
...
+          key: aarch64_be-unknown-linux-gnu
+      - name: Install stable Rust for use in 'cargo.sh'
+        uses: dtolnay/rust-toolchain@00b49be78f40fba4e87296b2ead62868750bdd83 # stable
+        with:
+          toolchain: stable
+      - name: Install Rust with nightly toolchain (${{ env.ZC_TOOLCHAIN }}) and target aarch64_be-unknown-linux-gnu
+        uses: dtolnay/rust-toolchain@00b49be78f40fba4e87296b2ead62868750bdd83 # stable
+        with:
+            toolchain: ${{ env.ZC_TOOLCHAIN }}
+            components: clippy, rust-src, miri
```
joshlf, line 610: Do you need `miri`?

joshlf (changes requested):

joshlf:
> @joshlf not sure what dependency is missing

You need to edit this line:

https://github.com/google/zerocopy/blob/a9f09d7afc526a3a54a7a37b991eca0f22358950/.github/workflows/ci.yml#L707

...to include `check_be_aarch64`

joshlf (approved):

joshlf:
Thanks again for your work on this!

workingjubilee:
Please fix this by fixing the intrinsics upstream. Fixing this in every single crate is not less effort than fixing it in stdarch: https://github.com/rust-lang/stdarch/issues/1484#issuecomment-1942934003

joshlf:
@workingjubilee is the lack of support in zerocopy breaking you? I'm hesitant for us to own a fix in stdarch unless there are users who actively need this feature.

joshlf:
> @joshlf Just to be clear, you're fine. If you want to accept this fix, that's okay. But I just don't think that fixing zerocopy, memchr, half, etc., in a fix that might be best-off reverted once stdarch is fixed, is better than fixing stdarch.

Okay gotcha, that makes senes.

PR #1793, opened 2024-10-02
[docs] Link to Discord

joshlf:

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,8 +25,13 @@ so you don't have to.
 check out our [release notes][release-notes], which include a step-by-step
 guide for upgrading from 0.7.*
 
+*Have questions? Need help? Ask the maintainers on [GitHub][github-q-a] or
+on [Discord][discord]!*
+
 [customer-request-issue]: https://github.com/google/zerocopy/issues/new/choose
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -25,8 +25,11 @@ so you don't have to.
 check out our [release notes][release-notes], which include a step-by-step
 guide for upgrading from 0.7.*
 
+*Have questions? Need help? Ask the maintainers [on Discord][discord]!*
```
jswrenn, line 28: Let's also link to [Github Discussions](https://github.com/google/zerocopy/discussions/categories/q-a)?

joshlf: What do you think the resulting text should look like? Maybe we should have at least a vague policy on what goes in Discussions and what goes on Discord?

jswrenn: > Ask the maintainers on Github or Discord

I think we have to meet people where they are / have an account. If it were up to me, we'd *only* be using Discord, but that's because I'm not hip with the kids.

jswrenn (approved):

README.md
joshlf: Done

PR #1795, opened 2024-10-02
src/impls.rs: don't attempt use of neon / SIMD on big-endian aarch64.

he32:
Neon / SIMD is known to be problematical in rust, ref. https://github.com/rust-lang/stdarch/issues/1484, even though the CPU itself supports it.


joshlf:
Looks like this is a duplicate of https://github.com/google/zerocopy/pull/1786. I'm gonna close it, but feel free to re-open if you disagree that it's a duplicate.

PR #1798, opened 2024-10-03
Add support for fmt-debug

bluec0re:
The unstable feature [fmt-debug](https://doc.rust-lang.org/beta/unstable-book/compiler-flags/fmt-debug.html) allows to influence `derive(Debug)` implementations, for example by making it return an empty string. As the standard library also does not make any guarantees on the outputs of `derive(Debug)` , this change does replace the reliance on it by an explicit conversion to the variant name.

### DIFF
--- a/zerocopy-derive/src/lib.rs
+++ b/zerocopy-derive/src/lib.rs
@@ -1114,7 +1114,26 @@ enum Trait {
 
 impl ToTokens for Trait {
     fn to_tokens(&self, tokens: &mut TokenStream) {
-        let ident = Ident::new(&format!("{:?}", self), Span::call_site());
+        // According to [1], the format of the derived `Debug`` output is not
+        // stable and therefore not guaranteed to represent the variant names.
+        // Indeed with the (unstable) `fmt-debug` compiler flag [2], it can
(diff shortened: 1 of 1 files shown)


zerocopy-derive/src/lib.rs
```diff
@@ -1076,7 +1076,17 @@ enum Trait {
 
 impl ToTokens for Trait {
     fn to_tokens(&self, tokens: &mut TokenStream) {
-        let ident = Ident::new(&format!("{:?}", self), Span::call_site());
+        let s = match self {
```
joshlf, line 1079: Can you add a comment here explaining why this verbosity is necessary and just using Debug isn't sufficient?

joshlf (changes requested):
Wow, had no idea, thanks! One small change and I'll approve this.

zerocopy-derive/src/lib.rs
bluec0re: Done. PTAL

joshlf (approved):
Awesome, thanks! I pushed a minor edit to this to make it consistent with our commit message style (including squashing everything into a single commit), and it should merge as soon as it goes through CI and the merge queue.

Issue #1792, opened 2024-10-02
Remove support for `#[derive(IntoBytes)]` on unions?


joshlf:
Support for `#[derive(IntoBytes)]` on unions was added in https://github.com/google/zerocopy/commit/9c19cbe1cf5c2621ae9a63179664db3470146c10 (reviewed at [fxrev.dev/639087](https://fuchsia-review.googlesource.com/c/fuchsia/+/639087)). This support is sound on the assumption that if every field of a union is `IntoBytes` and there is no extra padding before or after any field, then no bit-valid instance of that union can have uninitialized bytes.

However, this assumes too much. It is currently [up in the air](https://github.com/rust-lang/unsafe-code-guidelines/issues/438) whether this actually holds. It's not clear from reading the history why we were okay adding this implementation in the first place, but it now seems like it may have been premature.

It may eventually become the case that this assumption is guaranteed by Rust, so this may eventually become a non-issue. However, for the time being, it's a violation of our soundness policy, which [promises to be sound on all future Rust compilers](https://github.com/google/zerocopy/blob/main/POLICIES.md#forwards-compatibility).

Unfortunately, existing users rely on `#[derive(IntoBytes)]` support on unions, so removing this support will break users. This is likely not a problem on today's Rust, so forcing users to migrate to something else might be too drastic of a solution. Instead, I propose that we discourage further use by gating `#[derive(IntoBytes)]` on unions behind a `--cfg`. The reason for a `--cfg` instead of a Cargo feature is that Cargo features are footguns in cases like this - it's easy for crate A to enable the feature and then for crate B, which depends on A, to accidentally rely on that feature despite not enabling that feature itself. For something with soundness and stability implications, that's risky. By contrast, a `--cfg` requires the top-level crate to enable it for all downstream crates.

In parallel, we will try to get Rust to promise what we need in order for `#[derive(IntoBytes)]` on unions to be guaranteed sound on all future compilers.

joshlf:
Summarizing a discussion with @jswrenn 

The situation may be more permissive than I originally thought. In particular, while it *might* be the case at some point in the future that any union with all uninitialized bytes is considered bit-valid (ie, it is *sound* to produce such a union value), it is currently the case that creating such a value requires `unsafe` code. Safe code is only permitted to instantiate or modify a union value using its actual fields.

[...]

PR #1804, opened 2024-10-03
[derive] IntoBytes on unions requires --cfg

joshlf:
Makes progress on #1792


joshlf:
I've [asked for advice on the build failures](https://users.rust-lang.org/t/why-is-this-cfg-only-getting-passed-when-i-dont-cross-compile/119168) on URLO.

jswrenn (approved):

PR #1822, opened 2024-10-05
Add KnownLayout::MaybeUninit type

### LINKED ISSUE #1797: Support `MaybeUninit<T: ?Sized>` via `KnownLayout`
This could help unify many standard library APIs that are currently duplicated, and make them more powerful.

For example, consider this duplication:
- https://doc.rust-lang.org/std/boxed/struct.Box.html#method.new_uninit
- https://doc.rust-lang.org/std/boxed/struct.Box.html#method.new_uninit_slice

joshlf:
Closes #1797

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -349,7 +349,7 @@ use core::{
     fmt::{self, Debug, Display, Formatter},
     hash::Hasher,
     marker::PhantomData,
-    mem::{self, ManuallyDrop, MaybeUninit},
+    mem::{self, ManuallyDrop, MaybeUninit as CoreMaybeUninit},
     num::{
         NonZeroI128, NonZeroI16, NonZeroI32, NonZeroI64, NonZeroI8, NonZeroIsize, NonZeroU128,
         NonZeroU16, NonZeroU32, NonZeroU64, NonZeroU8, NonZeroUsize, Wrapping,
@@ -727,6 +727,14 @@ pub unsafe trait KnownLayout {
     /// This is `()` for sized types and `usize` for slice DSTs.
     type PointerMetadata: PointerMetadata;
 
+    /// A maybe-uninitialized analog of `Self`
+    ///
+    /// # Safety
+    ///
+    /// `Self::LAYOUT` and `Self::MaybeUninit::LAYOUT` are identical.
+    #[doc(hidden)]
+    type MaybeUninit: ?Sized + KnownLayout<PointerMetadata = Self::PointerMetadata>;
+
     /// The layout of `Self`.
     ///
     /// # Safety
@@ -859,6 +867,35 @@ unsafe impl<T> KnownLayout for [T] {
 
     type PointerMetadata = usize;
 
+    // SAFETY: `CoreMaybeUninit<T>::LAYOUT` and `T::LAYOUT` are identical
+    // because `CoreMaybeUninit<T>` has the same size and alignment as `T` [1].
+    // Consequently, `[CoreMaybeUninit<T>]::LAYOUT` and `[T]::LAYOUT` are
+    // identical, because they both lack a fixed-sized prefix and because they
+    // inherit the alignments of their inner element type (which are identical)
+    // [2][3].
+    //
+    // `[CoreMaybeUninit<T>]` admits uninitialized bytes at all positions
+    // because `CoreMaybeUninit<T>` admits uninitialized bytes at all positions
+    // and because the inner elements of `[CoreMaybeUninit<T>]` are laid out
+    // back-to-back [2][3].
+    //
+    // [1] Per https://doc.rust-lang.org/1.81.0/std/mem/union.MaybeUninit.html#layout-1:
+    //
+    //   `MaybeUninit<T>` is guaranteed to have the same size, alignment, and ABI as
+    //   `T`
+    //
+    // [2] Per https://doc.rust-lang.org/1.82.0/reference/type-layout.html#slice-layout:
+    //
+    //   Slices have the same layout as the section of the array they slice.
+    //
+    // [3] Per https://doc.rust-lang.org/1.82.0/reference/type-layout.html#array-layout:
+    //
+    //   An array of `[T; N]` has a size of `size_of::<T>() * N` and the same
+    //   alignment of `T`. Arrays are laid out so that the zero-based `nth`
+    //   element of the array is offset from the start of the array by `n *
+    //   size_of::<T>()` bytes.
+    type MaybeUninit = [CoreMaybeUninit<T>];
+
     const LAYOUT: DstLayout = DstLayout::for_slice::<T>();
 
     // SAFETY: `.cast` preserves address and provenance. The returned pointer
@@ -911,7 +948,7 @@ impl_known_layout!(
     T         => Option<T>,
     T: ?Sized => PhantomData<T>,
     T         => Wrapping<T>,
-    T         => MaybeUninit<T>,
+    T         => CoreMaybeUninit<T>,
     T: ?Sized => *const T,
     T: ?Sized => *mut T
 );
@@ -942,6 +979,21 @@ safety_comment! {
     unsafe_impl_known_layout!(#[repr([u8])] str);
     unsafe_impl_known_layout!(T: ?Sized + KnownLayout => #[repr(T)] ManuallyDrop<T>);
     unsafe_impl_known_layout!(T: ?Sized + KnownLayout => #[repr(T)] UnsafeCell<T>);
+}
+
+safety_comment! {
+    /// SAFETY:
+    /// - By consequence of the invariant on `T::MaybeUninit` that `T::LAYOUT`
+    ///   and `T::MaybeUninit::LAYOUT` are equal, `T` and `T::MaybeUninit`
+    ///   have the same:
+    ///   - Fixed prefix size
+    ///   - Alignment
+    ///   - (For DSTs) trailing slice element size
+    /// - By consequence of the above, referents `T::MaybeUninit` and `T` have
+    ///   the require the same kind of pointer metadata, and thus it is valid to
+    ///   perform an `as` cast from `*mut T` and `*mut T::MaybeUninit`, and this
+    ///   operation preserves referent size (ie, `size_of_val_raw`).
+    unsafe_impl_known_layout!(T: ?Sized + KnownLayout => #[repr(T::MaybeUninit)] MaybeUninit<T>);
 }
 
 /// Analyzes whether a type is [`FromZeros`].
@@ -2545,7 +2597,7 @@ pub unsafe trait TryFromBytes {
     where
         Self: Sized,
     {
-        let candidate = match MaybeUninit::<Self>::read_from_bytes(source) {
+        let candidate = match CoreMaybeUninit::<Self>::read_from_bytes(source) {
             Ok(candidate) => candidate,
             Err(e) => {
                 return Err(TryReadError::Size(e.with_dst()));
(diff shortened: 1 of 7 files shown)


build.rs
```diff
@@ -83,6 +83,7 @@ fn main() {
         println!(
             "cargo:rustc-check-cfg=cfg(__ZEROCOPY_INTERNAL_USE_ONLY_NIGHTLY_FEATURES_IN_TESTS)"
         );
+        println!("cargo:rustc-check-cfg=cfg(zerocopy_unstable)");
```
joshlf, line 86: How do we make sure we test both w/ and w/o this cfg in CI without duplicating every CI test?

jswrenn: I think our options here are to either:
- test on all our CI targets and eat the resource cost
- test on a subset of CI and eat the risk
- maintain a contract of `zerocopy_unstable` only being additive
- don't use zerocopy_unstable and use `doc(hidden)` instead

If our main motivation for landing `KnownLayout::MaybeUninit` is unblocking a generic value type, then I'm inclined to say we should be using `doc(hidden)`, because that API would subsume `MaybeUninit` *and* `UnalignUnsized`.

joshlf: Discussed offline. Will land in 0.9, upgrading the existing `Unalign` type to support unsized types rather than adding a new `UnalignUnsized`.

src/wrappers.rs
```diff
@@ -464,6 +464,191 @@ impl<T: Unaligned + Display> Display for Unalign<T> {
...
+#[cfg(zerocopy_unstable)]
+pub struct MaybeUninit<T: ?Sized + KnownLayout>(T::MaybeUninit);
+
+#[cfg(zerocopy_unstable)]
+impl<T: ?Sized + KnownLayout> MaybeUninit<T> {
+    /// Constructs a `MaybeUninit<T>` initialized with the given value.
+    #[inline(always)]
+    pub fn new(val: T) -> Self
+    where
+        T: Sized + KnownLayout<PointerMetadata = ()>,
```
joshlf, line 497:
```suggestion
        T: Sized,
```

src/wrappers.rs
```diff
@@ -464,6 +464,191 @@ impl<T: Unaligned + Display> Display for Unalign<T> {
...
+        // `T::MaybeUninit: Sized` bound to this function's `where` clause.
+        unsafe { crate::util::transmute_unchecked(val) }
+    }
+
+    /// Constructs an uninitialized `MaybeUninit<T>`.
+    #[must_use]
+    #[inline(always)]
+    pub fn uninit() -> Self
+    where
+        T: Sized + KnownLayout<PointerMetadata = ()>,
```
joshlf, line 523:
```suggestion
        T: Sized,
```

src/wrappers.rs
```diff
@@ -464,6 +464,191 @@ impl<T: Unaligned + Display> Display for Unalign<T> {
...
+        Self: Sized,
+    {
+        let uninit = CoreMaybeUninit::<T>::uninit();
+        // SAFETY: It is valid to transmute `CoreMaybeUninit<T>` to
+        // `MaybeUninit<T>` as both have the same size (i.e., that of `T`) and
```
joshlf, line 528: Prove the size equality and bit validity? In particular:
- `CoreMaybeUninit<T>` has the same size as `T`
- `CoreMaybeUninit<T>` has no bit validity
- `MaybeUninit<T>` has the same size as `T`
- `MaybeUninit<T>` has no bit validity

src/wrappers.rs
```diff
@@ -464,6 +464,191 @@ impl<T: Unaligned + Display> Display for Unalign<T> {
...
+        // bit validity (i.e., admitting uninitialized bytes).
+        unsafe { crate::util::transmute_unchecked(uninit) }
+    }
+
+    /// Creates a `Box<MaybeUninit<T>>`.
+    ///
+    /// This function is useful for allocating large, uninit values on the heap,
```
joshlf, line 535:
```suggestion
    /// This function is useful for allocating large, uninit values on the heap
```

src/wrappers.rs
```diff
@@ -464,6 +464,191 @@ impl<T: Unaligned + Display> Display for Unalign<T> {
...
+    /// without ever creating a temporary instance of `Self` on the stack.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error on allocation failure. Allocation failure is guaranteed
+    /// never to cause a panic or an abort.
+    #[cfg(feature = "alloc")]
+    #[inline]
+    pub fn new_boxed_uninit(meta: T::PointerMetadata) -> Result<Box<Self>, AllocError> {
```
joshlf, line 544: Can we refactor so this isn't duplicated from `FromZeros::new_boxed`?

src/wrappers.rs
```diff
@@ -464,6 +464,191 @@ impl<T: Unaligned + Display> Display for Unalign<T> {
...
+    /// Extracts the value from the `MaybeUninit<T>`` container.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that `self` is in an initialized state. Depending
+    /// on subsequent use, it may also need to be in a valid state.
+    #[inline(always)]
+    pub const unsafe fn assume_init(self) -> T
+    where
+        T: Sized + KnownLayout<PointerMetadata = ()>,
```
joshlf, line 636:
```suggestion
        T: Sized,
```

src/wrappers.rs
```diff
@@ -464,6 +464,191 @@ impl<T: Unaligned + Display> Display for Unalign<T> {
...
+    pub const fn as_mut_ptr(&self) -> *const T {
+        todo!()
+    }
+
+    /// Extracts the value from the `MaybeUninit<T>`` container.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that `self` is in an initialized state. Depending
+    /// on subsequent use, it may also need to be in a valid state.
```
joshlf, line 632: Let's be careful about bit validity vs safety here. It *does* need to be in a bit-valid state, but it is okay to not uphold (non-bit validity) safety invariants. Same in the safety comment in the body.

tools/cargo-zerocopy/src/main.rs
```diff
@@ -166,9 +166,9 @@ fn install_toolchain_or_exit(versions: &Versions, name: &str) -> Result<(), Erro
 fn get_rustflags(name: &str) -> &'static str {
     // See #1792 for context on zerocopy_derive_union_into_bytes.
     if name == "nightly" {
-        "--cfg __ZEROCOPY_INTERNAL_USE_ONLY_NIGHTLY_FEATURES_IN_TESTS --cfg zerocopy_derive_union_into_bytes "
+        "--cfg __ZEROCOPY_INTERNAL_USE_ONLY_NIGHTLY_FEATURES_IN_TESTS --cfg zerocopy_derive_union_into_bytes --cfg zerocopy_unstable "
     } else {
-        "--cfg zerocopy_derive_union_into_bytes "
+        "--cfg zerocopy_derive_union_into_bytes --cfg zerocopy_unstable "
```
joshlf, line 171: Obsolete; remove these

zerocopy-derive/src/lib.rs
```diff
@@ -1335,18 +1429,45 @@ fn impl_block<D: DataExt>(
...
+            pub struct __ZerocopyKnownLayoutMaybeUninit<#(#params),*>(
+                #(
+                    ::zerocopy::util::macro_util::core_reexport::mem::MaybeUninit<#leading_fields_tys>,
+                )*
+                <#trailing_field_ty as ::zerocopy::KnownLayout>::MaybeUninit
+            ) where #(#bounds,)*;
+        }
+    } else {
+        quote! {}
+    };
```
joshlf, line 1451: This is all a hack - this should live in `fn derive_known_layout`. It repeats a lot of the logic from that function.

src/wrappers.rs
```diff
@@ -464,6 +464,191 @@ impl<T: Unaligned + Display> Display for Unalign<T> {
...
+pub struct MaybeUninit<T: ?Sized + KnownLayout>(T::MaybeUninit);
+
+#[cfg(zerocopy_unstable)]
+impl<T: ?Sized + KnownLayout> MaybeUninit<T> {
+    /// Constructs a `MaybeUninit<T>` initialized with the given value.
+    #[inline(always)]
+    pub fn new(val: T) -> Self
+    where
+        T: Sized + KnownLayout<PointerMetadata = ()>,
+        Self: Sized,
```
joshlf, line 498: Before we merge this, let's try to make these bounds unnecessary or track somewhere that we want to eventually get rid of them. This will make it painful to use in a generic context.

src/lib.rs
```diff
@@ -941,6 +969,19 @@ safety_comment! {
...
+safety_comment! {
+    /// SAFETY:
+    /// - By invariant on `KnownLayout::MaybeUninit`, `T` and `T::MaybeUninit`
+    ///   have the same:
+    ///   - Fixed prefix size
+    ///   - Alignment
+    ///   - (For DSTs) trailing slice element size
+    /// - By invariant on `KnownLayout`, it is valid to perform an `as` cast
+    ///   from `*mut T` and `*mut T::MaybeUninit`, and this operation preserves
+    ///   referent size (ie, `size_of_val_raw`).
```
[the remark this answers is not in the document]
jswrenn: `unsafe_impl_known_layout` requires:
```
/// Implements `KnownLayout` for a type in terms of the implementation of
/// another type with the same representation.
///
/// # Safety
///
/// - `$ty` and `$repr` must have the same:
///   - Fixed prefix size
///   - Alignment
///   - (For DSTs) trailing slice element size
/// - It must be valid to perform an `as` cast from `*mut $repr` to `*mut $ty`,
///   and this operation must preserve referent size (ie, `size_of_val_raw`).
macro_rules! unsafe_impl_known_layout {
```
Does that final bullet point always follow from `$ty` and `$repr` having the same `KnownLayout::LAYOUT`, or is there more to prove?

src/util/mod.rs
```diff
@@ -683,6 +683,23 @@ pub(crate) unsafe fn copy_unchecked(src: &[u8], dst: &mut [u8]) {
...
 }
 
+/// Unsafely transmutes the given `src` into a type `Dst`.
+///
+/// # Safety
+///
+/// The value `src` must be a bit-valid instance of `Dst`.
+#[inline(always)]
+pub(crate) const unsafe fn transmute_unchecked<Src, Dst>(src: Src) -> Dst {
+    static_assert!(Src, Dst => core::mem::size_of::<Src>() == core::mem::size_of::<Dst>());
```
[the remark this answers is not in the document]
jswrenn: The safety precondition is sufficient, as it will allow us to eventually support padding-extension transmutations without updating the SAFETY comment of every other call-site. (We'll still need to update the SAFETY doc comment of `transmute_from`, of course.)

There's no point in making this a macro that desugars to `mem::transmute`. The whole point of this is to have something usable in generic contexts, since `mem::transmute` isn't.

build.rs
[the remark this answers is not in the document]
jswrenn: I kept this PR targeting 0.8, since — as a rule — I think it's easier to develop for 0.8 and remove MSRV cruft in the cherrypick to 0.9 — than it is to go the other way around. The `zerocopy_unstable` CFGs are nonetheless removed, since I've instead `doc(hidden)`'d the `MaybeUninit` machinery.

src/lib.rs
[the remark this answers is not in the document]
joshlf: Yeah I think that follows, but you do need to update the comment to explicitly state that the `::LAYOUT` equality is the premise you're using for this proof.

PR #1828, opened 2024-10-06
[wip] UnalignUnsized

jswrenn:
`todo!()`s remain. I suspect some of this would be easier, too, if we had `Ptr` integration with `UnalignUnsized`.


src/lib.rs
```diff
@@ -893,6 +930,57 @@ unsafe impl<T> KnownLayout for [T] {
         //   struct `Foo(i32, [u8])` or `(u64, Foo)`.
         slc.len()
     }
+
+    #[cfg(feature = "alloc")]
+    unsafe fn drop(slf: &mut UnalignUnsized<Self>) {
```
joshlf, line 935: Does this play nicely with `Pin`'d types?

jswrenn: Yes, Pin bars you from moving the referent before running its destructor. It doesn't place any restrictions on what you do in the destructor.

Issue #1655, opened 2024-09-15
Ensure all safety comments quote and cite a specific version of the documentation


joshlf:
All safety comments should be of the following form:

```rust
// SAFETY: The alignment cannot be greater than the size [1].
//
// [1] Per https://doc.rust-lang.org/1.81.0/reference/type-layout.html#size-and-alignment:
//
//   The size of a value is always a multiple of its alignment.
```

Note in particular:
- Cites a specific version of the documentation (in this case 1.81.0)
- Cites a stable version of the documentation (rather than nightly or beta)
- Quotes the relevant part of the citation

This issue tracks getting all existing safety comments up to this standard.

varun-doshi:
I'd like to take this up

joshlf:
Awesome, it'd be great to have the help! Let me know if you have any questions 😃

varun-doshi:
For example, this is a safety comment in the current code
```
/// Coverts `self` into a `&[u8]`.
    ///
    /// # Safety
    ///
    /// The returned reference has the same address and length as `self.deref()`
    /// and `self.deref_mut()`.
    ///
    /// Note that, combined with the safety invariant on [`ByteSlice`], this
    /// safety invariant implies that the returned reference is "stable" in the
    /// sense described in the `ByteSlice` docs.
    ```


What should the resultant structure be?

joshlf:
Great question; I should clarify a few things:

This issue is only concerned with `// SAFETY:` comments (ie, justifying why it's okay to call an `unsafe` function or implement an `unsafe` trait). That's distinct from the `# Safety` section of a doc comment, which describes what a *user* must guarantee in order to use an API that we define.

This is also only relevant to `// SAFETY:` comments that cite external documentation (some don't, and those don't need to be fixed). Here's an example:

https://github.com/google/zerocopy/blob/b35123b0f62e80d0d3ca7a11e38378abe488a558/src/impls.rs#L13-L22

This is very close since it already quotes the relevant documentation. We'd only need to change it to cite a specific *version* of the docs (note the added `1.81.0` in the URL):

```rust
 /// SAFETY: 
 /// Per the reference [1], "the unit tuple (`()`) ... is guaranteed as a 
 /// zero-sized type to have a size of 0 and an alignment of 1." 
 /// - `Immutable`: `()` self-evidently does not contain any `UnsafeCell`s. 
 /// - `TryFromBytes` (with no validator), `FromZeros`, `FromBytes`: There is 
[...]

PR #1835, opened 2024-10-07
feat: safety comments updated

varun-doshi:
Related issue https://github.com/google/zerocopy/issues/1655


src/impls.rs
```diff
@@ -264,8 +264,10 @@ safety_comment! {
     ///   clear how we'd prove it short of adding text to the stdlib docs that
     ///   says so explicitly, which likely wouldn't be accepted.
     ///
-    /// [1] https://doc.rust-lang.org/stable/std/num/struct.NonZeroU8.html
-    /// [2] https://doc.rust-lang.org/stable/std/num/struct.NonZeroI8.html
+    /// [1] https://doc.rust-lang.org/1.81.0/std/num/type.NonZeroU8.html
+    ///     `NonZeroU8 is guaranteed to have the same layout and bit validity as u8 with 
+    ///     the exception that 0 is not a valid instance`
+    /// [2] https://doc.rust-lang.org/stable/std/num/type.NonZeroI8.html
```
jswrenn, line 270:
```suggestion
    /// [2] https://doc.rust-lang.org/1.81.0/std/num/type.NonZeroI8.html
```

jswrenn (changes requested):
Thank you! There's one link you missed. Could you also run `./cargo.sh +stable fmt`?

src/impls.rs
```diff
@@ -264,8 +264,10 @@ safety_comment! {
     ///   clear how we'd prove it short of adding text to the stdlib docs that
     ///   says so explicitly, which likely wouldn't be accepted.
     ///
-    /// [1] https://doc.rust-lang.org/stable/std/num/struct.NonZeroU8.html
-    /// [2] https://doc.rust-lang.org/stable/std/num/struct.NonZeroI8.html
+    /// [1] https://doc.rust-lang.org/1.81.0/std/num/type.NonZeroU8.html
+    ///     `NonZeroU8 is guaranteed to have the same layout and bit validity as u8 with
+    ///     the exception that 0 is not a valid instance`
+    /// [2] https://doc.rust-lang.org/1.81.0/std/num/type.NonZeroI8.html
```
joshlf, line 270:
```suggestion
    /// [1] Per https://doc.rust-lang.org/1.81.0/std/num/type.NonZeroU8.html:
    ///
    ///     `NonZeroU8` is guaranteed to have the same layout and bit validity as `u8` with
    ///     the exception that 0 is not a valid instance
    ///
    /// [2] https://doc.rust-lang.org/1.81.0/std/num/type.NonZeroI8.html
```

The backticks are used around code items like `NonZeroU8` or `u8`, not around the entire quote. Also, add a newline before (and in the case where there's another citation afterward, a newline after as well).

joshlf (changes requested):
Thanks for this! One small nit.

joshlf (approved):

PR #1869, opened 2024-10-11
Cite stable docs for bit validity of numeric types

mrothNET:
The cited text is on the stable docs since a few months.

### DIFF
--- a/src/impls.rs
+++ b/src/impls.rs
@@ -38,14 +38,11 @@ safety_comment! {
     ///   - The only value >= 1 for which 1 is an integer multiple is 1
     ///   Therefore, the only possible alignment for `u8` and `i8` is 1.
     ///
-    /// [1] Per https://doc.rust-lang.org/beta/reference/types/numeric.html#bit-validity:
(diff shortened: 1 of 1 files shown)


src/impls.rs
```diff
@@ -38,15 +38,12 @@ safety_comment! {
     ///   - The only value >= 1 for which 1 is an integer multiple is 1
     ///   Therefore, the only possible alignment for `u8` and `i8` is 1.
     ///
-    /// [1] Per https://doc.rust-lang.org/beta/reference/types/numeric.html#bit-validity:
+    /// [1] Per https://doc.rust-lang.org/reference/types/numeric.html#bit-validity:
```
jswrenn, line 41: Per #1655, cite the exact version of the docs instead:
```suggestion
    /// [1] Per https://doc.rust-lang.org/1.81.0/reference/types/numeric.html#bit-validity:
```

jswrenn (changes requested):

jswrenn (approved):

Issue #1866, opened 2024-10-10
Overhaul `Ptr`'s validity invariant modeling


joshlf:
## Overview

Currently, the invariants enforced by `Ptr` are used strictly as lower bounds. We require code to prove soundness in order to increase the strictness of an invariant, for example to go from any alignment to `Aligned`:

https://github.com/google/zerocopy/blob/1b037c8145526351c96b09ca5bd34892c36844ec/src/pointer/ptr.rs#L779-L781

...but we permit the opposite direction without restriction:

https://github.com/google/zerocopy/blob/1b037c8145526351c96b09ca5bd34892c36844ec/src/pointer/ptr.rs#L936

For alignment, this is sound, but it is not sound for validity as recently illustrated by #2226.

This issue tracks updating `Ptr`'s validity modeling to make it fully sound.

## Proof of Concept

Currently, we model all invariants as lower bounds. Conceptually, an invariant describes what is known about a `Ptr` or its referent at a particular point in the program. Making stronger assertions (e.g., asserting that a `Ptr` with `Any` validity actually refers to a bit-valid value, and thus upgrading its validity to `Valid`) is not sound in the general case. By contrast, making weaker assertions is always sound because it simply entails "forgetting" what is known (e.g., going from `Valid` to `Any`).

This works for alignment, but does not work for aliasing (see #1889) or validity. Consider, for example, adding two seemingly innocuous methods to `Ptr`:
- `forget_valid` modifies the validity invariant on an existing `Ptr` to be `Any`
- `as_maybe_uninit` turns a `Ptr<T>` into a `Ptr<MaybeUninit<T>, (..., ..., Valid)>` (since `MaybeUninit` has no validity requirements)

Taken together, these can be used to produce UB:

https://github.com/google/zerocopy/blob/a3d4387d40e1f0e07f99fcecfa8c17580e066e4a/src/pointer/ptr.rs#L946-L974

Indeed, Miri detects this UB:

```text
test pointer::ptr::test_forget_valid ... error: Undefined Behavior: constructing invalid value: encountered 0x02, but expected a boolean
   --> src/pointer/ptr.rs:974:5
    |
974 |     assert_eq!(b, true);
    |     ^^^^^^^^^^^^^^^^^^^ constructing invalid value: encountered 0x02, but expected a boolean
    |
    = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
[...]

Issue #1183, opened 2024-05-03
Support generic pointer framework and generic pointer casts


joshlf:
*This design is prototyped in #1169.*

The idea is basically:

```rust
pub unsafe trait Pointer<'a, T: ?Sized> {
    type Pointer<U: 'a + ?Sized>: Pointer<'a, U, Aliasing = Self::Aliasing>;
    type Aliasing: invariant::Aliasing;
}
```

This trait (or one with a similar design) would permit us to abstract over the *type* of pointer used to reference a particular type. It could support shared and exclusive references, but also smart pointer types such as `Box`, `Rc`, `Arc`, and possibly even `Vec`.

This has a number of advantages:
- It allows us to remove much API duplication, where currently we have both shared and exclusive versions of many methods
- It allows generic methods to naturally support non-reference pointer types (`Box` et al)
- It allows us to possibly do the same via our `transmute_ref!` family of macros, or at least to introduce a new macro that permits "container transmutation" ([#114](https://github.com/google/zerocopy/issues/114))
- It may nearly or entirely replace the job currently done by `Ref`

## Aliasing

In order to do this, each pointer needs to be able to encode its `Aliasing` invariant so that it can interoperate with the `Ptr` type. For each smart pointer type, we would introduce a new aliasing invariant specific to that type (e.g., `Boxed`, `Arced`, etc) with the invariant being simply "this pointer supports the preconditions of `Box::from_raw`/`Arc::from_raw`/etc, with the exception of alignment and validity preconditions" (alignment and validity preconditions are handled separately). This implies that, roughly speaking, any pointer type with a `from_raw` method can be supported.

## `Vec`

It may be possible to support `Vec` directly. If not, we can at least use the fact that the standard library supports conversion between `Vec<T>` and `Box<[T]>` to build on top of our `Box` support.

PR #1889, opened 2024-10-12
[pointer] Clarify semantics of aliasing invariants

### LINKED ISSUE #1839: Audit variance of `Ptr`
What about `Ptr<T, (Exclusive, _, _)>`? It shouldn't be covariant over `T`, but invariant. What about the variance *of* `T` (e.g. `T = UnsafeCell<U>`)?

joshlf:
Previously, we supported the `AtLeast` bound, which was used to describe a subset relationship in which `I: AtLeast<J>` implied that `I` as at least as restrictive as `J`. However, as described in #1866, this incorrectly models invariants as monotonic. In reality, invariants both provide guarantees but also *require* guarantees.

This commit takes a step in the direction of resolving #1866 by removing `AtLeast`. Uses of `AtLeast<Shared>` are replaced by a new `Reference` trait, which is implemented for `Shared` and `Exclusive`. This serves two purposes: First, it makes it explicit what this bound means. Previously, `AtLeast<Shared>` had an ambiguous meaning, while `Reference` means precisely that an invariant is either `Shared` or `Exclusive` and nothing else. Second, it paves the way for #1183, in which we may add new aliasing invariants which convey ownership. In that case, it will be important for existing methods to add `Reference` bounds when those methods would not be sound in the face of ownership semantics.

We also inline the items in the `invariant` module, which were previously generated by macro. The addition of the `Reference` trait did not play nicely with that macro, and we will likely need to go further from the macro in order to fix #1839 – this fix will likely require making aliasing invariants meaningfully different than other invariants, for example by adding an associated type.

Makes progress on #1866

### DIFF
--- a/src/impls.rs
+++ b/src/impls.rs
@@ -657,9 +657,7 @@ unsafe impl<T: TryFromBytes + ?Sized> TryFromBytes for UnsafeCell<T> {
     }
 
     #[inline]
-    fn is_bit_valid<A: invariant::Aliasing + invariant::AtLeast<invariant::Shared>>(
-        candidate: Maybe<'_, Self, A>,
-    ) -> bool {
+    fn is_bit_valid<A: invariant::Reference>(candidate: Maybe<'_, Self, A>) -> bool {
         // The only way to implement this function is using an exclusive-aliased
         // pointer. `UnsafeCell`s cannot be read via shared-aliased pointers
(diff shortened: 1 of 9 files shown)


jswrenn (approved):

joshlf:
Backporting in #2378

PR #1894, opened 2024-10-12
[pointer] Match variance of references

joshlf:
When the aliasing mode is `Any`, `Ptr<'a, T>` is invariant in `'a` and `T`. When the aliasing mode is `Shared` or `Exclusive`, `Ptr` has the same variance as `&'a T` and `&'a mut T` respectively.

Makes progress on #1839


joshlf:
@jswrenn Any ideas on how to test this?

jswrenn (approved):

joshlf:
Backporting in #2382

PR #1895, opened 2024-10-12
[pointer] Make invariants opaque, more ergonomic

### LINKED ISSUE #1876: Make `Ptr` invariants opaque and more ergonomic
Once we publish 0.9, which will have a sufficiently high MSRV that we have access to GAT, we implement the following design. This will permit all concrete `Invariants` types to be private. This, in turn, means that we can preserve the ability to make future changes to how we represent invariants, or even add new invariants.

```rust
trait Invariants {
    // The invariants encoded by `Self`
    type Aliasing: Aliasing;
    type Alignment: Alignment;
    type Validity: Validity;

    // `Self` except with one invariant modified
    type WithAliasing<A: Aliasing>: Invariants<
        Aliasing = A,
        Alignment = Self::Alignment,
        Validity = Self::Validity,
    >;
    type WithAlignment<A: Alignment>: Invariants<
        Aliasing = Self::Aliasing,
        Alignment = A,
        Validity = Self::Validity,
    >;
    type WithValidity<A: Validity>: Invariants<
        Aliasing = Self::Aliasing,
        Alignment = Self::Alignment,
        Validity = V,
    >;
}

struct Ptr<'a, T, I: Invariants> { ... }

impl<'a, T, I: Invariants> Ptr<'a, T, I> {
    // Look! No concrete types!
    pub unsafe fn assume_aligned(self) -> Ptr<'a, T,  I::WithAlignment<Aligned>> { ... }
}
```

joshlf:
Closes #1876

### DIFF
--- a/src/pointer/ptr.rs
+++ b/src/pointer/ptr.rs
@@ -401,9 +401,7 @@ mod _conversions {
     {
         /// Converts a `Ptr` an unaligned `T` into a `Ptr` to an aligned
         /// `Unalign<T>`.
-        pub(crate) fn into_unalign(
-            self,
-        ) -> Ptr<'a, crate::Unalign<T>, (I::Aliasing, Aligned, I::Validity)> {
+        pub(crate) fn into_unalign(self) -> Ptr<'a, crate::Unalign<T>, I::WithAlignment<Aligned>> {
             // SAFETY:
             // - This cast preserves provenance.
             // - This cast preserves address. `Unalign<T>` promises to have the
@@ -421,7 +419,7 @@ mod _conversions {
             // SAFETY: `Unalign<T>` promises to have alignment 1, and so it is
             // trivially aligned.
             let ptr = unsafe { ptr.assume_alignment::<Aligned>() };
-            ptr
+            ptr.unify_invariants()
         }
     }
 }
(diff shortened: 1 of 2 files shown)


joshlf:
@jswrenn I didn't update any methods which modify two of three invariants. Any galaxy brain ideas on how to make those ergonomic?

The best I could come up with was yet another set of associated types e.g.:

```rust
type RetainAliasing<A: Alignment, V: Validity> = ...;
```

...but that's not forwards-compatible with adding new invariant types.

src/pointer/ptr.rs
```diff
@@ -140,12 +140,40 @@ pub mod invariant {
...
+            Validity = Self::Validity,
+        >;
+
+        /// Invariants identical to `Self` except with a different validity
+        /// invariant.
+        type WithValidity<V: Validity>: Invariants<
+            Aliasing = Self::Aliasing,
+            Alignment = Self::Alignment,
+            Validity = V,
+        >;
```
jswrenn, line 166: Should we hold off on this insofar that it might make backporting subsequent improvements to `Ptr` more difficult? It seems like there's a lot of cleanup to be done that doesn't involve GATs, and we'd benefit from having `Ptr` be as similar as possible between 0.8 and 0.9.

jswrenn (approved):

src/pointer/ptr.rs
joshlf: Discussed offline - decided to stick with this.

Issue #1122, opened 2024-04-18
`Ptr` invariant conversions could be smarter


joshlf:
*See also: #1359*

Consider `Ptr::transparent_wrapper_into_inner`:

https://github.com/google/zerocopy/blob/be56a3a79e6f010c5a9ef45128d6f7b27fe3c17b/src/pointer/ptr.rs#L609-L627

Currently, if the validity invariant is `Initialized` and the validity variance is `Invariant`, the returned pointer will have the validity invariant `Any`. However, this throws away information: regardless of the specific validity requirements of a type, if all of the outer type's bytes are initialized, then the same is true of the inner type's bytes. In other words, the `Initialized` validity invariant should be preserved *regardless* of the validity variance.

In the general case, what we really need is the ability to express a generic mapping `(Invariant, Variance) -> Invariant`, which would allow us to express this behavior.

In other conversion functions, each invariant can be mapped as either:
- Preserve the invariant
- Map to a fixed invariant

These conversion functions could similarly benefit from a generic mapping `Invariant -> Invariant`.

joshlf:
Closed by https://github.com/google/zerocopy/pull/1896

PR #1896, opened 2024-10-13
[pointer] Support generic invariant mapping

joshlf:
This commit adds a framework which supports encoding in the type system any `I -> I` mapping where `I` is any `Invariant` type. This permits us to make `cast_unsized`'s return value smarter, and as a result, allows us to remove a lot of `unsafe` code.

Makes progress on #1122


joshlf:
@jswrenn I'd love for you to take a crack at improving the ergonomics here.

jswrenn (approved):

jswrenn (approved):

PR #1898, opened 2024-10-13
[pointer][transmute] Migrate from util

joshlf:
This prepares us to rename `TransparentWrapper` to `TransmuteFrom`, which we'll do in a follow-up commit to preserve nice git diffs.

Makes progress on #1122


joshlf:
Obsoleted by #1911

PR #1906, opened 2024-10-14
[pointer][invariant] Move to separate file

joshlf:
This prepares us for future changes which will significantly increase the amount of code in the `invariant` module.

Also merge `aliasing_safety` into this new file.


jswrenn (approved):

joshlf:
Backporting in #2381

PR #1908, opened 2024-10-14
[pointer] Simplify AliasingSafe, rename to Read

joshlf:
`AliasingSafe` is really about whether a pointer permits unsynchronized reads - either because the referent contains no `UnsafeCell`s or because the aliasing mode is `Exclusive`. Previously, `AliasingSafe` was not named consistent with this meaning, and was a function of a *pair* of types rather than of a single type. This commit fixes both oversights.

While we're here, we also add `Read` bounds in some places, allowing us to simplify many safety comments.

### DIFF
--- a/src/impls.rs
+++ b/src/impls.rs
@@ -950,6 +950,7 @@ mod simd {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::pointer::invariant;
 
     #[test]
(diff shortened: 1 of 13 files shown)


jswrenn (approved):

joshlf:
Backporting in #2383

PR #1909, opened 2024-10-14
[pointer] Rename `Any` -> `Unknown`/`Inaccessible`

joshlf:
For aliasing, use `Inaccessible`. For alignment and validity, use `Unknown`.


jswrenn (approved):

joshlf:
Backporting in #2384

Issue #1817, opened 2024-10-04
Support slices in `transmute_ref!` et al


joshlf:
*This design is prototyped in ad777bc1fd9c53a65132fa48c4cb9307296b476b.*

Currently, `transmute_ref!` validates:
- `Src` and `Dst` have the same size
- `Src` has alignment at least as large as `Dst`

For slices, when transmuting `[Src]` into `[Dst]`, the conditions are identical:
- The fat pointer cast from `[Src]` to `[Dst]` preserves length if `Src` and `Dst` have the same size
- `[T]` has the same alignment as `T`

If we could figure out a way to get `transmute_ref!` to infer `Src` and `Dst` when those are the transmuted types and `Src` and `Dst` when the transmuted types are `[Src]` and `[Dst]`, then we could get it to support both sized and slice transmutations.

We could do this by introducing a trait like the following:

```rust
trait TransmuteRefHelper {
    type Elem;
}

impl<T: Sized> TransmuteRefHelper for T {
    type Elem = T;
}

impl<T: Sized> TransmuteRefHelper for [T] {
    type Elem = T;
}
```

We'd need to figure out some way of coaxing the correct inference. Currently we do the following:

https://github.com/google/zerocopy/blob/d204727bd370775c4ba4ce15d80ba673f5516299/src/macros.rs#L214-L217

Instead we need to make this more generic. Naively, we could do:

```rust
fn from_elem<T: TransmuteRefHelper>(e: T::Elem) -> &'static T
```

Then we could modify the code in `transmute_ref!` something like:

```rust
// Infers `t` as either equal to the sized source value or
// the element type of the source slice.
let mut t = loop {}; 
e = from_elem(t);
```

This works, but it means we can't support `const fn` on our MSRV, which doesn't permit trait bounds in `const fn`. If we can figure out another way to accomplish the same inference, then we could avoid breaking const support.

PR #1924, opened 2024-10-16
Teach transmute_{ref,mut}! to handle slice DSTs

joshlf:
TODO:
- Figure out how to make this backwards-compatible (since current macros work with `T: Sized`, but `T: Sized` does not imply `T: KnownLayout`)
- Make it so that `Sized` bounds are still enforced on older toolchains that don't support slice DST transmutes
- Write safety comments
- Get rid of now-unused macros and functions in util::macro_util

Makes progress on #1817

gherrit-pr-id: Ib4bc62202e0b3b09d155333b525087f7aa8f02c2

### DIFF
--- a/src/impls.rs
+++ b/src/impls.rs
@@ -680,6 +680,7 @@ mod atomics {
         // TODO(#170): Implement `FromBytes` and `IntoBytes` once we implement
         // those traits for `*mut T`.
         impl_for_transmute_from!(T => TryFromBytes for AtomicPtr<T> [UnsafeCell<*mut T>]);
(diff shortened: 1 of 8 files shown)


joshlf:
I wonder if we could use the autoref trick to get T: Sized to work. Specifically, we could provide one impl for T: Sized as the default, and then a fallback to T: KnownLayout

jswrenn:
> I wonder if we could use the autoref trick to get T: Sized to work. Specifically, we could provide one impl for T: Sized as the default, and then a fallback to T: KnownLayout

Yes, I believe we can: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=78af21ee715df8622ed1b28433a1d314

joshlf:
Closing in favor of #2428

PR #1929, opened 2024-10-17
[unsafe-fields] Initial commit

joshlf:
Makes progress on #1931

---

This PR is on branch [unsafe-fields](../tree/unsafe-fields).

- #1929

### DIFF
--- a/unsafe-fields/src/lib.rs
+++ b/unsafe-fields/src/lib.rs
@@ -1,0 +1,440 @@
+// Copyright 2024 The Fuchsia Authors
+//
+// Licensed under a BSD-style license <LICENSE-BSD>, Apache License, Version 2.0
+// <LICENSE-APACHE or https://www.apache.org/licenses/LICENSE-2.0>, or the MIT
+// license <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your option.
+// This file may not be copied, modified, or distributed except according to
+// those terms.
+
+//! Support for unsafe fields.
+//!
+//! This crate provides the [`unsafe_fields!`] macro, which can be used to mark
+//! fields as unsafe. Unsafe fields automatically have their types wrapped using
+//! the [`Unsafe`] wrapper type. An `Unsafe` is intended to be used to for
+//! struct, enum, or union fields which carry safety invariants. All accessors
+//! are `unsafe`, which requires any use of an `Unsafe` field to be inside an
+//! `unsafe` block.
+//!
+//! An unsafe field has the type `Unsafe<O, F, const NAME_HASH: u128>`. `O` is
+//! the enclosing type (struct, enum, or union), `F` is the type of the field,
+//! and `NAME_HASH` is the hash of the field's name. `O` prevents swapping
+//! unsafe fields of the same `F` type between different enclosing types, and
+//! `NAME_HASH` prevents swapping different fields of the same `F` type within
+//! the same enclosing type. Note that swapping the same field between instances
+//! of the same type [cannot be prevented](crate#limitations).
+//!
+//! # Examples
+//!
+//! ```
+//! use unsafe_fields::{unsafe_fields, Unsafe};
+//!
+//! unsafe_fields! {
+//!     /// A `usize` which is guaranteed to be even.
+//!     pub struct EvenUsize {
+//!         // INVARIANT: `n` is even.
+//!         #[unsafe]
+//!         n: usize,
+//!     }
+//! }
+//!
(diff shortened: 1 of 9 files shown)


unsafe-fields/src/lib.rs
```diff
@@ -0,0 +1,356 @@
...
+        //
+        // [1] Per https://doc.rust-lang.org/1.81.0/core/mem/struct.ManuallyDrop.html:
+        //
+        //   `ManuallyDrop<T>` is guaranteed to have the same layout and bit
+        //   validity as `T`
+        let dst = unsafe { Transmute { src: slf }.dst };
+
+        // SAFETY (satisfaction of `Unsafe`'s field invariant): This method is
+        // unsafe to call.
+        ManuallyDrop::into_inner(dst)
```
jswrenn, line 206: Could you just add a comment explaining why we need to go through all of this ceremony?

unsafe-fields/src/lib.rs
```diff
@@ -0,0 +1,356 @@
...
+    pub const unsafe fn into(self) -> F {
+        use core::mem::ManuallyDrop;
+
+        let slf = ManuallyDrop::new(self);
+
+        #[repr(C)]
+        union Transmute<O: ?Sized, F, const NAME_HASH: u128> {
+            src: ManuallyDrop<Unsafe<O, F, { NAME_HASH }>>,
+            dst: ManuallyDrop<F>,
+        }
```
jswrenn, line 191: Unless there's good reason to do otherwise, I'd prefer to see this written as:
```suggestion
        #[repr(C)]
        union Transmute<Src, Dst> {
            src: ManuallyDrop<Src>,
            dst: ManuallyDrop<Dst>,
        }
```

unsafe-fields/src/lib.rs
```diff
@@ -0,0 +1,356 @@
...
+    ///
+    /// # Safety
+    ///
+    /// The caller is responsible for upholding any safety invariants associated
+    /// with this field.
+    #[inline(always)]
+    pub const unsafe fn as_ref(&self) -> &F {
+        // SAFETY: This method is unsafe to call.
+        &self.field
+    }
```
jswrenn, line 147: With `Immutable`, we could provide `as_ref` safely. Perhaps — in the interest of leaving that name open for the safe accessor — this should be called something like `as_ref_unchecked`?

unsafe-fields/src/lib.rs
```diff
@@ -0,0 +1,356 @@
...
+// NOTE on design: It may seem counter-intuitive to offer an impl of traits that
+// don't require `unsafe` to call. Unfortunately, this is a fundamental
+// requirement if users want to be able to mark their types as `Copy`. Luckily,
+// we can implement `Copy` (and its unavoidable super-trait, `Clone`) without
+// invoking user code or opening up the possibility of modifying the field. We
+// do this by only implementing `Copy` and `Clone` when `F: Copy`. For `Clone`,
+// the user is still able to provide a manual impl, so this does not
+// fundamentally restrict what behavior can be supported.
+impl<O: ?Sized, F: Copy, const NAME_HASH: u128> Copy for Unsafe<O, F, { NAME_HASH }> {}
+impl<O: ?Sized, F: Copy, const NAME_HASH: u128> Clone for Unsafe<O, F, { NAME_HASH }> {
```
jswrenn, line 126: Food for thought: With `Immutable`, I think we could provide basically all of the standard trait impls. No need to take action on this in this PR.

jswrenn (changes requested):
A few minor nits.

unsafe-fields/src/lib.rs
joshlf: IMO we should keep this totally separate from zerocopy. We can also easily release breaking changes since this will never be in anyone's API, so I don't think we need to be too concerned about predicting future API changes.

unsafe-fields/src/lib.rs
joshlf: Ack.

unsafe-fields/src/lib.rs
joshlf: Done.

unsafe-fields/src/lib.rs
joshlf: Done.

unsafe-fields/src/lib.rs
jswrenn: Couldn't zerocopy be an option dependency of this?

jswrenn (approved):

PR #1930, opened 2024-10-17
Implement `Ref::{try_as_ref,try_into_ref,try_into_mut}`

### LINKED ISSUE #1865: Support creating a `Ref` of a type that implements `TryFromBytes`
Support fallible constructors on `Ref` for types that implement `TryFromBytes`

For example, would like to be able to do something like the following:

```rust
#[derive(TryFromBytes)]
#[repr(u8)]
enum MyEnum {
    V0 = 0,
    V1 = 1,
}

let (enum_ref, remainder) = Ref::<B, MyEnum>::try_from_prefix(bytes)?;
```

jswrenn:
Only `Ref::try_as_mut` remains missing, probably pending polonius landing in rustc.

Partially fixes #1865
Supersedes #1184

A restricted form of `Ref::try_as_mut` can be implemented now as:
```rust
impl<B, T> Ref<B, T>
where
    B: ByteSlice,
    T: KnownLayout + ?Sized,
{
    #[must_use = "has no side effects"]
    #[inline(always)]
    pub fn try_as_mut(r: &mut Self) -> Result<&mut T, ValidityError<&mut Self, T>>
    where
        T: TryFromBytes + Immutable,
        B: ByteSliceMut + CloneableByteSlice,
    {
        // Presumably unreachable, since we've guarded each constructor of `Ref`.
        static_assert_dst_is_not_zst!(T);

        // Due to a false-positive in Rust's current borrow checker, the same
        // `r` can't be used for both validation *and* be returned upon
        // validation failure. We use `r_tmp` for validation, and `r` as the
        // value returned upon failure.
        let mut r_tmp = r.clone();

        // SAFETY: We don't call any methods on `t` other than those provided by
        // `ByteSlice`.
        let b = unsafe { r_tmp.as_byte_slice_mut() };

        match Ptr::from_mut(b.deref_mut()).try_cast_into_no_leftover::<T, BecauseExclusive>(None) {
            Ok(candidate) => match candidate.try_into_valid() {
                Ok(valid) => {
                    // SAFETY: `valid` inherits the local lifetime of `r_tmp`,
                    // but its value is derived from `r_tmp`'s underlying bytes.
                    // By contract on `CloneableByteSlice`, `r_temp.deref()
                    // produces a byte slice whose address (and thus lifetime)
                    // is identicial to that produced by `r.deref()`. We can
                    // therefor soundly extend the lifetime of `valid` to that
                    // of `r`.
                    Ok(unsafe { valid.assume_lifetime() }.as_mut())
                },
                Err(e) => Err(e.with_src(r)),
            },
            Err(CastError::Validity(i)) => match i {},
            Err(CastError::Alignment(_) | CastError::Size(_)) => {
                // SAFETY: By invariant on `Ref::0`, the referenced byte slice
                // is aligned to `T`'s alignment and its size corresponds to a
                // valid size for `T`. Since properties are checked upon
                // constructing `Ref`, these failures are unreachable.
                unsafe { core::hint::unreachable_unchecked() }
            }
        }
    }
}
```
...but we don't yet implement `ClonableByteSlice` for anything except `&[u8]`, so the method is effectively untestable.

### DIFF
--- a/src/ref.rs
+++ b/src/ref.rs
@@ -595,10 +595,85 @@ where
     }
 }
 
+impl<B, T> Ref<B, T>
+where
+    B: ByteSlice,
+    T: KnownLayout + ?Sized,
+{
+    /// Attempts to dereference this `Ref<_, T>` into a `&T` without copying.
+    ///
+    /// If the bytes of `self` are a valid instance of `T`, this method returns
+    /// a reference to those bytes interpreted as `T`. If those bytes are not a
+    /// valid instance of `T`, this returns `Err`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use zerocopy::Ref;
+    /// # use zerocopy_derive::*;
+    ///
+    /// // The only valid value of this type is the byte `0xC0`
+    /// #[derive(TryFromBytes, KnownLayout, Immutable)]
+    /// #[repr(u8)]
+    /// enum C0 { xC0 = 0xC0 }
+    ///
+    /// // The only valid value of this type is the bytes `0xC0C0`.
+    /// #[derive(TryFromBytes, KnownLayout, Immutable)]
+    /// #[repr(C, packed)]
+    /// struct C0C0(C0, C0);
+    ///
+    /// #[derive(TryFromBytes, KnownLayout, Immutable)]
+    /// #[repr(C)]
+    /// struct Packet {
+    ///     magic_number: C0C0,
+    ///     mug_size: u8,
+    ///     temperature: u8,
+    ///     marshmallows: [[u8; 2]],
+    /// }
+    ///
+    /// let bytes = &[0xC0, 0xC0, 240, 77, 0, 1, 2, 3, 4, 5][..];
+    ///
+    /// let r = Ref::<_, Packet>::new(bytes).unwrap();
+    /// let packet = Ref::try_as_ref(&r).unwrap();
+    ///
+    /// assert_eq!(packet.mug_size, 240);
+    /// assert_eq!(packet.temperature, 77);
+    /// assert_eq!(packet.marshmallows, [[0, 1], [2, 3], [4, 5]]);
+    /// ```
+    #[must_use = "has no side effects"]
+    #[inline(always)]
+    pub fn try_as_ref(r: &Self) -> Result<&T, ValidityError<&Self, T>>
+    where
+        T: TryFromBytes + Immutable,
+    {
+        // Presumably unreachable, since we've guarded each constructor of `Ref`.
+        static_assert_dst_is_not_zst!(T);
+
+        // SAFETY: We don't call any methods on `r` other than those provided by
+        // `ByteSlice`.
+        let b = unsafe { r.as_byte_slice() };
+
+        match Ptr::from_ref(b.deref()).try_cast_into_no_leftover::<T, BecauseImmutable>(None) {
+            Ok(candidate) => match candidate.try_into_valid() {
+                Ok(valid) => Ok(valid.as_ref()),
+                Err(e) => Err(e.map_src(|_| r)),
+            },
+            Err(CastError::Validity(i)) => match i {},
+            Err(CastError::Alignment(_) | CastError::Size(_)) => {
+                // SAFETY: By invariant on `Ref::0`, the referenced byte slice
+                // is aligned to `T`'s alignment and its size corresponds to a
+                // valid size for `T`. Since properties are checked upon
+                // constructing `Ref`, these failures are unreachable.
+                unsafe { core::hint::unreachable_unchecked() }
+            }
+        }
+    }
+}
+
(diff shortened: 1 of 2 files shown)


src/ref.rs
```diff
@@ -627,12 +705,91 @@ where
...
+
+        // Extend the lifetime of `bytes` to `'a`. This gives us a reference
+        // `bytes` with the same lifetime as if we had called
+        // `r.into_byte_slice()`, but without consuming `r`. This is valuable,
+        // since we will need to return `r` if validation fails.
+        //
+        // SAFETY: This is sound because `bytes` lives for `'a`. `Self` is
+        // `IntoByteSlice`, whose `.into_byte_slice()` method is guaranteed to
+        // produce a `&'a [u8]` with the same address and length as the slice
+        // obtained by `.deref()` (which is how `bytes` is obtained).
```
joshlf, line 769: This wigs me out. Maybe we can just use `.into_byte_slice()` and then add an unsafe `Ref` constructor that reconstitutes the `Ref` from the byte slice? That way we can use it to reconstruct the `Ref` in the error case and we don't have to do lifetime shenanigans.

src/ref.rs
```diff
@@ -643,7 +800,10 @@ where
     /// there is no conflict with a method on the inner type.
     #[must_use = "has no side effects"]
     #[inline(always)]
-    pub fn into_mut(r: Self) -> &'a mut T {
+    pub fn into_mut(r: Self) -> &'a mut T
+    where
+        T: FromBytes,
+    {
```
joshlf, line 806: Why not just create a new impl block? Mixing bounds on the impl block and bounds on the function feels odd and error-prone (esp with future changes).

src/ref.rs
```diff
@@ -661,6 +821,86 @@ where
...
+        // Extend the lifetime of `bytes` to `'a`. This gives us a reference
+        // `bytes` with the same lifetime as if we had called
+        // `r.into_byte_slice_mut()`, but without consuming `r`. This is
+        // valuable, since we will need to return `r` if validation fails.
+        //
+        // SAFETY: This is sound because `bytes` lives for `'a`. `Self` is
+        // `IntoByteSliceMut`, whose `.into_byte_slice_mut()` method is
+        // guaranteed to produce a `&'a [u8]` with the same address and length
+        // as the slice obtained by `.deref()` (which is how `bytes` is
+        // obtained).
```
joshlf, line 886: Same here as above

joshlf (changes requested):

src/ref.rs
jswrenn: I don't think that's possible without changing the return type — `B` is convertible into a `&[u8]`, but it's not necessarily itself a `&[u8]`. 

Why does this wig you out? Lifetimes are fundamentally about the live extent of a referent, and our safety conditions on ByteSlice and IntoByteSlice are all about having a stable referent.

joshlf: Well `r` only lives as long as the body, so `bytes`'s lifetime also doesn't outlive the body. I grant that the underlying allocation we're pointing to lives for `'a`, but are we guaranteed that Rust won't make assumptions on the basis that `r` becomes unreachable after this scope ends? I'm not sure.

src/ref.rs
jswrenn: When I read bound-heavy documentation (e.g., nalgebra), I've always been grateful when methods with related bounds are grouped together — so much so, I [wrote a blog post about it](https://jack.wrenn.fyi/blog/rustdocing-nalgebra/).

joshlf: Ack

src/ref.rs
compiler-errors: Rust will never make optimizations based off of lifetimes because we specifically only computer lower-bound lifetimes in MIR and also erase them by the time we get to codegen.

joshlf: PoC of unsoundness: add the following to the bottom of this file and run `./cargo.sh +nightly miri test unsound --features alloc`:

```rust
#[cfg(test)]
#[cfg(feature = "alloc")]
mod unsound_demo {
    use alloc::boxed::Box;

    use super::*;

    #[test]
    fn test_try_into_ref_unsound() {
        unsafe impl ByteSlice for Box<[u8]> {}

        unsafe impl<'a> IntoByteSlice<'a> for Box<[u8]> {
            fn into_byte_slice(self) -> &'a [u8] {
                Box::leak(self)
            }
        }

        let b: Box<[u8]> = Box::new([0u8; 1]);
        let r = Ref::<_, bool>::from_bytes(b).unwrap();
        let b = Ref::try_into_ref(r).unwrap();
        assert_eq!(b, &false);
    }
}
```

The problem is that we're using the normal `Deref` here instead of going through `IntoByteSlice::into_byte_slice`, and so we're dropping the `B: ByteSlice` before returning. For `B` types which have ownership semantics, this can cause unsoundness like it does here.

PR #1933, opened 2024-10-18
[project] Initial commit

joshlf:
---

This PR is on branch [projection](../tree/projection).

- #1933

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -333,6 +333,8 @@ pub mod layout;
 mod macros;
 #[doc(hidden)]
 pub mod pointer;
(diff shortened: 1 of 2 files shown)


joshlf:
Unfortunately it appears that this design has a [soundness hole](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=978de749dca4f67ed24707ac049a5e4f) as described [here](https://github.com/google/zerocopy/pull/1933/files#diff-f36a5c348f86854b0177520d601d75ed1dc2ac736aebe2e6b928389485b75021R120-R124).

PR #1934, opened 2024-10-18
[project] Initial commit

joshlf:
---

This PR is on branch [projection-sized](../tree/projection-sized).

- #1934


joshlf:
This supersedes #1933, which has a [soundness hole](https://github.com/google/zerocopy/pull/1933#issuecomment-2421453843).

This only supports sized types for now, but maybe we'll figure out a way to lift that limitation eventually.

src/pointer/ptr.rs
```diff
@@ -1059,6 +1069,44 @@ mod _casts {
 mod _project {
     use super::*;
 
+    impl<'a, T, I> Ptr<'a, T, I>
+    where
+        T: 'a + ?Sized,
+        I: Invariants,
```
kupiakos, line 1075: Note that this can only ever be sound for invariants that structurally propagate to their fields. This doesn't apply for the `Pin` invariant.

src/project.rs
```diff
@@ -0,0 +1,186 @@
...
+where
+    P: Projectable,
+    P::Inner: Field<{ NAME_HASH }>,
+    A: Aliasing,
+    AA: Alignment,
+{
+    // NOTE: Doing the actual cast inside this function makes the `project!`
+    // macro simpler, but it also means that we can't handle both const code and
+    // unsized types at the same time. In order to be const, we can't call trait
+    // methods, but in order to support unsized types, we'd need to provide a
```
kupiakos, line 60: For slice DSTs, all you need is the offset of the last field, which is the same for all sizes. This could be stored as a `const`.

PR #1944, opened 2024-10-18
[doc] Render toolchain version detection in rustdoc

### LINKED ISSUE #894: Indicate in rustdoc that items are conditionally compiled based on Rust toolchain version
## Progress

- [x] Implement on v0.8.x - https://github.com/google/zerocopy/pull/1944
- [x] ~~Back-port [#1944](https://github.com/google/zerocopy/pull/1944) to main~~
  - Using the `needs-backport` label instead

joshlf:
Previously, we had avoided doing this because there wasn't any way to guarantee that we didn't make mistakes and render the wrong version for a particular named toolchain version.

In this commit, we take the pragmatic approach of renaming the named toolchain versions - in particular, we add a suffix which includes the toolchain version. This is fairly error-resistant for two reasons:
- At the definition site (in `Cargo.toml`), the name appears on the same line as the version, so mistakes should be obvious
- At the use site (in various `.rs` files), the name appears within a few lines of any `#[cfg_attr(doc_cfg, doc(cfg(...)))]` attribute, so mistakes should similarly be obvious

Release 0.8.7.

Closes #894

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,12 +54,12 @@ jobs:
           # These are the names of specific Rust versions detected in
           # `build.rs`. Each of these represents the minimum Rust version for
           # which a particular feature is supported.
-          "zerocopy-core-error",
-          "zerocopy-diagnostic-on-unimplemented",
-          "zerocopy-generic-bounds-in-const-fn",
-          "zerocopy-target-has-atomics",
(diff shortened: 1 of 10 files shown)


joshlf:
This PR generates rustdoc output like:

Screenshot 2024-10-18 at 1 36 05 PM

jswrenn (approved):

Issue #1931, opened 2024-10-17
Design for unsafe fields polyfill


joshlf:
Sometimes, struct (or enum or union) fields have safety invariants:

```rust
struct EvenUsize {
    // INVARIANT: `n` is even.
    n: usize,
}
```

However, Rust has no mechanism to ensure that reads from or writes to fields with invariants must happen inside an `unsafe` block as is required for `unsafe` functions or to implement `unsafe` traits. I propose that we can support this behavior outside the language, allowing a user to write:

```rust
struct EvenUsize {
    // INVARIANT: `n` is even.
    #[unsafe]
    n: usize,
}
```

The `#[unsafe]` attribute (implemented as a proc macro attribute) modifies the type of `n` to *something like* `Unsafe<usize>`, although as we'll see in a moment, it needs to be a tad more complex than that.

## Design take 1: `Unsafe<T>`

*Design take 1 is currently prototyped in #1929.*

Let's start with the basic design, though:

```rust
#[repr(transparent)]
pub struct Unsafe<T>(T);
```

An `Unsafe<T>` is a type whose constructors and accessors are all `unsafe` to call. We can't prevent code from moving an `Unsafe<T>` around, but we can prevent code from *doing* anything with it. Thus, for example, we might imagine the following constructor for `EvenUsize`, calling the unsafe `Unsafe::new` constructor:

```rust
impl EvenUsize {
    /// Constructs a new `EvenUsize`.
    ///
    /// Returns `None` if `n` is odd.
    pub fn new(n: usize) -> Option<EvenUsize> {
        if n % 2 != 0 {
            return None;
        }
        // SAFETY: We just confirmed that `n` is even.
        let n = unsafe { Unsafe::new(n) };
        Some(EvenUsize { n })
    }
}
```

While this gets us a step in the right direction, it has a soundness hole: There is nothing to stop `Unsafe`s from two different types being swapped:

```rust
struct OddUsize {
    // INVARIANT: `n` is odd.
    #[unsafe]
    n: usize,
}
```

Code operating on an `EvenUsize` and an `OddUsize` could swap the inner `Unsafe<usize>`s safely, which is obviously bad!

## Design take 2: `Unsafe<T, F>`

To prevent this from happening, we can instead design `Unsafe` to also take a parameter which is the type of the field's outer type:

```rust
#[repr(transparent)]
pub struct Unsafe<T, F>(PhantomData<T>, F);
```

The `#[unsafe]` attribute would then modify the code to something like:

[...]

PR #1960, opened 2024-10-21
[unsafe-fields] Add as_ref_checked

joshlf:
Release 0.2.0.

Makes progress on #1931

---

This PR is on branch [unsafe-fields-zerocopy](../tree/unsafe-fields-zerocopy).

- #1960

### DIFF
--- a/unsafe-fields/src/lib.rs
+++ b/unsafe-fields/src/lib.rs
@@ -13,7 +13,8 @@
 //! the [`Unsafe`] wrapper type. An `Unsafe` is intended to be used to for
 //! struct, enum, or union fields which carry safety invariants. All accessors
 //! are `unsafe`, which requires any use of an `Unsafe` field to be inside an
-//! `unsafe` block.
+//! `unsafe` block. One exception is [`Unsafe::as_ref`], which is available when
+//! the `zerocopy_0_8` feature is enabled. See its docs for more information.
 //!
 //! An unsafe field has the type `Unsafe<O, F, const NAME_HASH: u128>`. `O` is
 //! the enclosing type (struct, enum, or union), `F` is the type of the field,
@@ -22,6 +23,8 @@
 //! `NAME_HASH` prevents swapping different fields of the same `F` type within
 //! the same enclosing type. Note that swapping the same field between instances
 //! of the same type [cannot be prevented](crate#limitations).
+//!
+//! [immutable]: zerocopy_0_8::Immutable
 //!
 //! # Examples
 //!
@@ -170,6 +173,7 @@
     rustdoc::missing_crate_level_docs,
     rustdoc::private_intra_doc_links
 )]
+#![cfg_attr(doc_cfg, feature(doc_cfg))]
 
 use core::marker::PhantomData;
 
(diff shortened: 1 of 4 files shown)


unsafe-fields/src/lib.rs
```diff
@@ -220,6 +225,26 @@ impl<O: ?Sized, F: ?Sized, const NAME_HASH: u128> Unsafe<O, F, { NAME_HASH }> {
...
+    /// immutable.
+    ///
+    /// If [`F: Immutable`][immutable], then `F` does not permit interior
+    /// mutation, and so it is safe to return a reference to it.
+    ///
+    /// [immutable]: zerocopy_0_8::Immutable
+    #[inline(always)]
+    #[cfg(feature = "zerocopy_0_8")]
+    #[cfg_attr(doc_cfg, doc(cfg(feature = "zerocopy_0_8")))]
+    pub const fn as_ref_checked(&self) -> &F
```
jswrenn, line 238: My preference would be `as_ref_unchecked` for the unsafe method, and make the safe thing the shortest to type (`.as_ref`). Most types are `Immutable`, so I additionally expect the safe version will be the common case!

joshlf: Feels weird to have most of the API suffixed with `_unchecked`, but I agree that that's philosophically the right thing to do.

Done.

unsafe-fields/src/lib.rs
```diff
@@ -215,33 +219,53 @@ impl<O: ?Sized, F: ?Sized, const NAME_HASH: u128> Unsafe<O, F, { NAME_HASH }> {
...
+
     /// Gets a mutable reference to the inner value.
     ///
     /// # Safety
     ///
     /// The caller is responsible for upholding any safety invariants associated
     /// with this field.
     #[inline(always)]
-    pub unsafe fn as_mut(&mut self) -> &mut F {
+    pub unsafe fn as_mut_unchecked(&mut self) -> &mut F {
```
jswrenn, line 254: Same concern as `new`. I don't *think* there's a safe version we could possibly offer here that we'd want to distinguish the unsafe version from.

jswrenn (commented):
I don't think we necessarily need to name *everything* `_unchecked`. The `_unchecked` prefix in Rust is conventionally used in Rust when there's a safe option that the method needs to be distinguished from.

When there's no safe alternative, the standard library omits the `_unchecked` suffix.

unsafe-fields/src/lib.rs
joshlf: Done.

unsafe-fields/src/lib.rs
```diff
@@ -215,33 +219,53 @@ impl<O: ?Sized, F: ?Sized, const NAME_HASH: u128> Unsafe<O, F, { NAME_HASH }> {
...
-    /// Constructs a new `Unsafe<O, F, NAME_HASH>`.
+    /// Constructs a new `Unsafe`.
     ///
     /// # Safety
     ///
     /// The caller is responsible for upholding any safety invariants associated
     /// with this field.
     #[inline(always)]
-    pub const unsafe fn new(field: F) -> Unsafe<O, F, { NAME_HASH }> {
+    pub const unsafe fn new_unchecked(field: F) -> Unsafe<O, F, { NAME_HASH }> {
```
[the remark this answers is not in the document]
joshlf: Done.

unsafe-fields/src/lib.rs
```diff
@@ -253,7 +277,7 @@ impl<O: ?Sized, F, const NAME_HASH: u128> Unsafe<O, F, { NAME_HASH }> {
     /// The caller is responsible for upholding any safety invariants associated
     /// with this field.
     #[inline(always)]
-    pub const unsafe fn into(self) -> F {
+    pub const unsafe fn into_unchecked(self) -> F {
```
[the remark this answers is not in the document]
joshlf: Done.

jswrenn (approved):

PR #1978, opened 2024-10-24
[CI] skip installation step when cache hit

### LINKED ISSUE #1730: "Generate cache" CI step executes even when it has no effect
Consider [this execution](https://github.com/google/zerocopy/actions/runs/10976322510/job/30476954004?pr=1723). First, the cache [detects a cache hit](https://github.com/google/zerocopy/actions/runs/10976322510/job/30476954004?pr=1723#step:3:31). Next, it still [executes the "Download dependencies" step](https://github.com/google/zerocopy/actions/runs/10976322510/job/30476954004?pr=1723#step:4:1), which takes 1m55s. Finally, since there was a cache hit, it decides [not to save any of this work](https://github.com/google/zerocopy/actions/runs/10976322510/job/30476954004?pr=1723#step:7:3).

Instead, when there's a cache hit, we should just abort immediately so we don't block CI needlessly.

dacozai:
fix #1730

### DIFF
--- a/.github/actions/cache/action.yml
+++ b/.github/actions/cache/action.yml
@@ -8,12 +8,19 @@
 
 name: Cache
 description: 'Caches cargo dependencies'
+outputs:
+  cache-hit:
+    description: "Cache Hit"
+    value: ${{ steps.cache.outputs.cache-hit }}
 runs:
   using: composite
   steps:
     - uses: actions/cache@v4
+      id: cache
       with:
         path: |
           ~/.cargo/
         key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
+        restore-keys: |
+          ${{ runner.os }}-cargo-
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -659,7 +659,15 @@ jobs:
       - name: Populate cache
         uses: ./.github/actions/cache
 
+      - uses: actions/cache/restore@v4
+        id: restore-cache
+        with:
+          path: |
+            ~/.cargo/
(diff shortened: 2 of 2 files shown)


joshlf:
Thanks for this! Is this redundant with [.github/actions/cache](https://github.com/google/zerocopy/blob/main/.github/actions/cache/action.yml), which is also called from `generate_cache`? That was added in https://github.com/google/zerocopy/pull/835. I'll be honest that I'm not sure whether we need both the `generate_cache` job and the `.github/actions/cache` action - I'm a little rusty on the semantics of GitHub Actions in this case.

dacozai:
> Thanks for this! Is this redundant with [.github/actions/cache](https://github.com/google/zerocopy/blob/main/.github/actions/cache/action.yml), which is also called from `generate_cache`? That was added in #835. I'll be honest that I'm not sure whether we need both the `generate_cache` job and the `.github/actions/cache` action - I'm a little rusty on the semantics of GitHub Actions in this case.

No, you are RIGHT!!! I should use that composite.
There are two modifications here
1. Fetch the output from the steps. Accessing the output directly is not straightforward, but I understand the purpose of this extra layer of delegation—it ensures that the output remains consistent in the composite, regardless of the action packages used or the varying output names.
2. I added `restore-key` to the composite step. This allows for a broader cache key range. [See if this one is helpful?!](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#matching-a-cache-key)

joshlf:
Okay, that makes sense! And indeed, this seems to have considerably sped up CI:

Screenshot

Compare to this run for #1981:

Screenshot

joshlf (approved):

PR #1994, opened 2024-10-31
ci: enforce Pinned-Dependencies check in CI for PRs

### LINKED ISSUE #1579: Configure OpenSSF Scorecard's `Pinned-Dependencies` check block CI
OpenSSF Scorecard is [configured on this repository](https://github.com/google/zerocopy/blob/1552df2cd5f093c0f4d69de352fb642b1c42be3f/.github/workflows/scorecard.yml), but it only runs periodically and generates reports like this one (inserting screen shots since these alerts are not publicly viewable):

Screenshot 2024-08-08 at 8 33 31 AM

Screenshot 2024-08-08 at 8 34 01 AM

It would be better if we could block PRs if they fail this check.

## Mentoring instructions

- [ ] Figure out how to run the [Pinned-Dependency](https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies) check [in CI](https://github.com/google/zerocopy/blob/1552df2cd5f093c0f4d69de352fb642b1c42be3f/.github/workflows/ci.yml)
- [ ] Ensure all dependencies reported by this check are pinned

Aditya-PS-05:
closes #1579


joshlf (changes requested):
See [this CI failure](https://github.com/google/zerocopy/actions/runs/11607842937/job/32321986126?pr=1994); you'll also need to add a dependency on this job to [all-jobs-succeeded](https://github.com/google/zerocopy/blob/3eb2fb7681fa3a613223d7854f0ada4fda0b0305/.github/workflows/ci.yml#L737).

joshlf:
Thanks for doing this! Just a few small changes, but otherwise this looks good!

PR #2012, opened 2024-11-04
Relax `map_src` argument bound to `FnOnce`

### LINKED ISSUE #2009: Unnecessary `Fn` bound for `SizeError::map_src`.
I am implementing a small wrapper over `read_from_bytes` that allows working over types like `Bytes` instead of `&[u8]`. As such I want to change the `Src` bound to be `Bytes` instead of `&[u8]`.

However, the bound on `map_src` is unnecessarily a `Fn(Src) -> NewSrc` which implies it would be callable multiple times. This is inconsistent with other forms of `map` which use `FnOnce`. This is a problem for me as I cannot simply `map_src(|_| bytes_buf)` as that moves ownership, giving me the error

```rust
error[E0507]: cannot move out of `buf`, a captured variable in an `Fn` closure
   --> proxy/src/protocol2.rs:371:40
    |
363 |         let buf = if self.len() < len {
    |             --- captured outer variable
...
371 |         res.map_err(|e| e.map_src(|()| buf))
    |                                   ---- ^^^ move occurs because `buf` has type `bytes::BytesMut`, which does not implement the `Copy` trait
    |                                   |
    |                                   captured by this `Fn` closure
    |
help: consider cloning the value if the performance cost is acceptable
    |
371 |         res.map_err(|e| e.map_src(|()| buf.clone()))
    |                                           ++++++++
```

Easy to fix here with a clone, sure. But it's not ideal.

I have a feeling this can be fixed without a breaking change, but I'm not 100% certain.

jswrenn:
Fixes #2009

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@
 [package]
 edition = "2021"
 name = "zerocopy"
-version = "0.8.8"
+version = "0.8.9"
 authors = ["Joshua Liebow-Feeser <joshlf@google.com>"]
 description = "Zerocopy makes zero-cost memory manipulation effortless. We write \"unsafe\" so you don't have to."
 categories = ["embedded", "encoding", "no-std::no-alloc", "parsing", "rust-patterns"]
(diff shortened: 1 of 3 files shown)


joshlf (approved):

joshlf (changes requested):
Also release the next 0.8 version in this PR?

joshlf (approved):

dmgolembiowski:
This appears to be a semver breaking change. Could a new function name, like `map_with`, be created instead with the new relaxed bound?

jswrenn:
Why is it a breaking change? `FnOnce` is a supertrait of `Fn`, so all arguments that were previously accepted because they were `Fn` should still be accepted now that the bound has been relaxed to `FnOnce`.

dmgolembiowski:
Apologies @jswrenn, you're right

PR #2015, opened 2024-11-04
added backport-pr-yml file

### LINKED ISSUE #2014: Create GitHub Action to automatically back-port PRs between branches
For example, #1997 was originally submitted against `v0.8.x` and merged as https://github.com/google/zerocopy/commit/dc537e7ca058f4f7e31473b982ceab38514dab61. It was then manually backported to `main` in #2013.

Instead, we'd like to have a GitHub Action that we can manually trigger from the GitHub web UI that will create a new PR from the commit resulting from the old PR (in the above example, https://github.com/google/zerocopy/commit/dc537e7ca058f4f7e31473b982ceab38514dab61). This action will need to take:
- The commit hash to back-port
- The target branch for the new PR

It should be possible to adapt our existing [release-crate-version.yml](https://github.com/google/zerocopy/blob/255babe875f371a6e2a9e01c258ce5ebc54bd6bb/.github/workflows/release-crate-version.yml) workflow.

Aditya-PS-05:
closes #2014

### DIFF
--- a/.github/workflows/backport-pr.yml
+++ b/.github/workflows/backport-pr.yml
@@ -1,0 +1,82 @@
+name: Backport PR
+
+on:
+  workflow_dispatch:
+    inputs:
+      commit:
+        description: "Commit hash to backport"
+        required: true
+      target_branch:
+        description: "Target branch for the new PR"
+        required: true
+
+jobs:
+  backport:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set up Git user
+        run: |
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
+          git config --global user.name "github-actions[bot]"
+
+      - name: Backport Commit
+        id: backport
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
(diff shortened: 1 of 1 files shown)


joshlf:
Thanks for this! I think it should be possible to mostly just re-use the existing [release-crate-version.yml](https://github.com/google/zerocopy/blob/255babe875f371a6e2a9e01c258ce5ebc54bd6bb/.github/workflows/release-crate-version.yml) workflow rather than having to re-invent the wheel, so I took a stab at that in #2017. I'm going to test that and see if it does what we need it to - if not, I'll revisit this PR since we may need the complexity in this PR if it turns out that you can't just do it using the simpler approach.

Aditya-PS-05:
@joshlf , okay. From my side, I will once try to reuse the existing code and decrease the complexity.

Aditya-PS-05:
> Thanks for this! I think it should be possible to mostly just re-use the existing [release-crate-version.yml](https://github.com/google/zerocopy/blob/255babe875f371a6e2a9e01c258ce5ebc54bd6bb/.github/workflows/release-crate-version.yml) workflow rather than having to re-invent the wheel, so I took a stab at that in #2017. I'm going to test that and see if it does what we need it to - if not, I'll revisit this PR since we may need the complexity in this PR if it turns out that you can't just do it using the simpler approach.

Thanks, if you find any another issues, please tell me I would love to contribute.

joshlf:
Confirmed that the simpler approach in #2017 works as needed. Thanks again for putting this up, though!

Issue #158, opened 2023-02-16
Add `FromBytes` constructor which reads from an `io::Read`


joshlf:
Crosvm has a utility function called [zerocopy_from_reader](https://github.com/google/crosvm/blob/8a7550c/common/data_model/src/lib.rs#L12-L23):

```rust
pub fn zerocopy_from_reader<R: io::Read, T: FromBytes>(mut read: R) -> io::Result<T> {
    // Allocate on the stack via `MaybeUninit` to ensure proper alignment.
    let mut out = MaybeUninit::zeroed();

    // Safe because the pointer is valid and points to `size_of::<T>()` bytes of zeroes,
    // which is a properly initialized value for `u8`.
    let buf = unsafe { from_raw_parts_mut(out.as_mut_ptr() as *mut u8, size_of::<T>()) };
    read.read_exact(buf)?;

    // Safe because any bit pattern is considered a valid value for `T`.
    Ok(unsafe { out.assume_init() })
}
```

Maybe we should add something similar to `FromBytes` (and potentially a write analogue to `AsBytes`)?

joshlf:
@nwplayer123 has [also requested this feature](https://github.com/google/zerocopy/discussions/1288#discussioncomment-10642879).

jswrenn:
The obvious name, IMO, for this is `read_from`, but that name is currently squatted by a `#[doc(hidden)]` and `#[deprecated]`.

Since it's marked `#[doc(hidden)]`, it doesn't have any SemVer obligations, but in the interest of providing a smooth transition for users, perhaps we wait a release or two before revising it?

If this is the course we take, we should update its deprecation message to something like this:
```
#[deprecated(since = "0.8.0", note = "renamed to `FromBytes::read_from_bytes`. Migrate immediately. This item will soon be re-used for a different routine.")]
```

jswrenn:
Not sure what to call the `_from_prefix` variant, since `read_from_prefix` is already used, too (except for a visible, not deprecated function).

joshlf:
A few considerations:
- I don't mind `read_from_io_read` or similar. It's weird, but also consistent with our naming convention of `read_from_<place>`
- I don't think we need prefix/suffix variants since a reader vends single bytes at a time, and so it can always give us exactly `size_of::<Self>()` bytes
- We *will* want a `read_from_with_elems` for `Self: ?Sized + KnownLayout`
[...]

PR #2016, opened 2024-11-04
Add `FromBytes::read_from_io` and `IntoBytes::write_to_io`

jswrenn:
Makes progress on #158.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -359,6 +359,9 @@ use core::{
     slice,
 };
 
+#[cfg(feature = "std")]
+use std::io;
+
 use crate::pointer::{invariant, BecauseExclusive};
 
 #[cfg(any(feature = "alloc", test))]
@@ -3083,10 +3086,7 @@ pub unsafe trait FromZeros: TryFromBytes {
             // "exposed" provenance, and thus Rust may have to assume that this
             // may consume provenance from any pointer whose provenance has been
             // exposed.
-            #[allow(fuzzy_provenance_casts)]
-            unsafe {
-                NonNull::new_unchecked(dangling)
-            }
+            unsafe { NonNull::new_unchecked(dangling) }
         };
 
         let ptr = Self::raw_from_ptr_len(ptr, count);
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -4525,6 +4529,24 @@ pub unsafe trait FromBytes: FromZeros {
         }
     }
 
+    /// TODO
```
joshlf, line 4532: TODO: Doc comment

src/lib.rs
```diff
@@ -5188,6 +5210,16 @@ pub unsafe trait IntoBytes {
         Ok(())
     }
 
+    /// TODO
```
joshlf, line 5213: TODO: Doc comment

src/lib.rs
joshlf, line 1: TODO: Tests

joshlf (changes requested):

src/lib.rs
```diff
@@ -4525,6 +4528,48 @@ pub unsafe trait FromBytes: FromZeros {
         }
     }
 
+    /// Reads a copy of `self` from the IO byte source `src`.
```
joshlf, line 4531:
```suggestion
    /// Reads a copy of `self` from an `io::Read`.
```

src/lib.rs
```diff
@@ -4525,6 +4528,48 @@ pub unsafe trait FromBytes: FromZeros {
...
+    ///
+    /// This is useful for interfacing with operating system byte sinks (files,
+    /// sockets, etc.).
```
joshlf, line 4534: IMO this is more specific than we need - non-OS resources can implement `io::Read` too. I think we can just remove this sentence.

src/lib.rs
```diff
@@ -4525,6 +4528,48 @@ pub unsafe trait FromBytes: FromZeros {
...
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
```
joshlf, line 4538: If you want this to run, you could use a trick like we do here: https://github.com/google/zerocopy/blob/ace5bb0de76e150d2284984677892a5aaa34a36e/src/macros.rs#L711

src/lib.rs
```diff
@@ -5188,6 +5233,55 @@ pub unsafe trait IntoBytes {
...
     }
 
+    /// Writes a copy of `self` to the IO byte sink `dst`.
+    ///
+    /// This is a shorthand for `dst.write_all(self.as_bytes())`, and is useful
+    /// for interfacing with operating system byte sinks (files, sockets, etc.).
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
```
joshlf, line 5243: If you want this to run, you could use a trick like we do here: https://github.com/google/zerocopy/blob/ace5bb0de76e150d2284984677892a5aaa34a36e/src/macros.rs#L711

src/lib.rs
```diff
@@ -5188,6 +5233,55 @@ pub unsafe trait IntoBytes {
         Ok(())
     }
 
+    /// Writes a copy of `self` to the IO byte sink `dst`.
```
joshlf, line 5236:
```suggestion
    /// Writes a copy of `self` to an `io::Write`.
```

src/lib.rs
```diff
@@ -5188,6 +5233,55 @@ pub unsafe trait IntoBytes {
...
+    ///
+    /// This is a shorthand for `dst.write_all(self.as_bytes())`, and is useful
+    /// for interfacing with operating system byte sinks (files, sockets, etc.).
```
joshlf, line 5239: Same here - IMO we should remove the mention of OS byte sinks.

src/lib.rs
```diff
@@ -5188,6 +5233,55 @@ pub unsafe trait IntoBytes {
...
+    /// ```
+    ///
+    /// If the write fails, `write_to_io` returns `Err` and a partial write may
+    /// have occured; e.g.:
+    ///
+    /// ```
+    /// # use zerocopy::IntoBytes;
+    ///
+    /// let src = u128::MAX;
+    /// let mut dst = [0u8, 0];
```
joshlf, line 5268:
```suggestion
    /// let mut dst = [0u8; 2];
```

I misread this as `[0u8; 0]` and was confused by what was being tested. IMO this is less subject to being misread.

joshlf (changes requested):

src/lib.rs
jswrenn: This is here solely for the benefit of SEO/Cmd+F. I don't say that these are the *only* uses for this method, but while non-OS resources *can* implement `io::Read`, you're almost *always* better off using our other methods in those cases.

It's not a load bearing sentence, but I think it's a helpful one.

joshlf: Ack.

src/lib.rs
jswrenn: I don't think there's much sense in actually testing the IO. It'll make our tests more expensive to run, and will cause them to fail on platforms where file IO is stubbed out (e.g., wasm).

joshlf: Ack

joshlf (approved):

joshlf:
Backporting in https://github.com/google/zerocopy/pull/2046

PR #2024, opened 2024-11-05
added prefix transmute at transmute!

### LINKED ISSUE #1852: Support prefix transmutes in `transmute!` et al
### Syntax

We should be able to add special syntax so that existing uses of `transmute!` are unaffected, while users are able to opt-in to prefix transmute behavior:

```rust
// Current behavior, remains the default.
transmute!(a);
// Opt-in to prefix transmutation (exact syntax to be bikeshedded)
transmute!(@prefix a);
```

### Design

Currently, we require that `mem::transmute::<T, U>()` compiles, which ensures that `size_of::<T>() == size_of::<U>()`. If we instead introduce a `union MaxSizesOf<T, U>` type, then we can instead require that `mem::transmute::<MaxSizesOf<T, U>, T>()` compiles. This ensures that `size_of::<T>() >= size_of::<U>()`, which ensures that it's sound to perform a prefix transmute from `T` to `U`.

Aditya-PS-05:
closes #1852

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -51,18 +51,10 @@
 /// This macro can be invoked in `const` contexts.
 #[macro_export]
 macro_rules! transmute {
+    // Original behavior
     ($e:expr) => {{
-        // NOTE: This must be a macro (rather than a function with trait bounds)
-        // because there's no way, in a generic context, to enforce that two
-        // types have the same size. `core::mem::transmute` uses compiler magic
-        // to enforce this so long as the types are concrete.
-
         let e = $e;
         if false {
-            // This branch, though never taken, ensures that the type of `e` is
-            // `IntoBytes` and that the type of this macro invocation expression
-            // is `FromBytes`.
-
             struct AssertIsIntoBytes<T: $crate::IntoBytes>(T);
             let _ = AssertIsIntoBytes(e);
 
@@ -71,29 +63,56 @@ macro_rules! transmute {
             let u = AssertIsFromBytes(loop {});
             u.0
         } else {
+            #[allow(clippy::missing_transmute_annotations)]
             // SAFETY: `core::mem::transmute` ensures that the type of `e` and
             // the type of this macro invocation expression have the same size.
             // We know this transmute is safe thanks to the `IntoBytes` and
             // `FromBytes` bounds enforced by the `false` branch.
-            //
-            // We use this reexport of `core::mem::transmute` because we know it
-            // will always be available for crates which are using the 2015
-            // edition of Rust. By contrast, if we were to use
-            // `std::mem::transmute`, this macro would not work for such crates
-            // in `no_std` contexts, and if we were to use
-            // `core::mem::transmute`, this macro would not work in `std`
-            // contexts in which `core` was not manually imported. This is not a
-            // problem for 2018 edition crates.
             let u = unsafe {
-                // Clippy: We can't annotate the types; this macro is designed
-                // to infer the types from the calling context.
                 #[allow(clippy::missing_transmute_annotations)]
                 $crate::util::macro_util::core_reexport::mem::transmute(e)
             };
             $crate::util::macro_util::must_use(u)
         }
-    }}
+    }};
+
+    // New prefix transmute behavior
+    (@prefix $e:expr) => {{
+        let e = $e;
+        if false {
+            // Verify source type implements IntoBytes
+            struct AssertIsIntoBytes<T: $crate::IntoBytes>(T);
+            let _ = AssertIsIntoBytes(e);
+
+            // Verify destination type implements FromBytes
+            struct AssertIsFromBytes<U: $crate::FromBytes>(U);
+            #[allow(unused, unreachable_code)]
+            let u = AssertIsFromBytes(loop {});
+            u.0
+        } else {
+            // Create a union to verify size relationships at compile time
+
+            // SAFETY: `core::mem::transmute` is used here to ensure that
+            // the size of `e` is greater than or equal to the size of the
+            // destination type. We verify this through the union `MaxSizesOf`
+            // which enforces the size relationship at compile time. The
+            // transmute is safe because the original `e` is assumed to
+            // contain enough bytes to construct a value of the destination
+            // type.
+
+            let u = unsafe {
+                // First transmute to union to verify size relationships
+                let union_val = $crate::util::macro_util::core_reexport::mem::transmute::<_, $crate::util::macro_util::MaxSizesOf<_, _>>(e);
+
+                // Then transmute to final type, which is guaranteed to be prefix-compatible
+                $crate::util::macro_util::core_reexport::mem::transmute_copy(&union_val.source)
+            };
+            $crate::util::macro_util::must_use(u)
+        }
+    }};
 }
+
+// Helper trait to validate size relationships at compile time
 
 /// Safely transmutes a mutable or immutable reference of one type to an
 /// immutable reference of another type of the same size and compatible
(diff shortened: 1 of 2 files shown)


Aditya-PS-05:
@joshlf , please review it

src/macros.rs
```diff
@@ -462,6 +538,66 @@ macro_rules! try_transmute {
...
+            Ok(unsafe {
+                let union_val = $crate::util::macro_util::core_reexport::mem::transmute::<$src_type, $crate::util::macro_util::MaxSizesOf<$src_type, $dest_type>>(e);
+                $crate::util::macro_util::core_reexport::mem::transmute_copy(&union_val.source)
+            })
+        } else {
+            $crate::util::macro_util::try_transmute_prefix::<$src_type, $dest_type>(e)
+        }
+    }};
+}
+
```
jswrenn, line 600: I don't think we want to introduce another distinct macro here; rather, we want `try_transmute` to support the `@prefix` notation.

src/macros.rs
```diff
@@ -85,16 +121,56 @@ macro_rules! transmute {
...
+            #[allow(unused, unreachable_code)]
+            let u = AssertIsFromBytes(loop {});
+            u.0
+        } else {
+            // Create a union to verify size relationships at compile time
+            #[repr(C)]
+            union MaxSizesOf<T, U> {
+                source: T,
+                _dest: U,
+            }
```
jswrenn, line 150: You define `MaxSizesOf` this in `macro_util.rs`, so you don't need to redefine it here.

src/macros.rs
```diff
@@ -85,16 +121,56 @@ macro_rules! transmute {
...
+            // the size of `e` is greater than or equal to the size of the
+            // destination type. We verify this through the union `MaxSizesOf`
+            // which enforces the size relationship at compile time. The
+            // transmute is safe because the original `e` is assumed to
+            // contain enough bytes to construct a value of the destination
+            // type.
+
+            let u = unsafe {
+                // First transmute to union to verify size relationships
+                let union_val = $crate::util::macro_util::core_reexport::mem::transmute::<_, MaxSizesOf<_, _>>(e);
```
jswrenn, line 162: Can size verification go in a `else if false` branch so we don't pay the runtime cost of a double transmutation?

jswrenn (changes requested):

src/macros.rs
Aditya-PS-05: I added this just to test the changes and I thought it would be better to keep this.

src/macros.rs
Aditya-PS-05: Okay

src/macros.rs
Aditya-PS-05: Okay, I will resolve these changes.

src/macros.rs
```diff
@@ -49,20 +49,55 @@
...
+//                 // to infer the types from the calling context.
+//                 #[allow(clippy::missing_transmute_annotations)]
+//                 $crate::util::macro_util::core_reexport::mem::transmute(e)
+//             };
+//             $crate::util::macro_util::must_use(u)
+//         }
+//     }}
+// }
+///
 #[macro_export]
```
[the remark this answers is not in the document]
Aditya-PS-05: I am changing this also.

Aditya-PS-05:
@jswrenn , I request you to review this pr.

jswrenn:
This PR has huge swaths of commented-out code. As a rule, we don't leave commented-out code in the zerocopy codebase.

src/util/macro_util.rs
```diff
@@ -599,6 +601,104 @@ where
...
+// Implementation to verify size relationships
+impl<Src, Dst> StaticAssert for (Src, Dst)
+where
+    Src: Sized,
+    Dst: Sized,
+{
+    fn assert() -> bool {
+        mem::size_of::<Src>() >= mem::size_of::<Dst>()
+    }
+}
```
jswrenn, line 672: This is dead code; remove it.

src/util/macro_util.rs
```diff
@@ -599,6 +601,104 @@ where
...
+        Ok(ptr) => {
+            let dst = ptr.bikeshed_recall_aligned().as_mut();
+            // SAFETY: By shadowing `dst`, we ensure that `dst` is not re-used
+            // after taking its inner value.
+            let dst = unsafe { ManuallyDrop::take(dst) };
+            Ok(dst.into_inner())
+        }
+        Err(_) => Err(ValidityError::new(ManuallyDrop::into_inner(src))),
+    }
+}
```
jswrenn, line 700: We don't want to introduce any new public functions or macros in our API for prefix transmutation.

src/util/macro_util.rs
```diff
@@ -619,7 +719,7 @@ where
 {
     match try_cast_or_pme::<Src, Dst, _, BecauseImmutable>(Ptr::from_ref(src)) {
         Ok(ptr) => {
-            static_assert!(Src, Dst => mem::align_of::<Dst>() <= mem::align_of::<Src>());
+            assert!(mem::align_of::<Dst>() <= mem::align_of::<Src>());
```
jswrenn, line 722: This changes the semantics of the function. It's important that the assertion as at post-monomorphization time.

src/util/macro_util.rs
```diff
@@ -649,7 +749,7 @@ where
 {
     match try_cast_or_pme::<Src, Dst, _, BecauseExclusive>(Ptr::from_mut(src)) {
         Ok(ptr) => {
-            static_assert!(Src, Dst => mem::align_of::<Dst>() <= mem::align_of::<Src>());
```
jswrenn, line 652: This changes the semantics of the function. It's important that the assertion as at post-monomorphization time.

jswrenn (changes requested):

src/macros.rs
```diff
@@ -392,74 +411,109 @@ macro_rules! transmute_mut {
...
                 #[allow(clippy::missing_transmute_annotations)]
                 $crate::util::macro_util::core_reexport::mem::transmute(e)
             })
         } else {
             $crate::util::macro_util::try_transmute::<_, _>(e)
         }
-    }}
+    }};
+
+    (@prefix $src_type:ty, $dest_type:ty, $e:expr) => {{
```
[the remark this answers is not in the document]
Aditya-PS-05: I made try_transmute as 

```
#[macro_export]
macro_rules! try_transmute {
    ($e:expr) => {{
        let e = $e;
        if false {
            // SAFETY: `core::mem::transmute` ensures that the type of `e` and
            // the type of this macro invocation expression have the same size.
            // This check is enforced by the false branch, so the actual transmute
            // operation is safe.
            Ok(unsafe {
                #[allow(clippy::missing_transmute_annotations)]
                $crate::util::macro_util::core_reexport::mem::transmute(e)
            })
        } else {
            $crate::util::macro_util::try_transmute(e)
        }
    }};

    (@prefix $e:expr) => {{
        let e = $e;
        if false {
            #[allow(clippy::missing_transmute_annotations)]
            // SAFETY: This code is never executed, and only verifies size
            // compatibility at compile time using `MaxSizesOf`. Thus, no runtime
            // transmute is actually performed here.
            let _ = unsafe {
                $crate::util::macro_util::core_reexport::mem::transmute::<_, $crate::util::macro_util::MaxSizesOf<_, _>>(e)
            };

            // SAFETY: The above transmute check ensures that the prefix transmute
            // is safe, as `core::mem::transmute_copy` will only copy valid bytes.
            Ok(unsafe {
                $crate::util::macro_util::core_reexport::mem::transmute_copy(&e)
            })
        } else {
            $crate::util::macro_util::try_transmute(e)
        }
    }};
}
```
When I write a test to try it out `Type annotations needed cannot infer type` error is occuring.

```
let _ = unsafe {
                $crate::util::macro_util::core_reexport::mem::transmute::<_, $crate::util::macro_util::MaxSizesOf<_, _>>(e)
            };
```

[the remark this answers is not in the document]
Aditya-PS-05: @jswrenn , should I avoid writing a test?

Aditya-PS-05:
@jswrenn , please review this.

jswrenn:
I appreciate the work you've put into this PR, but it's still pretty far from where we could accept it. I'm going to close this PR, since I unfortunately don't have the capacity to give this the review and care it needs for the foreseeable future.

Issue #1957, opened 2024-10-21
[unsafe-fields] Support enums


joshlf:

joshlf:
@dacozai Would you like me to officially assign this issue to you?

PR #2034, opened 2024-11-08
[unsafe-fields] adding enum rule into unsafe-fields macro

dacozai:
#1957: support enum in [unsafe-fields]

### DIFF
--- a/unsafe-fields/src/lib.rs
+++ b/unsafe-fields/src/lib.rs
@@ -329,6 +329,44 @@ macro_rules! unsafe_fields {
             )*
         }
     };
+
+    (
+        $( #[$attr:meta] )*
+        $vis:vis enum $name:ident {
+            $(
+                $(#[$variant_attrs:meta])*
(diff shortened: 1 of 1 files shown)


dacozai:
I am sorry that I should clarify this in the first place.
I am not sure that the below are goals as well
```rust
unsafe_fields {
    enum FooEnum {
        #[unsafe]
        A,
        B,
        #[unsafe]
        C(i32, String),
    }
}
```
1. I am not sure unit variant is possible to do so. Maybe the input name will be `A` and the input type will be `FooEnum::A`.
2. I am still surveying the tuple variant. Although I don't know if it can be achieved or not, it seems achievable to me.

Thank you 🙏🏻

dacozai:
Hi @jswrenn,
After few days of surveying, I still failed to achieve 😭 
I posted a question in the forum to check if I missed anything. If there is any direction that you can guide, please let me know.
Appreciate your help. Many thanks.

PR #2055, opened 2024-11-13
Add initial support for unsized `MaybeUninit` wrapper type

jswrenn:
This is achieved by adding a `MaybeUninit` associated type to `KnownLayout`, whose layout is identical to `Self` except that it admits uninitialized bytes in all positions.

For sized types, this is bound to `mem::MaybeUninit<Self>`. For potentially unsized structs, we synthesize a doppelganger with the same `repr`, whose leading fields are wrapped in `mem::MaybeUninit` and whose trailing field is the `MaybeUninit` associated type of struct's original trailing field type. This type-level recursion bottoms out at `[T]`, whose `MaybeUninit` associated type is bound to `[mem::MaybeUninit<T>]`.

### `MaybeUninit<T: ?Sized>`

In the near term, we may remove `doc(hidden)` from our `MaybeUninit` wrapper. In doing so, we'd be quick-to-ship a gadget that extends the present capabilities of Rust. We might, then, be able to use this feature as a demonstration of our approach, potentially suitable for upstreaming into rustc.

### `UnalignUnsized<T: ?Sized>`

Presently, our `Unalign` wrapper requires `T: Sized`, because we could not figure out how to drop unsized values. For sized values, we copy them onto the stack, then run their destructor. With `MaybeUninit<T: ?Sized>`, we can extend unalign support to unsized values, by first copying them into a `Box<MaybeUninit<T>>`, casting the box to `Box<T>`, and dropping it. This process would be skipped for types with trivial drops, keeping the common case simple and efficient. 

### `Value<T, I> where I: Invariants<Validity = Any>`

Combining the above two approaches, we could create an invariant-parameterized `Ptr` analogue for values that generalizes over initialization and alignment.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -244,6 +244,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
+      with:
+        fetch-depth: 2
 
     - name: Populate cache
       uses: ./.github/actions/cache
@@ -490,7 +492,7 @@ jobs:
 
         if [ "${{ github.event_name }}" == "pull_request" ]; then
           # Invoked from a PR - get the PR body directly
-          MESSAGE="${{ github.event.pull_request.body }}"
+          MESSAGE="$(git log -1 --pretty=%B ${{ github.event.pull_request.head.sha }})"
         else
           # Invoked from the merge queue - get the commit message
           MESSAGE="$(git log -1 --pretty=%B ${{ github.sha }})"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -354,7 +354,7 @@ use core::{
     fmt::{self, Debug, Display, Formatter},
     hash::Hasher,
     marker::PhantomData,
-    mem::{self, ManuallyDrop, MaybeUninit},
+    mem::{self, ManuallyDrop, MaybeUninit as CoreMaybeUninit},
     num::{
         NonZeroI128, NonZeroI16, NonZeroI32, NonZeroI64, NonZeroI8, NonZeroIsize, NonZeroU128,
         NonZeroU16, NonZeroU32, NonZeroU64, NonZeroU8, NonZeroUsize, Wrapping,
@@ -732,6 +732,15 @@ pub unsafe trait KnownLayout {
     /// This is `()` for sized types and `usize` for slice DSTs.
     type PointerMetadata: PointerMetadata;
 
+    /// A maybe-uninitialized analog of `Self`
+    ///
+    /// # Safety
+    ///
+    /// `Self::LAYOUT` and `Self::MaybeUninit::LAYOUT` are identical.
+    /// `Self::MaybeUninit` admits uninitialized bytes in all positions.
+    #[doc(hidden)]
+    type MaybeUninit: ?Sized + KnownLayout<PointerMetadata = Self::PointerMetadata>;
+
     /// The layout of `Self`.
     ///
     /// # Safety
@@ -864,6 +873,35 @@ unsafe impl<T> KnownLayout for [T] {
 
     type PointerMetadata = usize;
 
+    // SAFETY: `CoreMaybeUninit<T>::LAYOUT` and `T::LAYOUT` are identical
+    // because `CoreMaybeUninit<T>` has the same size and alignment as `T` [1].
+    // Consequently, `[CoreMaybeUninit<T>]::LAYOUT` and `[T]::LAYOUT` are
+    // identical, because they both lack a fixed-sized prefix and because they
+    // inherit the alignments of their inner element type (which are identical)
+    // [2][3].
+    //
+    // `[CoreMaybeUninit<T>]` admits uninitialized bytes at all positions
+    // because `CoreMaybeUninit<T>` admits uninitialized bytes at all positions
+    // and because the inner elements of `[CoreMaybeUninit<T>]` are laid out
+    // back-to-back [2][3].
+    //
+    // [1] Per https://doc.rust-lang.org/1.81.0/std/mem/union.MaybeUninit.html#layout-1:
+    //
+    //   `MaybeUninit<T>` is guaranteed to have the same size, alignment, and ABI as
+    //   `T`
+    //
+    // [2] Per https://doc.rust-lang.org/1.82.0/reference/type-layout.html#slice-layout:
+    //
+    //   Slices have the same layout as the section of the array they slice.
+    //
+    // [3] Per https://doc.rust-lang.org/1.82.0/reference/type-layout.html#array-layout:
+    //
+    //   An array of `[T; N]` has a size of `size_of::<T>() * N` and the same
+    //   alignment of `T`. Arrays are laid out so that the zero-based `nth`
+    //   element of the array is offset from the start of the array by `n *
+    //   size_of::<T>()` bytes.
+    type MaybeUninit = [CoreMaybeUninit<T>];
+
     const LAYOUT: DstLayout = DstLayout::for_slice::<T>();
 
     // SAFETY: `.cast` preserves address and provenance. The returned pointer
@@ -916,9 +954,11 @@ impl_known_layout!(
     T         => Option<T>,
     T: ?Sized => PhantomData<T>,
     T         => Wrapping<T>,
-    T         => MaybeUninit<T>,
+    T         => CoreMaybeUninit<T>,
     T: ?Sized => *const T,
-    T: ?Sized => *mut T
+    T: ?Sized => *mut T,
+    T: ?Sized => &'_ T,
+    T: ?Sized => &'_ mut T,
 );
 impl_known_layout!(const N: usize, T => [T; N]);
 
@@ -947,6 +987,21 @@ safety_comment! {
     unsafe_impl_known_layout!(#[repr([u8])] str);
     unsafe_impl_known_layout!(T: ?Sized + KnownLayout => #[repr(T)] ManuallyDrop<T>);
     unsafe_impl_known_layout!(T: ?Sized + KnownLayout => #[repr(T)] UnsafeCell<T>);
+}
+
+safety_comment! {
+    /// SAFETY:
+    /// - By consequence of the invariant on `T::MaybeUninit` that `T::LAYOUT`
+    ///   and `T::MaybeUninit::LAYOUT` are equal, `T` and `T::MaybeUninit`
+    ///   have the same:
+    ///   - Fixed prefix size
+    ///   - Alignment
+    ///   - (For DSTs) trailing slice element size
+    /// - By consequence of the above, referents `T::MaybeUninit` and `T` have
+    ///   the require the same kind of pointer metadata, and thus it is valid to
+    ///   perform an `as` cast from `*mut T` and `*mut T::MaybeUninit`, and this
+    ///   operation preserves referent size (ie, `size_of_val_raw`).
+    unsafe_impl_known_layout!(T: ?Sized + KnownLayout => #[repr(T::MaybeUninit)] MaybeUninit<T>);
 }
 
 /// Analyzes whether a type is [`FromZeros`].
@@ -2550,7 +2605,7 @@ pub unsafe trait TryFromBytes {
     where
         Self: Sized,
     {
-        let candidate = match MaybeUninit::<Self>::read_from_bytes(source) {
+        let candidate = match CoreMaybeUninit::<Self>::read_from_bytes(source) {
             Ok(candidate) => candidate,
             Err(e) => {
                 return Err(TryReadError::Size(e.with_dst()));
@@ -2611,7 +2666,7 @@ pub unsafe trait TryFromBytes {
     where
         Self: Sized,
     {
-        let (candidate, suffix) = match MaybeUninit::<Self>::read_from_prefix(source) {
+        let (candidate, suffix) = match CoreMaybeUninit::<Self>::read_from_prefix(source) {
             Ok(candidate) => candidate,
             Err(e) => {
                 return Err(TryReadError::Size(e.with_dst()));
@@ -2673,7 +2728,7 @@ pub unsafe trait TryFromBytes {
     where
         Self: Sized,
     {
-        let (prefix, candidate) = match MaybeUninit::<Self>::read_from_suffix(source) {
+        let (prefix, candidate) = match CoreMaybeUninit::<Self>::read_from_suffix(source) {
             Ok(candidate) => candidate,
             Err(e) => {
                 return Err(TryReadError::Size(e.with_dst()));
@@ -2746,7 +2801,7 @@ fn swap<T, U>((t, u): (T, U)) -> (U, T) {
 #[inline(always)]
 unsafe fn try_read_from<S, T: TryFromBytes>(
     source: S,
-    mut candidate: MaybeUninit<T>,
+    mut candidate: CoreMaybeUninit<T>,
 ) -> Result<T, TryReadError<S, T>> {
     // We use `from_mut` despite not mutating via `c_ptr` so that we don't need
     // to add a `T: Immutable` bound.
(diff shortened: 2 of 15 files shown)


src/lib.rs
```diff
@@ -727,6 +727,14 @@ pub unsafe trait KnownLayout {
     /// This is `()` for sized types and `usize` for slice DSTs.
     type PointerMetadata: PointerMetadata;
 
+    /// A maybe-uninitialized analog of `Self`
+    ///
+    /// # Safety
+    ///
+    /// `Self::LAYOUT` and `Self::MaybeUninit::LAYOUT` are identical.
```
joshlf, line 734: Presumably it's also a safety invariant that `Self::MaybeUninit` has no bit validity requirement?

src/util/mod.rs
```diff
@@ -683,6 +683,114 @@ pub(crate) unsafe fn copy_unchecked(src: &[u8], dst: &mut [u8]) {
...
 }
 
+/// Unsafely transmutes the given `src` into a type `Dst`.
+///
+/// # Safety
+///
+/// The value `src` must be a bit-valid instance of `Dst`.
+#[inline(always)]
+pub(crate) const unsafe fn transmute_unchecked<Src, Dst>(src: Src) -> Dst {
+    static_assert!(Src, Dst => core::mem::size_of::<Src>() == core::mem::size_of::<Dst>());
```
joshlf, line 693: Mention this requirement in the doc comment?

src/util/mod.rs
```diff
@@ -683,6 +683,114 @@ pub(crate) unsafe fn copy_unchecked(src: &[u8], dst: &mut [u8]) {
...
+    #[repr(C)]
+    union Transmute<Src, Dst> {
+        src: ManuallyDrop<Src>,
+        dst: ManuallyDrop<Dst>,
+    }
+    // SAFETY: The caller promises that `src` is a bit-valid instance of `Dst`.
```
joshlf, line 699: Also argue that this is a transmute? Namely that `src` and `dst` are at the same byte offsets within `Transmute` and that `ManuallyDrop<T>` has the same layout as `T`.

src/util/mod.rs
```diff
@@ -683,6 +683,114 @@ pub(crate) unsafe fn copy_unchecked(src: &[u8], dst: &mut [u8]) {
...
+    unsafe { ManuallyDrop::into_inner(Transmute { src: ManuallyDrop::new(src) }.dst) }
+}
+
+/// Use `allocate` to create a `Box<T>`.
```
joshlf, line 703:
```suggestion
/// Uses `allocate` to create a `Box<T>`.
```

src/util/mod.rs
```diff
@@ -683,6 +683,114 @@ pub(crate) unsafe fn copy_unchecked(src: &[u8], dst: &mut [u8]) {
...
+///
+/// # Errors
+///
+/// Returns an error on allocation failure. Allocation failure is guaranteed
+/// never to cause a panic or an abort.
+///
+/// # Safety
```
joshlf, line 710: Presumably there should be invariants regarding the pre-conditions of calling `allocate`?

src/util/mod.rs
```diff
@@ -683,6 +683,114 @@ pub(crate) unsafe fn copy_unchecked(src: &[u8], dst: &mut [u8]) {
...
+///
+/// The referent of the pointer returned by `allocate` must be a bit-valid
+/// instance of `T`.
```
joshlf, line 713: I assume this should instead refer to the `Layout` argument to `allocate`? It may also be necessary to say something about `Layout` being related to `T`, but maybe not. I haven't thought too deeply about it.

src/util/mod.rs
joshlf, line 710: Presumably the safety post-condition (and maybe also the pre-condition) should mention `meta` in some way?

src/wrappers.rs
```diff
@@ -464,6 +464,137 @@ impl<T: Unaligned + Display> Display for Unalign<T> {
...
+#[repr(transparent)]
+#[doc(hidden)]
+pub struct MaybeUninit<T: ?Sized + KnownLayout>(
+    // SAFETY: `MaybeUninit<T>` has the same size as `T`, because (by invariant
+    // on `T::MaybeUninit`) `T::MaybeUninit` has `T::LAYOUT` identical to `T`,
+    // and because (invariant on `T::LAYOUT`) we can trust that `LAYOUT`
+    // accurately reflects the layout of `T`. By invariant on `T::MaybeUninit`,
+    // it admits uninitialized bytes in all positions. Because `MabyeUninit` is
+    // marked `repr(transparent)`, these properties additionally hold true for
+    // `MaybeUninit`.
```
joshlf, line 496:
```suggestion
    // `Self`.
```

There are a lot of `MaybeUninit`s flying around, so this strikes me as less ambiguous.

src/wrappers.rs
```diff
@@ -464,6 +464,137 @@ impl<T: Unaligned + Display> Display for Unalign<T> {
...
+    T::MaybeUninit,
+);
+
+#[doc(hidden)]
+impl<T: ?Sized + KnownLayout> MaybeUninit<T> {
+    /// Constructs a `MaybeUninit<T>` initialized with the given value.
+    #[inline(always)]
+    pub fn new(val: T) -> Self
+    where
+        T: Sized + KnownLayout<PointerMetadata = ()>,
```
joshlf, line 506: Is the `PointerMetadata = ()` bound required? It's obviously always true, but does it buy us anything to spell it out explicitly?

src/wrappers.rs
```diff
@@ -464,6 +464,137 @@ impl<T: Unaligned + Display> Display for Unalign<T> {
...
+        // `T::MaybeUninit: Sized` bound to this function's `where` clause.
+        unsafe { crate::util::transmute_unchecked(val) }
+    }
+
+    /// Constructs an uninitialized `MaybeUninit<T>`.
+    #[must_use]
+    #[inline(always)]
+    pub fn uninit() -> Self
+    where
+        T: Sized + KnownLayout<PointerMetadata = ()>,
```
joshlf, line 533: Is the `PointerMetadata = ()` bound required? It's obviously always true, but does it buy us anything to spell it out explicitly?

src/wrappers.rs
```diff
@@ -464,6 +464,137 @@ impl<T: Unaligned + Display> Display for Unalign<T> {
...
+    /// Extracts the value from the `MaybeUninit<T>`` container.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that `self` is in an bit-valid state. Depending
+    /// on subsequent use, it may also need to be in a library-valid state.
+    #[inline(always)]
+    pub unsafe fn assume_init(self) -> T
+    where
+        T: Sized + KnownLayout<PointerMetadata = ()>,
```
joshlf, line 583: Is the `PointerMetadata = ()` bound required? It's obviously always true, but does it buy us anything to spell it out explicitly?

zerocopy-derive/src/lib.rs
```diff
@@ -75,7 +75,8 @@ macro_rules! derive {
     ($trait:ident => $outer:ident => $inner:ident) => {
         #[proc_macro_derive($trait)]
         pub fn $outer(ts: proc_macro::TokenStream) -> proc_macro::TokenStream {
-            let ast = syn::parse_macro_input!(ts as DeriveInput);
+            let mut ast = syn::parse_macro_input!(ts as DeriveInput);
+            ast.generics.make_where_clause();
```
joshlf, line 79: Add a comment explaining why we do this?

joshlf (changes requested):
Haven't looked at most of zerocopy-derive yet, but looked at all of zerocopy.

src/util/mod.rs
jswrenn: It's present as a debugging safeguard, but not something callers may rely upon. The onus is entirely on callers to prove that `src` is a bit-valid instance of `Dst`. Those safety proofs will remain correct even if we remove this `static_assert` which we will probably eventually do to support truncating and padding-extension transmutations.

src/util/mod.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/96b54d1b8190c3b88a8d6d0315e14fd241a5a35a..9844c1f976bafef394033070eaf3cef0cf9e4c58#diff-6965df871359c91ee03b5564314f3e131c0877b8603aea626b5d8f9180eb4ee8R700-R717)

src/util/mod.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/96b54d1b8190c3b88a8d6d0315e14fd241a5a35a..9844c1f976bafef394033070eaf3cef0cf9e4c58#diff-6965df871359c91ee03b5564314f3e131c0877b8603aea626b5d8f9180eb4ee8L703-R721)

src/util/mod.rs
jswrenn: Yes, and also that the allocator used is the system allocator — otherwise the resulting `Box` could not be safely dropped.

[Done](https://github.com/google/zerocopy/compare/96b54d1b8190c3b88a8d6d0315e14fd241a5a35a..9844c1f976bafef394033070eaf3cef0cf9e4c58#diff-6965df871359c91ee03b5564314f3e131c0877b8603aea626b5d8f9180eb4ee8L712-R733)

src/wrappers.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/96b54d1b8190c3b88a8d6d0315e14fd241a5a35a..9844c1f976bafef394033070eaf3cef0cf9e4c58#diff-18618e7fc588bc81a0d47bd0e93c08213abdec19646cff8c8be2b654e3c9872dR484)

src/wrappers.rs
jswrenn: [Removed](https://github.com/google/zerocopy/compare/96b54d1b8190c3b88a8d6d0315e14fd241a5a35a..9844c1f976bafef394033070eaf3cef0cf9e4c58#diff-18618e7fc588bc81a0d47bd0e93c08213abdec19646cff8c8be2b654e3c9872dR494)

src/wrappers.rs
jswrenn: [Removed](https://github.com/google/zerocopy/compare/96b54d1b8190c3b88a8d6d0315e14fd241a5a35a..9844c1f976bafef394033070eaf3cef0cf9e4c58#diff-18618e7fc588bc81a0d47bd0e93c08213abdec19646cff8c8be2b654e3c9872dL533)

src/wrappers.rs
jswrenn: [Removed](https://github.com/google/zerocopy/compare/96b54d1b8190c3b88a8d6d0315e14fd241a5a35a..9844c1f976bafef394033070eaf3cef0cf9e4c58#diff-18618e7fc588bc81a0d47bd0e93c08213abdec19646cff8c8be2b654e3c9872dL533)

zerocopy-derive/src/lib.rs
jswrenn: This was vestigial; [removed](https://github.com/google/zerocopy/compare/96b54d1b8190c3b88a8d6d0315e14fd241a5a35a..9844c1f976bafef394033070eaf3cef0cf9e4c58#diff-18618e7fc588bc81a0d47bd0e93c08213abdec19646cff8c8be2b654e3c9872dL533).

src/util/mod.rs
jswrenn: [Post-condition added](https://github.com/google/zerocopy/compare/975118027605cf6b0b56c6e5fd6c95f7d50de5f2..ca928f5dce200a377e9b5ba85e74613d3eceb5ce). There's no safety pre-condition here; a too-large `T` or `meta` produces an `AllocError`.

jswrenn:
The remaining CI failures are from `cargo-semver-checks`; it complains that we've added a new item to `KnownLayout`:
```text
--- failure trait_associated_type_added: non-sealed public trait added associated type without default value ---

Description:
A non-sealed trait has gained an associated type without a default value, which breaks downstream implementations of the trait
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#trait-new-item-no-default
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.36.0/src/lints/trait_associated_type_added.ron

Failed in:
  trait associated type zerocopy::KnownLayout::MaybeUninit in file /home/runner/work/zerocopy/zerocopy/src/lib.rs:742
```

src/util/mod.rs
joshlf: Ack

zerocopy-derive/src/lib.rs
```diff
@@ -238,8 +210,112 @@ fn derive_known_layout_inner(ast: &DeriveInput, _top_level: Trait) -> Result<Tok
...
+                )
+                where
+                    #trailing_field_ty: ::zerocopy::KnownLayout,
+                    #predicates;
+
+                // SAFETY: We largely defer to the `KnownLayout` implementation on
+                // the derive target type (both by using the same tokens, and by
+                // deferring to impl via type-level indirection). This is sound,
+                // since  `__ZerocopyKnownLayoutMaybeUninit` is guaranteed to be
+                // identical to that of the derive target type, except that
```
joshlf, line 293:
```suggestion
                // since  `__ZerocopyKnownLayoutMaybeUninit` is guaranteed to have
                // the same layout as the derive target type, except that
```

joshlf (changes requested):
A few more nits, but otherwise LGTM!

src/util/mod.rs
```diff
@@ -683,6 +683,143 @@ pub(crate) unsafe fn copy_unchecked(src: &[u8], dst: &mut [u8]) {
...
+    static_assert!(Src, Dst => core::mem::size_of::<Src>() == core::mem::size_of::<Dst>());
+
+    #[repr(C)]
+    union Transmute<Src, Dst> {
+        src: ManuallyDrop<Src>,
+        dst: ManuallyDrop<Dst>,
+    }
+
+    // SAFETY: Since `Transmute<Src, Dst>` is `#[repr(C)]`, its `src` and `dst`
+    // fields both start at the safe offset and the types of those fields are
```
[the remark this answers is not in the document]
jswrenn: [Done](https://github.com/google/zerocopy/compare/bb140d07cfd265378c2a38132b7d3b0ba8dbd81a..b573d832ca02f2b735359c87dd92de151e7d8b33#diff-6965df871359c91ee03b5564314f3e131c0877b8603aea626b5d8f9180eb4ee8L702-R702)

src/util/mod.rs
```diff
@@ -683,6 +683,143 @@ pub(crate) unsafe fn copy_unchecked(src: &[u8], dst: &mut [u8]) {
...
+    // transparent wrappers around `Src` and `Dst` [1]. Consequently,
+    // initializng `Transmute` with with `src` and then reading out `dst` is
+    // equivalent to transmuting from `Src` to `Dst` [2].
```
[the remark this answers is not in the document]
jswrenn: [Done](https://github.com/google/zerocopy/compare/bb140d07cfd265378c2a38132b7d3b0ba8dbd81a..b573d832ca02f2b735359c87dd92de151e7d8b33#diff-6965df871359c91ee03b5564314f3e131c0877b8603aea626b5d8f9180eb4ee8R705-R707). Also note [this change](https://github.com/google/zerocopy/compare/bb140d07cfd265378c2a38132b7d3b0ba8dbd81a..b573d832ca02f2b735359c87dd92de151e7d8b33#diff-6965df871359c91ee03b5564314f3e131c0877b8603aea626b5d8f9180eb4ee8L690-R690).

zerocopy-derive/src/lib.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/bb140d07cfd265378c2a38132b7d3b0ba8dbd81a..b573d832ca02f2b735359c87dd92de151e7d8b33#diff-977c29e6cf079bd7307f811df24f99633df603962785c583145eb6250b0e12daL292-R293)

jswrenn:
> We should report this upstream. In particular, given that `KnownLayout` _already_ has `#[doc(hidden)]` items, we know that there can be no downstream implementations of `KnownLayout` which have not already opted out of semver stability.

Filed: https://github.com/obi1kenobi/cargo-semver-checks/issues/997

In the meantime, I think we should permit `cargo-semver-checks` to fail in CI. Otherwise, our only recourse is to bypass branch protections altogether, which would allow this PR to skip some of our more intensive CI steps that are run only in the merge queue.

src/util/mod.rs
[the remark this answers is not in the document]
joshlf: Is that latter change semantically meaningful? Or do you just prefer that wording?

joshlf (approved):

.github/workflows/ci.yml
```diff
@@ -490,7 +490,9 @@ jobs:
 
         if [ "${{ github.event_name }}" == "pull_request" ]; then
           # Invoked from a PR - get the PR body directly
-          MESSAGE="${{ github.event.pull_request.body }}"
+          read -r -d '' MESSAGE <<'EOF'
+            ${{ github.event.pull_request.body }}
+        EOF
```
joshlf, line 495: Indent to the same level as the rest of the `if` body?

```suggestion
          EOF
```

jswrenn: Testing locally, it looks like the EOF needs to be at the start of the line.

joshlf: Ack

joshlf:
Backporting in #2310

PR #2084, opened 2024-11-20
[ci] Permit skipping cargo-semver-checks step

### LINKED ISSUE #2083: [ci] Permit skipping `cargo-semver-checks` for a particular PR
Sometimes cargo-semver-checks has [false positives](https://github.com/google/zerocopy/pull/2055#issuecomment-2488978042). It's not desirable, in these cases, to simply use the GitHub web UI "Merge without waiting for requirements to be met" option:

Image

The reason is that this option also bypasses the merge queue, which we use to perform a number of checks which are disabled during a normal PR CI run.

Instead, we should teach the cargo-semver-checks CI step to check for an explicit directive in a PR's last commit message which instructs it to skip performing its checks. Something like:

```text
SKIP_CARGO_SEMVER_CHECKS=1
```

## Commit message

One important detail is that this directive should go in the last (ie most recent) commit message in the PR. Most of our PRs contain a single commit, but sometimes they contain multiple commits (e.g. when we use the web UI to merge recent changes from the target branch), and third-party contributors often don't use our development flow, and include multiple commits. The reason to use the commit message (as opposed to, for example, the PR message) is that:
- It's easier to check for in a GitHub Action
- It will be trivially available when the PR is running through the merge queue - by that time, multiple commits have been squashed, and their commit messages have been merged
- It's a minor inconvenience to users who get this wrong - CI will fail, and so they'll know that they need to fix something
- It will not admit false positives (ie, we'll never incorrectly skip cargo-semver-checks)

joshlf:
This is triggered when the most recent commit in a PR has a commit message which contains the string `SKIP_CARGO_SEMVER_CHECKS=1` on its own line.

SKIP_CARGO_SEMVER_CHECKS=1

Closes #2083

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -482,6 +482,25 @@ jobs:
         export RUSTDOCFLAGS="${{ matrix.toolchain == 'nightly' && '-Z unstable-options --document-hidden-items $METADATA_DOCS_RS_RUSTDOC_ARGS'|| '' }} $RUSTDOCFLAGS"
         ./cargo.sh +${{ matrix.toolchain }} doc --document-private-items --package ${{ matrix.crate }} ${{ matrix.features }}
 
+    # If the commit message contains the line `SKIP_CARGO_SEMVER_CHECKS=1`, then
+    # skip the cargo-semver-checks step.
+    - name: Check whether to skip cargo-semver-checks
+      run: |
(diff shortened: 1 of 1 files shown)


jswrenn (approved):

joshlf:
Backporting in #2309

PR #2085, opened 2024-11-20
Pin `libc` version to 0.2.163

jswrenn:
0.2.164 bumps the MSRV of libc beyond our crate's. By pinning libc to the last compatible version, we ensure that our CI can run without issue. Probably.


joshlf (approved):

joshlf:
Backporting in #2086.

PR #2086, opened 2024-11-21
Pin `libc` version to 0.2.163 (#2085)

joshlf:
0.2.164 bumps the MSRV of libc beyond our crate's. By pinning libc to the last compatible version, we ensure that our CI can run without issue.


joshlf:
Backporting #2085.

joshlf:
Nevermind, this isn't needed on `main`.

PR #2123, opened 2024-12-03
[derive] Don't emit #[cfg(coverage_nightly)]

### LINKED ISSUE #2117: The `KnownLayout` macro has an issue with the latest nightly
This is failing for me with `nightly-2024-11-26`
```
use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout, Unaligned};

#[derive(
    Clone, Copy, Hash, PartialEq, Eq, FromBytes, IntoBytes, KnownLayout, Immutable, Unaligned,
)]
#[repr(C)]
pub struct Test(pub [u8; 32]);

```
running `cargo +nightly-2024-11-26 clippy` gives
```
warning: unexpected `cfg` condition name: `coverage_nightly`
 --> src/main.rs:5:61
  |
5 |     Clone, Copy, Hash, PartialEq, Eq, FromBytes, IntoBytes, KnownLayout, Immutable, Unaligned,
  |                                                             ^^^^^^^^^^^
  |
  = help: expected names are: `clippy`, `debug_assertions`, `doc`, `docsrs`, `doctest`, `feature`, `fmt_debug`, `miri`, `overflow_checks`, `panic`, `proc_macro`, `relocation_model`, `rustfmt`, `sanitize`, `sanitizer_cfi_generalize_pointers`, `sanitizer_cfi_normalize_integers`, `target_abi`, `target_arch`, `target_endian`, `target_env`, `target_family`, `target_feature`, `target_has_atomic`, `target_has_atomic_equal_alignment`, `target_has_atomic_load_store`, `target_os`, `target_pointer_width`, `target_thread_local`, `target_vendor`, `test`, `ub_checks`, `unix`, and `windows`
  = help: consider using a Cargo feature instead
  = help: or consider adding in `Cargo.toml` the `check-cfg` lint config for the lint:
           [lints.rust]
           unexpected_cfgs = { level = "warn", check-cfg = ['cfg(coverage_nightly)'] }
  = help: or consider adding `println!("cargo::rustc-check-cfg=cfg(coverage_nightly)");` to the top of the `build.rs`
  = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
  = note: `#[warn(unexpected_cfgs)]` on by default
  = note: this warning originates in the derive macro `KnownLayout` (in Nightly builds, run with -Z macro-backtrace for more info)
```
which is causing issues in my strict CI

joshlf:
As of nightly-2024-11-20 - specifically [1] - this triggers an `unexpected_cfgs` lint even when emitted in derive-generated code.

[1] https://github.com/rust-lang/rust/pull/132577

Fixes #2117

### DIFF
--- a/zerocopy-derive/src/lib.rs
+++ b/zerocopy-derive/src/lib.rs
@@ -301,7 +301,6 @@ fn derive_known_layout_inner(ast: &DeriveInput, _top_level: Trait) -> Result<Tok
                     #predicates
                 {
                     #[allow(clippy::missing_inline_in_public_items)]
-                    #[cfg_attr(coverage_nightly, coverage(off))]
                     fn only_derive_is_allowed_to_implement_this_trait() {}
 
                     type PointerMetadata = <#ident #ty_generics as ::zerocopy::KnownLayout>::PointerMetadata;
--- a/zerocopy-derive/src/output_tests.rs
+++ b/zerocopy-derive/src/output_tests.rs
@@ -192,7 +192,6 @@ fn test_known_layout() {
                     <U as ::zerocopy::KnownLayout>::MaybeUninit: ::zerocopy::KnownLayout,
                 {
                     #[allow(clippy::missing_inline_in_public_items)]
-                    #[cfg_attr(coverage_nightly, coverage(off))]
                     fn only_derive_is_allowed_to_implement_this_trait() {}
                     type PointerMetadata = <Foo<T, U> as ::zerocopy::KnownLayout>::PointerMetadata;
                     type MaybeUninit = Self;
--- a/zerocopy-derive/tests/issue_2117.rs
+++ b/zerocopy-derive/tests/issue_2117.rs
@@ -1,0 +1,20 @@
+// Copyright 2024 The Fuchsia Authors
+//
+// Licensed under a BSD-style license <LICENSE-BSD>, Apache License, Version 2.0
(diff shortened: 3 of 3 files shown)


jswrenn (approved):

joshlf:
Backporting to `main` in #2126

taiki-e:
AFAIK, this is the kind of cfg that is set globally on the coverage tool side, so the warning itself is not something we need to be concerned about except that it will break the build if we deny the warning.

Therefore, I wonder if it would be fine to simply allow unexpected_cfg in the generated code.
(See https://github.com/taiki-e/pin-project-lite/issues/33#issuecomment-700494455 for why we believe that generated code should not be expected to be compatible with `forbid` except for certain lints.)

joshlf:
> AFAIK, this is the kind of cfg that is set globally on the coverage tool side, so the warning itself is not something we need to be concerned about except that it will break the build if we deny the warning.
> 
> Therefore, I wonder if it would be fine to simply allow unexpected_cfg in the generated code. (See [taiki-e/pin-project-lite#33 (comment)](https://github.com/taiki-e/pin-project-lite/issues/33#issuecomment-700494455) for why we believe that generated code should not be expected to be compatible with `forbid` except for certain lints.)

I tend to agree that this is bad CI hygiene. However, the benefit to us of using `cfg(coverage_nightly)` is nearly zero, so we decided to err on the side of unblocking users even if technically those users were assuming too much. But I do agree with your reasoning.

PR #2127, opened 2024-12-03
Practice good `Self` hygiene

### LINKED ISSUE #2116: Cycle detected when evaluating trait selection obligation
With `zerocopy` 0.8.11, the following code failed to compile:

```rust
use zerocopy::*;

#[derive(KnownLayout)]
#[repr(C)]
pub struct Dummy([u8; Self::CONST]);

impl Dummy {
    const CONST: usize = 42;
}
```

```plain
error[E0391]: cycle detected when evaluating trait selection obligation `_::__ZerocopyKnownLayoutMaybeUninit: core::iter::traits::collect::IntoIterator`
  |
note: ...which requires evaluating type-level constant...
 --> src/lib.rs:5:23
  |
5 | pub struct Dummy([u8; Self::CONST]);
  |                       ^^^^^^^^^^^
note: ...which requires const-evaluating + checking `_::__ZerocopyKnownLayoutMaybeUninit::0::{constant#0}`...
 --> src/lib.rs:5:23
  |
5 | pub struct Dummy([u8; Self::CONST]);
  |                       ^^^^^^^^^^^
note: ...which requires caching mir of `_::__ZerocopyKnownLayoutMaybeUninit::0::{constant#0}` for CTFE...
 --> src/lib.rs:5:23
  |
5 | pub struct Dummy([u8; Self::CONST]);
  |                       ^^^^^^^^^^^
note: ...which requires elaborating drops for `_::__ZerocopyKnownLayoutMaybeUninit::0::{constant#0}`...
 --> src/lib.rs:5:23
  |
5 | pub struct Dummy([u8; Self::CONST]);
  |                       ^^^^^^^^^^^
note: ...which requires borrow-checking `_::__ZerocopyKnownLayoutMaybeUninit::0::{constant#0}`...
 --> src/lib.rs:5:23
  |
5 | pub struct Dummy([u8; Self::CONST]);
  |                       ^^^^^^^^^^^
note: ...which requires promoting constants in MIR for `_::__ZerocopyKnownLayoutMaybeUninit::0::{constant#0}`...
 --> src/lib.rs:5:23
  |
5 | pub struct Dummy([u8; Self::CONST]);
  |                       ^^^^^^^^^^^
note: ...which requires const checking `_::__ZerocopyKnownLayoutMaybeUninit::0::{constant#0}`...
 --> src/lib.rs:5:23
  |
5 | pub struct Dummy([u8; Self::CONST]);
  |                       ^^^^^^^^^^^
note: ...which requires building MIR for `_::__ZerocopyKnownLayoutMaybeUninit::0::{constant#0}`...
 --> src/lib.rs:5:23
  |
5 | pub struct Dummy([u8; Self::CONST]);
  |                       ^^^^^^^^^^^
note: ...which requires match-checking `_::__ZerocopyKnownLayoutMaybeUninit::0::{constant#0}`...
 --> src/lib.rs:5:23
  |
5 | pub struct Dummy([u8; Self::CONST]);
  |                       ^^^^^^^^^^^
note: ...which requires type-checking `_::__ZerocopyKnownLayoutMaybeUninit::0::{constant#0}`...
 --> src/lib.rs:5:29
  |
5 | pub struct Dummy([u8; Self::CONST]);
  |                             ^^^^^
  = note: ...which again requires evaluating trait selection obligation `_::__ZerocopyKnownLayoutMaybeUninit: core::iter::traits::collect::IntoIterator`, completing the cycle
note: cycle used when type-checking `_::<impl at src/lib.rs:3:10: 3:21>::{constant#0}`
 --> src/lib.rs:5:29
  |
5 | pub struct Dummy([u8; Self::CONST]);
  |                             ^^^^^
  = note: see https://rustc-dev-guide.rust-lang.org/overview.html#queries and https://rustc-dev-guide.rust-lang.org/query.html for more information

For more information about this error, try `rustc --explain E0391`.
```

This compiles on 0.8.10.

```
> rustc --version
rustc 1.83.0 (90b35a623 2024-11-26)
> cargo --version
cargo 1.83.0 (5ffbef321 2024-10-29)
> cat Cargo.toml
[package]
name = "sandbox"
version = "0.1.0"
edition = "2021"

[dependencies]
#zerocopy = { version = "=0.8.11", features = ["derive"] }
zerocopy = { version = "=0.8.10", features = ["derive"] }
```

jswrenn:
This commit revises the `KnownLayout` derive on `repr(C)` target structs to preserve the correct resolution of `Self` when constructing `__ZerocopyKnownLayoutMaybeUninit`. This type contains a `MaybeUninit` version of each of the target struct's field types. Previously, it achieved this by simply copying the tokens corresponding to field types from the definition of the target struct into the definition of `__ZerocopyKnownLayoutMaybeUninit`

However, on types like this:

```rust
#[repr(C)]
struct Struct([u8; Self::N]);
```

…this approach is insufficient. Pasted into `__ZerocopyKnownLayoutMaybeUninit`, `Self` ceases to refer to the target struct and instead refers to `__ZerocopyKnownLayoutMaybeUninit`.

To preserve `Self` hygiene, this commit defines a struct for projecting the field types of the target struct based on their index:

```rust
pub unsafe trait Field<const N: usize> {
    type Type: ?Sized;
}
```

…then implements it for each of the field types of the target struct; e.g.:

```rust
impl Field<0> for Struct {
    type Type = [u8; Self::N];
}
```

With this, the fields of `__ZerocopyKnownLayoutMaybeUninit` can be defined hygienically; e.g., as `<Struct as Field<0>>::Type`.

Fixes #2116

### DIFF
--- a/src/util/macro_util.rs
+++ b/src/util/macro_util.rs
@@ -32,6 +32,20 @@ use crate::{
     Immutable, IntoBytes, Ptr, TryFromBytes, Unalign, ValidityError,
 };
 
+/// Projects the type of the field at `Index` in `Self`.
+///
+/// The `Index` parameter is any sort of handle that identifies the field; its
+/// definition is the obligation of the implementer.
+///
+/// # Safety
+///
+/// Unsafe code may assume that this accurately reflects the definition of
+/// `Self`.
+pub unsafe trait Field<Index> {
+    /// The type of the field at `Index`.
+    type Type: ?Sized;
+}
+
 #[cfg_attr(
     zerocopy_diagnostic_on_unimplemented_1_78_0,
     diagnostic::on_unimplemented(
--- a/zerocopy-derive/src/lib.rs
+++ b/zerocopy-derive/src/lib.rs
@@ -144,8 +144,8 @@ fn derive_known_layout_inner(ast: &DeriveInput, _top_level: Trait) -> Result<Tok
         Some((trailing_field, leading_fields)),
     ) = (is_repr_c_struct, fields.split_last())
     {
-        let (_name, trailing_field_ty) = trailing_field;
-        let leading_fields_tys = leading_fields.iter().map(|(_name, ty)| ty);
+        let (_vis, trailing_field_name, trailing_field_ty) = trailing_field;
+        let leading_fields_tys = leading_fields.iter().map(|(_vis, _name, ty)| ty);
 
         let core_path = quote!(::zerocopy::util::macro_util::core_reexport);
         let repr_align = repr
@@ -266,20 +266,66 @@ fn derive_known_layout_inner(ast: &DeriveInput, _top_level: Trait) -> Result<Tok
                 Default::default()
             };
 
+            // Generate a valid ident for a type-level handle to a field of a
+            // given `name`.
+            let field_index =
+                |name| Ident::new(&format!("__Zerocopy_Field_{}", name), ident.span());
+
+            let field_indices: Vec<_> =
+                fields.iter().map(|(_vis, name, _ty)| field_index(name)).collect();
+
+            // Define the collection of type-level field handles.
+            let field_defs = field_indices.iter().zip(&fields).map(|(idx, (vis, _, _))| {
+                quote! {
+                    #[allow(non_camel_case_types)]
+                    #vis struct #idx;
+                }
+            });
+
+            let field_impls = field_indices.iter().zip(&fields).map(|(idx, (_, _, ty))| quote! {
+                // SAFETY: `#ty` is the type of `#ident`'s field at `#idx`.
+                unsafe impl #impl_generics ::zerocopy::util::macro_util::Field<#idx> for #ident #ty_generics
+                where
+                    #predicates
+                {
+                    type Type = #ty;
+                }
+            });
+
+            let trailing_field_index = field_index(trailing_field_name);
+            let leading_field_indices =
+                leading_fields.iter().map(|(_vis, name, _ty)| field_index(name));
+
+            let trailing_field_ty = quote! {
+                <#ident #ty_generics as
+                    ::zerocopy::util::macro_util::Field<#trailing_field_index>
+                >::Type
+            };
+
             let methods = make_methods(&parse_quote! {
                 <#trailing_field_ty as ::zerocopy::KnownLayout>::MaybeUninit
             });
 
             quote! {
+                #(#field_defs)*
+
+                #(#field_impls)*
+
                 // SAFETY: This has the same layout as the derive target type,
-                // except that it admits uninit bytes. This is ensured by using the
-                // same repr as the target type, and by using field types which have
-                // the same layout as the target type's fields, except that they
-                // admit uninit bytes.
+                // except that it admits uninit bytes. This is ensured by using
+                // the same repr as the target type, and by using field types
+                // which have the same layout as the target type's fields,
+                // except that they admit uninit bytes. We indirect through
+                // `Field` to ensure that occurrences of `Self` resolve to
+                // `#ty`, not `__ZerocopyKnownLayoutMaybeUninit` (see #2116).
                 #repr
                 #[doc(hidden)]
                 #vis struct __ZerocopyKnownLayoutMaybeUninit<#params> (
-                    #(::zerocopy::util::macro_util::core_reexport::mem::MaybeUninit<#leading_fields_tys>,)*
+                    #(::zerocopy::util::macro_util::core_reexport::mem::MaybeUninit<
+                        <#ident #ty_generics as
+                            ::zerocopy::util::macro_util::Field<#leading_field_indices>
+                        >::Type
+                    >,)*
                     <#trailing_field_ty as ::zerocopy::KnownLayout>::MaybeUninit
                 )
                 where
@@ -295,9 +341,6 @@ fn derive_known_layout_inner(ast: &DeriveInput, _top_level: Trait) -> Result<Tok
                 unsafe impl #impl_generics ::zerocopy::KnownLayout for __ZerocopyKnownLayoutMaybeUninit #ty_generics
                 where
                     #trailing_field_ty: ::zerocopy::KnownLayout,
-                    // This bound may appear to be superfluous, but is required
-                    // on our MSRV (1.55) to avoid an ICE.
-                    <#trailing_field_ty as ::zerocopy::KnownLayout>::MaybeUninit: ::zerocopy::KnownLayout,
                     #predicates
                 {
                     #[allow(clippy::missing_inline_in_public_items)]
@@ -495,8 +538,8 @@ fn derive_try_from_bytes_struct(
 ) -> Result<TokenStream, Error> {
     let extras = try_gen_trivial_is_bit_valid(ast, top_level).unwrap_or_else(|| {
         let fields = strct.fields();
-        let field_names = fields.iter().map(|(name, _ty)| name);
-        let field_tys = fields.iter().map(|(_name, ty)| ty);
+        let field_names = fields.iter().map(|(_vis, name, _ty)| name);
+        let field_tys = fields.iter().map(|(_vis, _name, ty)| ty);
         quote!(
             // SAFETY: We use `is_bit_valid` to validate that each field is
             // bit-valid, and only return `true` if all of them are. The bit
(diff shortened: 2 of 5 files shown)


joshlf:
> Unfortunately, with the added `Field` indirection in `trailing_field_ty`, the hack here to avoid an ICE on our MSRV no longer does so: https://github.com/google/zerocopy/blob/e3c52768b0143100e09eec3aa55f8e77295bafe4/zerocopy-derive/src/lib.rs#L320-L327
> 
> 
> 
> This is the test case that exhibits the ICE: https://github.com/google/zerocopy/blob/ac7b8e8e98810fdc754fe79734a2235be3bbe3db/zerocopy-derive/tests/struct_known_layout.rs#L50-L59
> 
> 
> 
> The problematic quality here is the lifetimes in the trailing field. Lifetimes in leading fields are unproblematic.
> 
> 
> 
> Not yet sure how to work around this.

Hot take: what if we don't work around it? The likelihood of someone encountering this in practice is extremely low since our MSRV is so old. If someone does, we can explain that it's a compiler bug and thus not really our fault.

I'd want to figure out what toolchain fixed this ICE. If it's a recent toolchain, then I wouldn't want to go this route, but if it's like 1.58 or something, then I'm okay with it.

joshlf:
> Unfortunately, with the added `Field` indirection in `trailing_field_ty`, the hack here to avoid an ICE on our MSRV no longer does so: https://github.com/google/zerocopy/blob/e3c52768b0143100e09eec3aa55f8e77295bafe4/zerocopy-derive/src/lib.rs#L320-L327
> 
> 
> 
> This is the test case that exhibits the ICE: https://github.com/google/zerocopy/blob/ac7b8e8e98810fdc754fe79734a2235be3bbe3db/zerocopy-derive/tests/struct_known_layout.rs#L50-L59
> 
> 
> 
> The problematic quality here is the lifetimes in the trailing field. Lifetimes in leading fields are unproblematic.
> 
> 
> 
> Not yet sure how to work around this.

Hot take: what if we don't work around it? The likelihood of someone encountering this in practice is extremely low since our MSRV is so old. If someone does, we can explain that it's a compiler bug and thus not really our fault.

I'd want to figure out what toolchain fixed this ICE. If it's a recent toolchain, then I wouldn't want to go this route, but if it's like 1.58 or something, then I'm okay with it.

jswrenn:
The last problematic rustc version is 1.61, released May 19, 2022.

jswrenn:
> The problematic quality here is the lifetimes in the trailing field. Lifetimes in leading fields are unproblematic.

Static lifetimes are also unproblematic.

src/util/macro_util.rs
```diff
@@ -32,6 +32,17 @@ use crate::{
     Immutable, IntoBytes, Ptr, TryFromBytes, Unalign, ValidityError,
 };
 
+/// Projects the type of the field at `Index` in `Self`.
+///
+/// # Safety
+///
+/// Unsafe code may assume that this accurately reflects the definition of
+/// `Self`.
```
joshlf, line 40: Document what the `Index` type represents? IMO, "index" is an odd name since this will really end up being a type like `__Zerocopy_Field_a`. I assume this is a holdover from when it was a numerical const index?

zerocopy-derive/src/lib.rs
```diff
@@ -266,11 +266,48 @@ fn derive_known_layout_inner(ast: &DeriveInput, _top_level: Trait) -> Result<Tok
                 Default::default()
             };
 
+            let field_index =
+                |name| Ident::new(&format!("__Zerocopy_Field_{}", name), ident.span());
```
joshlf, line 270: Same comment here and below about the nomenclature "index". If you rename it in the `Field` trait, make sure to similarly rename it here.

zerocopy-derive/src/lib.rs
```diff
@@ -279,7 +316,11 @@ fn derive_known_layout_inner(ast: &DeriveInput, _top_level: Trait) -> Result<Tok
                 #repr
                 #[doc(hidden)]
                 #vis struct __ZerocopyKnownLayoutMaybeUninit<#params> (
-                    #(::zerocopy::util::macro_util::core_reexport::mem::MaybeUninit<#leading_fields_tys>,)*
+                    #(::zerocopy::util::macro_util::core_reexport::mem::MaybeUninit<
+                        <#ident #ty_generics as
+                            ::zerocopy::util::macro_util::Field<#leading_field_indices>
+                        >::Type
+                    >,)*
```
joshlf, line 323: Comment here explaining why we indirect via `Field` rather than just using the field types directly?

zerocopy-derive/tests/struct_known_layout.rs
```diff
@@ -46,16 +48,37 @@ util_assert_impl_all!(TypeParams<'static, (), imp::IntoIter<()>>: imp::KnownLayo
...
+        [T; N],
+        imp::PhantomData<&'a &'b ()>,
+    )
+    where
+        'a: 'b,
+        'b: 'a,
+        T: 'a + 'b + imp::KnownLayout;
+
+    util_assert_impl_all!(WithParams<'static, 'static, u8, 42>: imp::KnownLayout);
+};
```
joshlf, line 69: Also add a test which is similar but uses static lifetimes and is compiled on all toolchains?

zerocopy-derive/tests/struct_known_layout.rs
```diff
@@ -46,16 +48,37 @@ util_assert_impl_all!(TypeParams<'static, (), imp::IntoIter<()>>: imp::KnownLayo
...
+
+// Deriving `KnownLayout` should work if the struct references `Self`
```
joshlf, line 71: Mention the relevant issue number?

src/util/macro_util.rs
jswrenn: It's an index in the most general sense — a label used for lookups. When we use `Field` to project into enums (whose fields cannot have visibility modifiers), we'll use a single `#vis struct Index<const N: usize>` type.

zerocopy-derive/tests/struct_known_layout.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/c7cfc8df03d86f6f93d73c8bc37729ad12eebd96..2cfcc5a8c3baf2a66dbb18fe633ba234a411a155#diff-85b159abf94564dfae9563267afd74a8b55832be2174e5cfef9522c1a8b6955cR89-R90)

zerocopy-derive/tests/struct_known_layout.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/c7cfc8df03d86f6f93d73c8bc37729ad12eebd96..2cfcc5a8c3baf2a66dbb18fe633ba234a411a155#diff-85b159abf94564dfae9563267afd74a8b55832be2174e5cfef9522c1a8b6955cR71-R87)

zerocopy-derive/src/lib.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/c7cfc8df03d86f6f93d73c8bc37729ad12eebd96..2cfcc5a8c3baf2a66dbb18fe633ba234a411a155#diff-977c29e6cf079bd7307f811df24f99633df603962785c583145eb6250b0e12daR318-R320)

zerocopy-derive/src/lib.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/c7cfc8df03d86f6f93d73c8bc37729ad12eebd96..2cfcc5a8c3baf2a66dbb18fe633ba234a411a155#diff-977c29e6cf079bd7307f811df24f99633df603962785c583145eb6250b0e12daR269-R270)

src/util/macro_util.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/c7cfc8df03d86f6f93d73c8bc37729ad12eebd96..2cfcc5a8c3baf2a66dbb18fe633ba234a411a155#diff-2abf6454394ca2a0779f4bf7cc0631f23cf1dc484566b5937e8a1ad9a5b9e3e8R37-R39)

zerocopy-derive/src/lib.rs
joshlf: Also mention #2116 in this comment?

joshlf (approved):
One more change requested, but otherwise LGTM.

zerocopy-derive/src/lib.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/2cfcc5a8c3baf2a66dbb18fe633ba234a411a155..9f787224d7aa909c7c030422ded4f8495cacc35e)

PR #2140, opened 2024-12-08
Release the ownership of the src parameter in read_from_io.

lethargy123:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4527,7 +4527,7 @@ pub unsafe trait FromBytes: FromZeros {
     /// ```
     #[cfg(feature = "std")]
     #[inline(always)]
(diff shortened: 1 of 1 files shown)


jswrenn:
As you can see from the failing tests, this change will break code that depends on [the impl of `Read` for `&[u8]`](https://doc.rust-lang.org/1.83.0/std/io/trait.Read.html#impl-Read-for-%26%5Bu8%5D). If you don't want `read_from_io` to consume ownership of its argument, you can use the [impl of `Read` for `&mut impl Read`](https://doc.rust-lang.org/1.83.0/std/io/trait.Read.html#impl-Read-for-%26mut+R) by changing `read_from_io(arg)` to `read_from_io(&mut arg)` at your callsite.

PR #2159, opened 2024-12-14
[derive] Implement a IntoBytes-based Hash derive

### LINKED ISSUE #2075: Add fast, byte-oriented `Hash` derive
For types that are `IntoBytes`, we could derive optimized implementations of `Hash`.

Adding `derive(Hash)` to a struct generates a recursive descent into the fields of the struct. Instead, for a type that is `IntoBytes`, we could relatively easily have `derive(zerocopy::Hash)` expand to something like this:

```rust
impl core::hash::Hash for UserTy
where
    UserTy: ::zerocopy::IntoBytes,
{
    fn hash<H: core::hash::Hasher>(&self, state: &mut H) {
        core::hash::Hasher::write(
            state,
            ::zerocopy::IntoBytes::as_bytes(self)
        )
    }

    fn hash_slice<H: core::hash::Hasher>(data: &[Self], state: &mut H)
    where
        Self: Sized,
    {
        core::hash::Hasher::write(
            state,
            ::zerocopy::IntoBytes::as_bytes(data)
        )
    }
}
```

Alternatively, if we run with the assumption that `Hasher`s benefit from writing the greatest number of bytes at a time, we can leverage the known size and well-alignedness of the user type to emit calls to `state.write_u128`, `state.write_u64`, `state.write_u32`, and so on. A trailing DST suffix, if any, can be serialized with `state.write(trailing.as_bytes())`.

We should benchmark these approaches. I don't have a strong intuition for the performance characteristics of Rust's leading `Hasher` implementations. My guess is that we couldn't do better in the more complicated approach than a well optimized `Hasher::write` could do, so we're best off doing the simple thing.

max-heller:
The standard library's derive for `Hash` generates a recursive descent into the fields of the type it is applied to. This commit adds an alternative derive that generates an optimized, byte-oriented `Hash` implementation for types that implement `IntoBytes`. Instead of a recursive descent, the generated implementation makes a single call to `Hasher::write()` in both `Hash::hash()` and `Hash::hash_slice()`, feeding the hasher the bytes of the type or slice all at once.

Resolves #2075

### DIFF
--- a/zerocopy-derive/src/lib.rs
+++ b/zerocopy-derive/src/lib.rs
@@ -107,6 +107,7 @@ derive!(FromZeros => derive_from_zeros => derive_from_zeros_inner);
 derive!(FromBytes => derive_from_bytes => derive_from_bytes_inner);
 derive!(IntoBytes => derive_into_bytes => derive_into_bytes_inner);
 derive!(Unaligned => derive_unaligned => derive_unaligned_inner);
+derive!(ByteHash => derive_hash => derive_hash_inner);
 
 /// Deprecated: prefer [`FromZeros`] instead.
 #[deprecated(since = "0.8.0", note = "`FromZeroes` was renamed to `FromZeros`")]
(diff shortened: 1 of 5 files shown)


jswrenn:
Thank you for the contribution! I'm in awe you managed to pass our CI on your first commit! At a glance, this looks great; I'll give a more thorough review tomorrow.

I think the *right* choice is to advocate for the usage pattern of `derive(zerocopy::Hash)`, and thus preserve the property that `derive(Trait)` derives `Trait`. However, we need to be mindful that this could conflict with other imports of `core::hash::Hash` if the user writes `use zerocopy_derive::*`. 

The `Hash` macro (but not trait) is in the prelude. Would a subsequent glob import from `zerocopy_derive` overwrite it?

jswrenn:
Unfortunately, this can lead to errors. This code, for example:
```rust
use zerocopy::*;

#[derive(Hash)]
struct Foo;
```
...produces this error:
```
error[E0659]: `Hash` is ambiguous
   --> examples/foo.rs:3:10
    |
3   | #[derive(Hash)]
    |          ^^^^ ambiguous name
    |
    = note: ambiguous because of a conflict between a name from a glob import and an outer scope during import or macro resolution
note: `Hash` could refer to the derive macro imported here
   --> examples/foo.rs:1:5
    |
1   | use zerocopy::*;
    |     ^^^^^^^^^^^
    = help: consider adding an explicit import of `Hash` to disambiguate
    = help: or use `crate::Hash` to refer to this derive macro unambiguously
note: `Hash` could also refer to the derive macro defined here
   --> /home/ubuntu/.rustup/toolchains/nightly-2024-11-06-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/prelude/mod.rs:151:13
    |
151 |     pub use super::v1::*;
```

Unfortunately, we encourage users to glob-import from `zerocopy_derive`.  Given this, we should probably follow bytemuck's lead and call it something like `ByteHash`.

zerocopy-derive/src/lib.rs
```diff
@@ -528,6 +529,45 @@ fn derive_unaligned_inner(ast: &DeriveInput, _top_level: Trait) -> Result<TokenS
     }
 }
 
+fn derive_hash_inner(ast: &DeriveInput, _top_level: Trait) -> Result<TokenStream, Error> {
+    let type_ident = &ast.ident;
+    let (impl_generics, ty_generics, where_clause) = ast.generics.split_for_impl();
+    let where_predicates = where_clause.map(|clause| &clause.predicates);
+    Ok(quote! {
```
jswrenn, line 536: One last (I think) nit: Could you leave a comment why deferring to `impl_block` isn't appropriate here?

max-heller: Done!

jswrenn (approved):
This looks great to me! Thank you for the contribution!

PR #2182, opened 2024-12-23
Silence `private_bounds` lint in `KnownLayout` derive

### LINKED ISSUE #2177: Inconsistent visibility on some derive-synthesized types
Prompted by [this CI failure](https://github.com/rkyv/rend/actions/runs/12429476642/job/34703051308#step:4:1) in rend.

jswrenn:
For reasons not-yet-known, the `private_bounds` lint flags the output of `derive(KnownLayout)` on `repr(C)` structs generated by macros. This is likely a rustc bug, but to preserve a good experience for our users, we nonetheless, `allow(private_bounds)`.

Fixes #2177


joshlf:
Can we add a test for this?

jswrenn:
Hm. It's not straightforward to add it as a trybuild test, because trybuild expects compilation failures (which we avoid by `allow`ing the lint in `derive(KnownLayout)`). It's not straightforward to add it as a regular integration test, since those tests don't fail on warnings.

And we can't meaningfully pull shenanigans with `forbid(private_bounds)`, because rustc errors out early as soon as it sees the `allow(private_bounds)`, rather than upon the actual triggering of the lint. [`force-warn`](https://doc.rust-lang.org/rustc/lints/levels.html#force-warn) might do the trick, but it doesn't exist in attribute form, and trybuild doesn't have a `ui-test`-style way of specifying per-file rustflags.

My preference would be to merge this PR and cut a release sooner, rather than later, then put it on our roadmap to seek a general solution (or, at least, policy) for the problem of "our consumers might deny lints that trigger in certain invocations of our derives`.

joshlf (approved):
Approved, but let's file an issue to track either fixing the rustc bug or adding a test to prevent regressions.

PR #2229, opened 2025-01-12
Enforce `Src: FromBytes` in `try_transmute_mut!`

### LINKED ISSUE #2226: `try_transmute_mut`, `TryFromBytes::*mut*` soundness issue
## Progress

- [x] Add `Src: FromBytes` bound to `try_transmute_mut!`
- [x] Release new crate version (0.8.16)
- [x] Add `Self: IntoBytes` bound to `TryFromBytes::*mut*`
   - For why this is required, consider that a `MaybeUninit<u8>` is `TryFromBytes`, but permits writing uninitialized bytes that would invalidated the shadowed `src` reference.
- [x] Release new crate version (0.8.18)
- [ ] Do deeper surgery to make `try_cast_or_pme` sound and `TryFromBytes::*mut*` sound
   - #1866

## Original text

Using zerocopy 0.8.13:

```
use zerocopy::{TryFromBytes, IntoBytes, KnownLayout, Immutable, try_transmute_mut};

#[derive(TryFromBytes, IntoBytes, KnownLayout, Immutable)]
struct T {
    f: bool,
}

fn main() {
    let mut t = T { f: false };
    let slice: &mut [u8; 1] = try_transmute_mut!(&mut t).unwrap();

    slice[0] = u8::MAX;
    println!("f: {}", t.f);
}
```

`cargo +nightly miri run` caught an UB:

```
error: Undefined Behavior: constructing invalid value: encountered 0xff, but expected a boolean
    --> /home/zeling/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/fmt/mod.rs:2682:25
     |
2682 |         Display::fmt(if *self { "true" } else { "false" }, f)
     |                         ^^^^^ constructing invalid value: encountered 0xff, but expected a boolean
     |
     = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
     = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
     = note: BACKTRACE:
....
     |
13   |     println!("f: {}", t.f);
     |     ^^^^^^^^^^^^^^^^^^^^^^
     = note: this error originates in the macro `println` (in Nightly builds, run with -Z macro-backtrace for more info)

note: some details are omitted, run with `MIRIFLAGS=-Zmiri-backtrace=full` for a verbose backtrace

error: aborting due to 1 previous error
```

try_transmute_mut may not be a safe API that can be exposed because it only does validation when creating the reference but the user is free to write whatever bit pattern to the created reference.

jswrenn:
Ensures that the source reference remains valid after the transmuted (and possibly mutated)  destination is dropped.

Makes progress on #2226

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -577,7 +577,7 @@ macro_rules! try_transmute_ref {
 /// ```ignore
 /// fn try_transmute_mut<Src, Dst>(src: &mut Src) -> Result<&mut Dst, ValidityError<&mut Src, Dst>>
 /// where
-///     Src: IntoBytes,
+///     Src: FromBytes + IntoBytes,
 ///     Dst: TryFromBytes,
 ///     size_of::<Src>() == size_of::<Dst>(),
 ///     align_of::<Src>() >= align_of::<Dst>(),
@@ -888,9 +888,9 @@ mod tests {
     #[test]
     fn test_try_transmute_mut() {
         // Test that memory is transmuted with `try_transmute_mut` as expected.
-        let array_of_bools = &mut [false, true, false, true, false, true, false, true];
+        let array_of_u8s = &mut [0u8, 1, 0, 1, 0, 1, 0, 1];
         let array_of_arrays = &mut [[0u8, 1], [0, 1], [0, 1], [0, 1]];
-        let x: Result<&mut [[u8; 2]; 4], _> = try_transmute_mut!(array_of_bools);
+        let x: Result<&mut [[u8; 2]; 4], _> = try_transmute_mut!(array_of_u8s);
         assert_eq!(x, Ok(array_of_arrays));
 
         let array_of_bools = &mut [false, true, false, true, false, true, false, true];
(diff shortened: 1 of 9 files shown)


jswrenn:
(Have a killer headache tonight; I'll finish this up first thing tomorrow.)

src/macros.rs
```diff
@@ -577,7 +577,7 @@ macro_rules! try_transmute_ref {
 /// ```ignore
 /// fn try_transmute_mut<Src, Dst>(src: &mut Src) -> Result<&mut Dst, ValidityError<&mut Src, Dst>>
 /// where
-///     Src: IntoBytes,
+///     Src: FromByte + IntoBytes,
```
joshlf:
```suggestion
///     Src: FromBytes + IntoBytes,
```

src/macros.rs
```diff
@@ -888,11 +888,6 @@ mod tests {
     #[test]
     fn test_try_transmute_mut() {
         // Test that memory is transmuted with `try_transmute_mut` as expected.
-        let array_of_bools = &mut [false, true, false, true, false, true, false, true];
-        let array_of_arrays = &mut [[0u8, 1], [0, 1], [0, 1], [0, 1]];
-        let x: Result<&mut [[u8; 2]; 4], _> = try_transmute_mut!(array_of_bools);
-        assert_eq!(x, Ok(array_of_arrays));
```
joshlf, line 894: Can we keep this test and use `u8` instead of `bool`?

joshlf (changes requested):

src/macros.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/4fdd64642539808c5eedb73218db22d3539c60a9..cc95f58a14b3d8f2e130e86f80e04dfcba787938)

joshlf (approved):

joshlf:
Backporting in #2343

joshlf:
Backporting in #2343

PR #2250, opened 2025-01-21
Implement KnownLayout for f16 and f128

usamoi:


joshlf (approved):
Looks great, thanks for adding this!

PR #2285, opened 2025-02-01
[derive] Implement an IntoBytes-based PartialEq/Eq derive

### LINKED ISSUE #2274: `derive(ByteEq)`
Occasionally, replacing a `derive(PartialEq)` with a manual impl that compares the bytes of the types directly results in smaller code. A derive macro `ByteEq` would make this more ergonomic. ~This also applies to `Hash`~. Inspired by [bytemuck](https://docs.rs/bytemuck/latest/bytemuck/derive.ByteEq.html).

`derive(ByteEq)` should impl both `PartialEq` and `Eq`.

max-heller:
The standard library's derive for `PartialEq` generates a recursive descent into the fields of the type it is applied to. This commit adds a `ByteEq` derive that generates an optimized, byte-oriented `PartialEq` implementation -- and a corresponding `Eq` implementation -- for types that implement `IntoBytes`. Instead of a recursive descent, the generated implementation performs a single slice comparison of the bytes of the two values being compared.

Closes #2274

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5456,6 +5456,17 @@ pub unsafe trait Unaligned {
 #[cfg_attr(doc_cfg, doc(cfg(feature = "derive")))]
 pub use zerocopy_derive::ByteHash;
 
+/// Derives an optimized implementation of [`PartialEq`] and [`Eq`] for types
+/// that implement [`IntoBytes`] and [`Immutable`].
+///
+/// The standard library's derive for [`PartialEq`] generates a recursive
+///descent into the fields of the type it is applied to. Instead, the
+///implementation derived by this macro performs a single slice comparison of
+///the bytes of the two values being compared.
+#[cfg(any(feature = "derive", test))]
+#[cfg_attr(doc_cfg, doc(cfg(feature = "derive")))]
+pub use zerocopy_derive::ByteEq;
+
 #[cfg(feature = "alloc")]
 #[cfg_attr(doc_cfg, doc(cfg(feature = "alloc")))]
 #[cfg(zerocopy_panic_in_const_and_vec_try_reserve_1_57_0)]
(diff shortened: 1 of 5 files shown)


zerocopy-derive/tests/include.rs
```diff
@@ -26,7 +26,7 @@ mod imp {
     #[allow(unused)]
     pub use {
         ::core::{
-            assert_eq,
+            assert_eq, assert_ne,
```
jswrenn, line 29: Did rustfmt perhaps miss this file? My intuition is that these two names should be on different lines.

jswrenn (changes requested):

zerocopy-derive/tests/include.rs
max-heller: Surprised me too but that's how it seems to autoformat

max-heller: (If you put them on different lines it merges them)

jswrenn (approved):
Thanks!

src/lib.rs
```diff
@@ -5456,6 +5456,17 @@ pub unsafe trait Unaligned {
 #[cfg_attr(doc_cfg, doc(cfg(feature = "derive")))]
 pub use zerocopy_derive::ByteHash;
 
+/// Derives an optimized implementation of [`PartialEq`] and [`Eq`] for types
+/// that implement [`IntoBytes`] and [`Immutable`].
+///
+/// The standard library's derive for [`PartialEq`] generates a recursive
+///descent into the fields of the type it is applied to. Instead, the
+///implementation derived by this macro performs a single slice comparison of
+///the bytes of the two values being compared.
```
joshlf, line 5465:
```suggestion
/// descent into the fields of the type it is applied to. Instead, the
/// implementation derived by this macro performs a single slice comparison of
/// the bytes of the two values being compared.
```

joshlf (changes requested):
Awesome, thanks for doing this!

src/lib.rs
jswrenn, line 1: I'm surprised that rustfmt didn't catch this — something to dig into, I think.

jswrenn (approved):

joshlf (approved):

PR #2301, opened 2025-02-05
[pointer][invariant] Remove AliasingMapping, Inaccessible

joshlf:
We previously used `AliasingMapping`s and `Inaccessible` to model
`UnsafeCell` agreement. This abuses the notion of a mapping since one
doesn't ever actually want to change the aliasing of a pointer (and
certainly not to `Inaccessible`) - really this was meant to model
pointer casts which should never be performed. In addition to being an
awkward fit, the presence of `Inaccessible` meant that code could not
assume that any `Aliasing` invariant permitted reading, and so we had to
add extra machinery to work around this.

Future commits will use a different, simpler model for denoting
`UnsafeCell` agreement or disagreement.

While we're here, make `Read` slightly more permissive, implemented for
`A: Aliasing, T: Immutable` rather than just `A: Reference, T:
Immutable`.

Makes progress on #1122, #1866

---

This PR is on branch [ptr-overhaul-more-generic-issue-1945](../tree/ptr-overhaul-more-generic-issue-1945).

- #1970
- #1969
- #1911
- #2301

### DIFF
--- a/src/pointer/invariant.rs
+++ b/src/pointer/invariant.rs
@@ -55,6 +55,11 @@ impl<A: Aliasing, AA: Alignment, V: Validity> Invariants for (A, AA, V) {
 }
 
 /// The aliasing invariant of a [`Ptr`][super::Ptr].
+///
+/// All aliasing invariants must permit reading from the bytes of a pointer's
+/// referent which are not covered by [`UnsafeCell`]s.
+///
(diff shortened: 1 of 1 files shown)


jswrenn (approved):

joshlf:
Backporting in #2386

PR #2306, opened 2025-02-06
Improve consistency of `derive(Byte*)` documentation

jswrenn:
This edit brings the structure of these docs in line with that used for our other derives.


src/lib.rs
```diff
@@ -5443,14 +5442,36 @@ pub unsafe trait Unaligned {
...
+///     ...
+/// # */
+/// }
+/// ```
+///
+/// The standard library's [`derive(Hash)`][derive@Hash] produces hashes by
+/// individually hashing each field and combining the results. Instead, the
+/// implementations of [`Hash::hash()`] and [`Hash::hash_slice()`] generated by
+/// `derive(ByteHash)` convert the entirey of `self` to a byte slice and hashes
+/// it in a single call to [`Hasher::write()`].
```
joshlf, line 5474: Maybe mention something about how this is often faster or easier to optimize or something to that effect? After all, that's the only reason you'd use this derive. Currently the only mention of optimization or performance in this doc comment is in the summary sentence.

src/lib.rs
```diff
@@ -5459,13 +5480,35 @@ pub unsafe trait Unaligned {
...
+/// ```
 ///
-/// The standard library's derive for [`PartialEq`] generates a recursive
-/// descent into the fields of the type it is applied to. Instead, the
-/// implementation derived by this macro performs a single slice comparison of
-/// the bytes of the two values being compared.
+/// The standard library's [`derive(Eq, PartialEq)`][derive@PartialEq] computes
+/// equality by individually comparing each field. Instead, the implementation
+/// of [`PartialEq::eq`] emitted by `derive(ByteHash)` converts the entirey of
+/// `self` and `other` to byte slices and compares those slices for equality.
```
joshlf, line 5511: Same here re: mentioning performance

joshlf (changes requested):

src/lib.rs
jswrenn: The "only mention" being in the summary sentence is fairly prominent positioning. I'm hesitant to over-promise here, because this derive *isn't* necessarily a performance slam dunk. For example, comparing two `u64`s for equality is likely to optimize better than casting them to slices of dynamic length and comparing them byte-by-byte. I'm really not sure where the inflection point is. 

For that reason, I think the honest thing to do is to explain the mechanics, but not make specific promises about performance.

jswrenn: https://github.com/google/zerocopy/compare/f10667ea4dbd1a8d7802c1bbf09ba3d23bc81b08..a9faae2c1b9b5e45a02d52c24a9e984b44aa12b4

src/lib.rs
jswrenn: https://github.com/google/zerocopy/compare/f10667ea4dbd1a8d7802c1bbf09ba3d23bc81b08..a9faae2c1b9b5e45a02d52c24a9e984b44aa12b4

joshlf (approved):

Issue #2319, opened 2025-02-07
Transmuting `MaybeUninit::zeroed()` to `&mut [u8]` is unsound


kupiakos:
These days, `MaybeUninit::zeroed` is [pretty clear](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#method.zeroed) that you cannot expect all padding bytes to be zero in the output value (same for [`mem::zeroed`](https://doc.rust-lang.org/std/mem/fn.zeroed.html)). Zerocopy [mentions this](https://docs.rs/zerocopy/latest/zerocopy/trait.FromZeros.html#warning-padding-bytes), but it doesn't seem fully internalized by the library:

- `FromZeros::zero` contradicts this, stating that it's similar to `*self = Self::new_zeroed()` and that the difference is it doesn't semantically drop anything. This isn't the only difference! `zero` is capable of fully zeroing all of the bytes of an object until it is moved into, while `*self = Self::new_zeroed()` doesn't guarantee you can then soundly access as a byte slice unless that `Self: IntoBytes`.
- `FromZeros::new_zeroed` doesn't mention the same caveat as `MaybeUninit::zeroed` at all even though it applies
- The [current implementation](https://github.com/google/zerocopy/blob/17e7e4d98c9d2cd49c94307e2276b7ce377aa74a/src/lib.rs#L4551) of `FromBytes::read_from_io` is unsound, since there may still be padding bytes for `!IntoBytes` types from `new_zeroed`. It should be constructing from `MaybeUninit::uninit()`, accessing as `*mut u8`, then [zeroing its bytes](https://doc.rust-lang.org/std/ptr/fn.write_bytes.html) before casting to `&mut [u8]`.

PR #2320, opened 2025-02-07
Fix soundness of FromBytes::read_from_io

kupiakos:
See #2319.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4546,9 +4546,17 @@ pub unsafe trait FromBytes: FromZeros {
         Self: Sized,
         R: io::Read,
     {
-        let mut buf = CoreMaybeUninit::<Self>::zeroed();
+        // NOTE(#2319, #2320): We do `buf.zero()` separately rather than
+        // constructing `let buf = CoreMaybeUninit::zeroed()` because, if `Self`
+        // contains padding bytes, then a typed copy of `CoreMaybeUninit<Self>`
+        // will not necessarily preserve zeros written to those padding byte
+        // locations, and so `buf` could contain uninitialized bytes.
+        let mut buf = CoreMaybeUninit::<Self>::uninit();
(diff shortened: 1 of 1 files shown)


kupiakos:
I've confirmed that the new test fails Miri when run without the change in `read_from_io`

jswrenn (approved):
This looks reasonable to me. We really ought to update `FromZeros::new_zeroed` to match the disclaimers of `MaybeUninit::zeroed`, but that's a distinct issue from the one you're fixing and can be handled in a follow-up PR.

src/lib.rs
```diff
@@ -4546,7 +4546,8 @@ pub unsafe trait FromBytes: FromZeros {
         Self: Sized,
         R: io::Read,
     {
-        let mut buf = CoreMaybeUninit::<Self>::zeroed();
+        let mut buf = CoreMaybeUninit::<Self>::uninit();
+        buf.zero();
         let ptr = Ptr::from_mut(&mut buf);
         // SAFETY: `buf` consists entirely of initialized, zeroed bytes.
```
joshlf, line 4552: Maybe update this safety comment to explain why `buf.zero()` is necessary?

src/lib.rs
```diff
@@ -5905,6 +5906,29 @@ mod tests {
         assert_eq!(bytes, want);
     }
 
+    #[test]
+    #[cfg(feature = "std")]
+    fn test_read_io_with_padding_soundness() {
```
joshlf, line 5911: Could you add a brief comment here explaining what this test does?

joshlf (changes requested):
Awesome, thank you for catching this! Two small comment nits, but otherwise looks good.

joshlf (approved):

joshlf:
Backporting to 0.8 in #2358

PR #2351, opened 2025-02-18
[pointer] Fix Ptr[Inner] variance

joshlf:
Previously, `Ptr<'a, T>` and `PtrInner<'a, T>` documented themselves to
be covariant in both `'a` and `T`. This was true for `PtrInner`, but not
for `Ptr`, which used GATs, which are invariant. This is also not the
desired variance: for `Exclusive` aliasing, the desired variance matches
that of `&mut` references - namely, covariant in `'a` but invariant in
`T`.

This commit fixes this by making `Ptr<'a, T>` and `PtrInner<'a, T>`
unconditionally covariant in `'a` and invariant in `T`.

---

This PR is on branch [transmute-from](../tree/transmute-from).

- #2335
- #2351

### DIFF
--- a/src/pointer/inner.rs
+++ b/src/pointer/inner.rs
@@ -22,7 +22,7 @@ mod _def {
     use super::*;
     /// The inner pointer stored inside a [`Ptr`][crate::Ptr].
     ///
-    /// `Ptr<'a, T>` is [covariant] in `'a` and `T`.
+    /// `PtrInner<'a, T>` is [covariant] in `'a` and invariant in `T`.
     ///
     /// [covariant]: https://doc.rust-lang.org/reference/subtyping.html
     pub(crate) struct PtrInner<'a, T>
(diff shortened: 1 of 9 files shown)


src/pointer/inner.rs
```diff
@@ -22,7 +22,7 @@ mod _def {
     use super::*;
     /// The inner pointer stored inside a [`Ptr`][crate::Ptr].
     ///
-    /// `Ptr<'a, T>` is [covariant] in `'a` and `T`.
+    /// `PtrInner<'a, T>` is [covariant] in `'a` and invariant in `T`.
```
joshlf, line 25: We should probably create a test for this (and for `Ptr`). @jswrenn any ideas?

jswrenn: Morally, we want the variance to be that of a &'a mut T, not &'a T, regardless of aliasing (unfortunately). So we want two tests:
 - for exclusive aliasing, we want an example that fails to compile, but would succeed if we used &T variance
 - for shared aliasing, we want an example that fails to compiled, but would succeed if we used &T variance

joshlf: Done.

joshlf (commented):
Tests look good, thanks for adding!

jswrenn (approved):

joshlf:
Backporting in #2393

PR #2369, opened 2025-02-20
Add `FromBytes::zero_and_get_bytes`

### LINKED ISSUE #2318: Accessing `&mut MaybeUninit<T>` as `&mut [u8]` by zeroing
I have a `&mut MaybeUninit<T>`. `T: FromBytes` and is very large. I want to initialize it with data from an `io::Read::read` method in a closure and output the initialized `&mut T`.

Today, my options are:

- `FromBytes::read_from_io`, unpack the `Result`, and `write` that to the `&mut MaybeUninit<T>`. However, I've already proven that return-value optimization doesn't optimize away the temporary objects and memory copies, which is why I'm using a `&mut MaybeUninit<T>` in the first place. We're trying to do zero copies after all 😄 
- Use the `FromZeros::zero` method to zero out all of the bytes, and then `assume_init_mut().as_mut_bytes()` to pass into `read`. This requires using `unsafe`.

There are two improvements to `zerocopy` I can consider to solve this problem:

- Provide a zeroing method on `FromBytes` that outputs a `&mut [u8]`. This would be valuable for my above situation and for an extra reason: it allows users to get a `[u8]` out of a non-`IntoBytes` type by first ensuring all of the padding bytes are zeroed. While the `&mut` exists, the object cannot be overwritten with a `T` containing padding. 

- As part of #1797, add a `zero` method to the zerocopy `MaybeUninit` that outputs `&mut T` just like `write` does. That'd solve this particular problem.

joshlf:
Closes #2318

gherrit-pr-id: I38a568578de67f067fab936d17ba57332c6cdfed

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3453,6 +3453,44 @@ pub unsafe trait FromBytes: FromZeros {
     where
         Self: Sized;
 
+    /// Zeroes the bytes of `self` and returns them as a `&mut [u8]`.
+    ///
+    /// This permits accessing the bytes of a type which may not implement
+    /// [`IntoBytes`] by first zeroing any padding bytes which would be unsound
+    /// to expose via [`IntoBytes::as_bytes`].
+    ///
+    /// # Examples
+    ///
+    /// TODO
+    #[inline]
+    fn zero_and_get_bytes(&mut self) -> &mut [u8] {
+        self.zero();
+
+        let len = mem::size_of_val(self);
+        let data: *mut Self = self;
+        let data = data.cast::<u8>();
+
+        // SAFETY:
+        // - `data` is non-null by invariant on `&mut self`.
+        // - `data` is valid for reads and writes of `len` bytes:
+        //   - `data` refers to a block of `len` bytes within a single allocated
+        //     object by invariant on `&mut self`
+        //   - TODO: Others? https://doc.rust-lang.org/std/ptr/index.html#safety
+        // - Thanks to the preceding `self.zero()`, it is guaranteed that `data`
+        //   points to `len` bytes with the value `0u8`, which is a valid value
+        //   for `u8`
+        // - Because `self: &mut Self`, `self` is guaranteed to be the only live
+        //   reference to this data. Because `self` and the returned `&mut [u8]`
+        //   have the same lifetime, `self` is guaranteed not to be live so long
+        //   as the returned `&mut [u8]` is live, and thus it is the only
+        //   reference which is permitted to access the referent bytes for the
+        //   duration of its lifetime.
+        // - By invariant on `&mut self`, the referent is not larger than
+        //   `isize::MAX` bytes, and the referent does not wrap around the
+        //   address space
+        unsafe { core::slice::from_raw_parts_mut(data, len) }
+    }
+
     /// Interprets the given `source` as a `&Self`.
     ///
     /// This method attempts to return a reference to `source` interpreted as a


joshlf:
@kupiakos does this accomplish what you need for #2318?

src/lib.rs
```diff
@@ -3453,6 +3453,44 @@ pub unsafe trait FromBytes: FromZeros {
...
 
+    /// Zeroes the bytes of `self` and returns them as a `&mut [u8]`.
+    ///
+    /// This permits accessing the bytes of a type which may not implement
+    /// [`IntoBytes`] by first zeroing any padding bytes which would be unsound
+    /// to expose via [`IntoBytes::as_bytes`].
+    ///
+    /// # Examples
+    ///
+    /// TODO
```
joshlf, line 3464: TODO

src/lib.rs
```diff
@@ -3453,6 +3453,44 @@ pub unsafe trait FromBytes: FromZeros {
...
+        let len = mem::size_of_val(self);
+        let data: *mut Self = self;
+        let data = data.cast::<u8>();
+
+        // SAFETY:
+        // - `data` is non-null by invariant on `&mut self`.
+        // - `data` is valid for reads and writes of `len` bytes:
+        //   - `data` refers to a block of `len` bytes within a single allocated
+        //     object by invariant on `&mut self`
+        //   - TODO: Others? https://doc.rust-lang.org/std/ptr/index.html#safety
```
joshlf, line 3478: TODO

src/lib.rs
```diff
@@ -3453,6 +3453,44 @@ pub unsafe trait FromBytes: FromZeros {
...
+    ///
+    /// This permits accessing the bytes of a type which may not implement
+    /// [`IntoBytes`] by first zeroing any padding bytes which would be unsound
+    /// to expose via [`IntoBytes::as_bytes`].
+    ///
+    /// # Examples
+    ///
+    /// TODO
+    #[inline]
+    fn zero_and_get_bytes(&mut self) -> &mut [u8] {
```
joshlf, line 3466: TODO: Bikeshed name

src/lib.rs
```diff
@@ -3453,6 +3453,44 @@ pub unsafe trait FromBytes: FromZeros {
...
+        //   for `u8`
+        // - Because `self: &mut Self`, `self` is guaranteed to be the only live
+        //   reference to this data. Because `self` and the returned `&mut [u8]`
+        //   have the same lifetime, `self` is guaranteed not to be live so long
+        //   as the returned `&mut [u8]` is live, and thus it is the only
+        //   reference which is permitted to access the referent bytes for the
+        //   duration of its lifetime.
+        // - By invariant on `&mut self`, the referent is not larger than
+        //   `isize::MAX` bytes, and the referent does not wrap around the
+        //   address space
```
kupiakos, line 3490: nit: period consistency

src/lib.rs
kupiakos: My thoughts on the name:

- `zero_and_get_bytes` is a good and clear name, though its length is right on that border of being verbose.
- `mut` could be mentioned but is not necessary when the name `zero` implies mutation.
- `fn zero(&mut self)` on `FromZeros` could yield `&[u8]` of all zeros, and that could be given the same name. I struggle to consider a situation where it's practically useful to have access to an all-zeros `&[u8]` with the same address as some `&mut T`, so 🤷🏻‍♀️

kupiakos:
> @kupiakos does this accomplish what you need for #2318?

Yes, this is excellent! I'm glad to see it's DST-compatible as well.

PR #2393, opened 2025-02-25
[pointer] Fix Ptr[Inner] variance (#2351)

joshlf:
Previously, `Ptr<'a, T>` and `PtrInner<'a, T>` documented themselves to be covariant in both `'a` and `T`. This was true for `PtrInner`, but not for `Ptr`, which used GATs, which are invariant. This is also not the desired variance: for `Exclusive` aliasing, the desired variance matches that of `&mut` references - namely, covariant in `'a` but invariant in `T`.

This commit fixes this by making `Ptr<'a, T>` and `PtrInner<'a, T>` unconditionally covariant in `'a` and invariant in `T`.

gherrit-pr-id: I29f8429d9d7b14026313f030f8dc1e895a98ad56

### DIFF
--- a/src/pointer/inner.rs
+++ b/src/pointer/inner.rs
@@ -22,7 +22,7 @@ mod _def {
     use super::*;
     /// The inner pointer stored inside a [`Ptr`][crate::Ptr].
     ///
-    /// `Ptr<'a, T>` is [covariant] in `'a` and `T`.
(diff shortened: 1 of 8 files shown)


jswrenn (approved):

joshlf:
Backporting #2351

PR #2397, opened 2025-02-25
[pointer] Improve soundness of invariant modeling

joshlf:
This commit makes the following improvements:
- Removes the `Inaccessible` aliasing mode. This mode was not unsound,
  but it was unnecessary in practice.
- Replaces `Unknown` with `Unaligned` for alignment.
- Replaces `Unknown` with `Uninit` for validity.

Finally, with the exception of `transparent_wrapper_into_inner`, this
commit ensures that all `Ptr` methods which modify the type or validity
of a `Ptr` are sound.

Previously, we modeled validity as "knowledge" about the `Ptr`'s
referent (see #1866 for a more in-depth explanation). In particular, we
assumed that it was always sound to "forget" information about a `Ptr`'s
referent in the same way in which it is sound to "forget" that a `Ptr`
is validly-aligned, converting a `Ptr<T, (_, Aligned, _)>` to a
`Ptr<T, (_, Unaligned, _)>`.

The problem with this approach is that validity doesn't just specify
what bit patterns can be expected to be read from a `Ptr`'s referent, it
also specifies what bit patterns are permitted to be *written* to the
referent. Thus, "forgetting" about validity and expanding the set of
expected bit patterns also expands the set of bit patterns which can be
written.

Consider, for example, "forgetting" that a `Ptr<bool, (_, _, Valid)>` is
bit-valid, and downgrading it to a `Ptr<bool, (_, _, Initialized)>`. If
the aliasing is `Exclusive`, then the resulting `Ptr` would permit
writing arbitrary `u8`s to the referent, violating the bit validity of
the `bool`.

This commit moves us in the direction of a new model, in which changes
to the referent type or the validity of a `Ptr` must abide by the
following rules:
- If the resulting `Ptr` permits mutation of its referent (either via
  interior mutation or `Exclusive` aliasing), then the set of allowed
  bit patterns in the referent may not expand
- If the original `Ptr` permits mutation of its referent while the
  resulting `Ptr` is also live (i.e., via interior mutation on a
  `Shared` `Ptr`), then the set of allowed bit patterns in the referent
  may not shrink

This commit does not fix `transparent_wrapper_into_inner`, which will
require an overhaul or replacement of the `TransparentWrapper` trait.

Makes progress on #2226, #1866

---

This PR is on branch [v0.8.x-ptr-invariant-mapping](../tree/v0.8.x-ptr-invariant-mapping).

- #2397

### DIFF
--- a/src/util/macro_util.rs
+++ b/src/util/macro_util.rs
@@ -26,7 +26,7 @@ use core::ptr::{self, NonNull};
 
 use crate::{
     pointer::invariant::{self, BecauseExclusive, BecauseImmutable, Invariants, ReadReason},
-    FromBytes, Immutable, IntoBytes, Ptr, TryFromBytes, Unalign, ValidityError,
+    FromBytes, Immutable, IntoBytes, Ptr, TryFromBytes, ValidityError,
 };
 
 /// Projects the type of the field at `Index` in `Self`.
@@ -529,9 +529,14 @@ pub unsafe fn transmute_mut<'dst, 'src: 'dst, Src: 'src, Dst: 'dst>(
 ///
 /// # Safety
 ///
-/// Unsafe code may assume that, if `try_cast_or_pme(src)` returns `Some`,
+/// Unsafe code may assume that, if `try_cast_or_pme(src)` returns `Ok`,
 /// `*src` is a bit-valid instance of `Dst`, and that the size of `Src` is
 /// greater than or equal to the size of `Dst`.
+///
+/// Unsafe code may assume that, if `try_cast_or_pme(src)` returns `Err`, the
+/// encapsulated `Ptr` value is the original `src`. `try_cast_or_pme` cannot
+/// guarantee that the referent has not been modified, as it calls user-defined
+/// code (`TryFromBytes::is_bit_valid`).
 ///
 /// # Panics
 ///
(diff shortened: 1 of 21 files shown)


src/util/macro_util.rs
```diff
@@ -611,19 +611,40 @@ where
...
+
+    // SAFETY: Since `Src: IntoBytes`, and since `size_of::<Src>() ==
+    // size_of::<Dst>()` by the preceding assertion, all of `mu_dst`'s bytes are
+    // initialized.
+    let ptr = unsafe { ptr.assume_validity::<invariant::Initialized>() };
+
+    // SAFETY: `MaybeUninit<T>` and `T` have the same size [1], so this cast
+    // preserves the referent's size. This cast preserves provenance.
+    //
+    // [1] TODO
```
joshlf, line 636: TODO

src/util/macro_util.rs
```diff
@@ -611,19 +611,40 @@ where
...
+    let ptr = unsafe { ptr.cast_unsized(|mu: *mut mem::MaybeUninit<Dst>| mu.cast()) };
```
joshlf, line 637: Add explicit type here so it's clear from reading it what type we're casting into.

src/util/macro_util.rs
joshlf: Done

jswrenn (approved):
Reviewed in-person.

PR #2403, opened 2025-02-28
[pointer] Document validity safety invariant

joshlf:
Document the semantics of `Ptr` validity invariants on the `Validity`
trait.

TODO:
- [ ] Prove that this safety invariant is upheld at all transmute sites
- [ ] Consume this safety invariant wherever it's relevant

Makes progress on #2226, #1866

gherrit-pr-id: Ie66db9044be1dc310a6b7280a73652a357878376


joshlf:
This draft lays the groundwork for implementing the semantics described in the commit message of #2397. As of this writing, all I've added is a safety comment on the `Validity` trait. This is clearly insufficient on its own - in order to be completely sound, there must be places where this invariant is consumed. (It needs to be proven to be upheld as well, but that's more obvious.) My hunch is that this will need to be at places where we actually do something with the raw pointer inside a `Ptr` (read from it, convert it to a vanilla reference, etc).

src/pointer/invariant.rs
```diff
@@ -42,6 +42,46 @@ pub trait Aliasing: Sealed {
...
+///
+/// In this section, we will use `Ptr<T, V>` as a shorthand for `Ptr<T, I:
+/// Invariants<Validity = V>>` for brevity.
+///
+/// Each `V: Validity` defines a set of bit values which may appear in the
+/// referent of a `Ptr<T, V>`. Each `V: Validity`, in its documentation,
+/// provides a definition of this set in terms of any referent type (in other
+/// words, this definition holds for all `T: ?Sized`).
+///
+/// Let `S(T, V)` be the set of bit values permitted to appear in the referent
```
jswrenn, line 56: The set of things that can be in the referent of a `NonNull<T>` is different than the set of things that can be in a `T`. I think you mean the latter, not the former.

src/pointer/ptr.rs
```diff
@@ -971,8 +1016,13 @@ mod _casts {
...
             //    invariant of `I::Alignment` because `self` does.
-            // 8. By the above lemma, `slice` conforms to the validity invariant
-            //    of `I::Validity` because `self` does.
+            // 8. Since `[T; N]` and `[T]` have the same bit validity [1], and
+            //    since `self` and the returned `Ptr` have the same validity
+            //    invariant, neither `self` nor the returned `Ptr` can be used
+            //    to write a value to the referent which violates the other's
+            //    validity invariant.
+            //
+            // [1] TODO
```
joshlf, line 1025: TODO

jswrenn (approved):

Issue #1940, opened 2024-10-18
Teach generic transmutability to reason about size inequality and unsized types


joshlf:
We would like to support all combinations of the following axes
- Src type sized or unsized
- Dst type sized or unsized
- Src larger than, same size as, or smaller than Dst type for the same fat pointer metadata
  - Includes when Src is sized and Dst is unsized and vice-versa
- Dst same size as Src given a metadata fix-up operation

Some specific operations we could support:
- Sizes same w/o metadata fix-up: permits shared and exclusive reference casts
- For same metadata, Dst smaller than Src: supports shared reference casts
  - When both types are sized: supports by-value transmute
- For same metadata, Dst larger than Src: supports by-value transmute if uninit bytes are valid at the end of Dst
- Sizes same given metadata fix-up: shared and exclusive casts (with metadata fix-up)

PR #2408, opened 2025-02-28
[pointer] Support generic `TransmuteFrom` framework

### LINKED ISSUE #1226: Make `AliasingSafe` more powerful
Currently, `AliasingSafe` has the following safety invariants:

```
/// `U: AliasingSafe<T, A, R>` if either of the following conditions holds:
/// - `A` is [`Exclusive`]
/// - `T` and `U` both implement [`Immutable`]
```

However, we could relax this to:

```
/// `U: AliasingSafe<T, A, R>` if either of the following conditions holds:
/// - `A` is [`Exclusive`]
/// - `T` and `U` contain [`UnsafeCell`]s at the same locations
```

In particular, "`T` and `U` implement `Immutable`" is just a special case of " `T` and `U` contain `UnsafeCell`s at the same locations." If we did this, we could add a new `AliasingSafeReason` called `BecauseUnsafeCellsAgree` or something to that effect.

We could then piggy-back on our existing `TransparentWrapper` machinery to provide a blanket impl. In particular, wherever `W: TransparentWrapper<Inner = T, UnsafeCellVariance = Covariant>`, we can infer that `W: AliasingSafe<T, _, BecauseUnsafeCellsAgree>`.

Unfortunately, this construction is less powerful than we'd like. In particular, it is not true in the general case that `W: TransparentWrapper<Inner = T>` implies `T: TransparentWrapper<Inner = W>`, but `AliasingSafe` *is* reflexive in this way. In practice, this causes problems for `transparent_wrapper_into_inner`: If we try to add an `AliasingSafe` bound to `cast_unsized` (a natural thing to do, and it removes a safety precondition), `transparent_wrapper_into_inner`'s `TransparentWrapper` bound isn't sufficient to satisfy `cast_unsized`'s bound. This is prototyped [here](https://github.com/google/zerocopy/compare/aliasing-safe-transparent-wrapper-prototype). Try to build that code and see its compilation errors from `transparent_wrapper_into_inner`.

---

In order to work around this, we should split `TransparentWrapper` into multiple traits. Some of its properties - namely, size equality and `UnsafeCell` agreement - are reflexive, and can go in a separate trait which guarantees this reflexivity (and maybe encodes it in the type system somehow).

### LINKED ISSUE #1359: Generic transmutability
*See also: #1122*

Currently, we have a number of traits that express some aspect of transmutability:
- [`TransparentWrapper`](https://github.com/google/zerocopy/blob/76915d0978e68404d4e53181d462fcc341a71a39/src/util.rs#L45)
- [`AliasingSafe`](https://github.com/google/zerocopy/blob/76915d0978e68404d4e53181d462fcc341a71a39/src/pointer/aliasing_safety.rs#L27)

These, in turn, are used to bound certain `Ptr` transmutation methods, e.g.:
- [`transparent_wrapper_into_inner`](https://github.com/google/zerocopy/blob/76915d0978e68404d4e53181d462fcc341a71a39/src/pointer/ptr.rs#L560)
- [`as_bytes`](https://github.com/google/zerocopy/blob/76915d0978e68404d4e53181d462fcc341a71a39/src/pointer/ptr.rs#L1006)
- [`try_cast_into`](https://github.com/google/zerocopy/blob/76915d0978e68404d4e53181d462fcc341a71a39/src/pointer/ptr.rs#L1129)

Both `transparent_wrapper_into_inner` and `as_bytes` implement special cases of a generic `Ptr<T>` to `Ptr<U>` transmutation. This suggests that we could in theory support a more generic `Ptr::transmute` method, supported by a new `TransmutableFrom` trait. This trait would share some similarities with the [built-in transmutability trait](https://github.com/rust-lang/rust/issues/99571). However, as `TransparentWrapper`'s support for invariant variance shows, it might be *more* powerful than that trait in certain ways.

In its most basic form, this API would look something like:

```rust
unsafe trait TransmutableFrom<T> {}

impl<'a, T, A: Aliasing> Ptr<'a, T, (A, Aligned, Valid)> {
    fn transmute<U: TransmutableFrom<T>>(self) -> Ptr<'a, U, (A, Aligned, Valid)>
}
```

However, we will likely want to relax the `Aligned` and `Valid` requirements and post-conditions, and support a generic framework for mapping source invariants (on `T`) to destination invariants (on `U`) - essentially a generalization of what's currently supported with `TransparentWrapper`'s invariant variance concept.

There may be multiple overlapping reasons to support transmuting a particular pair of types. For example, we could imagine the following impls:

```rust
unsafe impl<T: IntoBytes, U: FromBytes> TransmutableFrom<T> for U {}
unsafe impl<T> TransmutableFrom<T> for MaybeUninit<T> {}
```

Currently, these would result in an impl conflict. If [`#[marker]` traits](https://github.com/rust-lang/rust/issues/29864) are stabilized, this won't be an issue - we can just mark `TransmutableFrom` as a `#[marker]` trait. Alternatively, we could use a dummy "reason" parameter to disambiguate, [as we do with `AliasingSafe` today](https://github.com/google/zerocopy/blob/76915d0978e68404d4e53181d462fcc341a71a39/src/pointer/aliasing_safety.rs#L27):

```rust
unsafe trait TransmutableFrom<T, R> {}

enum BecauseIntoBytesFromBytes {}
enum BecauseMaybeUninit {}

unsafe impl<T: IntoBytes, U: FromBytes> TransmutableFrom<T, BecauseIntoBytesFromBytes> for U {}
unsafe impl<T> TransmutableFrom<T> for MaybeUninit<T, BecauseMaybeUninit> {}
```

### LINKED ISSUE #1866: Overhaul `Ptr`'s validity invariant modeling

### LINKED ISSUE #1945: Support arbitrary invariant mappings
*This design is in progress in #1969.*

In https://github.com/google/zerocopy/pull/1896, we added support for "invariant mappings". For a given `I` (where `I = trait Alignment` or `I = trait Validity`), and for all `X: I`, a mapping maps `X` to some `Y: I`.

The idea behind these mappings is that some operations modify the invariants of a `Ptr`. Historically, we had to be conservative and return a new `Ptr` with a single, concrete invariant. For example, `Ptr::cast_unsized` casts to a `Ptr` of a different type, and so it must conservatively assume that the resulting validity is `Unknown`. With mappings, we can encode that, regardless of the types involved, an `Initialized` validity will always map to an `Initialized` validity (because a `Ptr` whose referent bytes are all initialized retains this invariant regardless of what type is has).

However, even invariant mapping is more limited than it could be. In full generality, some operations are valid over multiple possible invariant mappings.

Consider, for example, performing a `Ptr` transmute from `T` to `MaybeUninit<T>`. We can observe that *any* of the following mappings would be valid:
- `Unknown` -> `Unknown`, `AsInitialized`, `Valid`
- `Initialized` -> `Unknown`, `Initialized`, `AsInitialized`, `Valid`
- `AsInitialized` -> `Unknown`, `AsInitialized`, `Valid`
- `Valid` -> `Unknown`, `AsInitialized`, `Valid`

This suggests that we in fact want to encode a *family* of mappings. In this case, there are 3 * 4 * 3 * 3 = 108 valid mappings in this family!

Of course we don't actually want to encode 108 mappings. Not only would that be patently ridiculous, it would also be *too* specific. The real information we need is merely that a given *pair* of invariants constitutes a valid transition. For example, that it's valid to map `Valid` to `AsInitialized`.

Likely the most natural way to encode this is as a trait:

```rust
// Implemented for pairs of `(Type, Validity)`. This won't actually work because it doesn't
// support `Type: ?Sized`, but something similar would.
unsafe trait FromValidity<T: ?Sized, V: Validity> {}

unsafe impl<T> FromValidity<T, Valid> for (MaybeUninit<T>, AsInitialized) {}
// ...
```

This would permit us to support an API like:

```rust
impl<'a, T, I> Ptr<'a, T, I>
where
    T: 'a + ?Sized,
    I: Invariants,
{
    pub fn transmute<U, A, V>(
        self,
    ) -> Ptr<'a, U, (I::Aliasing, A, V)>
    where
        U: ?Sized + FromAlignment<T, I::Alignment> + FromValidity<T, I::Validity>
    { ... }
```

*This API as written might be unsound! We also need to make sure to handle bidirectional validity if the aliasing invariant permits writing to the returned `Ptr`. But the basic idea remains the same, so I haven't bothered to work out the details here.*

This is maximally powerful, but will also cause inference ambiguities in many cases, requiring the user to manually specify the desired invariants. It may be useful to retain the more conservative existing notion of a mapping in order to provide reasonable defaults.

### LINKED ISSUE #2226: `try_transmute_mut`, `TryFromBytes::*mut*` soundness issue

joshlf:
This commit removes the `TransparentWrapper` trait and the
`Ptr::transparent_wrapper_into_inner` method. It replaces them with a
new family of transmutation traits which encode more generic
transmutation (from any `T` to any `U`) and a set of `Ptr` methods which
use those traits to bound transmutation operations.

In particular:
- `Dst: TransmuteFrom<Src>` denotes that a by-value transmutation is
  sound
- `Dst: TryTransmuteFromPtr<Src>` denotes that a transmutation is sound
  so long as it can be guaranteed that the source is bit-valid for the
  destination; this is used by e.g. `Ptr::try_into_valid`, which
  performs runtime validation of bit validity
- `Dst: TransmuteFromPtr<Src>` is equivalent to `TransmuteFrom<Src> +
  TryTransmuteFromPtr<Src>`

Some type arguments are omitted in this summary. In particular, all
three traits also take validity invariant parameters for both the source
and destination types. Also, the `[Try]TransmuteFromPtr` traits take an
aliasing parameter.

In order to support these traits, we introduce a generalization of
`Read` known as `MutationCompatible`. `T: MutationCompatible<U, A>`
denotes that *either* `T: Read<A>` and `U: Read<A>` *or* `T` and `U`
have the same interior mutation semantics (formally, it is sound for
`&T` and `&U` to reference the same referent - safe code operating on
these references cannot cause undefined behavior). This is a refinement
of the "`UnsafeCell` agreement" concept that we have used before, but it
supports types which store but don't actually use `UnsafeCell`s. For
example, given a hypothetical `ReadOnly<T>`, the following bound holds:

  usize: MutationCompatible<ReadOnly<AtomicUsize>, Exclusive>

This commit also takes a different approach from the one originally
envisioned in #1945. In particular, it turns out that we don't need a
full type-level mapping concept. Instead, we need a *predicate* over
transitions to determine which ones are valid (e.g., it is valid to go
from a `Valid` `MaybeUninit<T>` to an `Uninit` `MaybeUninit<T>`). By
contrast, the invariant mapping concept suggests that each source
validity has *exactly one* destination validity.

This commit makes progress on #1940 by supporting unsized
transmutations, but we don't yet support size shrinking or expanding
transmutations.

This commit obsoletes #1359, as that issue was predicated upon the
existence of `TransparentWrapper`, which this commit removes.

This commit closes #1226, which suggests supporting `UnsafeCell`
agreement.

Closes #1945
Closes #1359
Closes #2226
Closes #1226
Closes #1866
Makes progress on #1359

---

This PR is on branch [ptr-validity](../tree/ptr-validity).

- #2408


joshlf:
I think we may have to handle `UnsafeCell` agreement after all. In particular, in order to implement `TryFromBytes` for atomics, we need to be able to go from e.g. `Ptr<AtomicU8>` to `Ptr<UnsafeCell<u8>>` so that we can call `<UnsafeCell<u8> as TryFromBytes>::is_bit_valid`. This will require care since this is unsound for shared aliasing. We know in practice that we'll always be using exclusive aliasing, but I'm not sure how to express that generically.

src/wrappers.rs
```diff
@@ -394,14 +394,14 @@ impl<T: Unaligned> Deref for Unalign<T> {
 
     #[inline(always)]
     fn deref(&self) -> &T {
-        Ptr::from_ref(self).transparent_wrapper_into_inner().bikeshed_recall_aligned().as_ref()
+        Ptr::from_ref(self).transmute().bikeshed_recall_aligned().as_ref()
```
joshlf, line 397: Bad news: This fails because it requires `T: KnownLayout`, which we wanted to avoid. We might be able to work around it with the `SizedKnownLayout` trick, or we might need to reconsider the `T: KnownLayout` bound.

src/pointer/ptr.rs
```diff
@@ -843,24 +818,24 @@ mod _transitions {
...
             I::Aliasing: Reference,
             I: Invariants<Validity = Initialized>,
         {
             // This call may panic. If that happens, it doesn't cause any soundness
             // issues, as we have not generated any invalid state which we need to
             // fix before returning.
             if T::is_bit_valid(self.reborrow().forget_aligned()) {
-                // SAFETY: If `T::is_bit_valid`, code may assume that `self`
-                // contains a bit-valid instance of `Self`.
+                // TODO: Complete this safety comment.
```
joshlf, line 835: TODO

src/pointer/ptr.rs
```diff
@@ -902,9 +877,10 @@ mod _casts {
         /// - `u` has the same provenance as `p`
         /// - If `I::Aliasing` is [`Shared`], `UnsafeCell`s in `*u` must exist
         ///   at ranges identical to those at which `UnsafeCell`s exist in `*p`
+        /// TODO: UnsafeCell compatibility
```
joshlf, line 880: TODO

src/pointer/ptr.rs
```diff
@@ -940,20 +916,29 @@ mod _casts {
...
+            //   - `I::Aliasing` is `Exclusive`
+            //   - `T` and `U` are both `Immutable`
+            // - `T` and `U` contain `UnsafeCell`s at identical locations
+            //
+            // In the first case, `I::Aliasing` is `Exclusive`, and in the
+            // second and third case, `T` and `U` contain `UnsafeCell`s at
+            // identical locations (in the second case, this is because `T` and
+            // `U` contain no `UnsafeCell`s at all).
+            //
+            // TODO: This should also promise UnsafeCell compatibility
```
joshlf, line 941: TODO

zerocopy-derive/src/lib.rs
```diff
@@ -662,8 +664,14 @@ fn derive_try_from_bytes_struct(
...
                     //   as they do in `*slf`
                     let field_candidate = unsafe {
-                        let project = |slf: *mut Self|
-                            ::zerocopy::util::macro_util::core_reexport::ptr::addr_of_mut!((*slf).#field_names);
+                        let project = |slf: core_reexport::ptr::NonNull<Self>| {
+                            let slf = slf.as_ptr();
+                            let field = core_reexport::ptr::addr_of_mut!((*slf).#field_names);
+                            // TODO: Safety comment. Needs precondition that the
+                            // original `slf` pointer doesn't wrap around
+                            // address space.
```
joshlf, line 672: TODO

zerocopy-derive/src/lib.rs
```diff
@@ -720,8 +730,14 @@ fn derive_try_from_bytes_union(
...
                     //   returned pointer's referent contain any `UnsafeCell`s
                     let field_candidate = unsafe {
-                        let project = |slf: *mut Self|
-                            ::zerocopy::util::macro_util::core_reexport::ptr::addr_of_mut!((*slf).#field_names);
+                        let project = |slf: core_reexport::ptr::NonNull<Self>| {
+                            let slf = slf.as_ptr();
+                            let field = core_reexport::ptr::addr_of_mut!((*slf).#field_names);
+                            // TODO: Safety comment. Needs precondition that the
+                            // original `slf` pointer doesn't wrap around
+                            // address space.
```
joshlf, line 738: TODO

src/pointer/transmute.rs
```diff
@@ -0,0 +1,312 @@
...
+///
+/// Given `Dst: TryTransmuteFromPtr<Src, A, _>`, callers may assume the
+/// following:
+///
+/// Given `src: Ptr<Src, SI>` where `SI: Invariants<Aliasing = A, Validity =
+/// SV>`, if the referent of `src` contains a `Dst` which conforms to the
+/// validity `DV`, then it is sound to transmute `src` into `dst: Ptr<Dst, DI>`
+/// whre `DI: Invariants<Aliasing = A, Validity = DV>`.
+///
+/// TODO: Mention alignment
```
joshlf, line 29: TODO

src/pointer/transmute.rs
```diff
@@ -0,0 +1,312 @@
...
+/// - Forwards transmutation: Any of the following hold:
+///     - So long as `dst` is active, no mutation of `dst`'s referent is allowed
+///       except via `dst` itself
+///     - The set of `DV`-valid `Dst`s is a superset of the set of `SV`-valid
+///       `Src`s
+/// - Reverse transmutation: Any of the following hold:
+///     - `dst` does not permit mutation of its referent
+///     - The set of `DV`-valid `Dst`s is a subset of the set of `SV`-valid
+///       `Src`s
+/// - Interior mutation: TODO (ie, at least one of `Exclusive` or `Immutable`
```
joshlf, line 46: TODO

src/pointer/transmute.rs
```diff
@@ -0,0 +1,312 @@
...
+///   required)
+///
+/// ## Proof
+///
+/// TODO: Prove that the pre-conditions imply the post-conditions.
```
joshlf, line 51: TODO

src/pointer/transmute.rs
```diff
@@ -0,0 +1,312 @@
...
+}
+
+unsafe impl<Src: ?Sized, Dst: ?Sized, A: Aliasing, SV: Validity, DV: Validity, R>
+    TransmuteFromPtr<Src, A, SV, DV, R> for Dst
+where
+    Dst: TransmuteFrom<Src, SV, DV> + TryTransmuteFromPtr<Src, A, SV, DV, R>,
+{
+}
+
+/// TODO
```
joshlf, line 131: TODO

src/pointer/transmute.rs
```diff
@@ -0,0 +1,312 @@
...
+where
+    Src: ?Sized,
+    Dst: FromBytes + ?Sized,
+    Src: SizeEq<Dst>,
+{
+}
+
+// TODO: This seems like a smell - the soundness of this bound has nothing to do
+// with `Src` or `Dst` - we're basically just saying `[u8; N]` is transmutable
+// into `[u8; N]`.
```
joshlf, line 186: TODO

src/pointer/transmute.rs
```diff
@@ -0,0 +1,312 @@
...
+where
+    Src: ?Sized,
+    Dst: ?Sized,
+    Src: SizeEq<Dst>,
+{
+}
+
+// TODO: This seems like a smell - the soundness of this bound has nothing to do
+// with `Src` or `Dst` - we're basically just saying `MaybeUninit<[u8; N]>` is
+// transmutable into `MaybeUninit<[u8; N]>`.
```
joshlf, line 202: TODO

src/impls.rs
```diff
@@ -468,10 +470,8 @@ mod atomics {
             unsafe_impl!(AtomicI8: Unaligned);
             assert_unaligned!(AtomicBool, AtomicU8, AtomicI8);
 
-            /// SAFETY:
-            /// All of these pass an atomic type and that type's native equivalent, as
-            /// required by the macro safety preconditions.
-            unsafe_impl_transparent_wrapper_for_atomic!(AtomicU8 [u8], AtomicI8 [i8], AtomicBool [bool]);
+            /// SAFETY: TODO
```
joshlf, line 473: TODO

src/impls.rs
```diff
@@ -482,13 +482,11 @@ mod atomics {
...
 
-        impl_traits_for_atomics!(AtomicU16, AtomicI16);
+        impl_traits_for_atomics!(AtomicU16[u16], AtomicI16[i16]);
 
         safety_comment! {
-            /// SAFETY:
-            /// All of these pass an atomic type and that type's native equivalent, as
-            /// required by the macro safety preconditions.
-            unsafe_impl_transparent_wrapper_for_atomic!(AtomicU16 [u16], AtomicI16 [i16]);
+            /// SAFETY: TODO
```
joshlf, line 488: TODO

src/impls.rs
```diff
@@ -499,13 +497,11 @@ mod atomics {
...
 
-        impl_traits_for_atomics!(AtomicU32, AtomicI32);
+        impl_traits_for_atomics!(AtomicU32[u32], AtomicI32[i32]);
 
         safety_comment! {
-            /// SAFETY:
-            /// All of these pass an atomic type and that type's native equivalent, as
-            /// required by the macro safety preconditions.
-            unsafe_impl_transparent_wrapper_for_atomic!(AtomicU32 [u32], AtomicI32 [i32]);
+            /// SAFETY: TODO
```
joshlf, line 503: TODO

src/impls.rs
```diff
@@ -516,13 +512,11 @@ mod atomics {
...
 
-        impl_traits_for_atomics!(AtomicU64, AtomicI64);
+        impl_traits_for_atomics!(AtomicU64[u64], AtomicI64[i64]);
 
         safety_comment! {
-            /// SAFETY:
-            /// All of these pass an atomic type and that type's native equivalent, as
-            /// required by the macro safety preconditions.
-            unsafe_impl_transparent_wrapper_for_atomic!(AtomicU64 [u64], AtomicI64 [i64]);
+            /// SAFETY: TODO
```
joshlf, line 518: TODO

src/impls.rs
```diff
@@ -533,21 +527,27 @@ mod atomics {
...
-        impl_for_transparent_wrapper!(T => FromZeros for AtomicPtr<T>);
+        impl_for_transmute_from!(T => TryFromBytes for AtomicPtr<T> [UnsafeCell<*mut T>]);
 
         safety_comment! {
-            /// SAFETY:
-            /// This passes an atomic type and that type's native equivalent, as
-            /// required by the macro safety preconditions.
-            unsafe_impl_transparent_wrapper_for_atomic!(AtomicUsize [usize], AtomicIsize [isize]);
-            unsafe_impl_transparent_wrapper_for_atomic!(T => AtomicPtr<T> [*mut T]);
+            /// SAFETY: TODO
```
joshlf, line 548: TODO

src/impls.rs
```diff
@@ -577,39 +650,76 @@ safety_comment! {
...
-impl_for_transparent_wrapper!(T: IntoBytes => IntoBytes for Wrapping<T>);
-impl_for_transparent_wrapper!(T: Unaligned => Unaligned for Wrapping<T>);
+impl_for_transmute_from!(T: TryFromBytes => TryFromBytes for Wrapping<T>[T]);
+impl_for_transparent_wrapper!(T: FromZeros => FromZeros for Wrapping<T>[T]);
+impl_for_transparent_wrapper!(T: FromBytes => FromBytes for Wrapping<T>[T]);
+impl_for_transparent_wrapper!(T: IntoBytes => IntoBytes for Wrapping<T>[T]);
 assert_unaligned!(Wrapping<()>, Wrapping<u8>);
 
+safety_comment! {
+    /// SAFETY: TODO
```
joshlf, line 660: TODO

src/impls.rs
```diff
@@ -577,39 +650,76 @@ safety_comment! {
...
+}
+
 safety_comment! {
     /// SAFETY:
     /// `TryFromBytes` (with no validator), `FromZeros`, `FromBytes`:
     /// `MaybeUninit<T>` has no restrictions on its contents.
     unsafe_impl!(T => TryFromBytes for CoreMaybeUninit<T>);
     unsafe_impl!(T => FromZeros for CoreMaybeUninit<T>);
     unsafe_impl!(T => FromBytes for CoreMaybeUninit<T>);
+    /// SAFETY: TODO
```
joshlf, line 672: TODO

src/impls.rs
```diff
@@ -577,39 +650,76 @@ safety_comment! {
...
 assert_unaligned!(CoreMaybeUninit<()>, CoreMaybeUninit<u8>);
 
-impl_for_transparent_wrapper!(T: ?Sized + Immutable => Immutable for ManuallyDrop<T>);
-impl_for_transparent_wrapper!(T: ?Sized + TryFromBytes => TryFromBytes for ManuallyDrop<T>);
-impl_for_transparent_wrapper!(T: ?Sized + FromZeros => FromZeros for ManuallyDrop<T>);
-impl_for_transparent_wrapper!(T: ?Sized + FromBytes => FromBytes for ManuallyDrop<T>);
-impl_for_transparent_wrapper!(T: ?Sized + IntoBytes => IntoBytes for ManuallyDrop<T>);
-impl_for_transparent_wrapper!(T: ?Sized + Unaligned => Unaligned for ManuallyDrop<T>);
+safety_comment! {
+    /// SAFETY: TODO
```
joshlf, line 679: TODO

jswrenn (approved):
Developed and reviewed in tandem!

PR #2415, opened 2025-03-05
Inline try_cast_into()

korran:
This significantly improves codegen on embedded targets, allowing many checks to be optimized out.


joshlf:
Published this change in [0.8.22](https://crates.io/crates/zerocopy/0.8.22).

PR #2428, opened 2025-03-07
Teach transmute_{ref,mut}! to handle slice DSTs

joshlf:
TODO:
- Figure out how to make this backwards-compatible (since current macros
  work with `T: Sized`, but `T: Sized` does not imply `T: KnownLayout`)
- Make it so that `Sized` bounds are still enforced on older toolchains
  that don't support slice DST transmutes
- Write safety comments
- Get rid of now-unused macros and functions in util::macro_util

Makes progress on #1817

gherrit-pr-id: Ib4bc62202e0b3b09d155333b525087f7aa8f02c2


src/layout.rs
```diff
@@ -92,6 +92,136 @@ pub(crate) enum MetadataCastError {
...
+    /// If `Src: KnownLayout` and `Dst: KnownLayout`, and if `self` was computed
+    /// as `Src::LAYOUT.try_compute_cast_params(&Dst::LAYOUT)`, then the
+    /// following is guaranteed:
+    ///
+    /// Let `src_meta` be the pointer metadata from a `src: &Src`, and compute
+    /// `let dst_meta = self.compute_cast(src_meta)`. If `dst: &Dst` is
+    /// constructed using the address of `src` and using `dst_meta` as its
+    /// pointer metadata, then `dst` will address the same bytes as `src`.
+    ///
+    /// TODO: Mention that post-padding may not be preserved.
```
joshlf, line 136: TODO

src/layout.rs
```diff
@@ -92,6 +92,136 @@ pub(crate) enum MetadataCastError {
...
+        //      return value)
+        //
+        // (13) `dst_len = dl.offset + (dl.elem_size * ((src_len - dl.offset) /
+        //      dl.elem_size))` (substituting (10))
+        //
+        // (14) `dst_len = dl.offset + (src_len - dl.offset)`
+        //
+        // (15) `dst_len = src_len`
+        //
+        // [1] TODO
```
joshlf, line 216: TODO

src/layout.rs
```diff
@@ -92,6 +92,136 @@ pub(crate) enum MetadataCastError {
...
+        //
+        // TODO: Clarify that this algorithm is agnostic to post-padding, and
```
joshlf, line 218: TODO

src/layout.rs
```diff
@@ -202,6 +332,48 @@ impl DstLayout {
...
+        // PANICS: The preceding `slf.elem_size.checked_rem(other.elem_size)`
+        // could only return `Some` if `other.elem_size > 0`, so this division
+        // will not divide by zero.
+        #[allow(clippy::arithmetic_side_effects)]
+        let offset_delta_elems = offset_delta / other.elem_size;
+        // PANICS: See previous panics comment.
+        #[allow(clippy::arithmetic_side_effects)]
+        let elem_multiple = slf.elem_size / other.elem_size;
+
+        // INVARIANTS: TODO
```
joshlf, line 373: TODO

src/macros.rs
```diff
@@ -206,36 +199,13 @@ macro_rules! transmute_ref {
...
-            //   `AssertDstIsImmutable` above.
-            // - We know that `size_of::<Src>() == size_of::<Dst>()` thanks to
-            //   the use of `assert_size_eq!` above.
-            // - We know that `align_of::<Src>() >= align_of::<Dst>()` thanks to
-            //   the use of `assert_align_gt_eq!` above.
-            let u = unsafe { $crate::util::macro_util::transmute_ref(e) };
-            $crate::util::macro_util::must_use(u)
+            use $crate::util::macro_util::TransmuteRefDst;
+            let t = $crate::util::macro_util::Wrap(e);
+            // SAFETY: todo
```
joshlf, line 205: TODO

src/macros.rs
```diff
@@ -362,32 +324,13 @@ macro_rules! transmute_mut {
...
-            // SAFETY: For source type `Src` and destination type `Dst`:
-            // - We know that `size_of::<Src>() == size_of::<Dst>()` thanks to
-            //   the use of `assert_size_eq!` above.
-            // - We know that `align_of::<Src>() >= align_of::<Dst>()` thanks to
-            //   the use of `assert_align_gt_eq!` above.
-            let u = unsafe { $crate::util::macro_util::transmute_mut(e) };
-            $crate::util::macro_util::must_use(u)
+            use $crate::util::macro_util::TransmuteMutDst;
+            let t = $crate::util::macro_util::Wrap(e);
+            // SAFETY: todo
```
joshlf, line 330: TODO

src/util/macros.rs
```diff
@@ -779,26 +789,142 @@ macro_rules! impl_transitive_transmute_from {
...
+
+        use crate::KnownLayout;
+
+        #[repr(transparent)]
+        struct $src<T: ?Sized>(T);
+
+        #[repr(transparent)]
+        struct $dst<U: ?Sized>(U);
+
+        // SAFETY: TODO
```
joshlf, line 865: TODO

src/util/macros.rs
```diff
@@ -779,26 +789,142 @@ macro_rules! impl_transitive_transmute_from {
...
+        unsafe_impl_for_transparent_wrapper!(T: ?Sized => $src<T>);
+
+        // SAFETY: TODO
```
joshlf, line 868: TODO

src/util/macros.rs
```diff
@@ -779,26 +789,142 @@ macro_rules! impl_transitive_transmute_from {
...
+        unsafe_impl_for_transparent_wrapper!(T: ?Sized => $dst<T>);
+
+        // SAFETY: TODO
```
joshlf, line 871: TODO

src/util/macros.rs
```diff
@@ -779,26 +789,142 @@ macro_rules! impl_transitive_transmute_from {
...
+        unsafe impl<T: ?Sized> InvariantsEq<$src<T>> for T {}
+
+        // SAFETY: TODO
```
joshlf, line 874: TODO

src/util/macros.rs
```diff
@@ -779,26 +789,142 @@ macro_rules! impl_transitive_transmute_from {
...
+        unsafe impl<T: ?Sized> InvariantsEq<$dst<T>> for T {}
+
+        // SAFETY: TODO
```
joshlf, line 877: TODO

src/util/macros.rs
```diff
@@ -779,26 +789,142 @@ macro_rules! impl_transitive_transmute_from {
...
+        impl_for_transmute_from!(T: ?Sized + FromBytes => FromBytes for $src<T>[<T>]);
+        impl_for_transmute_from!(T: ?Sized + FromZeros => FromZeros for $src<T>[<T>]);
+        impl_for_transmute_from!(T: ?Sized + IntoBytes => IntoBytes for $src<T>[<T>]);
+
+        impl_for_transmute_from!(U: ?Sized + TryFromBytes => TryFromBytes for $dst<U>[<U>]);
+        impl_for_transmute_from!(U: ?Sized + FromBytes => FromBytes for $dst<U>[<U>]);
+        impl_for_transmute_from!(U: ?Sized + FromZeros => FromZeros for $dst<U>[<U>]);
+        impl_for_transmute_from!(U: ?Sized + IntoBytes => IntoBytes for $dst<U>[<U>]);
+
+        // SAFETY: TODO
```
joshlf, line 922: TODO

src/util/macros.rs
```diff
@@ -779,26 +789,142 @@ macro_rules! impl_transitive_transmute_from {
...
+        unsafe_impl!(T: ?Sized + Immutable => Immutable for $src<T>);
+
+        // SAFETY: TODO
```
joshlf, line 925: TODO

src/pointer/transmute.rs
```diff
@@ -274,27 +273,59 @@ where
...
+/// Let `dsize` be the smallest valid size of `Dst` such that `dsize >=
+/// ssize`.\* For all `DV`-valid values of `Dst` with size `dsize`, `dst`, and
+/// for all `SV`-valid values of `Src` with size `ssize`, `src`, let `dst'` be
+/// constructed by concatenating `src` with the trailing `dsize - ssize` bytes
+/// of `dst`. It must be the case that `dst'` is a `DV`-valid `Dst`.
+///
+/// Now, instead let `dsize` be the largest valid size of `Dst` such that `dsize
+/// <= ssize`.\* For all `SV`-valid values of `Src` with size `ssize`, `src`, it
+/// must be the case that the leading `dsize` bytes of `src` constitute a
+/// `DV`-valid `Dst`.
```
joshlf, line 295: The motivation behind this invariant is case analysis:
- When going from `&Src` to `&Dst`, we require that `ssize >= dsize`, or else we'd produce a `&Dst` that addresses bytes past the end of the original referent. So long as we always implement such transmutes by choosing the largest `dsize` such that `dsize <= ssize`, then the second condition ensures that this is sound.
- If we have already gone from `&mut Dst` to `&mut Src` (or `&` with interior mutation), then we need it to be the case that writing a `src` to this reference will not invalidate the original `Dst`. Again, so long as we always implement reference transmutes by choosing the largest possible valid size for the produced reference, then the first condition ensures that it is sound for the resulting `&mut Src` to be modified, overwriting the prefix of `Dst`.

My intention is that this should handle by-reference transmutations for all four combinations of sized or unsized `Src` to sized or unsized `Dst`:
- Unsized to unsized is the fully general case described by this prose.
- Sized to unsized means that there's exactly one value of `ssize`, but the subsequent condition still needs to hold (regarding the largest and smallest values of `dsize` on either size of `ssize`).
- Unsized to sized means that there's exactly one value of `dsize` to consider in the analysis
  - For all values of `ssize` such that `dsize <= ssize`, we consider truncating transmutes
  - For all values of `ssize` such that `dsize >= ssize`, we consider tearing transmutes
- Sized to sized is trivial, since there are no "foralls" or "there exists" involved. There are again two sub-cases:
  - If `dsize <= ssize`, we must consider truncating transmutes
  - If `dsize >= ssize`, we must consider tearing transmutes

jswrenn:
Got the autoref specialization trick working: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=d4848523c0aa963d0b716147289044e4

src/layout.rs
```diff
@@ -608,6 +608,216 @@ impl DstLayout {
...
+    ///
+    /// # PME
+    ///
+    /// Generates a post-monomorphization error if it is not possible to satisfy
+    /// the soundness conditions of [`SizeCompat::cast_from_raw`] for `Src` and
+    /// `Dst`.
+    ///
+    /// [cast_from_raw]: crate::pointer::SizeCompat::cast_from_raw
+    //
+    // TODO: Support Sized->Unsized and Unsized->Sized casts
```
jswrenn, line 625: Issue?

src/util/macro_util.rs
```diff
@@ -733,7 +674,183 @@ where
...
+            })
+        }
+    }
+}
+
+pub trait TransmuteMutDst<'a> {
+    type Dst: ?Sized;
+    #[must_use]
+    fn transmute_mut(self) -> &'a mut Self::Dst;
+}
```
jswrenn, line 821: Can we collapse `TransmuteRefDst` and `TransmuteMutDst` into a single trait?

jswrenn (changes requested):

src/util/macro_util.rs
joshlf: This doesn't work - we have to specifically implement for `Wrap<&Src>` and `Wrap<&mut Src>` separately - these can't be unified.

src/layout.rs
joshlf: Done.

src/pointer/transmute.rs
```diff
@@ -269,75 +267,133 @@ where
...
-/// etc) and have the same size. In particular:
-/// - If `T: Sized` and `Self: Sized`, then their sizes must be equal
-/// - If `T: ?Sized` and `Self: ?Sized`, then it must be the case that, given
-///   any `t: *mut T`, `t as *mut Self` produces a pointer which addresses the
-///   same number of bytes as `t`.
-pub unsafe trait SizeEq<T: ?Sized> {
-    fn cast_from_raw(t: NonNull<T>) -> NonNull<Self>;
+/// Implementations of `cast_from_raw` must satisfy that method's safety
+/// post-condition.
+pub unsafe trait SizeCompat<Src: ?Sized> {
```
jswrenn, line 325: We don't need to do this *now*, but it'd be more idiomatic to name this `CastFrom`, after its method (and source of safety conditions). That name also conveys directionality, unlike `SizeCompat`.

src/util/macros.rs
```diff
@@ -779,26 +805,148 @@ macro_rules! impl_transitive_transmute_from {
...
+///
+/// This macro emits code which implements `SizeComapt`, and ensures that the
+/// impl is sound via PME.
+///
+/// # Safety
+///
+/// Inside of `$blk`, the caller must only use `$src` and `$dst` as `$src<$t>`
+/// and `$dst<$u>`. The caller must not use `$src` or `$dst` to wrap any other
+/// types.
+macro_rules! unsafe_with_size_eq {
```
jswrenn, line 861: `unsafe_with_size_compat`?

joshlf: Done

jswrenn (dismissed):

src/pointer/transmute.rs
```diff
@@ -269,75 +267,133 @@ where
...
-unsafe impl<T: ?Sized> SizeEq<T> for T {
-    fn cast_from_raw(t: NonNull<T>) -> NonNull<T> {
+unsafe impl<T: ?Sized> SizeCompat<T> for T {
+    #[inline(always)]
+    fn cast_from_raw<'a>(t: PtrInner<'a, T>) -> PtrInner<'a, T> {
         t
     }
 }
 
+// TODO: Update all `TransmuteFrom` safety proofs.
```
joshlf, line 348: TODO

src/util/macros.rs
```diff
@@ -758,30 +772,153 @@ macro_rules! impl_transitive_transmute_from {
...
+/// `SizeEq`.
+///
+/// This macro emits code which implements `SizeEq`, and ensures that the impl
+/// is sound via PME.
+///
+/// # Safety
+///
+/// Inside of `$blk`, the caller must only use `$src` and `$dst` as `$src<$t>`
+/// and `$dst<$u>`. The caller must not use `$src` or `$dst` to wrap any other
+/// types.
```
jswrenn, line 827: Curious if we can make this safe by doing `type $src = Src<$src>;` above the block.

src/util/macros.rs
```diff
@@ -758,30 +772,153 @@ macro_rules! impl_transitive_transmute_from {
...
+
+        // SAFETY: `$src<T>` is a `#[repr(transparent)]` wrapper around `T`, and
+        // so permits interior mutation exactly when `T` does.
+        unsafe_impl!(T: ?Sized + Immutable => Immutable for $src<T>);
+
+        // SAFETY: `$dst<T>` is a `#[repr(transparent)]` wrapper around `T`, and
+        // so permits interior mutation exactly when `T` does.
+        unsafe_impl!(T: ?Sized + Immutable => Immutable for $dst<T>);
+
+        $blk
```
jswrenn, line 919: Wonder if the outer `unsafe` required to currently call this macro leaks into cover `$blk`.

src/util/macros.rs
```diff
@@ -758,30 +772,153 @@ macro_rules! impl_transitive_transmute_from {
...
+        if 1 == 0 {
+            let ptr = <$t as KnownLayout>::raw_dangling();
+            #[allow(unused_unsafe)]
+            // SAFETY: This call is never executed.
+            let ptr = unsafe { crate::pointer::PtrInner::new(ptr) };
+            #[allow(unused_unsafe)]
+            // SAFETY: This call is never executed.
+            let ptr = unsafe { cast!(ptr) };
+            let _ = <$dst<$u> as SizeEq<$src<$t>>>::cast_from_raw(ptr);
+        }
```
jswrenn, line 899: We should drop a comment as to why this block exists.

jswrenn: Is this okay? Can we rely on this PME occuring despite this code being dead?

src/util/macros.rs
joshlf: Tried offline; turned out to be very difficult, so we're going to punt for now.

src/util/macros.rs
joshlf: Tried offline; turned out to be very difficult, so we're going to punt for now.

src/util/macros.rs
joshlf: Comment: Done.

Can we rely on the PME: Maybe not, but the follow-up PR will ensure that it's sound since `SizeEq` will no longer convey a safety post-condition that can be consumed by unsafe code - the only way to use it will be via `SizeEq::cast_from_raw`, which will guarantee that the PME is triggered.

jswrenn (approved):

PR #2436, opened 2025-03-11
[layout] Move Kani proofs from crate root

joshlf:
All moved proofs test code from the `layout` module, and so belong there
rather than in the crate root.

---

This PR is on branch [cell-traits](../tree/cell-traits).

- #2428
- #2435
- #2436


joshlf:
Replaced by #2430.

PR #2472, opened 2025-03-28
Generalize SizeEq to SizeFrom, support shrinking

joshlf:
Generalize `SizeEq`, renaming it to `SizeFrom`, and supporting casts
which shrink rather than preserve referent size.

This requires fairly deep surgery in our transmute taxonomy. We can no
longer rely on `SizeFrom::cast_from_raw` preserving referent size, which
means that all pointer transmutes must now reason about two types of
transmutes:
- Size-shrinking transmutes (in which the source value is larger than
  the destination value)
- Overwriting transmutes (in which the source value is larger than the
  destination value *and* the destination permits mutation). In an
  overwriting transmute, the resulting value is a concatenation of a
  valid `T` and the *suffix* of a valid `U`.

(Note that size-preserving transmutes can be viewed as a special case of
size-shrinking transmutes.)

In order to support these semantics, we split `TransmuteFrom` into two
traits: `TransmuteFrom` (which supports shriking transmutes) and
`TransmuteOverwrite` (which supports overwriting transmutes).

While we're here, we simplify the relationship between
`TryTransmuteFromPtr`, `MutationCompatible`, and `Read`. In particular,
`TryTransmuteFromPtr` now has a single blanket impl which is bounded by
`MutationCompatible`, and `MutationCompatible` now directly supports
three cases (and no longer delegates to `Read`):
- Exclusive reference transmutes in which the destination type may be
  transmuted into the source type
- Shared reference transmutes in which both the source and the
  destination type are `Immutable`
- Shared reference transmutes which permit interior mutation, in which
  the source and destination types may both be transmuted into one
  another, and in which the source and destination types support
  compatible interior mutation

Makes progress on #1817

---

This PR is on branch [transmute-ref-dst](../tree/transmute-ref-dst).

- #2487
- #2472
- #2570

### DIFF
--- a/src/impls.rs
+++ b/src/impls.rs
@@ -105,11 +105,13 @@ assert_unaligned!(bool);
 //   pattern 0x01.
 const _: () = unsafe {
     unsafe_impl!(=> TryFromBytes for bool; |byte| {
-        let byte = byte.transmute::<u8, invariant::Valid, _>();
+        let mut byte = byte;
+        let byte = byte.reborrow().into_shared().transmute::<u8, invariant::Valid, _>();
         *byte.unaligned_as_ref() < 2
     })
 };
-impl_size_eq!(bool, u8);
+
+impl_size_from!(bool, u8);
 
 // SAFETY:
 // - `Immutable`: `char` self-evidently does not contain any `UnsafeCell`s.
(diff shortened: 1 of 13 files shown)


jswrenn (approved):
🎉

Issue #1290, opened 2024-05-17
Support slicing a `KnownLayout` type


joshlf:
Given a `&[T]`, Rust supports creating a new `&[T]` referring to a subset of the original referent, and also supports splitting into two non-overlapping `&[T]`s that, together, cover the original referent. To my knowledge, this is not supported for slice DSTs.

We could add APIs for these slicing operations to `KnownLayout<PointerMetadata = usize>`.

One use case for this design would be to support packet formats with explicit length fields - it would be a possibly lighter-weight way of supporting #1289 without requiring solving the general problem described in that issue.

joshlf:
If we also encode the trailing slice element type in `KnownLayout` (e.g. via an associated type), then we could support a more general `split_at` operation which would return `(&Self, &[Self::Elem])` or `(&mut Self, &mut [Self::Elem])`.

joshlf:
> If we also encode the trailing slice element type in `KnownLayout` (e.g. via an associated type), then we could support a more general `split_at` operation which would return `(&Self, &[Self::Elem])` or `(&mut Self, &mut [Self::Elem])`.

I wonder if there's some way that we could provide an associated element type *only* when `PointerMetadata = usize`. Could we teach Rust to only expect such a type to exist when `PointerMetadata = usize`?

jswrenn:
> I wonder if there's some way that we could provide an associated element type _only_ when `PointerMetadata = usize`.

We can probably solve this problem alongside #2149 in 0.9.

In the meantime, I've been experimenting on a branch with having `TrailingElem = ()` for sized types, and that works totally fine. If we want, our surface-level API can require `KnownLayout<PointerMetadata = usize>` for `split_at`.

---

The interaction of dynamic padding and splitting is weird. Not insurmountably weird, but the semantics are going to be tricky to document. In particular, `.split_at(0)` cannot always mean *split at the beginning of the trailing slice*. To illustrate, let's assume that invocation *does* have this meaning and mentally execute this code:

```rust
use zerocopy::*; // 0.8.16

#[derive(FromBytes, KnownLayout, Immutable)]
#[repr(C)]
struct Foo(u16, u8, [u8]);

fn main() {
    let bytes = [0, 0, 1, 2];
    let foo = Foo::ref_from_bytes(&bytes).unwrap();
    let (bar, trailing) = foo.split_at(0);
    assert_eq!(trailing, &[2]);
}
```

[...]

PR #2473, opened 2025-03-28
Separate `SplitAt`'s bounds checking and overlap checking

jswrenn:
Previously, `SplitAt`'s methods returned `Option<(&(mut) Self, &(mut) Self::Trailing)>`, where `None` reflects an index that's either out of bounds, or results in the split parts overlapping in a problematic manner. This design conflated two failure conditions into one and, worse, presented an overly expensive and conservative API: dynamically checking for overlap is expensive, and there are many cases in which overlapping parts are *not* problematic, and this API could not scale to accommodate them all. 

`SplitAt`'s methods now return `Option<Split<&(mut) Self>>`, where `None` reflects an out-of-bounds index, and `Split<&(mut) Self>` encodes the possibly-overlapping parts. `Split<&(mut) Self>` provides methods for producing references to the split parts via both dynamic overlap checking and via static bounds (e.g., `Self: Immutable`). This API is more both cheaper and more expressive.

This API is, in fact, slightly more expressive than is possible to demonstrate within zerocopy's current capabilities. `Split::via_runtime_check` is overly conservative when `T: Immutable`, but it is not currently possible to construct a DST which would have overlapping parts *without* `T: Immutable`. We intend to follow up with mechanisms for DST construction that don't have this limitation.

Makes progress towards #1290.

### DIFF
--- a/src/split_at.rs
+++ b/src/split_at.rs
@@ -8,8 +8,7 @@
 // those terms.
 
 use super::*;
-#[cfg(doc)]
-use invariant::Exclusive;
+use crate::pointer::invariant::{Aligned, Exclusive, Invariants, Shared, Valid};
 
 /// Types that can be split in two.
 ///
@@ -54,7 +53,8 @@ use invariant::Exclusive;
 )]
 // # Safety
 //
-// The trailing slice is well-aligned for its element type.
+// The trailing slice is well-aligned for its element type. `Self` is `[T]`, or
+// a `repr(C)` or `repr(transparent)` slice DST.
 pub unsafe trait SplitAt: KnownLayout<PointerMetadata = usize> {
     /// The element type of the trailing slice.
     type Elem;
@@ -72,22 +72,9 @@ pub unsafe trait SplitAt: KnownLayout<PointerMetadata = usize> {
     /// `self`'s trailing slice.
     #[inline]
     #[must_use]
-    unsafe fn split_at_unchecked(&self, l_len: usize) -> (&Self, &[Self::Elem])
-    where
-        Self: Immutable,
-    {
-        // SAFETY: `&self` is an instance of `&Self` for which the caller has
-        // promised that `l_len` is not greater than the length of `self`'s
-        // trailing slice.
-        let l_len = unsafe { MetadataOf::new_unchecked(l_len) };
-        let ptr = Ptr::from_ref(self);
-        // SAFETY:
-        // 0. The caller promises that `l_len` is not greater than the length of
-        //    `self`'s trailing slice.
-        // 1. `ptr`'s aliasing is `Shared` and does not permit interior mutation
-        //    because `Self: Immutable`.
-        let (left, right) = unsafe { ptr_split_at_unchecked(ptr, l_len) };
-        (left.as_ref(), right.as_ref())
+    unsafe fn split_at_unchecked(&self, l_len: usize) -> Split<&Self> {
+        // SAFETY: By precondition on the caller, `l_len <= self.len()`.
+        unsafe { Split::<&Self>::new(self, l_len) }
     }
 
     /// Attempts to split `self` in two.
@@ -116,52 +103,41 @@ pub unsafe trait SplitAt: KnownLayout<PointerMetadata = usize> {
     /// assert_eq!(packet.length, 4);
     /// assert_eq!(packet.body, [1, 2, 3, 4, 5, 6, 7, 8, 9]);
     ///
-    /// let (packet, rest) = packet.split_at(packet.length as usize).unwrap();
+    /// // Attempt to split `packet` at `length`.
+    /// let split = packet.split_at(packet.length as usize).unwrap();
+    ///
+    /// // Use the `Immutable` bound on `Packet` to prove that it's okay to
+    /// // return concurrent references to `packet` and `rest`.
+    /// let (packet, rest) = split.via_immutable();
+    ///
     /// assert_eq!(packet.length, 4);
     /// assert_eq!(packet.body, [1, 2, 3, 4]);
     /// assert_eq!(rest, [5, 6, 7, 8, 9]);
     /// ```
     #[inline]
     #[must_use = "has no side effects"]
-    fn split_at(&self, l_len: usize) -> Option<(&Self, &[Self::Elem])>
-    where
-        Self: Immutable,
-    {
-        if l_len <= Ptr::from_ref(self).len() {
-            // SAFETY: We have checked that `l_len` is not greater than the
-            // length of `self`'s trailing slice.
-            Some(unsafe { self.split_at_unchecked(l_len) })
-        } else {
-            None
-        }
+    fn split_at(&self, l_len: usize) -> Option<Split<&Self>> {
+        MetadataOf::new_in_bounds(self, l_len).map(
+            #[inline(always)]
+            |l_len| {
+                // SAFETY: We have ensured that `l_len <= self.len()` (by
+                // post-condition on `MetadataOf::new_in_bounds`)
+                unsafe { Split::new(self, l_len.get()) }
+            },
+        )
     }
 
     /// Unsafely splits `self` in two.
     ///
     /// # Safety
     ///
-    /// The caller promises that:
-    /// 0. `l_len` is not greater than the length of `self`'s trailing slice.
-    /// 1. The trailing padding bytes of the left portion will not overlap the
-    ///    right portion. For some dynamically sized types, the padding that
-    ///    appears after the trailing slice field [is a dynamic function of the
-    ///    trailing slice length](KnownLayout#slice-dst-layout). Thus, for some
-    ///    types, this condition is dependent on the value of `l_len`.
+    /// The caller promises that `l_len` is not greater than the length of
+    /// `self`'s trailing slice.
     #[inline]
     #[must_use]
-    unsafe fn split_at_mut_unchecked(&mut self, l_len: usize) -> (&mut Self, &mut [Self::Elem]) {
-        // SAFETY: `&mut self` is an instance of `&mut Self` for which the
-        // caller has promised that `l_len` is not greater than the length of
-        // `self`'s trailing slice.
-        let l_len = unsafe { MetadataOf::new_unchecked(l_len) };
-        let ptr = Ptr::from_mut(self);
-        // SAFETY:
-        // 0. The caller promises that `l_len` is not greater than the length of
-        //    `self`'s trailing slice.
-        // 1. `ptr`'s aliasing is `Exclusive`; the caller promises that
-        //    `l_len.padding_needed_for() == 0`.
-        let (left, right) = unsafe { ptr_split_at_unchecked(ptr, l_len) };
-        (left.as_mut(), right.as_mut())
+    unsafe fn split_at_mut_unchecked(&mut self, l_len: usize) -> Split<&mut Self> {
+        // SAFETY: By precondition on the caller, `l_len <= self.len()`.
+        unsafe { Split::<&mut Self>::new(self, l_len) }
     }
 
     /// Attempts to split `self` in two.
@@ -178,7 +154,7 @@ pub unsafe trait SplitAt: KnownLayout<PointerMetadata = usize> {
     /// use zerocopy::{SplitAt, FromBytes};
     /// # use zerocopy_derive::*;
     ///
-    /// #[derive(SplitAt, FromBytes, KnownLayout, Immutable, IntoBytes)]
+    /// #[derive(SplitAt, FromBytes, KnownLayout, IntoBytes)]
     /// #[repr(C)]
     /// struct Packet<B: ?Sized> {
     ///     length: u8,
@@ -194,7 +170,13 @@ pub unsafe trait SplitAt: KnownLayout<PointerMetadata = usize> {
     /// assert_eq!(packet.body, [1, 2, 3, 4, 5, 6, 7, 8, 9]);
     ///
     /// {
-    ///     let (packet, rest) = packet.split_at_mut(packet.length as usize).unwrap();
+    ///     // Attempt to split `packet` at `length`.
+    ///     let split = packet.split_at_mut(packet.length as usize).unwrap();
+    ///
+    ///     // Use the `IntoBytes` bound on `Packet` to prove that it's okay to
+    ///     // return concurrent references to `packet` and `rest`.
+    ///     let (packet, rest) = split.via_into_bytes();
+    ///
     ///     assert_eq!(packet.length, 4);
     ///     assert_eq!(packet.body, [1, 2, 3, 4]);
     ///     assert_eq!(rest, [5, 6, 7, 8, 9]);
@@ -206,17 +188,15 @@ pub unsafe trait SplitAt: KnownLayout<PointerMetadata = usize> {
     /// assert_eq!(packet.body, [1, 2, 3, 4, 0, 0, 0, 0, 0]);
     /// ```
     #[inline]
-    fn split_at_mut(&mut self, l_len: usize) -> Option<(&mut Self, &mut [Self::Elem])> {
-        match MetadataOf::new_in_bounds(self, l_len) {
-            Some(l_len) if l_len.padding_needed_for() == 0 => {
-                // SAFETY: We have ensured both that:
-                // 0. `l_len <= self.len()` (by post-condition on
-                //    `MetadataOf::new_in_bounds`)
-                // 1. `l_len.padding_needed_for() == 0` (by guard on match arm)
-                Some(unsafe { self.split_at_mut_unchecked(l_len.get()) })
-            }
-            _ => None,
-        }
+    fn split_at_mut(&mut self, l_len: usize) -> Option<Split<&mut Self>> {
+        MetadataOf::new_in_bounds(self, l_len).map(
+            #[inline(always)]
+            |l_len| {
+                // SAFETY: We have ensured that `l_len <= self.len()` (by
+                // post-condition on `MetadataOf::new_in_bounds`)
+                unsafe { Split::new(self, l_len.get()) }
+            },
+        )
     }
 }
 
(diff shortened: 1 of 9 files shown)


zerocopy-derive/src/lib.rs
```diff
@@ -702,15 +702,20 @@ fn derive_split_at_inner(
...
         trailing_field
     } else {
         return Err(Error::new(Span::call_site(), "must at least one field"));
     };
 
-    // SAFETY: `#ty`, per the above check, is not packed; its trailing slice
-    // field is guaranteed to be well-aligned for its type.
+    // SAFETY: `#ty`, per the above checks, is `repr(C)` or `repr(transparent)`
+    // and is not packed; its trailing slice field is guaranteed to be
+    // well-aligned for its type.
```
joshlf, line 718: Does this need to also argue that the trailing field is `SplitAt`, which guarantees the alignment property recursively?

src/split_at.rs
```diff
@@ -116,52 +103,41 @@ pub unsafe trait SplitAt: KnownLayout<PointerMetadata = usize> {
...
-        if l_len <= Ptr::from_ref(self).len() {
-            // SAFETY: We have checked that `l_len` is not greater than the
-            // length of `self`'s trailing slice.
-            Some(unsafe { self.split_at_unchecked(l_len) })
-        } else {
-            None
-        }
+    fn split_at(&self, l_len: usize) -> Option<Split<&Self>> {
+        MetadataOf::new_in_bounds(self, l_len).map(
+            #[inline(always)]
```
joshlf, line 121: TIL you could use `#[inline(always)]` on a closure...

src/split_at.rs
```diff
@@ -233,59 +213,571 @@ unsafe impl<T> SplitAt for [T] {
...
+    ///
+    /// # Safety
+    ///
+    /// `source` is a slice DST whose length no less than `l_len`.
+    source: T,
+    /// The length of the future left half of `source`.
+    ///
+    /// # Safety
+    ///
+    /// `l_len` is no greater than `source`'s length.
```
joshlf, line 240: IMO you don't need to repeat the safety condition. I'd just put it on `l_len`. It may still make sense to say "`source` is a slice DST."

src/split_at.rs
```diff
@@ -233,59 +213,571 @@ unsafe impl<T> SplitAt for [T] {
...
+    /// it is sound to have concurrent references to both parts.
+    #[inline(always)]
+    fn via_unaligned(self) -> (Ptr<'a, T, I>, Ptr<'a, [T::Elem], I>)
+    where
+        T: Unaligned,
+    {
+        // SAFETY: By `T: SplitAt + Unaligned`, `T` is either `[T]`, or a
+        // `repr(C)` or `repr(transparent)` slice DST, that is well-aligned at
+        // any address and length. Consequently, `T` can be split into strictly
+        // non-overlapping parts any any index.
```
joshlf, line 304: Can you more explicitly argue that `Unaligned + repr(C)` implies that there will be no padding? IMO you have to be quite familiar with our prior discussions to understand the logical step happening here.

joshlf:
cc @kupiakos

src/split_at.rs
```diff
@@ -233,59 +213,571 @@ unsafe impl<T> SplitAt for [T] {
...
+    {
+        let (l, r) = self.into_ptr().via_unaligned();
+        (l.as_ref(), r.as_ref())
+    }
+
+    /// Produces the split parts of `self`, using a dynamic check to ensure that
+    /// it is sound to have concurrent references to both parts. Note, this
+    /// check is overly conservative if `T` is [`Immutable`]. You should prefer
+    /// using [`Self::via_immutable`], [`Self::via_into_bytes`], or
+    /// [`Self::via_unaligned`], which have no runtime cost.
```
joshlf, line 529: I'd recommend a paragraph break after "is overly conservative if `T` is [`Immutable`]." As written, it sounds like the following sentence is meant to flow logically from that, but actually it stands on its own (regardless of `T: Immutable`, this function performs runtime checks).

src/split_at.rs
```diff
@@ -233,59 +213,571 @@ unsafe impl<T> SplitAt for [T] {
...
+    /// ```
+    /// use zerocopy::{SplitAt, FromBytes};
+    /// # use zerocopy_derive::*;
+    ///
+    /// #[derive(SplitAt, FromBytes, KnownLayout, Immutable)]
+    /// #[repr(C)]
+    /// struct Packet {
+    ///     length: u8,
+    ///     body: [u8],
+    /// }
```
joshlf, line 542: Maybe do `length: u16` so that you can also demonstrate failure?

src/split_at.rs
```diff
@@ -233,59 +213,571 @@ unsafe impl<T> SplitAt for [T] {
...
+    /// ```
+    /// use zerocopy::{SplitAt, FromBytes};
+    /// # use zerocopy_derive::*;
+    ///
+    /// #[derive(SplitAt, FromBytes, KnownLayout, IntoBytes)]
+    /// #[repr(C)]
+    /// struct Packet<B: ?Sized> {
+    ///     length: u8,
+    ///     body: B,
+    /// }
```
joshlf, line 725: Same here re: `length: u16`.

src/split_at.rs
```diff
@@ -233,59 +213,571 @@ unsafe impl<T> SplitAt for [T] {
...
+    /// prefer using [`Self::via_into_bytes`] or [`Self::via_unaligned`], which
+    /// have no runtime cost.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use zerocopy::{SplitAt, FromBytes};
+    /// # use zerocopy_derive::*;
+    ///
+    /// #[derive(SplitAt, FromBytes, KnownLayout, IntoBytes)]
```
joshlf, line 720: Presumably you don't want to derive `IntoBytes`? As a consequence, you presumably don't need `Packet` to be generic over `B`?

src/split_at.rs
```diff
@@ -335,4 +827,32 @@ mod tests {
         test_split_at::<1, 17>();
         test_split_at::<2, 18>();
     }
+
+    #[cfg(feature = "derive")]
+    #[test]
+    #[allow(clippy::as_conversions)]
+    fn test_split_at_overlapping() {
```
joshlf, line 834: Also add a test (probably in a different function than this one) for `via_runtime_check` failing?

joshlf (changes requested):
🎉

src/split_at.rs
jswrenn: What do you mean? The test here is that `via_runtime_checks` succeeds for odd indices and fails for even ones.

src/split_at.rs
jswrenn: It's required by `FromBytes::mut_from_bytes`.

src/split_at.rs
joshlf: Oops, I just misread this test.

src/split_at.rs
joshlf: Ack. I wonder if this is a hint that users will ~never need this API lol. In particular, since slice DSTs are unsized, you can't ever construct one by value, and so the only way to construct a `&mut MySliceDst` is via reference transmute, but all of our support for reference transmutes requires `Self: IntoBytes`. I suppose you could have constructed one yourself (e.g. via unsized coercion), but it's certainly niche. Oh well.

jswrenn: A bunch of these look a little silly when situated right next to `FromByte::ref_from_bytes` (which requires `Immutable`), and `FromBytes::mut_from_bytes` (which requires `IntoBytes`), but I can see these being useful in generic contexts.

joshlf: Example?

jswrenn: Let's say you have a generic function that needs to split a dst that gets passed in. This function is generic over `&` and `&mut` references to a DST. Maybe you have some `&mut` DSTs that aren't `Immutable`, and some `&` DSTs that aren't `IntoBytes`. In such a case, a `T: Unaligned` bound might be the common denominator that lets you write this generic function.

kupiakos: I see `SplitAt` as extending the toolkit of DST manipulation that's dependent on zerocopy traits and expertise to get right, towards a goal of working with them nearly as easily as sized structs. So I do generally support whatever flexible trait bounds are possible for converting, even though I don't have a "key use case" example either.

> you can't ever construct one by value

I've been thinking that `Box<Dst>` construction could be done with a macro and struct init syntax. This'd fit well in zerocopy with the knowledge from `KnownLayout`.

joshlf: > In such a case, a `T: Unaligned` bound might be the common denominator that lets you write this generic function.

Sure, but this comment thread is about `via_runtime_check`, not about `via_unaligned`. My point is that, if you *always* need to have an `IntoBytes` bound anyway, then you may never actually need `via_runtime_check`.

---

To be clear, I still think we should keep `via_runtime_check` for completeness. I just think it's an interesting observation that we should think more about at some point.

joshlf: > I've been thinking that `Box<Dst>` construction could be done with a macro and struct init syntax. This'd fit well in zerocopy with the knowledge from `KnownLayout`/`SplitAt`.

Oh this could be really cool! If you have a prototype at some point, let us know! I could see adding this to zerocopy if it's a simple enough API (which I imagine it would be).

src/split_at.rs
jswrenn: https://github.com/google/zerocopy/compare/15ed1325a438c4e6cc71ff047903987802d6c8f6..b8695e14b15b1a3b27d0bb758375bb572b6f13b8

src/split_at.rs
jswrenn: The `IntoBytes` blocks us from writing a neat test here that shows off overlapping trailing padding.

src/split_at.rs
```diff
@@ -233,59 +213,603 @@ unsafe impl<T> SplitAt for [T] {
...
+    /// let idx = packet.length.get() - 1;
+    /// let split = packet.split_at(idx as usize).unwrap();
+    ///
+    /// // Attempt (and fail) to use a dynamic check to prove that it's okay
+    /// // to return concurrent references to `packet` and `rest`.
+    /// assert!(split.via_runtime_check().is_err());
+    /// ```
+    #[must_use = "has no side effects"]
+    #[inline(always)]
+    pub fn via_runtime_check(self) -> Result<(&'a T, &'a [T::Elem]), Self> {
```
jswrenn, line 599: I [changed](https://github.com/google/zerocopy/compare/b8695e14b15b1a3b27d0bb758375bb572b6f13b8..b342e5116805790c3e7533a7369635820c7e3492) `via_runtime_check` to return a `Result` instead of an `Option`, since `via_runtime_check` consumes the `Split`, but there are still things you can do with a `Split` that isn't concurrently splittable. Namely: you can *serially* access the left and right portions.

zerocopy-derive/src/lib.rs
jswrenn: https://github.com/google/zerocopy/compare/fd5f6b551b68939c61ea1d9bbcdc802d7d4823cb..9562f5e1784e47002f2e20edbdc6cfd7cb753afe

src/split_at.rs
jswrenn: https://github.com/google/zerocopy/compare/b342e5116805790c3e7533a7369635820c7e3492..7bcbc9177934ec7ad7051cfc1d8cdbfbfbe092e9

src/split_at.rs
jswrenn: https://github.com/google/zerocopy/compare/9562f5e1784e47002f2e20edbdc6cfd7cb753afe..5ae44a7b575613bc5ff777c1d84649fe55679889

src/split_at.rs
jswrenn: Like this: https://github.com/google/zerocopy/compare/75c9ce7c837f6b3cdbf688aa84ec4ca7a7037313..1bc06d113bbf932bdcbd1e61e47601b91e809a70

src/split_at.rs
```diff
@@ -233,59 +213,832 @@ unsafe impl<T> SplitAt for [T] {
...
+        let (l, r) = self.into_ptr().via_unaligned();
+        (l.as_ref(), r.as_ref())
+    }
+
+    /// Produces the split parts of `self`, using a dynamic check to ensure that
+    /// it is sound to have concurrent references to both parts. You should
+    /// prefer using [`Self::via_immutable`], [`Self::via_into_bytes`], or
+    /// [`Self::via_unaligned`], which have no runtime cost.
+    ///
+    /// Note, this check is overly conservative if `T` is [`Immutable`].
```
joshlf, line 411:
```suggestion
    /// Note, this check is overly conservative if `T` is [`Immutable`];
    /// for some `Immutable` types, this check will reject some `Split`
    /// locations which [`via_immutable`] will accept.
```

...or something like that.

src/split_at.rs
```diff
@@ -233,59 +213,832 @@ unsafe impl<T> SplitAt for [T] {
...
+    ///
+    /// // Project the left part of the split.
+    /// let packet = split.left();
+    ///
+    /// assert_eq!(packet.length, 3);
+    /// assert_eq!(packet.body, [0, 1, 0]);
+    /// ```
+    #[must_use = "has no side effects"]
+    #[inline(always)]
+    pub fn left(self) -> &'a T {
```
joshlf, line 528: A limitation of consuming `self` by value is that you can't actually borrow left *and then* right from the same `Split`.

Maybe we could have `left`, `right`, `into_left`, and `into_right`, where `left` and `right` operate on `&self`?

Same for `Split<&mut _>`, except `left` and `right` would operate on `&mut self`.

joshlf (changes requested):

src/split_at.rs
jswrenn: https://github.com/google/zerocopy/compare/57acb73aed79835ff693fde6d7c116ce503ab1f7..02b927ec93f33331e5e178bc8071db6d3bd366e0

src/split_at.rs
```diff
@@ -233,59 +213,834 @@ unsafe impl<T> SplitAt for [T] {
...
-///    `l_len.padding_needed_for() == 0`.
-#[inline(always)]
-unsafe fn ptr_split_at_unchecked<'a, T, I, R>(
-    ptr: Ptr<'a, T, I>,
-    l_len: MetadataOf<T>,
-) -> (Ptr<'a, T, I>, Ptr<'a, [T::Elem], I>)
+/// If is impossible to concurrently reference both parts, you can convert a
+/// `Split` into a reference to only the left part (via [`Self::left`] or to
+/// only right part (via [`Self::right`]).
+#[derive(Clone, Copy, Debug)]
```
jswrenn, line 232:
```suggestion
#[derive(Debug)]
```
The `Clone` and `Copy` impls are, in principle, unsound. Since a `Split` represents two possibly overlapping parts, it should not be possible to obtain concurrent borrows to both parts of the split by copying the `Split`.

Annoying, demonstrating the issue with a test case is seemingly impossible.  The only way to currently create a `Split<&T>` requires an `Immutable` bound, so the problem of overlapping mutable parts cannot be demonstrated with a `&T`. And the only way to currently create a `Split<&mut T>` requires an `IntoBytes` bound, which precludes the existence of trailing padding that would result in overlapping parts.

src/split_at.rs
```diff
@@ -233,59 +213,832 @@ unsafe impl<T> SplitAt for [T] {
...
+    /// assert_eq!(packet.body, [0, 1, 0, 2]);
+    /// assert_eq!(rest, [0, 3, 0, 4]);
+    ///
+    /// // Attempt to split `packet` at `length - 1`.
+    /// let idx = packet.length.get() - 1;
+    /// let split = packet.split_at(idx as usize).unwrap();
+    ///
+    /// // Attempt (and fail) to use a dynamic check to prove that it's okay
+    /// // to return concurrent references to `packet` and `rest`.
+    /// assert!(split.via_runtime_check().is_err());
```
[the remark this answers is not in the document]
jswrenn: https://github.com/google/zerocopy/compare/135e87f6f3b6534727f99c3f0d09ab78a473eb1e..33d3de83eecc3db6b07f514368ad2e34a12861ab

src/split_at.rs
jswrenn: Derives removed: https://github.com/google/zerocopy/compare/02b927ec93f33331e5e178bc8071db6d3bd366e0..135e87f6f3b6534727f99c3f0d09ab78a473eb1e#diff-394d6458153031257865a7aaddcda3ac55a4ae61c642dee241136b08a66fcf90L232

src/split_at.rs
```diff
@@ -72,113 +72,855 @@ pub unsafe trait SplitAt: KnownLayout<PointerMetadata = usize> {
...
+    #[inline(always)]
+    pub unsafe fn via_unchecked(self) -> (&'a T, &'a [T::Elem]) {
+        // SAFETY: The aliasing of `self.into_ptr()` is not `Exclusive`, but the
+        // caller has promised that if `T` permits interior mutation then the
+        // left and right portions of `self` split at `l_len` do not overlap.
+        let (l, r) = unsafe { self.into_ptr().via_unchecked() };
+        (l.as_ref(), r.as_ref())
+    }
+
+    /// Reborrow `&self`, allowing the left and right parts to be borrowed
```
joshlf, line 495:
```suggestion
    /// Reborrows `&self`, allowing the left and right parts to be borrowed
```

src/split_at.rs
```diff
@@ -72,113 +72,855 @@ pub unsafe trait SplitAt: KnownLayout<PointerMetadata = usize> {
...
+    pub unsafe fn via_unchecked(self) -> (&'a mut T, &'a mut [T::Elem]) {
+        // SAFETY: The aliasing of `self.into_ptr()` is `Exclusive`, and the
+        // caller has promised that the left and right portions of `self` split
+        // at `l_len` do not overlap.
+        let (l, r) = unsafe { self.into_ptr().via_unchecked() };
+        (l.as_mut(), r.as_mut())
     }
 
-    /// Attempts to split `self` in two.
+    /// Reborrow `&mut self`, allowing the left and right parts to be borrowed
```
joshlf, line 843:
```suggestion
    /// Reborrows `&mut self`, allowing the left and right parts to be borrowed
```

src/split_at.rs
```diff
@@ -72,113 +72,855 @@ pub unsafe trait SplitAt: KnownLayout<PointerMetadata = usize> {
...
+    /// let left = split.reborrow().left();
+    /// let right = split.reborrow().right();
+    ///
+    /// println!("{:?}", left);
+    /// println!("{:?}", right);
+    /// ```
+    #[must_use = "has no side effects"]
+    #[inline(always)]
+    pub fn reborrow<'b>(&'b mut self) -> Split<&'b T> {
+        // SAFETY: `&*self.source` points to exactly `self.source` and thus
```
joshlf, line 563: It wigs me out that this safety proof doesn't mention lifetimes at all, despite the fact that we explicitly advertise that you can't reborrow twice overlapping (and that that's a soundness requirement). Maybe we need to revisit the safety invariant on `Split` or the safety precondition on `Split::new`?

Same with the other `fn reborrow`.

joshlf (changes requested):

src/split_at.rs
jswrenn: https://github.com/google/zerocopy/compare/5eedd1e424b9266334d29e77b36bcfa843ad2948..a872160893f3db1faa6742575ee834d2d6944db8#diff-394d6458153031257865a7aaddcda3ac55a4ae61c642dee241136b08a66fcf90R495

src/split_at.rs
jswrenn: https://github.com/google/zerocopy/compare/5eedd1e424b9266334d29e77b36bcfa843ad2948..a872160893f3db1faa6742575ee834d2d6944db8#diff-394d6458153031257865a7aaddcda3ac55a4ae61c642dee241136b08a66fcf90R843

src/split_at.rs
jswrenn: Yeah, I agree something is fishy here. Similarly, it would be nice if `SplitAt`'s invariant obviously precluded implementing `Copy`/`Clone`. 

It's not quite about lifetimes, I don't think, but rather about shadowing; e.g., on `SplitAt::reborrow`:
```rust
        // SAFETY: `&mut *self.source` points to exactly `self.source` and thus
        // maintains the invariants of `self` with respect to `l_len`. The
        // reborrowed `&mut *self.source` shadows `&mut self`'s source, ensuring
        // that it is the sole active reference from which the left and right
        // parts may be accessed.
```

...but I'm not immediately sure what aspects of this need to be uplifted into `SplitAt`.

src/split_at.rs
jswrenn: https://github.com/google/zerocopy/compare/708346e8b2599757d96f3e483386d830ed6e8e09..c8a657d1740bfb846580483b3f277a7a3ee58757

src/split_at.rs
joshlf: Discussed offline, and decided that `left`, `right`, and `reborrow` are unsound for both `&` and `&mut`. In particular:
- In the `&` case, `left` can overlap with pre-existing `&` references to `T`, and so if there's padding and `T: !Immutable`, then this exposes unsoundness
- In both the `&` case (where `T: !Immutable`) and the `&mut` case, if you use `left` to overwrite padding, then once you have dropped the reference, you un-shadow the original, which now spuriously points to uninit bytes where there should be none

Thus, we're going to remove `left`, `right`, and `reborrow`.

src/split_at.rs
```diff
@@ -233,59 +213,611 @@ unsafe impl<T> SplitAt for [T] {
...
+    /// // to return concurrent references to `packet` and `rest`. Note that
+    /// // this is a case of `via_runtime_check` being overly conservative.
+    /// // Although the left and right parts indeed overlap, the `Immutable`
+    /// // bound ensures that concurrently referencing these overlapping
+    /// // parts is sound.
+    /// assert!(split.via_runtime_check().is_err());
+    /// ```
+    #[must_use = "has no side effects"]
+    #[inline(always)]
+    pub fn via_runtime_check(self) -> Result<(&'a T, &'a [T::Elem]), Self> {
```
jswrenn, line 463: Do we want to rename this to accommodate that we might make this a const fn in the future?

src/split_at.rs
```diff
@@ -233,59 +213,611 @@ unsafe impl<T> SplitAt for [T] {
...
+/// [`Self::via_immutable`], [`Self::via_into_bytes`], or
+/// [`Self::via_unaligned`], or with a dynamic check by using
+/// [`Self::via_runtime_check`].
+#[derive(Debug)]
+pub struct Split<T> {
+    /// A pointer to the source slice DST.
+    ///
+    /// # Safety
+    ///
+    /// `source` is a pointer to a slice DST; see also `Self::l_len`.
```
joshlf, line 234: This is redundant since we know at each impl site that `T = &U` or `T = &mut U`, and that `U: SplitAt`, which implies `U: KnownLayout<PointerMetadata = usize>`. Instead, we should remove this and have `l_len` promse that "If `source` is a pointer to a slice DST, then...".

src/split_at.rs
```diff
@@ -233,59 +213,611 @@ unsafe impl<T> SplitAt for [T] {
...
-    //    `right: [T::Elem]` is derived from the trailing slice of `ptr`, which,
-    //    by contract on `T: SplitAt::Elem`, has type `[T::Elem]`.
-    let right = unsafe { Ptr::from_inner(right) };
-
-    (left, right)
+    #[inline(always)]
+    fn into_ptr(self) -> Split<Ptr<'a, T, (Shared, Aligned, Valid)>> {
+        // SAFETY: `Ptr::from_ref(self.source)` points to exactly `self.source`
+        // and thus maintains the invariants of `self` with respect to `l_len`.
+        unsafe { Split::new(Ptr::from_ref(self.source), self.l_len) }
```
joshlf, line 264: Assign `Ptr::from_ref(self.source)` to a variable outside of the `unsafe` block so it's clear what the `unsafe` block is guarding?

src/split_at.rs
```diff
@@ -233,59 +213,611 @@ unsafe impl<T> SplitAt for [T] {
...
+        let (l, r) = self.into_ptr().via_unaligned();
+        (l.as_ref(), r.as_ref())
+    }
+
+    /// Produces the split parts of `self`, using a dynamic check to ensure that
+    /// it is sound to have concurrent references to both parts. You should
+    /// prefer using [`Self::via_immutable`], [`Self::via_into_bytes`], or
+    /// [`Self::via_unaligned`], which have no runtime cost.
+    ///
+    /// Note, this check is overly conservative if `T` is [`Immutable`]; for
```
joshlf, line 407:
```suggestion
    /// Note that this check is overly conservative if `T` is [`Immutable`]; for
```

src/split_at.rs
joshlf: Discussed offline: We'll keep it as `via_runtime_check`.

src/split_at.rs
```diff
@@ -233,59 +213,611 @@ unsafe impl<T> SplitAt for [T] {
...
+
+impl<'a, T> Split<&'a mut T>
+where
+    T: ?Sized + SplitAt,
+{
+    #[inline(always)]
+    fn into_ptr(self) -> Split<Ptr<'a, T, (Exclusive, Aligned, Valid)>> {
+        // SAFETY: `Ptr::from_mut(self.source)` points to exactly `self.source`,
+        // and thus maintains the invariants of `self` with respect to `l_len`.
+        unsafe { Split::new(Ptr::from_mut(self.source), self.l_len) }
```
joshlf, line 499: Same here as above regarding moving `Ptr::from_mut` into a variable out of the `unsafe` block.

src/split_at.rs
```diff
@@ -233,59 +213,611 @@ unsafe impl<T> SplitAt for [T] {
...
+        unsafe { Split::new(self.source.unify_invariants().as_mut(), self.l_len) }
+    }
+
+    /// Produces the length of `self`'s left part.
+    #[inline(always)]
+    fn l_len(&self) -> MetadataOf<T> {
+        // SAFETY: By invariant on `Split`, `self.l_len` is not greater than the
+        // length of `self.source`.
+        unsafe { MetadataOf::<T>::new_unchecked(self.l_len) }
+    }
```
joshlf, line 710: Maybe we could just make `l_len` be a `MetadataOf<T>`?

src/split_at.rs
```diff
@@ -233,59 +213,611 @@ unsafe impl<T> SplitAt for [T] {
...
+
+    /// Produces the split parts of `self`, using [`Unaligned`] to ensure that
+    /// it is sound to have concurrent references to both parts.
+    #[inline(always)]
+    fn via_unaligned(self) -> (Ptr<'a, T, I>, Ptr<'a, [T::Elem], I>)
+    where
+        T: Unaligned,
+    {
+        // SAFETY: By `T: SplitAt + Unaligned`, `T` is either a slice or a
+        // `repr(C)` or `repr(transparent)` slice DST, that is well-aligned at
```
joshlf, line 745:
```suggestion
        // `repr(C)` or `repr(transparent)` slice DST that is well-aligned at
```

src/split_at.rs
```diff
@@ -233,59 +213,611 @@ unsafe impl<T> SplitAt for [T] {
...
+        // SAFETY:
+        // 0. `right` conforms to the aliasing invariant of `I::Aliasing`, by Lemma
+        //    0.
+        // 1. `right` conforms to the alignment invariant of `I::Alignment, because
+        //    if `ptr` with `I::Alignment = Aligned`, then by invariant on `T:
+        //    SplitAt`, the trailing slice of `ptr` (from which `right` is derived)
+        //    will also be well-aligned.
+        // 2. `right` conforms to the validity invariant of `I::Validity`, because
+        //    `right: [T::Elem]` is derived from the trailing slice of `ptr`, which,
+        //    by contract on `T: SplitAt::Elem`, has type `[T::Elem]`.
```
joshlf, line 816: Also need to prove that `left` won't be used to invalidate validity of `right`

joshlf (changes requested):

src/split_at.rs
jswrenn: Discussed offline. Regrettably impossible without adding additional bounds.

src/split_at.rs
jswrenn: https://github.com/google/zerocopy/compare/b21d06c408367bc92de75f99017fa67dad9e8aa4..187c1794711c6487fe424506c9132cf9b92f8576#diff-394d6458153031257865a7aaddcda3ac55a4ae61c642dee241136b08a66fcf90R236

src/split_at.rs
jswrenn: https://github.com/google/zerocopy/compare/b21d06c408367bc92de75f99017fa67dad9e8aa4..187c1794711c6487fe424506c9132cf9b92f8576#diff-394d6458153031257865a7aaddcda3ac55a4ae61c642dee241136b08a66fcf90R259-R262

src/split_at.rs
jswrenn: https://github.com/google/zerocopy/compare/b21d06c408367bc92de75f99017fa67dad9e8aa4..187c1794711c6487fe424506c9132cf9b92f8576#diff-394d6458153031257865a7aaddcda3ac55a4ae61c642dee241136b08a66fcf90R405

src/split_at.rs
jswrenn: https://github.com/google/zerocopy/compare/b21d06c408367bc92de75f99017fa67dad9e8aa4..187c1794711c6487fe424506c9132cf9b92f8576#diff-394d6458153031257865a7aaddcda3ac55a4ae61c642dee241136b08a66fcf90R744

src/split_at.rs
jswrenn: https://github.com/google/zerocopy/compare/b21d06c408367bc92de75f99017fa67dad9e8aa4..187c1794711c6487fe424506c9132cf9b92f8576#diff-394d6458153031257865a7aaddcda3ac55a4ae61c642dee241136b08a66fcf90R813-R819

joshlf (approved):

PR #2574, opened 2025-06-09
Support x86 AVX-12 SIMD types on 1.89.0 stable

joshlf:
Previously, these types were gated by the `stdarch_x86_avx512` nightly
Rust feature. As of 1.89.0, they are supported on stable.

gherrit-pr-id: I6907ba8f144bd90d917b85c9687443fc7cf19de2

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,7 @@ jobs:
           # These are the names of specific Rust versions detected in
           # `build.rs`. Each of these represents the minimum Rust version for
           # which a particular feature is supported.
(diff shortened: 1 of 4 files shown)


jswrenn (approved):

joshlf:
Blocked on #2583.

usamoi:
In these four types `__m512bh`, `__m512`, `__m512d`, and `__m512i`, only `__m512bh` was stabilized in Rust 1.89, while the other three were stabilized in Rust 1.72. Does it make sense to lower the required Rust version to 1.72?

See also: https://github.com/rust-lang/stdarch/pull/1436

PR #2577, opened 2025-06-09
Simplify some Ptr operations

joshlf:
Some raw pointer casts now have the signature `fn(PtrInner) -> PtrInner`
instead of `fn(PtrInner) -> NonNull`. Simplify the implementations of
some `Ptr` methods.

---

This PR is on branch [transmute-ref-dst](../tree/transmute-ref-dst).

- #2570
- #2487
- #2472
- #2577

### DIFF
--- a/src/pointer/inner.rs
+++ b/src/pointer/inner.rs
@@ -6,7 +6,7 @@
 // This file may not be copied, modified, or distributed except according to
 // those terms.
 
-use core::{marker::PhantomData, ops::Range, ptr::NonNull};
+use core::{marker::PhantomData, mem, ops::Range, ptr::NonNull};
 
 #[allow(unused_imports)]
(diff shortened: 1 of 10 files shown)


src/pointer/inner.rs
```diff
@@ -6,7 +6,7 @@
 // This file may not be copied, modified, or distributed except according to
 // those terms.
 
-use core::{marker::PhantomData, ops::Range, ptr::NonNull};
+use core::{marker::PhantomData, mem, ops::Range, ptr::NonNull};
```
jswrenn, line 9: You'll need a `#[cfg(doc)] use core::ptr::NonNull;` to get CI to pass.

joshlf: Done

PR #2581, opened 2025-06-09
[macros] Support shrinking value transmutes

joshlf:
In `transmute!`, support an `#![allow(shrink)]` attribute which is
invoked as follows:

  transmute!(#![allow(shrink)] src);

When this attribute is provided, `transmute!` will permit shrinking
transmutes, in which the destination value may be smaller than the
source value.

gherrit-pr-id: I46b18b4b1d10507b7e1d2e01b09dc4960cfcdce1

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -51,39 +51,58 @@
 /// This macro can be invoked in `const` contexts.
 #[macro_export]
 macro_rules! transmute {
-    ($e:expr) => {{
-        // NOTE: This must be a macro (rather than a function with trait bounds)
-        // because there's no way, in a generic context, to enforce that two
-        // types have the same size. `core::mem::transmute` uses compiler magic
-        // to enforce this so long as the types are concrete.
+    // NOTE: This must be a macro (rather than a function with trait bounds)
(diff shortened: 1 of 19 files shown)


src/macros.rs
```diff
@@ -51,39 +51,109 @@
 /// This macro can be invoked in `const` contexts.
 #[macro_export]
 macro_rules! transmute {
```
jswrenn, line 64: Need to document `allow(shrink)`.

joshlf: Done.

src/macros.rs
```diff
@@ -35,6 +35,17 @@
 /// original `Src` will be forgotten, and the value of type `Dst` will be
 /// returned.
 ///
+/// # `#![allow(shrink)]`
+///
+/// If `#![allow(shrink)]` is provided, `transmute!` additionally supports
+/// transmutations that shrink the size of the referent; e.g.:
```
jswrenn:
```suggestion
/// transmutations that shrink the size of the value; e.g.:
```

jswrenn (changes requested):

jswrenn (approved):

PR #2602, opened 2025-06-21
Fix typo in error messages

kilavvy:
### **Description:**

This pull request corrects a recurring typo in our UI test error messages, changing "implementor" to "implementer".

This is a minor fix to improve the clarity and quality of our diagnostic output.


joshlf:
The `.stderr` files are generated from compiler output, so they need to exactly match the compiler. If the compiler uses that spelling, then that's what the `.stderr` files need to contain (that's what's causing the CI test failures).

PR #2616, opened 2025-07-01
[wip] Implement traits for tuples up to arity 12

### LINKED ISSUE #274: Implement traits for tuples
Although we do not do so currently, it's valid to implement `FromZeroes` and `FromBytes` for tuples of any size, so long as all of the elements of those tuples also implement those traits. Use this as a model for how to provide generic implementations of these traits: https://github.com/google/zerocopy/blob/0ff2dd15c958ab615b8094cf273f6b37d113c8fa/src/lib.rs#L1094-L1108

Note: We ***cannot*** implement `Unaligned` nor `AsBytes` for any tuple type other than `()` — all other tuple types have the [default representation](https://doc.rust-lang.org/reference/type-layout.html#the-default-representation), and so a) there is no upper bound on their alignment and, b) there are no guarantees made about the absence of padding.

Additionally, add tests for the presence/absence of these impls for various concrete tuple types (where the tuple element types are both sufficient and insufficient for generating the impl for the entire tuple). The `assert_impls!` macro may be helpful for this: https://github.com/google/zerocopy/blob/b083f1f12839b91a54460ab5d2518105f3fc5355/src/lib.rs#L4512

Following the lead of the standard library, we should implement these traits for tuples up to size 12.

jswrenn:
Fixes #274

### DIFF
--- a/src/impls.rs
+++ b/src/impls.rs
@@ -924,6 +924,207 @@ const _: () = unsafe {
 // [1] https://doc.rust-lang.org/core/option/enum.Option.html
 const _: () = unsafe { unsafe_impl!(T: Immutable => Immutable for Option<T>) };
 
+macro_rules! impl_traits_for_tuples {
+    ($(($($i:tt : $t:ident,)*),)*) => {
+        $(
+            // SAFETY: todo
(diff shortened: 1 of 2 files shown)


joshlf:
@jswrenn is this made obsolete by your recent tuple work? If so, close?

PR #2619, opened 2025-07-02
Add TryFromBytes::try_read_from_io

joshlf:
Makes progress on #158

gherrit-pr-id: Ia15467145b06dd57e4dfd9e127fdd6d7a313ecba

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2696,7 +2696,7 @@ pub unsafe trait TryFromBytes {
         };
         // SAFETY: `candidate` was copied from from `source: &[u8]`, so all of
         // its bytes are initialized.
-        unsafe { try_read_from(source, candidate) }
+        unsafe { try_read_from(source, candidate) }.map_err(TryReadError::Validity)
     }
 
     /// Attempts to read a `Self` from the prefix of the given `source`.
@@ -2757,7 +2757,10 @@ pub unsafe trait TryFromBytes {
         };
         // SAFETY: `candidate` was copied from from `source: &[u8]`, so all of
         // its bytes are initialized.
-        unsafe { try_read_from(source, candidate).map(|slf| (slf, suffix)) }
+        match unsafe { try_read_from(source, candidate) } {
+            Ok(slf) => Ok((slf, suffix)),
+            Err(err) => Err(TryReadError::Validity(err)),
+        }
     }
 
     /// Attempts to read a `Self` from the suffix of the given `source`.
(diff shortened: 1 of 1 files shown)


jswrenn:
I'm still working on exploring the problem space, presently by considering what a `try_mut_from_io` would look like. I *think* we could provide this API:
```rust
fn try_mut_from_io<R>(mut source: R, buffer: &mut CoreMaybeUninit<Self>) -> io::Result<Result<&mut Self, ValidityError<&[u8], Self>>>
    where
        Self: KnownLayout + Sized,
        R: io::Read,
```
Note the lack of an `IntoBytes` bound! If we can find a way to express this, them I'm less concerned about maximizing the expressivity of `try_read_from_io`. The ideal API there is one in which we can return a `Initialized<T>` in the error type, and until that happens, we'll just have to be satisfied with a sub-optimal API. Those wishing to have access to the buffer in the error branch or who wish to pass a buffer in can just use `try_mut_from_io`.

src/lib.rs
```diff
@@ -2819,7 +2822,77 @@ pub unsafe trait TryFromBytes {
...
+    ///
+    /// let mut file = File::open("image.bin").unwrap();
+    /// let header = BitmapFileHeader::try_read_from_io(&mut file)
+    ///     .expect("read failed")
+    ///     .expect("invalid header");
+    /// ```
+    // TODO: Write tests for this method
+    #[cfg(feature = "std")]
+    #[inline(always)]
+    fn try_read_from_io<R>(mut src: R) -> io::Result<Result<Self, ValidityError<[u8; 0], Self>>>
```
joshlf: Why not just `ValidityError<(), Self>`? `[u8; 0]` seems a bit unnecessary. I feel like `()` conveys "we're not providing a source value."

jswrenn: We'll want it to be something that's `Deref` so if it's embedded in a `ConvertError`, the whole package implements `Error`. So how about `&'static ()`.

joshlf: Eh, between `&'static ()` and `[u8; 0]`, I'd prefer `[u8; 0]`.

jswrenn: It would need to be `&'static [u8; 0]`

jswrenn: Which I think is worse, tbh.

joshlf: Oh I see. Yeah, agreed that `&'static ()` is better.

Issue #2541, opened 2025-05-16
Expose layout `fn`s in `KnownLayout`


kupiakos:
Until solutions like #2490 are present for DST construction (and even after), it'd be extremely convenient to get the correct `Layout` for a given DST with specific tail slice length, considering all of the challenges present in doing that calculation correctly.

Consider this utility type

```rust
```

kupiakos:
The safest and stablest approach I've been able to take is using the `TryFromBytes` machinery to construct the pointers, but it still requires that the user already know a valid size to use:

```rust

impl<T: TpmvRequest + FromZeros + KnownLayout<PointerMetadata = usize> + ?Sized>
    TpmRequestMessage<T>
{
    /// Constructs a `TpmRequestMessage` with zeroed `request` bytes and a correct header.
    ///
    /// `request_size` does not include the header.
    ///
    /// # Panics
    /// Panics if `size` is not a correct data size for `T` or if the total size exceed `isize::MAX`.
    pub fn new_zeroed(request_size: usize) -> Box<Self> {
        let total_size = size_of::<TpmRequestHeader>().checked_add(request_size).unwrap();
        assert!(total_size < isize::MAX as usize);
        let mut box_bytes: Box<[u8]> = Box::from(vec![0; total_size]);
        box_bytes[..size_of::<TpmRequestHeader>()]
            .copy_from_slice(TpmRequestHeader::for_request::<T>().as_bytes());
        match <Self as TryFromBytes>::try_mut_from_bytes(Box::leak(box_bytes)) {
            // zerocopy determines the correct trailing slice len
            Ok(x) => unsafe { Box::from_raw(x) },
            Err(e) => {
                let _ = unsafe { Box::<[u8]>::from_raw(e.into_src()) };
                panic!("request_size is not valid for the request type");
            }
        }
    }
}
```

PR #2674, opened 2025-08-13
Add `KnownLayout::size_for_metadata`

jswrenn:
Computes the size of an object of type `Self` with the given pointer metadata.

Makes progress towards #2541

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -95,6 +95,8 @@ zerocopy-derive = { version = "=0.8.26", path = "zerocopy-derive" }
 either = "=1.13.0"
 # FIXME(#381) Remove this dependency once we have our own layout gadgets.
 elain = "0.3.0"
+# More recent versions of `glob` have an MSRV higher than ours.
+glob = "=0.3.2"
 itertools = "0.11"
 rand = { version = "0.8.5", default-features = false, features = ["small_rng"] }
 rustversion = "1.0"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -821,7 +821,7 @@ pub unsafe trait KnownLayout {
         let meta = Self::pointer_to_metadata(ptr.as_ptr());
         // SAFETY: `size_for_metadata` promises to only return `None` if the
         // resulting size would not fit in a `usize`.
-        meta.size_for_metadata(Self::LAYOUT)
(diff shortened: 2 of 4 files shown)


src/lib.rs
```diff
@@ -831,6 +831,32 @@ pub unsafe trait KnownLayout {
...
+
+    /// Computes the size of an object of type `Self` with the given pointer
+    /// metadata.
+    ///
+    /// # Safety
+    ///
+    /// `size_for_metadata` promises to return `None` if and only if the
+    /// resulting size would not fit in a `usize`. Note that the returned size
+    /// could exceed the actual maximum valid size of an allocated object,
+    /// `isize::MAX`.
```
joshlf, line 843: Can you add an example that demonstrates this?

joshlf (changes requested):

src/lib.rs
jswrenn: Besides `assert_eq!(<[u16]>::size_for_metadata(usize::MAX), None);`?

joshlf: Nah I mean something where it returns `Some(x)` where `x > isize::MAX`. It'd be good to have such an example and call out (with a code comment or prose in the doc comment) that it's an example of a size which is invalid for any real object.

jswrenn: [Done](https://github.com/google/zerocopy/compare/353b5cbf72415fdedb5c1a9e05c65595e94876b1..bbbfc79988c84864de77b75a89d9dc0a6689430b#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R855-R856)

Cargo.toml
```diff
@@ -95,6 +95,8 @@ zerocopy-derive = { version = "=0.8.26", path = "zerocopy-derive" }
 either = "=1.13.0"
 # FIXME(#381) Remove this dependency once we have our own layout gadgets.
 elain = "0.3.0"
+# More recent versions of `glob` have an MSRV higher than ours.
+glob = "=0.3.2"
```
joshlf, line 99: Does this belong in this change?

jswrenn: Yeah, CI was breaking because a recent `glob` update increased its MSRV beyond ours.

Issue #1112, opened 2024-04-15
Improve `#[derive(IntoBytes)]` padding analysis on DSTs


jswrenn:
Many dynamically sized types can be `derive(FromBytes)` but not `derive(IntoBytes)` because of shortcomings in our padding analysis. For example:

```rust
#[derive(KnownLayout, FromBytes, IntoBytes)]
#[repr(C)]
struct Example {
    header: Header,
    body: [Body],
}
```

Presently, a user must either mark `Example` as `repr(C, packed)`, or they must model `Example` instead as a composite struct containing `Ref`s to the `Header` and `[Body]`. 

Can we make our analysis more sophisticated in this case?

joshlf:
In theory, the impl of `KnownLayout` contains enough information to determine whether it's guaranteed that there will never be any padding between any of the fields or after `body`. However, relying on that would mean adding the assumption that `KnownLayout` is also derived. For non-`repr(C)` types, this is fine (since `KnownLayout` isn't supported for such types, we would know not to even try). For `repr(C)` types, we would have to choose between either assuming `KnownLayout` or sticking with the current analysis, which doesn't permit types like this.

Note that even the problem of detecting whether or not a type is unsized is hard - there's no syntactic way to determine whether a concrete type is unsized.

joshlf:
See also: https://github.com/google/zerocopy/issues/1566

korran:
The inability to use IntoBytes derives with non-packed DSTs is significantly impacting code size on embedded targets that don't support unaligned-loads. In particular, some rv32imc CPUS trap on all unaligned loads, and require correct alignment.

I'm very interesting in solving this issue; any pointers / guidance?

joshlf:
Knowing that this is an issue for you, we can prioritize it on our side. Could you provide an example of a type that you're having this issue with?

korran:
Here's one example from the opentitan ownership config structs:

```
#[derive(Debug, Eq, PartialEq, FromBytes, Immutable, KnownLayout)]
#[repr(C)]
pub struct FlashConfig {
    pub header: TlvHeader,
    pub regions: [FlashRegion],
}
unsafe impl IntoBytes for FlashConfig {
    fn only_derive_is_allowed_to_implement_this_trait() {
        // TODO: Remove this when https://github.com/google/zerocopy/issues/1112 is fixed.
        const {
            assert!(offset_of!(FlashConfig, header) == 0);
            assert!(
[...]

PR #2679, opened 2025-08-15
`derive(IntoBytes)`: support `repr(C)` structs with explicit trailing slices

jswrenn:
For example:

    #[repr(C)]
    struct Example {
        leading: u8,
        trailing: [Trailing]
    }

Makes progress towards #1112

### DIFF
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -26,10 +26,16 @@ const POINTER_WIDTH_BITS: usize = mem::size_of::<usize>() * 8;
 /// the layout of real Rust types or instances of types.
 #[doc(hidden)]
 #[allow(missing_debug_implementations, missing_copy_implementations)]
-#[cfg_attr(any(kani, test), derive(Copy, Clone, Debug, PartialEq, Eq))]
+#[cfg_attr(any(kani, test), derive(Debug, PartialEq, Eq))]
+#[derive(Copy, Clone)]
 pub struct DstLayout {
     pub(crate) align: NonZeroUsize,
     pub(crate) size_info: SizeInfo,
+    // Is it guaranteed statically (without knowing a value's runtime metadata)
+    // that the top-level type contains no padding? This does *not* apply
+    // recursively - for example, `[(u8, u16)]` has `statically_shallow_unpadded
+    // = true` even though this type likely has padding inside each `(u8, u16)`.
+    pub(crate) statically_shallow_unpadded: bool,
 }
 
 #[cfg_attr(any(kani, test), derive(Debug, PartialEq, Eq))]
@@ -130,6 +136,22 @@ impl DstLayout {
         None => const_unreachable!(),
     };
 
+    /// Assumes that this layout lacks static shallow padding.
+    ///
+    /// # Panics
+    ///
+    /// This method does not panic.
+    ///
+    /// # Safety
+    ///
+    /// If `self` describes the size and alignment of type that lacks static
+    /// shallow padding, unsafe code may assume that the result of this method
+    /// accurately reflects the size, alignment, and lack of static shallow
+    /// padding of that type.
+    const fn assume_shallow_unpadded(self) -> Self {
+        Self { statically_shallow_unpadded: true, ..self }
+    }
+
     /// Constructs a `DstLayout` for a zero-sized type with `repr_align`
     /// alignment (or 1). If `repr_align` is provided, then it must be a power
     /// of two.
@@ -152,10 +174,15 @@ impl DstLayout {
 
         const_assert!(align.get().is_power_of_two());
 
-        DstLayout { align, size_info: SizeInfo::Sized { size: 0 } }
+        DstLayout {
+            align,
+            size_info: SizeInfo::Sized { size: 0 },
+            statically_shallow_unpadded: true,
+        }
     }
 
-    /// Constructs a `DstLayout` which describes `T`.
+    /// Constructs a `DstLayout` which describes `T` and assumes `T` may contain
+    /// padding.
     ///
     /// # Safety
     ///
@@ -166,14 +193,29 @@ impl DstLayout {
     pub const fn for_type<T>() -> DstLayout {
         // SAFETY: `align` is correct by construction. `T: Sized`, and so it is
         // sound to initialize `size_info` to `SizeInfo::Sized { size }`; the
-        // `size` field is also correct by construction.
+        // `size` field is also correct by construction. `unpadded` can safely
+        // default to `false`.
         DstLayout {
             align: match NonZeroUsize::new(mem::align_of::<T>()) {
                 Some(align) => align,
                 None => const_unreachable!(),
             },
             size_info: SizeInfo::Sized { size: mem::size_of::<T>() },
+            statically_shallow_unpadded: false,
         }
+    }
+
+    /// Constructs a `DstLayout` which describes a `T` that does not contain
+    /// padding.
+    ///
+    /// # Safety
+    ///
+    /// Unsafe code may assume that `DstLayout` is the correct layout for `T`.
+    #[doc(hidden)]
+    #[must_use]
+    #[inline]
+    pub const fn for_unpadded_type<T>() -> DstLayout {
+        Self::for_type::<T>().assume_shallow_unpadded()
     }
 
     /// Constructs a `DstLayout` which describes `[T]`.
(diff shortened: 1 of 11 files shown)


src/layout.rs
```diff
@@ -202,10 +203,71 @@ impl DstLayout {
         }
     }
 
+    /// Constructs a complete `DstLayout` reflecting a `repr(C)` struct with the
+    /// given alignment modifiers and fields.
+    ///
+    /// The accompanying boolean is `true` if the resulting composition of
+    /// fields necessitated static (as opposed to dynamic) padding; otherwise
+    /// false.
```
joshlf: Can we make this an enum so its meaning is clearer at the call site?

src/layout.rs
```diff
@@ -202,10 +203,71 @@ impl DstLayout {
...
+    ///
+    /// This method cannot be used to match the layout of a record with the
+    /// default representation, as that representation is mostly unspecified.
+    ///
+    /// # Safety
+    ///
+    /// For any definition `repr(C)` struct, if this method is invoked with
```
joshlf:
```
definition `repr(C)` struct
```

Typo?

src/layout.rs
```diff
@@ -202,10 +203,71 @@ impl DstLayout {
...
+    }
+
     /// Like `Layout::extend`, this creates a layout that describes a record
     /// whose layout consists of `self` followed by `next` that includes the
     /// necessary inter-field padding, but not any trailing padding.
     ///
+    /// The accompanying boolean is `true` if the resulting composition of
+    /// fields necessitated static (as opposed to dynamic) padding; otherwise
+    /// false.
+    ///
```
joshlf: Same here

src/layout.rs
```diff
@@ -295,59 +357,66 @@ impl DstLayout {
...
     }
 
     /// Like `Layout::pad_to_align`, this routine rounds the size of this layout
     /// up to the nearest multiple of this type's alignment or `repr_packed`
     /// (whichever is less). This method leaves DST layouts unchanged, since the
     /// trailing padding of DSTs is computed at runtime.
     ///
+    /// The accompanying boolean is `true` if the resulting composition of
+    /// fields necessitated static (as opposed to dynamic) padding; otherwise
+    /// false.
```
joshlf: Same here

src/layout.rs
```diff
@@ -844,7 +931,7 @@ mod tests {
                     core::iter::once(None).chain((0..29).map(|p| NonZeroUsize::new(2usize.pow(p))));
 
                 for pack in packs {
-                    let composite = base.extend(trailing_field, pack);
+                    let composite = base.extend(trailing_field, pack).1;
```
joshlf: To make this a bit more self-documenting:
```suggestion
                    let (_padding, composite) = base.extend(trailing_field, pack);
```

Same below in various places that `.1` is used.

src/layout.rs
```diff
@@ -1703,6 +1790,44 @@ mod proofs {
...
+    #[kani::proof]
+    fn prove_requires_dynamic_padding() {
+        let layout: DstLayout = kani::any();
+
+        let SizeInfo::SliceDst(size_info) = layout.size_info else {
+            kani::assume(false);
+            loop {}
+        };
+
+        //kani::assume(!layout.requires_dynamic_padding());
```
joshlf: Remove?

src/util/macro_util.rs
```diff
@@ -359,6 +359,31 @@ macro_rules! struct_has_padding {
...
 }
 
+/// Does the struct type `$t` have padding?
+///
+/// `$ts` is the list of the type of every field in `$t`. `$t` must be a
+/// `repr(C)` struct type, or else `struct_has_padding!`'s result may be
+/// meaningless.
+#[doc(hidden)] // `#[macro_export]` bypasses this module's `#[doc(hidden)]`.
+#[macro_export]
+macro_rules! cstruct_has_padding {
```
joshlf:
```suggestion
macro_rules! repr_c_struct_has_padding {
```

I found `cstruct` to be confusing until I saw `repr(C)` in the doc comment.

src/util/macro_util.rs
```diff
@@ -359,6 +359,31 @@ macro_rules! struct_has_padding {
     };
 }
 
+/// Does the struct type `$t` have padding?
```
joshlf:
```suggestion
/// Does the `repr(C)` struct type `$t` have padding?
```

zerocopy-derive/src/lib.rs
```diff
@@ -1288,6 +1289,8 @@ fn derive_into_bytes_struct(
         // check for in this branch's condition.
         (None, false)
     } else if ast.generics.params.is_empty() {
+        let is_syntactic_dst =
+            strct.fields().last().map(|(_, _, ty)| matches!(ty, Type::Slice(_))).unwrap_or(false);
```
joshlf, line 1294: Can you add a comment explaining this line? IME this won't be easy to read in the future when we've forgotten the context.

zerocopy-derive/src/lib.rs
```diff
@@ -1475,6 +1482,8 @@ enum PaddingCheck {
     /// Check that the sum of the fields' sizes exactly equals the struct's
     /// size.
     Struct,
+    /// Check that a `repr(C)` struct has no padding.
+    CStruct,
```
joshlf:
```suggestion
    ReprCStruct,
```

Same justification as the macro name

joshlf (changes requested):
Wooo!!!

Can you add UI tests that succeed but wouldn't have succeeded before this change? We may also want some new failure ones to make sure we're properly exercising the edge cases of our logic (namely, what if you have a non-syntactically-DST-but-actually-DST `repr(C)` type? What about a syntactically-DST non-`repr(C)` type? etc).

jswrenn:
> Can you add UI tests that succeed but wouldn't have succeeded before this change? We may also want some new failure ones to make sure we're properly exercising the edge cases of our logic (namely, what if you have a non-syntactically-DST-but-actually-DST `repr(C)` type? What about a syntactically-DST non-`repr(C)` type? etc).

[Done](https://github.com/google/zerocopy/compare/30192700eb00f4f13aeff12d9b48e135e8ff4295..07fd645369aa3b5dff4c73b64890c9c4176f479f)

src/layout.rs
jswrenn: For the two lines in the form here, I'm in favor of this change ([done](https://github.com/google/zerocopy/compare/07fd645369aa3b5dff4c73b64890c9c4176f479f..a692413d419371dccd6313bc0575f525066b31e3#diff-7937a85ae16def66de428ea5dc339277fc8ca7bb7e40c247dc0b485acc39da6cR934-R1002)), but we have eight instances that look basically exactly like this:
```rust
let expected = DstLayout::new_zst(None)
            .extend(DstLayout::for_type::<NotKnownLayout<AU32>>(), None)
            .1
            .extend(<[u8] as KnownLayout>::LAYOUT, None)
            .1
            .pad_to_align()
            .1;
```
I'm not sure much is gained by expanding each of those out and tripling the number of `let`s. Are you sure you want me to go forward with that change here?

src/layout.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/07fd645369aa3b5dff4c73b64890c9c4176f479f..a692413d419371dccd6313bc0575f525066b31e3#diff-7937a85ae16def66de428ea5dc339277fc8ca7bb7e40c247dc0b485acc39da6cR218)

src/util/macro_util.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/07fd645369aa3b5dff4c73b64890c9c4176f479f..a692413d419371dccd6313bc0575f525066b31e3#diff-2abf6454394ca2a0779f4bf7cc0631f23cf1dc484566b5937e8a1ad9a5b9e3e8R342-R362)

src/util/macro_util.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/07fd645369aa3b5dff4c73b64890c9c4176f479f..a692413d419371dccd6313bc0575f525066b31e3#diff-2abf6454394ca2a0779f4bf7cc0631f23cf1dc484566b5937e8a1ad9a5b9e3e8R369)

zerocopy-derive/src/lib.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/07fd645369aa3b5dff4c73b64890c9c4176f479f..a692413d419371dccd6313bc0575f525066b31e3#diff-977c29e6cf079bd7307f811df24f99633df603962785c583145eb6250b0e12daR1292-R1294)

src/layout.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/07fd645369aa3b5dff4c73b64890c9c4176f479f..a692413d419371dccd6313bc0575f525066b31e3#diff-7937a85ae16def66de428ea5dc339277fc8ca7bb7e40c247dc0b485acc39da6cL1802-L1803)

src/layout.rs
jswrenn: What's the typo here?

joshlf: Sorry, I meant - can we make this an enum so its meaning is clearer at the call site?

src/layout.rs
joshlf: Clarifying: Can we make this an enum so its meaning is clearer at the call site?

src/layout.rs
joshlf: Agreed; let's keep those as-is.

src/layout.rs
jswrenn: We do boolean operations on the 'is statically padded' component in both implementation (the first two snippets) and call-site (the final snippet):

- https://github.com/google/zerocopy/blob/a692413d419371dccd6313bc0575f525066b31e3/src/layout.rs#L241
- https://github.com/google/zerocopy/blob/a692413d419371dccd6313bc0575f525066b31e3/src/layout.rs#L260
- https://github.com/google/zerocopy/blob/a692413d419371dccd6313bc0575f525066b31e3/src/util/macro_util.rs#L376

I think these sites would be made less clear using a bespoke enum, since we'd need to compensate for the lack of `|=` and `||`. We could add `BitOr` and `Or` implementations, but that would require keeping a mental mapping of which variant of our bespoke enum was truthy and which is falsey.

I'm in favor of bespoke enums in argument position or in a tuple return position that would otherwise carry multiple tuples, but here — as the sole boolean produced by this method (and used in a _very_ boolean fashion) — I think just using a `bool` is clearer.

jswrenn: I know there's something of an anti-bool sentiment in engineering right now, but the two mostly widely-read anti-bool manifestos cite argument-position bools (not return position bools):
- https://www.luu.io/posts/2019-dont-use-booleans
- https://alexkondov.com/should-you-pass-boolean-to-functions/

joshlf: My concern is that the function names don't make it clear what the meaning of the `bool` is (`requires_dynamic_padding` is an obvious exception, and I'm fine with that returning a `bool`).

If we returned an enum, we could do a few things to make the usage site more concise:
- Implement `Eq`, permitting e.g. `static_padding |= (interfield_padding == Static)`
- Add a method, permitting e.g. `static_padding |= interfield_padding.is_static()`
- Probably others I'm not thinking of

joshlf: Obsoleted by recent changes.

src/layout.rs
joshlf: Obsoleted by recent changes.

src/layout.rs
joshlf: Obsoleted by recent changes.

Issue #1717, opened 2024-09-21
IntoBytes padding error message should include the number of padding bytes


joshlf:
https://github.com/google/zerocopy/pull/1712#pullrequestreview-2319746468

PR #2699, opened 2025-08-29
[derive] `IntoBytes` padding error says number of bytes

joshlf:
While we're here, always say "padding" instead of "inter-field padding",
as all kinds (structs, enums, and unions) can have trailing padding too.

As an example, the diff for
`zerocopy-derive/tests/ui-stable/enum.stderr` contains:

-error[E0277]: `IntoBytes3` has inter-field padding
+error[E0277]: `IntoBytes3` has 2 total byte(s) of padding

Closes https://github.com/google/zerocopy/issues/1717


jswrenn (approved):
Very cool!

Issue #2752, opened 2025-11-03
Optimize `Ref<_, T>` methods when `T: Sized`


joshlf:
## Progress

- [x] Add benchmarks for `Ref` methods
- [x] Optimize methods when `T: Sized`

## Details

As described in #368, `Ref<B, T>` currently stores a `B` with the invariant that the referent has a valid size and alignment for `T`. This allows us to infallibly produce a `&(mut) T` at runtime (given an existing `Ref` as a witness to the size and alignment guarantees). Currently, we use `Ptr::try_cast_into_no_leftover` to re-compute the `&(mut) T` and `.expect` on its result since we know it can't fail:

https://github.com/google/zerocopy/blob/e8eb59576a50076b0dcd493d48fbdcb7f23b3a40/src/ref.rs#L624-L626

This is expensive. In the long run, we should be able to remove _most_ of this computation by storing the `&(mut) T` itself rather than the `B` (see #368), although that only works for some `B` types.

For the time being, and in order to support _all_ `B` types, we can at least optimize this for the `T: Sized` case. In particular, since a `&(mut) T` has no pointer metadata when `T: Sized`, we can just short-circuit and cast the `B` pointer directly to a `&(mut) T` without performing any runtime checks. In other words, this entirely bypasses `Ptr::try_cast_into_no_leftover`.

We should also add benchmarks so we can measure the performance impact of this change, which I expect will be substantial.

PR #2758, opened 2025-11-08
Optimize Ref<_, T> methods when T: Sized (#2752)

bmberger:
With optimization:
test r#ref::benches::bench_deref_mut_sized  ... bench:           0.64 ns/iter (+/- 0.25)
test r#ref::benches::bench_deref_sized      ... bench:           0.54 ns/iter (+/- 0.20)
test r#ref::benches::bench_from_bytes_sized ... bench:           5.71 ns/iter (+/- 1.59)
test r#ref::benches::bench_into_mut_sized   ... bench:           0.66 ns/iter (+/- 0.05)
test r#ref::benches::bench_into_ref_sized   ... bench:           0.88 ns/iter (+/- 0.08)

Without optimization:
test r#ref::benches::bench_deref_mut_sized  ... bench:           0.74 ns/iter (+/- 0.09)
test r#ref::benches::bench_deref_sized      ... bench:           0.69 ns/iter (+/- 0.10)
test r#ref::benches::bench_from_bytes_sized ... bench:           5.09 ns/iter (+/- 0.42)
test r#ref::benches::bench_into_mut_sized   ... bench:           0.68 ns/iter (+/- 0.08)
test r#ref::benches::bench_into_ref_sized   ... bench:           0.88 ns/iter (+/- 0.01)

Closes https://github.com/google/zerocopy/issues/2752

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -319,12 +319,16 @@
     any(__ZEROCOPY_INTERNAL_USE_ONLY_NIGHTLY_FEATURES_IN_TESTS, miri),
     feature(layout_for_ptr)
 )]
+#![cfg_attr(all(test, __ZEROCOPY_INTERNAL_USE_ONLY_NIGHTLY_FEATURES_IN_TESTS), feature(test))]
 
 // This is a hack to allow zerocopy-derive derives to work in this crate. They
 // assume that zerocopy is linked as an extern crate, so they access items from
 // it as `zerocopy::Xxx`. This makes that still work.
 #[cfg(any(feature = "derive", test))]
 extern crate self as zerocopy;
+
+#[cfg(all(test, __ZEROCOPY_INTERNAL_USE_ONLY_NIGHTLY_FEATURES_IN_TESTS))]
+extern crate test;
 
 #[doc(hidden)]
 #[macro_use]
--- a/src/ref.rs
+++ b/src/ref.rs
@@ -6,8 +6,10 @@
 // license <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your option.
 // This file may not be copied, modified, or distributed except according to
 // those terms.
-
 use super::*;
+use crate::pointer::{
+    BecauseInvariantsEq, BecauseMutationCompatible, MutationCompatible, TransmuteFromPtr,
+};
 
 mod def {
     use core::marker::PhantomData;
@@ -193,6 +195,11 @@ mod def {
 #[allow(unreachable_pub)] // This is a false positive on our MSRV toolchain.
 pub use def::Ref;
 
+use crate::pointer::{
+    invariant::{Aligned, BecauseExclusive, Initialized, Unaligned, Valid},
+    BecauseRead, PtrInner,
+};
+
 impl<B, T> Ref<B, T>
 where
     B: ByteSlice,
@@ -272,7 +279,7 @@ where
     /// `T` may be a sized type, a slice, or a [slice DST][slice-dst].
     ///
     /// [valid-size]: crate::KnownLayout#what-is-a-valid-size
-    /// [t-unaligned]: Unaligned
+    /// [t-unaligned]: crate::Unaligned
     /// [size-error-from]: error/struct.SizeError.html#method.from-1
     /// [slice-dst]: KnownLayout#dynamically-sized-types
     ///
@@ -326,7 +333,7 @@ where
     /// `T` may be a sized type, a slice, or a [slice DST][slice-dst].
     ///
     /// [valid-size]: crate::KnownLayout#what-is-a-valid-size
-    /// [t-unaligned]: Unaligned
+    /// [t-unaligned]: crate::Unaligned
     /// [size-error-from]: error/struct.SizeError.html#method.from-1
     /// [slice-dst]: KnownLayout#dynamically-sized-types
     ///
@@ -391,7 +398,7 @@ where
     /// `T` may be a sized type, a slice, or a [slice DST][slice-dst].
     ///
     /// [valid-size]: crate::KnownLayout#what-is-a-valid-size
-    /// [t-unaligned]: Unaligned
+    /// [t-unaligned]: crate::Unaligned
     /// [size-error-from]: error/struct.SizeError.html#method.from-1
     /// [slice-dst]: KnownLayout#dynamically-sized-types
     ///
@@ -454,7 +461,7 @@ where
     /// aligned, this returns `Err`. If [`T: Unaligned`][t-unaligned], you can
     /// [infallibly discard the alignment error][size-error-from].
     ///
-    /// [t-unaligned]: Unaligned
+    /// [t-unaligned]: crate::Unaligned
     /// [size-error-from]: error/struct.SizeError.html#method.from-1
     ///
     /// # Compile-Time Assertions
@@ -505,7 +512,7 @@ where
     /// Unaligned`][t-unaligned], you can [infallibly discard the alignment
     /// error][size-error-from].
     ///
-    /// [t-unaligned]: Unaligned
+    /// [t-unaligned]: crate::Unaligned
     /// [size-error-from]: error/struct.SizeError.html#method.from-1
     ///
     /// # Compile-Time Assertions
@@ -551,7 +558,7 @@ where
     /// Unaligned`][t-unaligned], you can [infallibly discard the alignment
     /// error][size-error-from].
     ///
-    /// [t-unaligned]: Unaligned
+    /// [t-unaligned]: crate::Unaligned
     /// [size-error-from]: error/struct.SizeError.html#method.from-1
     ///
     /// # Compile-Time Assertions
@@ -616,6 +623,20 @@ where
         // SAFETY: We don't call any methods on `b` other than those provided by
         // `IntoByteSlice`.
         let b = unsafe { r.into_byte_slice() };
+        let b = b.into_byte_slice();
+
+        if let crate::layout::SizeInfo::Sized { .. } = T::LAYOUT.size_info {
+            let ptr = Ptr::from_ref(b);
+            // SAFETY: We just checked that `T: Sized`. By invariant on `r`,
+            // `b`'s size is equal to `size_of::<T>()`.
+            let ptr = unsafe { cast_for_sized::<T, _, _, _>(ptr) };
+
+            // SAFETY: None of the preceding transformations modifies the
+            // address of the pointer, and by invariant on `r`, we know that it
+            // is validly-aligned.
+            let ptr = unsafe { ptr.assume_alignment::<Aligned>() };
+            return ptr.as_ref();
+        }
 
         // PANICS: By post-condition on `into_byte_slice`, `b`'s size and
         // alignment are valid for `T`. By post-condition, `b.into_byte_slice()`
@@ -650,6 +671,27 @@ where
         // SAFETY: We don't call any methods on `b` other than those provided by
         // `IntoByteSliceMut`.
         let b = unsafe { r.into_byte_slice_mut() };
+        let b = b.into_byte_slice_mut();
+
+        if let crate::layout::SizeInfo::Sized { .. } = T::LAYOUT.size_info {
+            let ptr = Ptr::from_mut(b);
+            // SAFETY: We just checked that `T: Sized`. By invariant on `r`,
+            // `b`'s size is equal to `size_of::<T>()`.
+            let ptr = unsafe {
+                cast_for_sized::<
+                    T,
+                    _,
+                    (BecauseRead, (BecauseExclusive, BecauseExclusive)),
+                    (BecauseMutationCompatible, BecauseInvariantsEq),
+                >(ptr)
+            };
+
+            // SAFETY: None of the preceding transformations modifies the
+            // address of the pointer, and by invariant on `r`, we know that it
+            // is validly-aligned.
+            let ptr = unsafe { ptr.assume_alignment::<Aligned>() };
+            return ptr.as_mut();
+        }
 
         // PANICS: By post-condition on `into_byte_slice_mut`, `b`'s size and
         // alignment are valid for `T`. By post-condition,
@@ -763,11 +805,25 @@ where
         // SAFETY: We don't call any methods on `b` other than those provided by
         // `ByteSlice`.
         let b = unsafe { self.as_byte_slice() };
+        let b = b.deref();
+
+        if let crate::layout::SizeInfo::Sized { .. } = T::LAYOUT.size_info {
+            let ptr = Ptr::from_ref(b);
+            // SAFETY: We just checked that `T: Sized`. By invariant on `r`,
+            // `b`'s size is equal to `size_of::<T>()`.
+            let ptr = unsafe { cast_for_sized::<T, _, _, _>(ptr) };
+
+            // SAFETY: None of the preceding transformations modifies the
+            // address of the pointer, and by invariant on `r`, we know that it
+            // is validly-aligned.
+            let ptr = unsafe { ptr.assume_alignment::<Aligned>() };
+            return ptr.as_ref();
+        }
 
         // PANICS: By postcondition on `as_byte_slice`, `b`'s size and alignment
         // are valid for `T`, and by invariant on `ByteSlice`, these are
         // preserved through `.deref()`, so this `unwrap` will not panic.
-        let ptr = Ptr::from_ref(b.deref())
+        let ptr = Ptr::from_ref(b)
             .try_cast_into_no_leftover::<T, BecauseImmutable>(None)
             .expect("zerocopy internal error: Deref::deref should be infallible");
         let ptr = ptr.recall_validity();
@@ -791,12 +847,33 @@ where
         // SAFETY: We don't call any methods on `b` other than those provided by
         // `ByteSliceMut`.
         let b = unsafe { self.as_byte_slice_mut() };
+        let b = b.deref_mut();
+
+        if let crate::layout::SizeInfo::Sized { .. } = T::LAYOUT.size_info {
+            let ptr = Ptr::from_mut(b);
+            // SAFETY: We just checked that `T: Sized`. By invariant on `r`,
+            // `b`'s size is equal to `size_of::<T>()`.
+            let ptr = unsafe {
+                cast_for_sized::<
+                    T,
+                    _,
+                    (BecauseRead, (BecauseExclusive, BecauseExclusive)),
+                    (BecauseMutationCompatible, BecauseInvariantsEq),
+                >(ptr)
+            };
+
+            // SAFETY: None of the preceding transformations modifies the
+            // address of the pointer, and by invariant on `r`, we know that it
+            // is validly-aligned.
+            let ptr = unsafe { ptr.assume_alignment::<Aligned>() };
+            return ptr.as_mut();
+        }
 
         // PANICS: By postcondition on `as_byte_slice_mut`, `b`'s size and
         // alignment are valid for `T`, and by invariant on `ByteSlice`, these
         // are preserved through `.deref_mut()`, so this `unwrap` will not
         // panic.
-        let ptr = Ptr::from_mut(b.deref_mut())
+        let ptr = Ptr::from_mut(b)
             .try_cast_into_no_leftover::<T, BecauseExclusive>(None)
             .expect("zerocopy internal error: DerefMut::deref_mut should be infallible");
         let ptr = ptr.recall_validity::<_, (_, (_, (BecauseExclusive, BecauseExclusive)))>();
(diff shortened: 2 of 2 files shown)


src/ref.rs
```diff
@@ -879,6 +999,48 @@ mod tests {
 
     use super::*;
     use crate::util::testutil::*;
+    use test::{self, Bencher};
+
+    #[bench]
+    fn bench_from_bytes_sized(b: &mut Bencher) {
+        let buf = Align::<[u8; 8], AU64>::default();
+        // `buf.t` should be aligned to 8, so this should always succeed.
+        b.iter(|| test::black_box(Ref::<_, AU64>::from_bytes(&buf.t[..]).unwrap()));
```
joshlf: I think this may be allowing more optimization than you want. Try running the _input_ through `black_box` as well?


```suggestion
        let bytes = &buf.t[..];
        b.iter(|| test::black_box(Ref::<_, AU64>::from_bytes(test::black_box(bytes)).unwrap()));
```

(Same w/ other benchmarks)

src/ref.rs
joshlf, line 645: Since you added `let b = b.into_bytes_slice()` above.

```suggestion
        let ptr = Ptr::from_ref(b)
```

Same in other methods.

src/ref.rs
```diff
@@ -616,6 +616,36 @@ where
...
+        let b = b.into_byte_slice();
+
+        if let crate::layout::SizeInfo::Sized { .. } = T::LAYOUT.size_info {
+            let ptr = Ptr::from_ref(b);
+            // SAFETY: `T::raw_from_ptr_len` promises to preserve address and provenance in the referent
+            let ptr: Ptr<'_, T, _> = unsafe {
+                ptr.transmute_unchecked(|ptr| {
+                    let ptr = T::raw_from_ptr_len(
+                        ptr.as_non_null().cast(),
+                        <T::PointerMetadata as crate::PointerMetadata>::from_elem_count(0),
```
joshlf: Can you add a comment here explaining that this is redundant (since we've checked that `T: Sized` in this if branch's conditional), but we have to do it because the type system doesn't know that `T: Sized`?

Same in other functions.

src/ref.rs
```diff
@@ -616,6 +616,36 @@ where
...
+                    //    `ptr` is guaranteed to live for at least `'a`, because `ptr`
+                    //    is entirely contained in `self`, which lives for at least `'a`
+                    //    by invariant on `Ptr`.
+                    unsafe { crate::pointer::PtrInner::new(ptr) }
+                })
+            };
+            // SAFETY: By invariant on `r`, we know that size and alignment are valid.
+            let ptr = unsafe { ptr.assume_alignment() };
+            return ptr.as_ref();
+        }
```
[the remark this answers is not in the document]
joshlf: This is broadly a good suggestion, but we should be able to do it via a function instead of a macro. Off the top of my head, I'd expect this to have roughly the signature:

```rust
unsafe fn try_sized_fast_path<T, B>(bytes: &[u8]) -> Result<Ptr<T>, B> { ... }
```

You'll need to propagate the size and alignment safety invariants on `Ref` to be safety preconditions on this function.

We'll also need to figure out a way to be generic over `&[u8]` vs `&mut [u8]`, and to have the resulting `Ptr` similarly be generic over `Shared` vs `Exclusive` aliasing.

src/ref.rs
```diff
@@ -616,6 +616,36 @@ where
         // SAFETY: We don't call any methods on `b` other than those provided by
         // `IntoByteSlice`.
         let b = unsafe { r.into_byte_slice() };
+        let b = b.into_byte_slice();
+
+        if let crate::layout::SizeInfo::Sized { .. } = T::LAYOUT.size_info {
+            let ptr = Ptr::from_ref(b);
+            // SAFETY: `T::raw_from_ptr_len` promises to preserve address and provenance in the referent
```
joshlf: Unfortunately the safety precondition of `transmute_unchecked` is [quite gnarly](https://google.github.io/zerocopy/zerocopy/pointer/ptr/def/struct.Ptr.html#method.transmute_unchecked). I think it'll make sense for us to discuss and figure out a way to not have to use `transmute_unchecked`.

src/ref.rs
```diff
@@ -879,6 +993,53 @@ mod tests {
...
+    }
+
+    #[bench]
+    fn bench_into_ref_sized(b: &mut Bencher) {
+        let buf = Align::<[u8; 8], AU64>::default();
+        let bytes = &buf.t[..];
+        b.iter(|| {
+            let r = Ref::<_, AU64>::from_bytes(test::black_box(bytes)).unwrap();
+            test::black_box(Ref::into_ref(r))
+        });
```
joshlf: So that only the `into_ref` part is being benchmarked:

```suggestion
        let r = Ref::<_, AU64>::from_bytes(bytes).unwrap();
        b.iter(|| {
            test::black_box(Ref::into_ref(test::black_box(r)))
        });
```

src/ref.rs
```diff
@@ -879,6 +993,53 @@ mod tests {
...
+    }
+
+    #[bench]
+    fn bench_into_mut_sized(b: &mut Bencher) {
+        let mut buf = Align::<[u8; 8], AU64>::default();
+        // TODO change these
+        b.iter(|| {
+            let r = Ref::<_, AU64>::from_bytes(test::black_box(&mut buf.t[..])).unwrap();
+            test::black_box(Ref::into_mut(r));
+        });
```
joshlf: Normally I'd suggest doing something like this:

```suggestion
        let r = Ref::<_, AU64>::from_bytes(test::black_box(&mut buf.t[..])).unwrap();
        b.iter(move || {
            test::black_box(Ref::into_mut(test::black_box(r)));
        });
```

Unfortunately, that won't work because you can't create multiple `Ref`s which mutably reference the same memory. @jswrenn may have ideas here, but I think for a benchmark it'd be fine to just unsafely construct a clone of `r` in the closure body – it's sound since we don't do anything with the value returned from `Ref::into_mut`.

src/ref.rs
```diff
@@ -879,6 +993,53 @@ mod tests {
...
+    }
+
+    #[bench]
+    fn bench_deref_sized(b: &mut Bencher) {
+        let buf = Align::<[u8; 8], AU64>::default();
+        let bytes = &buf.t[..];
+        let r = Ref::<_, AU64>::from_bytes(test::black_box(bytes)).unwrap();
+        b.iter(|| test::black_box(r.deref()));
```
joshlf: `black_box` is only important inside of `b.iter`, but it *is* important that we `black_box` `r` itself.

```suggestion
        let r = Ref::<_, AU64>::from_bytes(bytes).unwrap();
        b.iter(|| test::black_box(test::black_box(r).deref()));
```

src/ref.rs
```diff
@@ -879,6 +993,53 @@ mod tests {
...
+    }
+
+    #[bench]
+    fn bench_deref_mut_sized(b: &mut Bencher) {
+        let mut buf = Align::<[u8; 8], AU64>::default();
+        // TODO change these
+        b.iter(|| {
+            let mut r = Ref::<_, AU64>::from_bytes(test::black_box(&mut buf.t[..])).unwrap();
+            test::black_box(r.deref_mut());
+        });
```
joshlf: Same as above, but also moving construction outside of `b.iter` so we don't benchmark it.

```suggestion
        let mut r = Ref::<_, AU64>::from_bytes(&mut buf.t[..]).unwrap();
        b.iter(|| {
            test::black_box(test::black_box(r).deref_mut());
        });
```

joshlf (changes requested):

src/lib.rs
```diff
@@ -334,13 +334,17 @@
     __ZEROCOPY_INTERNAL_USE_ONLY_NIGHTLY_FEATURES_IN_TESTS,
     feature(layout_for_ptr, coverage_attribute)
 )]
+#![cfg_attr(test, feature(test))]
```
joshlf:
```suggestion
#![cfg_attr(all(test, __ZEROCOPY_INTERNAL_USE_ONLY_NIGHTLY_FEATURES_IN_TESTS), feature(test))]
```

This ensures that we only compile benchmarks on the nightly compiler, which is required since `feature(test)` is an unstable (ie, nightly-only) Rust feature.

src/lib.rs
```diff
@@ -334,13 +334,17 @@
...
 
 // This is a hack to allow zerocopy-derive derives to work in this crate. They
 // assume that zerocopy is linked as an extern crate, so they access items from
 // it as `zerocopy::Xxx`. This makes that still work.
 #[cfg(any(feature = "derive", test))]
 extern crate self as zerocopy;
 
+#[cfg(test)]
```
joshlf:
```suggestion
#[cfg(all(test, __ZEROCOPY_INTERNAL_USE_ONLY_NIGHTLY_FEATURES_IN_TESTS))]
```

src/ref.rs
```diff
@@ -650,6 +686,58 @@ where
...
+            let ptr: Ptr<'_, [u8], (Exclusive, Unaligned, Initialized)> = ptr.transmute::<
+                [u8],
+                Initialized,
+                (
+                    BecauseMutationCompatible,
+                    (
+                        BecauseRead,
+                        (BecauseExclusive, BecauseExclusive),
+                    ),
+                ),
```
joshlf: You should be able to replace some of these with underscores (at the most extreme, `(_, (_, (_, _)))`) and inference will still work – that'll make this more concise and format more nicely.

src/ref.rs
```diff
@@ -650,6 +686,58 @@ where
...
+            let ptr = unsafe { ptr.assume_alignment::<Aligned>() };
+            let ptr = ptr.recall_validity::<
+                Valid,
+                (
+                    BecauseMutationCompatible,
+                    (
+                        BecauseRead,
+                        (BecauseExclusive, BecauseExclusive),
+                    ),
+                ),
```
joshlf: Same here re: underscores

src/ref.rs
```diff
@@ -791,12 +912,64 @@ where
...
+            let ptr: Ptr<'_, [u8], (Exclusive, Unaligned, Initialized)> = ptr.transmute::<
+                [u8],
+                Initialized,
+                (
+                    BecauseMutationCompatible,
+                    (
+                        BecauseRead,
+                        (BecauseExclusive, BecauseExclusive),
+                    ),
+                ),
```
joshlf: Same here re: underscores

src/ref.rs
```diff
@@ -791,12 +912,64 @@ where
...
+            let ptr = unsafe { ptr.assume_alignment::<Aligned>() };
+            let ptr = ptr.recall_validity::<
+                Valid,
+                (
+                    BecauseMutationCompatible,
+                    (
+                        BecauseRead,
+                        (BecauseExclusive, BecauseExclusive),
+                    ),
+                ),
```
joshlf: Same here re: underscores

src/ref.rs
```diff
@@ -616,6 +619,39 @@ where
...
         let b = unsafe { r.into_byte_slice() };
+        let b = b.into_byte_slice();
+
+        if let crate::layout::SizeInfo::Sized { .. } = T::LAYOUT.size_info {
+            let ptr = Ptr::from_ref(b);
+            let ptr: Ptr<'_, [u8], (Shared, Unaligned, Initialized)> = ptr.transmute::<_, _, (_, (_, _))>();
+            // SAFETY: `T::raw_from_ptr_len` promises to preserve address and provenance in the referent.
+            let ptr: Ptr<'_, T, (Shared, Unaligned, Initialized)> = unsafe {
+                ptr.cast_unsized(|ptr| {
+                    let ptr = T::raw_from_ptr_len(
```
joshlf: Maybe call this `p` so you can distinguish between `ptr` (the closure argument) and `p` in the safety comment below?

src/ref.rs
```diff
@@ -616,6 +619,39 @@ where
...
+                    // 0. If `ptr`'s referent is not zero sized, then `ptr` has valid
+                    //    provenance for its referent, because it derived from `self`
+                    //    using a series of provenance-preserving operations, and
+                    //    because `self` has valid provenance for its referent. By the
+                    //    same argument, `ptr`'s referent is entirely contained within
+                    //    the same allocated object as `self`'s referent.
+                    // 1. If `ptr`'s referent is not zero sized, then the allocation of
+                    //    `ptr` is guaranteed to live for at least `'a`, because `ptr`
+                    //    is entirely contained in `self`, which lives for at least `'a`
+                    //    by invariant on `Ptr`.
```
joshlf: Assuming you do the rename suggested above:

```suggestion
                    // SAFETY: The safety invariants of `PtrInner::new` (see definition) are
                    // satisfied:
                    // 0. If `p`'s referent is not zero sized, then `p` has valid
                    //    provenance for its referent, because it derived from `ptr`
                    //    using a series of provenance-preserving operations, and
                    //    because `ptr` has valid provenance for its referent. By the
                    //    same argument, `p`'s referent is entirely contained within
                    //    the same allocated object as `ptr`'s referent.
                    // 1. If `p`'s referent is not zero sized, then the allocation of
                    //    `p` is guaranteed to live for at least `'a`, because `p`
                    //    is entirely contained in `ptr`, which lives for at least `'a`
                    //    by invariant on `PtrInner`.
```

src/ref.rs
```diff
@@ -616,6 +619,39 @@ where
...
+                    unsafe { PtrInner::new(ptr) }
+                })
+            };
+
+            // SAFETY: By invariant on `r`, we know that alignment is valid.
```
joshlf:
```suggestion
            // SAFETY: None of the preceding transformations modifies the address of the pointer, and by invariant on `r`, we know that it is validly-aligned.
```

src/ref.rs
```diff
@@ -650,6 +686,58 @@ where
...
+                    ),
+                ),
+            >();
+            // SAFETY: `T::raw_from_ptr_len` promises to preserve address and provenance in the referent.
+            let ptr: Ptr<'_, T, (Exclusive, Unaligned, Initialized)> = unsafe {
+                ptr.cast_unsized::<_, _, (BecauseRead, (BecauseExclusive, BecauseExclusive))>(|ptr| {
+                    let ptr = T::raw_from_ptr_len(
+                        ptr.as_non_null().cast(),
+                        <T::PointerMetadata as crate::PointerMetadata>::from_elem_count(0),
+                    );
```
joshlf:
```suggestion
                    let p = T::raw_from_ptr_len(
                        ptr.as_non_null().cast(),
                        <T::PointerMetadata as crate::PointerMetadata>::from_elem_count(0),
                    );
```

src/ref.rs
```diff
@@ -650,6 +686,58 @@ where
...
+                    // 0. If `ptr`'s referent is not zero sized, then `ptr` has valid
+                    //    provenance for its referent, because it derived from `self`
+                    //    using a series of provenance-preserving operations, and
+                    //    because `self` has valid provenance for its referent. By the
+                    //    same argument, `ptr`'s referent is entirely contained within
+                    //    the same allocated object as `self`'s referent.
+                    // 1. If `ptr`'s referent is not zero sized, then the allocation of
+                    //    `ptr` is guaranteed to live for at least `'a`, because `ptr`
+                    //    is entirely contained in `self`, which lives for at least `'a`
+                    //    by invariant on `Ptr`.
```
joshlf: Same as above

src/ref.rs
```diff
@@ -763,11 +851,44 @@ where
...
+        if let crate::layout::SizeInfo::Sized { .. } = T::LAYOUT.size_info {
+            let ptr = Ptr::from_ref(b);
+            let ptr: Ptr<'_, [u8], (Shared, Unaligned, Initialized)> = ptr.transmute::<_, _, (_, (_, _))>();
+            // SAFETY: `T::raw_from_ptr_len` promises to preserve address and provenance in the referent.
+            let ptr: Ptr<'_, T, (Shared, Unaligned, Initialized)> = unsafe {
+                ptr.cast_unsized(|ptr| {
+                    let ptr = T::raw_from_ptr_len(
+                        ptr.as_non_null().cast(),
+                        <T::PointerMetadata as crate::PointerMetadata>::from_elem_count(0),
+                    );
```
joshlf:
```suggestion
                    let p = T::raw_from_ptr_len(
                        ptr.as_non_null().cast(),
                        <T::PointerMetadata as crate::PointerMetadata>::from_elem_count(0),
                    );
```

src/ref.rs
```diff
@@ -763,11 +851,44 @@ where
...
+                    // 0. If `ptr`'s referent is not zero sized, then `ptr` has valid
+                    //    provenance for its referent, because it derived from `self`
+                    //    using a series of provenance-preserving operations, and
+                    //    because `self` has valid provenance for its referent. By the
+                    //    same argument, `ptr`'s referent is entirely contained within
+                    //    the same allocated object as `self`'s referent.
+                    // 1. If `ptr`'s referent is not zero sized, then the allocation of
+                    //    `ptr` is guaranteed to live for at least `'a`, because `ptr`
+                    //    is entirely contained in `self`, which lives for at least `'a`
+                    //    by invariant on `Ptr`.
```
joshlf: Same as above

src/ref.rs
```diff
@@ -791,12 +912,64 @@ where
...
+                    ),
+                ),
+            >();
+            // SAFETY: `T::raw_from_ptr_len` promises to preserve address and provenance in the referent.
+            let ptr: Ptr<'_, T, (Exclusive, Unaligned, Initialized)> = unsafe {
+                ptr.cast_unsized::<_, _, (BecauseRead, (BecauseExclusive, BecauseExclusive))>(|ptr| {
+                    let ptr = T::raw_from_ptr_len(
+                        ptr.as_non_null().cast(),
+                        <T::PointerMetadata as crate::PointerMetadata>::from_elem_count(0),
+                    );
```
joshlf:
```suggestion
                    let p = T::raw_from_ptr_len(
                        ptr.as_non_null().cast(),
                        <T::PointerMetadata as crate::PointerMetadata>::from_elem_count(0),
                    );
```

src/ref.rs
```diff
@@ -791,12 +912,64 @@ where
...
+                    // 0. If `ptr`'s referent is not zero sized, then `ptr` has valid
+                    //    provenance for its referent, because it derived from `self`
+                    //    using a series of provenance-preserving operations, and
+                    //    because `self` has valid provenance for its referent. By the
+                    //    same argument, `ptr`'s referent is entirely contained within
+                    //    the same allocated object as `self`'s referent.
+                    // 1. If `ptr`'s referent is not zero sized, then the allocation of
+                    //    `ptr` is guaranteed to live for at least `'a`, because `ptr`
+                    //    is entirely contained in `self`, which lives for at least `'a`
+                    //    by invariant on `Ptr`.
```
joshlf: Same as above

src/ref.rs
```diff
@@ -879,6 +1052,61 @@ mod tests {
...
+        let mut buf = Align::<[u8; 8], AU64>::default();
+        let _ = Ref::<_, AU64>::from_bytes(&mut buf.t[..]).unwrap();
+        b.iter(|| {
+            // SAFETY: The preceding `from_bytes` succeeded, and so we know that
+            // `&mut buf.t[.]            
+            let r = unsafe { Ref::<&mut [u8], AU64>::new_unchecked(&mut buf.t[..]) };
+            let mut temp = test::black_box(r);
+            test::black_box(temp.deref_mut());
+        });
+    }
```
joshlf: Let's move these into a new `#[cfg(all(test, __ZEROCOPY_INTERNAL_USE_ONLY_NIGHTLY_FEATURES_IN_TESTS))] mod benches` below `mod tests`?

src/ref.rs
```diff
@@ -616,6 +619,39 @@ where
...
+                    //    because `self` has valid provenance for its referent. By the
+                    //    same argument, `ptr`'s referent is entirely contained within
+                    //    the same allocated object as `self`'s referent.
+                    // 1. If `ptr`'s referent is not zero sized, then the allocation of
+                    //    `ptr` is guaranteed to live for at least `'a`, because `ptr`
+                    //    is entirely contained in `self`, which lives for at least `'a`
+                    //    by invariant on `Ptr`.
+                    unsafe { PtrInner::new(ptr) }
+                })
+            };
```
joshlf: You should be able to factor this entire thing out into a private, `unsafe` bare function below with its own separate safety preconditions. If you make it generic over the aliasing type parameter, you should be able to have a single function that's used by all four methods.

src/ref.rs
```diff
@@ -650,6 +668,22 @@ where
...
         let b = unsafe { r.into_byte_slice_mut() };
+        let b = b.into_byte_slice_mut();
+
+        if let crate::layout::SizeInfo::Sized { .. } = T::LAYOUT.size_info {
+            let ptr = Ptr::from_mut(b);
+            let ptr: Ptr<'_, [u8], (Exclusive, Unaligned, Initialized)> = ptr
+                .transmute::<[u8], Initialized, (_, (_, (BecauseExclusive, BecauseExclusive)))>();
+            let ptr = unsafe {
+                cast_for_sized::<T, _, (BecauseRead, (BecauseExclusive, BecauseExclusive))>(ptr)
+            };
```
joshlf, line 687: Safety comment?

src/ref.rs
```diff
@@ -650,6 +668,22 @@ where
...
+
+            // SAFETY: None of the preceding transformations modifies the address of the pointer, and by invariant on `r`, we know that it is validly-aligned.
```
joshlf: Rewrap at 80 columns?

I'd recommend [this VSCode extension](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap).

src/ref.rs
```diff
@@ -763,11 +797,24 @@ where
         // SAFETY: We don't call any methods on `b` other than those provided by
         // `ByteSlice`.
         let b = unsafe { self.as_byte_slice() };
+        let b = b.deref();
+
+        if let crate::layout::SizeInfo::Sized { .. } = T::LAYOUT.size_info {
+            let ptr = Ptr::from_ref(b);
+            let ptr: Ptr<'_, [u8], (Shared, Unaligned, Initialized)> =
+                ptr.transmute::<_, _, (_, (_, _))>();
+            let ptr = unsafe { cast_for_sized::<T, _, _>(ptr) };
```
joshlf: Safety comment?

src/ref.rs
```diff
@@ -763,11 +797,24 @@ where
...
+
+            // SAFETY: None of the preceding transformations modifies the address of the pointer, and by invariant on `r`, we know that it is validly-aligned.
```
joshlf: Reflow to 80 columns.

src/ref.rs
```diff
@@ -791,12 +838,28 @@ where
...
+
+        if let crate::layout::SizeInfo::Sized { .. } = T::LAYOUT.size_info {
+            let ptr = Ptr::from_mut(b);
+            let ptr: Ptr<'_, [u8], (Exclusive, Unaligned, Initialized)> = ptr
+                .transmute::<[u8], Initialized, (_, (_, (BecauseExclusive, BecauseExclusive)))>();
+            let ptr = unsafe {
+                cast_for_sized::<T, _, (BecauseRead, (BecauseExclusive, BecauseExclusive))>(ptr)
+            };
+
+            // SAFETY: None of the preceding transformations modifies the address of the pointer, and by invariant on `r`, we know that it is validly-aligned.
```
joshlf: Reflow to 80 columns.

src/ref.rs
```diff
@@ -791,12 +838,28 @@ where
         // SAFETY: We don't call any methods on `b` other than those provided by
         // `ByteSliceMut`.
         let b = unsafe { self.as_byte_slice_mut() };
+        let b = b.deref_mut();
+
+        if let crate::layout::SizeInfo::Sized { .. } = T::LAYOUT.size_info {
+            let ptr = Ptr::from_mut(b);
+            let ptr: Ptr<'_, [u8], (Exclusive, Unaligned, Initialized)> = ptr
+                .transmute::<[u8], Initialized, (_, (_, (BecauseExclusive, BecauseExclusive)))>();
+            let ptr = unsafe {
```
joshlf, line 856: Safety comment?

joshlf (changes requested):

joshlf:
Okay, I've rebased on `main` and force-pushed. @bmberger you should be able to download these changes and continue working on them.

joshlf (changes requested):
Please update the commit message with new benchmark numbers, but otherwise LGTM!

PR #2804, opened 2025-11-22
Invert build.rs version detection flags and update CI

### LINKED ISSUE #2259: Consider reversing `--cfg` flags for feature detection
Currently, your `--cfg` flags are set up such that if you pass no flags, then you get a build where all version-conditional features are disabled. This makes it difficult to use the crate with build systems that don't run the `build.rs` file, since you need to manually put together the list of `--cfg` flags to pass. If you forget to do this, the build will still pass, so you don't catch the mistake.

[In Tokio](https://github.com/tokio-rs/tokio/blob/a58beb3aca18e6ec4cb444b6c78d5a3373742998/tokio/build.rs), we were able to avoid this problem by reversing the meaning of the flags. For example, you would have a `--cfg no_zerocopy_target_has_atomics_1_60_0` that you pass on compilers *older* than 1.60.0. 

For projects with a non-cargo build system that does not run `build.rs`, this has the following advantages:

* When using the newest compiler, you get the correct behavior by default, as the correct set of `--cfg` flags is the empty list. You do not need to adjust the list of cfg flags on every crate update.
* When using an older compiler, the build will fail if you don't do anything. This ensures that you do not forget to look at the list of `--cfg` flags so that you pass the right set of flags.

joshlf:
Inverts version detection flags to default to enabled on modern compilers. Renames `zerocopy-foo` to `no-zerocopy-foo` in `Cargo.toml`, `build.rs`, source code, and `.github/workflows/ci.yml`. Ensures `check_all_toolchains_tested.sh` passes by syncing CI config. Verified unit tests on all affected toolchains.

Closes #2259

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,13 +50,13 @@ jobs:
           # These are the names of specific Rust versions detected in
           # `build.rs`. Each of these represents the minimum Rust version for
           # which a particular feature is supported.
-          "zerocopy-simd-x86-avx12-1-89-0",
-          "zerocopy-core-error-1-81-0",
-          "zerocopy-diagnostic-on-unimplemented-1-78-0",
-          "zerocopy-generic-bounds-in-const-fn-1-61-0",
-          "zerocopy-target-has-atomics-1-60-0",
-          "zerocopy-aarch64-simd-1-59-0",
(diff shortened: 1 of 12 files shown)


jswrenn:
What's the motivation here?

jswrenn (approved):

PR #2806, opened 2025-11-22
Add zizmor CI job and fix security findings

joshlf:
- Adds `zizmor` job to `ci.yml` (strict mode, blocking).
- Pins `codecov-action` to SHA in `ci.yml`.
- Fixes template injection vulnerabilities by using environment variables in `backport-pr.yml`, `docs.yml`, `release-crate-version.yml`, and `roll-pinned-toolchain-versions.yml`.
- Adds `persist-credentials: false` to `actions/checkout` calls across all workflows.
- Tightens permissions in `ci.yml` and suppresses necessary permission warnings in release workflows.
- Updates `all-jobs-succeed` to depend on `zizmor`.

### DIFF
--- a/.github/workflows/backport-pr.yml
+++ b/.github/workflows/backport-pr.yml
@@ -29,7 +29,7 @@ jobs:
   release:
     permissions:
       contents: write # Need to write to repo to cherry-pick
-      pull-requests: write # Need to create PR
+      pull-requests: write # zizmor: ignore[excessive-permissions]
     runs-on: ubuntu-latest
     name: Backport PR
     steps:


.github/workflows/backport-pr.yml
```diff
@@ -18,10 +18,18 @@ on:
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 
 jobs:
   release:
+    permissions:
+      contents: write # Need to write to repo to cherry-pick
+      pull-requests: write # Need to create PR
```
joshlf: Why did this succeed before without this permission? Maybe it's not needed?

PR #2809, opened 2025-11-23
Migrate zerocopy-derive from trybuild to ui-test

joshlf:
Replaces the `trybuild`-based UI testing infrastructure with `ui-test`.
This allows for better support of multiple toolchains (MSRV, Stable,
Nightly) by using separate `.stderr` files for each, rather than
relying on complex inline annotations.

Key changes:
- Replace `zerocopy-derive/tests/trybuild.rs` with a new `ui-test` harness
  in `zerocopy-derive/tests/ui.rs`.
- Add `tools/ui-runner` to support running tests on a newer toolchain
  while building dependencies on the target toolchain (trampoline pattern).
- Update UI test files to remove inline annotations and use standard
  `.stderr` files.
- Preserve existing symlink structure in UI test directories.

Future work:
- Re-introduce inline annotations using `ui-test`'s support for
  toolchain-specific annotations (e.g., `//~[stable]`).
- Migrate the top-level `tests` directory to `ui-test` as well.


zerocopy-derive/tests/ui-msrv/union_into_bytes_cfg/union_into_bytes_cfg.stderr
```diff
@@ -1,8 +1,14 @@
-error: requires --cfg zerocopy_derive_union_into_bytes;
```
joshlf, line 1: TODO: Figure out why this is regressing

PR #2815, opened 2025-11-24
Add serde optional feature to serialize primitive types

ElouanPetereau:
This pull request add a feature named `serde` to enable serializing and deserializing the `byteorder` numeric  wrappers using [serde](https://github.com/serde-rs/serde) derives.

# Why 
The reason for adding this is that I need to be able to convert a struct to either:
- a raw binary blob in the most performance efficient way (using `zerocopy` on my struct)
- a serialized format like `CSV` or `json` which doesn't need to be as fast (using `serde` on my struct)

Instead of having to manually write the [`serde:Serialize`](https://docs.rs/serde/latest/serde/trait.Serialize.html) and [`serde:Deserialize`](https://docs.rs/serde/latest/serde/trait.Deserialize.html) traits every time for my structs and calling the `get()`/`from()` function to get the inner type, e.g. `f64` for an `F64`, it would have been easier to use the `Serialize` and `Deserialize` [derives](https://serde.rs/derive.html) provided by the serde crate.

# How

This simply update the `define_type!()` macro to add the `Serialize` and `Deserialize` trait implementations using the `get()`/`from()` function to get the inner type when the `serde` feature is enabled.

---

Thank you for your work in maintaining this project.

### DIFF
--- a/zerocopy/src/byteorder.rs
+++ b/zerocopy/src/byteorder.rs
@@ -652,6 +652,36 @@ example of how it can be used for parsing UDP packets.
             }
         }
 
+        #[cfg_attr(doc_cfg, doc(cfg(feature = "serde")))]
+        #[cfg(feature = "serde")]
+        impl<O: ByteOrder> serde::Serialize for $name<O>
+        where
+            $native: serde::Serialize,
+        {
+            #[inline(always)]
+            fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+            where
+                S: serde::Serializer,
+            {
+                $native::serialize(&self.get(), serializer)
+            }
+        }
+
+        #[cfg_attr(doc_cfg, doc(cfg(feature = "serde")))]
+        #[cfg(feature = "serde")]
+        impl<'de, O: ByteOrder> serde::Deserialize<'de> for $name<O>
+        where
+            $native: serde::Deserialize<'de>,
+        {
+            #[inline(always)]
+            fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+            where
+                D: serde::Deserializer<'de>,
+            {
+                $native::deserialize(deserializer).map(Self::from)
+            }
+        }
+
         $(
             impl<O: ByteOrder> From<$name<O>> for $larger_native {
                 #[inline(always)]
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -106,6 +113,7 @@ rand = { version = "0.8.5", default-features = false, features = ["small_rng"] }
 rustversion = "1.0"
 static_assertions = "1.1"
 testutil = { path = "testutil" }
+serde_json = "1.0.0"
```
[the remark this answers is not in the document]
ElouanPetereau: Tested and works with MSRV, applied

Cargo.toml
```diff
@@ -83,9 +83,16 @@ std = ["alloc"]
...
+__internal_use_only_features_that_work_on_stable = [
+    "alloc",
+    "derive",
+    "serde",
+    "simd",
+    "std",
+]
 
 [dependencies]
+serde = { version = "1.0.0", default-features = false, optional = true }
```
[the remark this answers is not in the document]
ElouanPetereau: Tested and works with MSRV, applied

src/lib.rs
```diff
@@ -121,6 +121,10 @@
 //!   derives as `use zerocopy_derive::*` rather than by name (e.g., `use
 //!   zerocopy_derive::FromBytes`).
 //!
+//! - **`serde`**
+//!   Provides `Serialize` and `Deserialize` impls for the `byteorder` numeric
```
joshlf:
```suggestion
//!   Provides [`serde`] [`Serialize`] and [`Deserialize`] impls for the [`byteorder`] numeric
```

(or something along these lines – just want to mention serde, and want to make things into doc links)

Cargo.toml
```diff
@@ -106,6 +113,7 @@ rand = { version = "0.8.5", default-features = false, features = ["small_rng"] }
 rustversion = "1.0"
 static_assertions = "1.1"
 testutil = { path = "testutil" }
+serde_json = "1.0.107"
```
joshlf: Why is this needed when we already have an optional non-dev dependency?

zerocopy/src/byteorder.rs
```diff
@@ -1530,3 +1560,82 @@ mod tests {
...
+                U16::<LittleEndian>::new(primitive_value_u16),
+            ],
+        };
+        let serialized_value =
+            serde_json::to_string(&primitive_value).expect("Serialization to json should succeed");
+        let deserialized_wrapper: SerializableStruct = serde_json::from_str(&serialized_value)
+            .expect("Deserialization from json should succeed");
+
+        assert_eq!(primitive_value, deserialized_wrapper);
+    }
```
joshlf, line 1757: Would it be possible to make this type generic and test w/ all types in this module?

joshlf (changes requested):
Wow, thanks for this! I'm sure others will find this useful too.

I'm broadly on board with the implementation – just a few small changes requested.

src/lib.rs
ElouanPetereau: I cannot use the doc links directly as the feature is optional and when it is not used `serde` is not imported. I guarded this doc part with a config attribute and updated slightly the doc when not using the `serde` feature to specify that the implemented traits come from `serde`, wdyt? 

Additionally, I fixed the links for the `simd` feature just below, or rather added the links that where missing:
```diff
-//!   When the `simd` feature is enabled, `FromZeros`, `FromBytes`, and
-//!   `IntoBytes` impls are emitted for all stable SIMD types which exist on the
+//!   When the `simd` feature is enabled, [`FromZeros`], [`FromBytes`], and
+//!   [`IntoBytes`] impls are emitted for all stable SIMD types which exist on the
```

Cargo.toml
ElouanPetereau: I use [`serde_json`](https://docs.rs/serde_json/latest/serde_json/) in the tests to simplify the serialization/deserialization and avoid creating my own.    
Since this crate is the one showcased in the [`README.md of serde`](https://github.com/serde-rs/serde#serde-in-action) it is probably one of the most used as well (a quick check at crates.io seems to confirm that)

zerocopy/src/byteorder.rs
ElouanPetereau: Yes completely, done.

src/lib.rs
ElouanPetereau: It seems using the config attribute breaks the `readme` option of cargo and I did not realize that the `README.md` and the doc comment need to be a 1:1 match every time so I cannot use a conditional doc anyway.

I switched to add a link to the serde crate repository but I had to update the `generate-readme` crate because it was removing the square brackets on proper links with urls (`[]()`) where it should only do it for Rust doc links (`[]`).

Wdyt?

zerocopy/src/byteorder.rs
joshlf: Great, thanks!

Cargo.toml
joshlf: Ah I just can't read apparently – I misread that this and the optional dependency on serde were the same crate.

src/lib.rs
ElouanPetereau: Hello @joshlf , did you had any time to check my proposition here?
I am trying to keep this branch up to date every time main changes, hence my commits in the past days.

PR #2836, opened 2025-11-30
[derive] Fix IntoBytes on repr(C) DST

### LINKED ISSUE #2835: `derive(IntoBytes)` fails on `repr(C)` DSTs with parameterized leading fields
For example, compiling:
```rust
use zerocopy::IntoBytes;

#[derive(IntoBytes)]
#[repr(C)]
struct Struct {
    leading: Unalign<u32>,
    trailing: [u8]
}
```
...fails with this error:
```
error: no rules expected `<`
   --> src/lib.rs:6:18
    |
  6 |     addr: Unalign<u32>,
    |                  ^ no rules expected this token in macro call
    |
note: while trying to match `]`
   --> /playground/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/zerocopy-0.8.27/src/util/macro_util.rs:388:25
    |
388 |     ($t:ty, [$($ts:tt),*]) => {{
```

The failure arises because our internal `repr_c_struct_has_padding` macro accepts field types as comma delimited `tt`s, but `Unalign<u32>` is not a single `tt`.

We should see if there's a minor patchup we can do to make this work. `Unalign<u32>` can be made to parse as a `tt` by parenthesizing it.

joshlf:
Closes #2835

### DIFF
--- a/src/util/macro_util.rs
+++ b/src/util/macro_util.rs
@@ -393,6 +393,12 @@ macro_rules! repr_c_struct_has_padding {
         );
         layout.requires_static_padding() || layout.requires_dynamic_padding()
     }};
+    (@field ([$t:ty])) => {
+        <[$t] as $crate::KnownLayout>::LAYOUT
+    };
+    (@field ($t:ty)) => {
+        $crate::DstLayout::for_unpadded_type::<$t>()
+    };
     (@field [$t:ty]) => {
         <[$t] as $crate::KnownLayout>::LAYOUT
     };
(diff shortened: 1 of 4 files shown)


src/util/macro_util.rs
```diff
@@ -393,6 +393,12 @@ macro_rules! repr_c_struct_has_padding {
         );
         layout.requires_static_padding() || layout.requires_dynamic_padding()
     }};
+    (@field ([$t:ty])) => {
+        <[$t] as $crate::KnownLayout>::LAYOUT
+    };
+    (@field ($t:ty)) => {
+        $crate::DstLayout::for_unpadded_type::<$t>()
+    };
```
joshlf, line 401: @jswrenn lmk if you can think of something here – it'd be great to be able to *merge* these with the following branches rather than just *add* them so we have branch duplication. I tried to move the parentheses up to the top branch's match, but that didn't work (that said, I didn't try that hard to get it to work). This is very minor, so we can just leave it, but it'd be nice to merge these if it turns out there's something easy we can do.

jswrenn (approved):

PR #2842, opened 2025-12-03
Introduce `HasField` trait

jswrenn:
The `HasField` trait permits privacy-preserving field projections
from structs, enums and unions.

`HasField` preserves privacy by leveraging the [scoped trait pattern];
for each field `f`, `HasField` is implemented with the `Field` parameter
set to a type with identical visibility as `f`. In many useful contexts,
this `Field` parameter is inferable, and need not be actually reachable.
Consequently, `HasField` can be used to index a field from only the name
of a field. See `zerocopy-derive/tests/ui-nightly/privacy.rs` for
examples.

[scoped trait pattern]: https://jack.wrenn.fyi/blog/private-trait-impls/

---

- #2844
- #2843
- #2842

**Latest Update:** v2 — [Compare vs v1](https://github.com/google/zerocopy/pull/2842/compare/gherrit/G2238bc341570838db412d880017b3f0c25ac09fa/v1..gherrit/G2238bc341570838db412d880017b3f0c25ac09fa/v2)

<details>

| Version | Base | v1 |
| :--- | :--- | :--- |
| v2 | [vs Base](https://github.com/google/zerocopy/pull/2842/compare/main..gherrit/G2238bc341570838db412d880017b3f0c25ac09fa/v2) | [vs v1](https://github.com/google/zerocopy/pull/2842/compare/gherrit/G2238bc341570838db412d880017b3f0c25ac09fa/v1..gherrit/G2238bc341570838db412d880017b3f0c25ac09fa/v2) |
| v1 | [vs Base](https://github.com/google/zerocopy/pull/2842/compare/main..gherrit/G2238bc341570838db412d880017b3f0c25ac09fa/v1) | |

</details>


zerocopy-derive/src/enum.rs
```diff
@@ -245,6 +250,34 @@ pub(crate) fn derive_is_bit_valid(
...
+            assert!(matches!(vis, syn::Visibility::Inherited));
+            ImplBlockBuilder::new(
+                ast,
+                data,
+                Trait::HasField {
+                    variant_id: parse_quote!({ #zerocopy_crate::ident_id!(#variant_ident) }),
+                    // Since Rust does not presently support explicit visibility
+                    // modifiers on enum fields, any public type is suitable
+                    // here; we use `()`.
+                    field: parse_quote!(()),
```
joshlf: Feel free to disregard, but it might be worth considering whether we could put this in a `const` or `static` or something to avoid recomputing it for every field.

zerocopy-derive/src/lib.rs
```diff
@@ -464,11 +467,12 @@ fn derive_known_layout_inner(
 
     Ok(match &ast.data {
         Data::Struct(strct) => {
-            let require_trait_bound_on_field_types = if self_bounds == SelfBounds::SIZED {
-                FieldBounds::None
-            } else {
-                FieldBounds::TRAILING_SELF
-            };
+            let require_trait_bound_on_field_types =
+                if matches!(self_bounds, SelfBounds::All(&[Trait::Sized])) {
```
joshlf, line 471: Is this a breaking change?

zerocopy-derive/src/lib.rs
```diff
@@ -737,6 +741,52 @@ fn derive_split_at_inner(
...
+fn derive_has_field_variant(
+    ast: &DeriveInput,
+    data: &dyn DataExt,
+    zerocopy_crate: &Path,
+) -> TokenStream {
+    let fields = ast.data.fields();
+
+    if fields.is_empty() {
+        return quote! {};
+    }
```
joshlf: Style uber-nit; feel free to disregard.

```suggestion
    if fields.is_empty() {
        return quote! {};
    }
```

zerocopy-derive/src/lib.rs
```diff
@@ -737,6 +741,52 @@ fn derive_split_at_inner(
...
+
+    let field_tokens = fields.iter().map(|(vis, ident, _)| {
+        let ident = Ident::new(&format!("ẕ{}", ident), ident.span());
```
joshlf, line 755:
```suggestion
        let ident = Ident::new(&format!("__Zerocopy_{}", ident), ident.span());
```

(or something along those lines)

zerocopy-derive/src/lib.rs
```diff
@@ -737,6 +741,52 @@ fn derive_split_at_inner(
...
+        quote!(
+            #vis enum #ident {}
+        )
+    });
+
+    let has_fields = fields.iter().map(|(_, ident, ty)| {
+        let field_token = Ident::new(&format!("ẕ{}", ident), ident.span());
```
joshlf, line 762: Same here re: `__Zerocopy_` prefix.

zerocopy-derive/src/lib.rs
```diff
@@ -737,6 +741,52 @@ fn derive_split_at_inner(
...
+        ImplBlockBuilder::new(
+            ast,
+            data,
+            Trait::HasField {
+                variant_id: parse_quote!({ #zerocopy_crate::ident_id!(0) }),
```
joshlf, line 768: If this is for a variant, why is it always 0? Shouldn't it be the ID of the variant in question?

zerocopy-derive/tests/ui-nightly/privacy.rs
```diff
@@ -0,0 +1,102 @@
+// Copyright 2019 The Fuchsia Authors
```
joshlf:
```suggestion
// Copyright 2025 The Fuchsia Authors
```

zerocopy-derive/tests/ui-nightly/privacy.rs
```diff
@@ -0,0 +1,102 @@
...
+#![allow(unused_braces)]
+
+#[macro_use]
+extern crate zerocopy;
+
+fn main() {}
+
+// These tests cause errors which are generated by a later compilation pass than
+// the other errors we generate, and so if they're compiled in the same file,
+// the compiler will never get to that pass, and so we won't get the errors.
```
joshlf, line 18: That comment only makes sense in the `late_compile_pass.rs` test IMO. Is it true for this file too?

zerocopy-derive/src/lib.rs
jswrenn: Unlike our generated idents prior to this PR, the ones we generate here actually can appear in user-facing error messages. Given this, we can do one of two things:

1. Use the same identifier as the one provided by the user. This has the best UX for privacy-violating projections, but raises something between a soundness and a UX issue for the case where the field name conflicts with a type name.
2. Modify the identifier in a way that's exceedingly to conflict with a user-defined type. The UX degradation in this case is proportional how much the original identifier is degraded.

I think we need to at least _try_ to mitigate the name conflict issue, so this PR selects the second option. `__Zerocopy_` is a tolerable prefix for things that never see the light of day, but it's over-the-top for something using-facing. Especially for user-facing things, I'd like us to adopt something that's just as exceedingly unlikely to conflict with user-defined types but less intrusive. I selected `ẕ`, because it's:

- unintrusive; just a single character
- extremely uncommon
- gives a nod to zerocopy
- in the Latin script and thus does not trigger `uncommon_codepoints`, `mixed_script_confusables` or any other identifier lints

joshlf: Discussed offline – we considered an approach where we detect based on the input tokens whether a conflict exists and only falling back to this scheme in that case. We determined that it was too risky and error-prone for now, although it remains on the table for the future (this would be a backwards-compatible change).

Sticking with this for now.

zerocopy-derive/src/lib.rs
jswrenn: No, these are identical. This is just the long-form expansion of that const.

zerocopy-derive/src/lib.rs
jswrenn: Discussed offline.

PR #2851, opened 2025-12-15
Exclude various files from the published packages

weiznich:
During a dependency review we noticed that the zerocopy crate includes various development scripts and also test output for the compile tests in the published packages. Neither of that content is required for building zerocopy as dependency. Especially the development scripts shouldn't be there as they might, at some point become problematic. As of now they prevent any downstream user from enabling the `[bans.build.interpreted]` option of `cargo deny`.

I opted for using an explicit include list instead of an exclude list to prevent these files from beeing included in the published packages to make sure that everything that's included is an conscious choice.


weiznich:
@joshlf This is still waiting for a review. I still would like to see the script removed from the published package.

PR #2860, opened 2025-12-23
[pointer] Add generic projection/cast framework

joshlf:
Add `unsafe trait Project` and `unsafe trait Cast: Project`. `Cast` is
implemented for any address-preserving cast, while `Project` generalizes
to conversions which may not preserve the address of the referent (ie,
field projections). Use these (mostly `Cast`, but some `Project`) to
unify `PtrInner`/`Ptr` casts, field projections, and `SizeEq` casts.
Replace a good amount of unsafe derive-generated code with uses of this
machinery.

Makes progress on #196
Closes #2856

---

- &#x3000; #2866
- &#x3000; #2867
- 👉 #2860

**Latest Update:** v20 — [Compare vs v19](/google/zerocopy/compare/gherrit/Gdeb4f5a282b778a82175703218739a93074d0cc4/v19..gherrit/Gdeb4f5a282b778a82175703218739a93074d0cc4/v20)

<details>
<summary><strong>📚 Full Patch History</strong></summary>

*Links show the diff between the row version and the column version.*

|Version| v19 | v18 | v17 | v16 | v15 | v14 | v13 | v12 | v11 | v10 | v9 | v8 | v7 | v6 | v5 | v4 | v3 | v2 | v1 |Base|
|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|
[...]

PR #2862, opened 2025-12-30
Generate trait implementations for tuples

### LINKED ISSUE #274: Implement traits for tuples

jswrenn:
Emit implementations of `HasField`, `Immutable`, `TryFromBytes`,
`FromZeros`, and `FromBytes` for tuples of length up to 26.

Closes #274

---

- #2862

**Latest Update:** v7 — [Compare vs v6](https://github.com/google/zerocopy/compare/gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v6..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v7)

<details>

| Version | Base | v1 | v2 | v3 | v4 | v5 | v6 |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| v7 | [vs Base](https://github.com/google/zerocopy/compare/main..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v7) | [vs v1](https://github.com/google/zerocopy/compare/gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v1..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v7) | [vs v2](https://github.com/google/zerocopy/compare/gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v2..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v7) | [vs v3](https://github.com/google/zerocopy/compare/gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v3..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v7) | [vs v4](https://github.com/google/zerocopy/compare/gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v4..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v7) | [vs v5](https://github.com/google/zerocopy/compare/gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v5..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v7) | [vs v6](https://github.com/google/zerocopy/compare/gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v6..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v7) |
| v6 | [vs Base](https://github.com/google/zerocopy/compare/main..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v6) | [vs v1](https://github.com/google/zerocopy/compare/gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v1..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v6) | [vs v2](https://github.com/google/zerocopy/compare/gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v2..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v6) | [vs v3](https://github.com/google/zerocopy/compare/gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v3..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v6) | [vs v4](https://github.com/google/zerocopy/compare/gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v4..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v6) | [vs v5](https://github.com/google/zerocopy/compare/gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v5..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v6) | |
| v5 | [vs Base](https://github.com/google/zerocopy/compare/main..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v5) | [vs v1](https://github.com/google/zerocopy/compare/gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v1..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v5) | [vs v2](https://github.com/google/zerocopy/compare/gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v2..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v5) | [vs v3](https://github.com/google/zerocopy/compare/gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v3..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v5) | [vs v4](https://github.com/google/zerocopy/compare/gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v4..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v5) | | |
| v4 | [vs Base](https://github.com/google/zerocopy/compare/main..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v4) | [vs v1](https://github.com/google/zerocopy/compare/gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v1..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v4) | [vs v2](https://github.com/google/zerocopy/compare/gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v2..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v4) | [vs v3](https://github.com/google/zerocopy/compare/gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v3..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v4) | | | |
| v3 | [vs Base](https://github.com/google/zerocopy/compare/main..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v3) | [vs v1](https://github.com/google/zerocopy/compare/gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v1..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v3) | [vs v2](https://github.com/google/zerocopy/compare/gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v2..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v3) | | | | |
| v2 | [vs Base](https://github.com/google/zerocopy/compare/main..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v2) | [vs v1](https://github.com/google/zerocopy/compare/gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v1..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v2) | | | | | |
| v1 | [vs Base](https://github.com/google/zerocopy/compare/main..gherrit/G992a50c1ccf344487e1cd764afe0b17e566bc620/v1) | | | | | | |

</details>

### DIFF
--- a/src/impls.rs
+++ b/src/impls.rs
@@ -1005,6 +1005,137 @@ const _: () = unsafe {
 // [1] https://doc.rust-lang.org/core/option/enum.Option.html
 const _: () = unsafe { unsafe_impl!(T: Immutable => Immutable for Option<T>) };
 
+mod tuples {
+    use super::*;
+
+    /// Generates various trait implementations for tuples.
+    ///
+    /// # Safety
+    ///
+    /// `impl_tuple!` should be provided name-number pairs, where each number is
+    /// the ordinal of the preceding type name.
+    macro_rules! impl_tuple {
+        // Entry point.
+        ($($T:ident $I:tt),+ $(,)?) => {
+            crate::util::macros::__unsafe();
+            impl_tuple!(@all [] [$($T $I)+]);
+        };
+
+        // Build up the set of tuple types (i.e., `(A,)`, `(A, B)`, `(A, B, C)`,
+        // etc.) Trait implementations that do not depend on field index may be
+        // added to this branch.
+        (@all [$($head_T:ident $head_I:tt)*] [$next_T:ident $next_I:tt $($tail:tt)*]) => {
+            // SAFETY: If all fields of the tuple `Self` are `Immutable`, so too is `Self`.
+            unsafe_impl!($($head_T: Immutable,)* $next_T: Immutable => Immutable for ($($head_T,)* $next_T,));
+
+            // SAFETY: If all fields in `c` are `is_bit_valid`, so too is `c`.
+            unsafe_impl!($($head_T: TryFromBytes,)* $next_T: TryFromBytes => TryFromBytes for ($($head_T,)* $next_T,); |c| {
+                let mut c = c;
+                $(TryFromBytes::is_bit_valid(c.reborrow().project::<_, { crate::ident_id!($head_I) }>()) &&)*
+                    TryFromBytes::is_bit_valid(c.reborrow().project::<_, { crate::ident_id!($next_I) }>())
+            });
+
+            // SAFETY: If all fields in `Self` are `FromZeros`, so too is `Self`.
+            unsafe_impl!($($head_T: FromZeros,)* $next_T: FromZeros => FromZeros for ($($head_T,)* $next_T,));
+
+            // SAFETY: If all fields in `Self` are `FromBytes`, so too is `Self`.
+            unsafe_impl!($($head_T: FromBytes,)* $next_T: FromBytes => FromBytes for ($($head_T,)* $next_T,));
+
+            // Generate impls that depend on tuple index.
+            impl_tuple!(@variants
+                [$($head_T $head_I)* $next_T $next_I]
+                []
+                [$($head_T $head_I)* $next_T $next_I]
+            );
+
+            // Recurse to next tuple size
+            impl_tuple!(@all [$($head_T $head_I)* $next_T $next_I] [$($tail)*]);
+        };
+        (@all [$($head_T:ident $head_I:tt)*] []) => {};
+
+        // Emit trait implementations that depend on field index.
+        (@variants
+            // The full tuple definition in type–index pairs.
+            [$($AllT:ident $AllI:tt)+]
+            // Types before the current index.
+            [$($BeforeT:ident)*]
+            // The types and indices at and after the current index.
+            [$CurrT:ident $CurrI:tt $($AfterT:ident $AfterI:tt)*]
+        ) => {
+            // SAFETY:
+            // - `Self` is a struct (albeit anonymous), so `VARIANT_ID` is
+            //   `STRUCT_VARIANT_ID`.
+            // - `$CurrI` is the field at index `$CurrI`, so `FIELD_ID` is
+            //   `zerocopy::ident_id!($CurrI)`
+            // - `()` has the same visibility as the `.$CurrI` field (ie, `.0`,
+            //   `.1`, etc)
+            // - `Type` has the same type as `$CurrI`; i.e., `$CurrT`.
+            unsafe impl<$($AllT),+> crate::HasField<
+                (),
+                { crate::STRUCT_VARIANT_ID },
+                { crate::ident_id!($CurrI)}
+            > for ($($AllT,)+) {
+                #[inline]
+                fn only_derive_is_allowed_to_implement_this_trait()
+                where
+                    Self: Sized
+                {}
+
+                type Type = $CurrT;
+
+                #[inline(always)]
+                fn project(slf: crate::PtrInner<'_, Self>) -> *mut Self::Type {
+                    let slf = slf.as_non_null().as_ptr();
+                    // SAFETY: `PtrInner` promises it references either a zero-sized
+                    // byte range, or else will reference a byte range that is
+                    // entirely contained within an allocated object. In either
+                    // case, this guarantees that `(*slf).$CurrI` is in-bounds of
+                    // `slf`.
+                    unsafe { core::ptr::addr_of_mut!((*slf).$CurrI) }
+                }
+            }
+
+            // Recurse to the next index.
+            impl_tuple!(@variants [$($AllT $AllI)+] [$($BeforeT)* $CurrT] [$($AfterT $AfterI)*]);
+        };
+        (@variants [$($AllT:ident $AllI:tt)+] [$($BeforeT:ident)*] []) => {};
+    }
+
+    // SAFETY: `impl_tuple` is provided name-number pairs, where number is the
+    // ordinal of the name.
+    #[allow(clippy::multiple_unsafe_ops_per_block)]
+    const _: () = unsafe {
+        impl_tuple! {
+            A 0,
+            B 1,
+            C 2,
+            D 3,
+            E 4,
+            F 5,
+            G 6,
+            H 7,
+            I 8,
+            J 9,
+            K 10,
+            L 11,
+            M 12,
+            N 13,
+            O 14,
+            P 15,
+            Q 16,
+            R 17,
+            S 18,
+            T 19,
+            U 20,
+            V 21,
+            W 22,
+            X 23,
+            Y 24,
+            Z 25,
+        };
+    };
+}
+
 // SIMD support
 //
 // Per the Unsafe Code Guidelines Reference [1]:
--- a/zerocopy-derive/src/lib.rs
+++ b/zerocopy-derive/src/lib.rs
@@ -897,15 +897,11 @@ fn derive_try_from_bytes_struct(
                 where
                     ___ZerocopyAliasing: #zerocopy_crate::pointer::invariant::Reference,
                 {
-                    use #zerocopy_crate::util::macro_util::core_reexport;
-                    use #zerocopy_crate::pointer::PtrInner;
-
                     true #(&& {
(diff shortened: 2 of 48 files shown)


src/impls.rs
```diff
@@ -1021,6 +1021,155 @@ const _: () = unsafe {
...
+            [$($BeforeT:ident)*]
+            // The types and indices at and after the current index.
+            [$CurrT:ident $CurrI:tt $($AfterT:ident $AfterI:tt)*]
+        ) => {
+            // SAFETY:
+            // - `Self` is a struct (albeit anonymous), so `VARIANT_ID` is
+            //   `STRUCT_VARIANT_ID`.
+            // - `$CurrI` is the field at index `$CurrI`, so `FIELD_ID` is
+            //   `zerocopy::ident_id!($CurrI)`
+            // - `()` has the same visibility as `$CurrI`
```
joshlf:
```suggestion
            // - `()` has the same visibility as the `.$CurrI` field (ie, `.0`, `.1`, etc)
            //   since tuple fields are globally public, and so is the `()` type
```

joshlf: I know it's technically fine to just say `0`, `1`, etc in the same way that it'd be fine for named fields to refer to them without the leading dot (`foo`, `bar`, etc), but at least to me, this makes it clearer that we're talking about the visibility of tuple fields, which is not something most people are used to doing.

src/impls.rs
```diff
@@ -1021,6 +1021,155 @@ const _: () = unsafe {
...
+                    // 0. `ptr` addresses a subset of the bytes of
+                    //    `slf`, so by invariant on `slf: PtrInner`,
+                    //    if `ptr`'s referent is not zero sized,
+                    //    then `ptr` has valid provenance for its
+                    //    referent, which is entirely contained in
+                    //    some Rust allocation, `A`.
+                    // 1. By invariant on `slf: PtrInner`, if
+                    //    `ptr`'s referent is not zero sized, `A` is
+                    //    guaranteed to live for at least `'a`.
+                    unsafe { crate::PtrInner::new(ptr) }
```
joshlf: TODO(me): I haven't reviewed this yet because I'm assuming we'll rebase on #2860, which will allow us to significantly simplify this body.

zerocopy-derive/src/lib.rs
```diff
@@ -845,28 +844,8 @@ fn derive_try_from_bytes_struct(
...
-                        //   at the same byte ranges in the returned pointer's
-                        //   referent as they do in `*slf`
-                        let field_candidate = unsafe {
-                            candidate.reborrow().cast_unsized_unchecked(project)
-                        };
-
-                        <#field_tys as #zerocopy_crate::TryFromBytes>::is_bit_valid(field_candidate)
-                    })*
+                    true
+                        #(&& #zerocopy_crate::util::macro_util::is_field_valid::<_, _, _, { #zerocopy_crate::ident_id!(#field_names) }>(candidate.reborrow()))*
```
joshlf: 🥳

zerocopy-derive/src/lib.rs
```diff
@@ -901,16 +896,12 @@ fn derive_try_from_bytes_struct(
...
-                    use #zerocopy_crate::pointer::PtrInner;
-
                     true #(&& {
                         let field_candidate = candidate.reborrow().project::<
                             _,
                             { #zerocopy_crate::ident_id!(#field_names) }
                         >();
-
-                        <#field_tys as #zerocopy_crate::TryFromBytes>::is_bit_valid(field_candidate)
+                        #zerocopy_crate::TryFromBytes::is_bit_valid(field_candidate)
```
joshlf: It may be slightly more robust/reliable to keep the `<#field_tys as ...>` syntax since it makes the self type completely unambiguous. Currently not a problem, but I could theoretically see this becoming an issue if we ever change the signature of `is_bit_valid` to something that makes the type inference ambiguous (or worse, makes it infer as the wrong function so this silently compiles but is unsound).

src/impls.rs
jswrenn: [Done](https://github.com/google/zerocopy/compare/8303c302db5a22149005258d92dd60b698f9c693..42a62515e9d986b9be7f733a2c8fa232d9152cc3#diff-ce23348aa7be0caf33ab14158088aa649a46ef3f12537b70f5583c90f36e13a1R1070-R1071

zerocopy-derive/src/lib.rs
jswrenn: I think the risk is negligible and outweighed by clarity benefits of removing a turbofish. The type of `field_candidate` flows directly from projection. The salient soundness condition of `is_bit_valid` is that every field is validated (and so we splat over `#field_names`); not that projection is correct. I agree that projection being incorrect would cause a soundness issue, but it'd be silly to program defensively on the assumption that `project` has confused the field type. And if projection is sound, there cannot be any ambiguity, since the field type is an associated type of `HasField`, not a parameter.

jswrenn: And I think we should apply this guiding principle to most other parts of our codebase. Like here, why are we programming defensively against `cast` returning the wrong type?
https://github.com/google/zerocopy/blob/4cfb5c979891222bf61323690bfbcdf795d1315e/zerocopy-derive/src/enum.rs#L343-L350

joshlf: My concern is less about projection returning the wrong type, and more about `is_bit_valid` itself no longer taking a `Maybe<Self>` and instead taking something that's not *directly* a `Self` (an associated type, or a type parameter on `is_bit_valid` itself, etc).

Generally, since this code is very convoluted, my take is that explicitness is more important than concision. I tend to have the opposite opinion in "normal" code, but auto-generated code is very tricky to reason about (and you don't have the normal niceties like rust-analyzer showing you the types of local variables), so more explicit annotations reduces the likelihood of code behaving unexpectedly.

jswrenn: If we change the argument of `is_bit_valid`, the inference issue will arise between that argument and projection, not between projection and field validation. Even if we managed to pull off making a change that somehow triggered an issue between projection and validation, we'd have to do so in a way that evaded all of our tests in order for it to affect our users.

I've [restored the turbofish](https://github.com/google/zerocopy/compare/42a62515e9d986b9be7f733a2c8fa232d9152cc3..247dd33f85c6ad68234e7731524c6a9d561d5942#diff-977c29e6cf079bd7307f811df24f99633df603962785c583145eb6250b0e12daR905) but it's still my opinion that this extreme degree of defensive programming is counterproductive. The turbofishes in both this and the `cast` example above add considerable boat to our `is_bit_valid` codegen (contributing to it looking "very convoluted"), and there's something to be said for minimizing the "what the fuck" factor of zerocopy's derives.

PR #3073, opened 2026-02-27
[zerocopy] Re-enable big endian aarch64 types

197g:
When initially released the LLVM semantics did not agree with the intrinsics in the order of lanes and hence they got removed from stable on big endian targets. The fix of these semantics was shipped with Rust 1.87 in March 2025 (<https://github.com/rust-lang/rust/pull/136831>).

To keep the library working in these previous compiler versions the intrinsics are enabled by their prior condition, a little endian target is detected, or the feature detection cfg of the more recent rust release is present.

Interesting point: the tests were never gated on `target_endian = "little"` so it seems that CI is not configured for any big endian aarch64 platform.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -79,6 +79,11 @@ no-zerocopy-target-has-atomics-1-60-0 = "1.60.0"
 # versions, these types require the "simd-nightly" feature.
 no-zerocopy-aarch64-simd-1-59-0 = "1.59.0"
 
+# Include SIMD types from `core::arch::aarch64` on big endian targets. Prior to
+# 1.87.0 (https://github.com/rust-lang/rust/pull/136831) the types were only
+# correct on little endian and backed off from stable in a breaking change.
+no-zerocopy-aarch64-simd-be-1-87-0 = "1.87.0"
+
 # Permit panicking in `const fn`s and calling `Vec::try_reserve`.
 no-zerocopy-panic-in-const-and-vec-try-reserve-1-57-0 = "1.57.0"
 
--- a/src/impls.rs
+++ b/src/impls.rs
@@ -1350,12 +1350,26 @@ mod simd {
             #[cfg(all(feature = "simd-nightly", target_arch = "powerpc64"))]
             powerpc64, powerpc64, vector_bool_long, vector_double, vector_signed_long, vector_unsigned_long
         );
+        // NOTE: NEON intrinsics were broken on big-endian platforms from their stabilization up to
+        // Rust 1.87. (Context in https://github.com/rust-lang/stdarch/issues/1484). Support is
+        // split in two different version ranges on top of the base configuration, requiring either
+        // little endian or the more recent version to be detected as well.
         #[cfg(not(no_zerocopy_aarch64_simd_1_59_0))]
         simd_arch_mod!(
-            // NOTE(https://github.com/rust-lang/stdarch/issues/1484): NEON intrinsics are currently
-            // broken on big-endian platforms.
-            #[cfg(all(target_arch = "aarch64", target_endian = "little"))]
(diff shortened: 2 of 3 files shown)


src/impls.rs
```diff
@@ -1352,10 +1352,14 @@ mod simd {
         );
         #[cfg(not(no_zerocopy_aarch64_simd_1_59_0))]
         simd_arch_mod!(
-            // NOTE(https://github.com/rust-lang/stdarch/issues/1484): NEON intrinsics are currently
-            // broken on big-endian platforms.
```
jswrenn, line 1356: Can you keep some sort of comment here, that explains the fairly complex `cfg` below?

src/impls.rs
```diff
@@ -1352,10 +1352,14 @@ mod simd {
...
-            #[cfg(all(target_arch = "aarch64", target_endian = "little"))]
-            #[cfg_attr(doc_cfg, doc(cfg(rust = "1.59.0")))]
+            #[cfg(any(
+                all(target_arch = "aarch64", target_endian = "little"),
+                all(target_arch = "aarch64", not(no_zerocopy_aarch64_simd_be_1_87_0))
+            ))]
+            #[cfg_attr(
+                all(doc_cfg, target_endian = "little"),
+                doc(cfg(rust = "1.59.0"))
+            )]
```
jswrenn, line 1372:
```suggestion
            #[cfg_attr(
                doc_cfg,
                doc(cfg(all(target_arch = "aarch64", any(
                    all(rust = "1.59.0", target_endian = "little"),
                    rust = "1.87.0",
                ))))
            )]
```

src/impls.rs
```diff
@@ -1352,10 +1352,14 @@ mod simd {
...
         #[cfg(not(no_zerocopy_aarch64_simd_1_59_0))]
         simd_arch_mod!(
-            // NOTE(https://github.com/rust-lang/stdarch/issues/1484): NEON intrinsics are currently
-            // broken on big-endian platforms.
-            #[cfg(all(target_arch = "aarch64", target_endian = "little"))]
-            #[cfg_attr(doc_cfg, doc(cfg(rust = "1.59.0")))]
+            #[cfg(any(
+                all(target_arch = "aarch64", target_endian = "little"),
+                all(target_arch = "aarch64", not(no_zerocopy_aarch64_simd_be_1_87_0))
+            ))]
```
jswrenn: Factor out the `target_arch = "aarch64"`?

```suggestion
            #[cfg(all(
                target_arch = "aarch64",
                any(
                    target_endian = "little",
                    not(no_zerocopy_aarch64_simd_be_1_87_0)
                )
            ))]
```

Cargo.toml
```diff
@@ -79,6 +79,11 @@ no-zerocopy-target-has-atomics-1-60-0 = "1.60.0"
 # versions, these types require the "simd-nightly" feature.
 no-zerocopy-aarch64-simd-1-59-0 = "1.59.0"
 
+# Include SIMD types from `core::arch::aarch64` on big endian targets. Prior to
+# 1.87.0 (#136831) the types were only correct on little endian and backed off
```
jswrenn:
```suggestion
# 1.87.0 (https://github.com/rust-lang/rust/pull/136831) the types were only correct on little endian and backed off
```

src/impls.rs
197g: Since I took those recommendations verbatim, I wanted to mark the commit with you in Co-Authored-By. It seems the CLA tool does not like it. How would you like me to refer to you?

joshlf: Try `jswrenn@google.com` or `jack@wrenn.fyi`? Not sure why since the email you used is in the recent Git history 🤷‍♂️

Issue #2445, opened 2025-03-18
Support serializing this Fuchsia format


joshlf:
https://fuchsia-review.git.corp.google.com/c/fuchsia/+/1227944/19/src/lib/diagnostics/log/message/rust/src/lib.rs

PR #3153, opened 2026-03-27
  [tests] Add regression tests for Fuchsia packet types with large arrays (#2445)

unseen2004:
This PR adds regression tests to verify that `zerocopy` correctly handles `repr(C)` structs containing large computed array fields (e.g., `[u8; 2000]`), as seen in Fuchsia's `FxLogPacket` and similar datagram types.
    
    Specifically, it ensures that these types correctly derive and implement:
    - `KnownLayout`
    - `Immutable`
    - `IntoBytes`
 
   These tests confirm that `zerocopy`'s const-generic implementation for arrays correctly scales to large sizes without hitting compiler limits or logic errors in the derive macros.
    
   ### Verification:
   Tests were verified locally on the `stable` toolchain:
   - `./cargo.sh +stable test -p zerocopy-derive --test struct_known_layout`
   - `./cargo.sh +stable test -p zerocopy-derive --test struct_no_cell`
   - `./cargo.sh +stable test -p zerocopy-derive --test struct_to_bytes`
   
   The tests passed successfully and were confirmed to fail if the derives were manually removed from the test cases.

### DIFF
--- a/zerocopy-derive/tests/struct_no_cell.rs
+++ b/zerocopy-derive/tests/struct_no_cell.rs
@@ -107,7 +107,6 @@ util_assert_impl_all!(WithParams<'static, 'static, u8, 42>: imp::Immutable);
 // Regression test for #2445: Fuchsia has `repr(C)` packet types with a large
 // computed array field, and they should still be `Immutable`.
 const FX_LOG_MAX_DATAGRAM_LEN: usize = 2032;
-const FX_LOG_METADATA_SIZE: usize = 32;
 
 #[derive(imp::Immutable)]
 #[zerocopy(crate = "zerocopy_renamed")]
@@ -119,6 +118,8 @@ struct FxLogMetadata {
     severity: i32,
     dropped_logs: u32,
 }
+
+const FX_LOG_METADATA_SIZE: usize = imp::core::mem::size_of::<FxLogMetadata>();
 
 #[derive(imp::Immutable)]
 #[zerocopy(crate = "zerocopy_renamed")]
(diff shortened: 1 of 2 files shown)


zerocopy-derive/tests/struct_no_cell.rs
```diff
@@ -103,3 +103,30 @@ where
...
+    severity: i32,
+    dropped_logs: u32,
+}
+
+#[derive(imp::Immutable)]
+#[zerocopy(crate = "zerocopy_renamed")]
+#[repr(C)]
+struct FxLogPacket {
+    metadata: FxLogMetadata,
+    data: [u8; FX_LOG_MAX_DATAGRAM_LEN - FX_LOG_METADATA_SIZE],
```
joshlf, line 129: Can you make `FX_LOG_METADATA_SIZE` `size_of::<FxLogMetadata>()` instead of hard-coding its size?

zerocopy-derive/tests/struct_to_bytes.rs
```diff
@@ -227,3 +227,30 @@ pub struct IndexEntry<const SIZE_BLOCK_ID: usize> {
...
+    severity: i32,
+    dropped_logs: u32,
+}
+
+#[derive(imp::IntoBytes)]
+#[zerocopy(crate = "zerocopy_renamed")]
+#[repr(C)]
+struct FxLogPacket {
+    metadata: FxLogMetadata,
+    data: [u8; FX_LOG_MAX_DATAGRAM_LEN - FX_LOG_METADATA_SIZE],
```
joshlf, line 253: Same here re: `size_of`.

PR #3253, opened 2026-04-13
[simd] support more simd types for powerpc(64) and s390x

usamoi:
They're missing probably because they're not supported by rust when they're added to `zerocopy`.


usamoi:
Hi @joshlf, could you review this?

PR #3261, opened 2026-04-14
[anneal] Add lean package management (lake) files

mdittmer:
Bootstrap code intelligence for `anneal/src/**/*.lean` by adding a `lakefile.toml` and the additional versioning files generated by `lake update`.


joshlf:
I've seen this failure with other PRs – I think it's a permissions issue with external contributors not being allowed to push to our container registry cache. I'll need to fix this issue before we can land this (unless you want me to just re-push from my account).

mdittmer:
I'm in no rush. We can wait until that issue is fixed.

Issue #10, opened 2022-09-19
Support deriving AsBytes on types with type parameters


joshlf:
*Migrated from https://fxbug.dev/100235*

In order to accomplish this, we need to be able to emit an impl which is bounded by a const expr, which requires the unstable feature [`generic_const_exprs`](https://github.com/rust-lang/rust/issues/76560). This issue is blocked until that feature is stabilized.

bjorn3:
Would it be possible to have a special case if the type parameter has a bound on `ByteOrder` and only on `ByteOrder` by checking all implementations of `ByteOrder`? `ByteOrder` is a sealed trait, so all implementations of it are known. This would make it much easier to write code that handles file formats that can be both little and big endian.

joshlf:
I'm not sure what the relevance of `ByteOrder` is here? The byteorder-aware types in the [byteorder module](https://docs.rs/zerocopy/0.7.0-alpha.2/zerocopy/byteorder/index.html) implement `AsBytes`, but a type can satisfy the rules of `AsBytes` even if its type parameters do not implement `ByteOrder`. What's the connection you're getting at?

bjorn3:
I did like to have types like

```rust
#[derive(FromBytes)]
struct Foo<B: ByteOrder> {
    a: U32<B>,
    b: U32<B>,
}
```

to work with both little endian and big endian using the same type definition. However the restriction of not having any type parameters when deriving FromBytes prevents this. I thought it did be easier to allow FromBytes just for type parameters bound by ByteOrder than implementing the general case.

joshlf:
That code should work fine. It's just `#[derive(AsBytes)]` that doesn't currently support generics (actually, on ToT, it supports generics on a struct so long as that struct is also `#[repr(transparent)]` or `#[repr(packed)]`).

bjorn3:
> (actually, on ToT, it supports generics on a struct so long as that struct is also #[repr(transparent)] or #[repr(packed)]).

I see. Adding `#[repr(packed)]` would work for my case.

InternetOfTofu:
I wonder if single field `#[repr(C)]` generic structs can be treated the same way as `#[repr(transparent)]` for this purpose?

joshlf:
They could, but I'm not sure what the point would be. If you're considering adding `#[derive(AsBytes)]` to such a type, then:
- You own the code, and so you can change it
- You could replace `#[repr(C)]` with `#[repr(transparent)]` and not change the behavior
[...]

PR #3324, opened 2026-04-22
Support structs with same type

z-tech:
## What

Let `derive(IntoBytes)` accept `#[repr(C)]` generic structs when every
field has the same type. Only requires `T: IntoBytes` — no `Unaligned`
bound.

## Why it's sound

`#[repr(C)]` struct, N fields all of type `T`, no `#[repr(align)]`:
field `i` sits at `i * size_of::<T>()`, size is `N * size_of::<T>()`,
struct alignment is `align_of::<T>()`. Everything's a multiple of
alignment, so no padding. If `T: IntoBytes`, so is the struct.

## Why I want it

arkworks has things like:

```rust
#[repr(C)]
pub struct QuadExtField<P: QuadExtConfig> {
    pub c0: P::BaseField,
    pub c1: P::BaseField,
}
```
Can't use repr(transparent) (two fields). Existing generic branch
wants P::BaseField: Unaligned, which it isn't (Montgomery form,
align 8). So today the only way to treat &[Fq2] as bytes is a
hand-written unsafe transmute.

With this, elems.as_bytes() just works.

Partially addresses [#10](https://github.com/google/zerocopy/issues/10)

### DIFF
--- a/zerocopy-derive/src/derive/into_bytes.rs
+++ b/zerocopy-derive/src/derive/into_bytes.rs
@@ -114,8 +114,8 @@ fn derive_into_bytes_struct(ctx: &Ctx, strct: &DataStruct) -> Result<TokenStream
 }
 
 fn all_fields_same_type(strct: &DataStruct) -> bool {
-    let mut fields =
-        strct.fields().into_iter().map(|(_, _, ty)| ty.into_token_stream().to_string());
+    let fields = strct.fields();
+    let mut fields = fields.into_iter().map(|(_, _, ty)| ty.into_token_stream().to_string());
     if let Some(first) = fields.next() {
         fields.all(|field| field == first)
     } else {
--- a/zerocopy-derive/src/derive/into_bytes.rs
+++ b/zerocopy-derive/src/derive/into_bytes.rs
@@ -1,5 +1,5 @@
 use proc_macro2::{Span, TokenStream};
-use quote::quote;
+use quote::{quote, ToTokens};
 use syn::{Data, DataEnum, DataStruct, DataUnion, Error, Type};
 
 use crate::{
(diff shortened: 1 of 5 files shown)


zerocopy-derive/tests/struct_to_bytes.rs
```diff
@@ -160,6 +160,37 @@ struct ReprCGenericOneField<T: ?imp::Sized> {
...
+#[zerocopy(crate = "zerocopy_renamed")]
+#[repr(C)]
+struct ReprCGenericHomogeneousTuple<T>(T, T);
+
+util_assert_impl_all!(ReprCGenericHomogeneousTuple<u8>: imp::IntoBytes);
+util_assert_impl_all!(ReprCGenericHomogeneousTuple<u64>: imp::IntoBytes);
+// `AU16` has alignment > 1, so the non-homogeneous branch would have
+// rejected it (see `ReprCGenericMultipleFields<_, AU16>` below). The
+// homogeneous case accepts it because `N * size_of::<AU16>()` is a
+// multiple of `align_of::<AU16>()`, so there's still no padding.
```
jswrenn:
```suggestion
```
No need to say so much. :) Comments in tests that reference the current implementation details of zerocopy are likely to fall out of date.

z-tech:
Concrete example: https://github.com/arkworks-rs/algebra/pull/1097/changes#diff-2469de9bf2aa1061df2b63a13ddc2a145bc719a177c77cf1320a0d17b51aaeccR87

zerocopy-derive/src/derive/into_bytes.rs
```diff
@@ -96,6 +114,30 @@ fn derive_into_bytes_struct(ctx: &Ctx, strct: &DataStruct) -> Result<TokenStream
...
+/// same syntactic type token tree.
+///
+/// Comparison is syntactic (tokens only), not semantic: `T` and `Self::T`
+/// compare unequal even if they resolve to the same type. That's
+/// deliberate — a proc macro has no type resolution, and the soundness
+/// argument we rely on ("N copies of the same type ⇒ no padding under
+/// `repr(C)`") only holds when the types are actually the same at the
+/// Rust-layout level. Being strict here keeps the rule trivially sound;
+/// users wanting the optimisation for type-alias-equal fields can spell
+/// the fields with a single alias.
```
jswrenn:
```suggestion
```
This is said at the use-site of `all_fields_same_type` and does not need to be repeated here.

zerocopy-derive/src/derive/into_bytes.rs
```diff
@@ -96,6 +114,30 @@ fn derive_into_bytes_struct(ctx: &Ctx, strct: &DataStruct) -> Result<TokenStream
...
+    let fields = strct.fields();
+    if fields.len() < 2 {
+        // Zero- and one-field cases are already handled by the
+        // `num_fields <= 1` branch above. Return `false` so this branch
+        // is never mistaken for that one.
+        return false;
+    }
+    let mut tokens = fields.iter().map(|(_, _, ty)| ty.to_token_stream().to_string());
+    let first = tokens.next().expect("len >= 2");
+    tokens.all(|t| t == first)
```
jswrenn:
```suggestion
    let mut fields =
        strct.fields().into_iter().map(|(_, _, ty)| ty.into_token_stream().to_string());
    if let Some(first) = fields.next() {
        fields.all(|field| field == first)
    } else {
        true
    }
```
I'd rather avoid the `len >= 2` condition so this predicate can be reasoned about independently of how it's *currently* used.

jswrenn (commented):
This is a long-overdue addition to zerocopy! Here's some preliminary feedback.

z-tech:
Hi @jswrenn 👋. Happy to hear! Accepted suggestions. Let me know how else I can help. Thanks

PR #3364, opened 2026-05-13
Only use `layout_for_ptr` feature for zerocopy tests

erickt:
This puts the use of the unstable feature `layout_for_ptr` behind `all(test, miri)`, since it is only necessary for zerocopy tests, not for downstream users.


erickt:
The anneal test failures seem to be some unrelated docker issue as far as I can tell.

joshlf:
Superseded by #3365 to work around Docker registry permissions issues

PR #3429, opened 2026-06-03
[byteorder] Add cfg no_fp_fmt_parse

### LINKED ISSUE #3426: Linux kernel: `cfg(no_fp_fmt_parse)`
*Context: For our incoming use of `zerocopy` and `zerocopy-derive` in the Linux kernel, I only needed a couple small patches to adapt the crates. It would be nice (but they are not blockers at the moment) if we could have these upstream to simplify upgrades later on.*

---

The most important one is avoiding the `Display` impls for floating point: please see https://lore.kernel.org/rust-for-linux/20260602172920.30342-11-ojeda@kernel.org/ for the single line Linux kernel patch that implements the idea there.

Essentially, the idea would be to have in upstream Rust a `cfg` similar to `core`'s `no_fp_fmt_parse` which we would enable in our builds.

eiffel-fl:
This cfg deactivates Debug and Display for floatting point numbers, this is particluarly useful in kernel context.
The implementation is inspired by:
https://github.com/rust-lang/rust/commit/ec7292ad3c35

Fixes #3426

### DIFF
--- a/zerocopy/build.rs
+++ b/zerocopy/build.rs
@@ -95,6 +95,7 @@ fn main() {
         println!("cargo:rustc-check-cfg=cfg(coverage_nightly)");
         println!("cargo:rustc-check-cfg=cfg(zerocopy_inline_always)");
         println!("cargo:rustc-check-cfg=cfg(zerocopy_unstable_ptr)");
+        println!("cargo:rustc-check-cfg=cfg(no_fp_fmt_parse)");
     }
 
     for version_cfg in version_cfgs {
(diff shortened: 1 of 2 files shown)


joshlf (approved):
Looks good, thanks for this! If you rebase on `main`, the CI errors should go away (fixed by #3430)

ojeda:
Thanks for this!

eiffel-fl:
Welcome! Thanks for the review and merge!

PR #3432, opened 2026-06-04
zerocopy: add SPDX-License-Identifier to source files

### LINKED ISSUE #3428: Linux kernel: SPDX license identifiers
*Context: For our incoming use of `zerocopy` and `zerocopy-derive` in the Linux kernel, I only needed a couple small patches to adapt the crates. It would be nice (but they are not blockers at the moment) if we could have these upstream to simplify upgrades later on.*

---

This is a trivial one, and we can definitely handle it on our side at upgrade time, but for completeness: we also add SPDX license identifiers at the top of every file. These one line comments could be added to upstream `zerocopy` which could help codebases that vendor `zerocopy` and at the same time care about SPDX tagging.

A couple notes, though:

  - I didn't add them in our vendored copy on `benches/` because those files get embedded. But I did so for `src/` as well as the `rustdoc` stylesheet.
  - The current Linux kernel SPDX checking script does not like chained `OR` conditions, so I had to add parentheses in one of them. I am not sure yet if there is a particular reason for that in the script, but for upstream `zerocopy` it probably makes sense to match `Cargo.toml`'s `license` field.

Please see https://lore.kernel.org/rust-for-linux/20260602172920.30342-10-ojeda@kernel.org/ and https://lore.kernel.org/rust-for-linux/20260602172920.30342-15-ojeda@kernel.org/ for the Linux kernel patches that add the lines on our side.

Sagartaunk:
Fixes #3428

### DIFF
--- a/zerocopy/src/pointer/inner.rs
+++ b/zerocopy/src/pointer/inner.rs
@@ -1,3 +1,5 @@
+// SPDX-License-Identifier: BSD-2-Clause OR Apache-2.0 OR MIT
+//
 // Copyright 2024 The Fuchsia Authors
 //
 // Licensed under a BSD-style license <LICENSE-BSD>, Apache License, Version 2.0
--- a/zerocopy/src/byte_slice.rs
+++ b/zerocopy/src/byte_slice.rs
@@ -1,3 +1,5 @@
+// SPDX-License-Identifier: BSD-2-Clause OR Apache-2.0 OR MIT
+//
 // Copyright 2024 The Fuchsia Authors
 //
 // Licensed under a BSD-style license <LICENSE-BSD>, Apache License, Version 2.0
--- a/zerocopy/src/byteorder.rs
+++ b/zerocopy/src/byteorder.rs
@@ -1,3 +1,5 @@
+// SPDX-License-Identifier: BSD-2-Clause OR Apache-2.0 OR MIT
+//
 // Copyright 2019 The Fuchsia Authors
 //
 // Licensed under a BSD-style license <LICENSE-BSD>, Apache License, Version 2.0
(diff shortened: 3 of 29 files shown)


zerocopy/src/pointer/inner.rs
```diff
@@ -1,3 +1,4 @@
+// SPDX-License-Identifier: BSD-2-Clause OR Apache-2.0 OR MIT
```
joshlf, line 1: Here and in other `.rs` files:

```suggestion
// SPDX-License-Identifier: BSD-2-Clause OR Apache-2.0 OR MIT
//
```

joshlf (changes requested):
Tiny nit, but otherwise looks good, thanks! You'll also need to rebase on `main`.

joshlf:
Oh and also, you'll need to run `./tools/update-expected-test-output.sh` after you've made any changes.

Sagartaunk:
> Oh and also, you'll need to run `./tools/update-expected-test-output.sh` after you've made any changes.

Made the changes and rebased on `main`. Also, are there any good first issues you'd recommend to someone looking to contribute more?

Sagartaunk:
I have removed the accidental generated-output churn, and the branch now contains only the SPDX changes in .rs files. I haven’t committed the UI expected-output changes because running `./tools/update-expected-test-output.sh` also modifies the bench output on my machine, I am not sure why. Since those bench changes seem unrelated to this PR, I left all generated output out for now.

joshlf:
Looks like I can't push to your branch, but I put up a new PR with your commit (but rebased and having re-run the UI test generation): #3455.

Sagartaunk:
> Looks like I can't push to your branch, but I put up a new PR with your commit (but rebased and having re-run the UI test generation): #3455.

Thanks, should i close this one ?

joshlf:
Yeah I was waiting to see whether the tests passed on the other one, which they now have, so closing this.

PR #3449, opened 2026-06-08
[derive] Remove unnecessary clones

FranciscoTGouveia:
If I am not overlooking anything, many of these clones can be removed as the receiver's type implements `Copy`, or can simply be moved.

CC @nunoplopes


FranciscoTGouveia:
Looking at the CI report, I believe this was a spurious failure.

joshlf (approved):

joshlf:
Thanks! The CI failure is a bug on our side. I'll fix that first and then land this.

joshlf:
Once https://github.com/google/zerocopy/pull/3452 lands on `main`, you can rebase and re-push and this should automatically get merged (assuming no other test failures).

joshlf:
Looks like I can rebase for you through the web UI, so I did that.

PR #3459, opened 2026-06-12
[exocrate] use `XDG` cache dir instead of home dir

### LINKED ISSUE #3420: `[anneal][v2][exocrate]` Stop polluting user home directories, adopt standard base directory conventions
**Note**: This is a _generated issue description_ based on a chat transcript and related documentation.

## Context

Currently, the `exocrate` tool (located in `anneal/v2/exocrate`) may automatically store files, caches, or configuration directly in the root of the user's home directory (`~`). Writing data straight to `~` pollutes user systems and goes against standard desktop/operating system environment practices.

## Proposal

We should ensure that `exocrate` respects platform-specific guidelines for storing application-related data, configurations, and caches. 

Specifically:
- **Linux/Unix**: Respect the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir/).
  - Caches should default to `$XDG_CACHE_HOME` (falling back to `~/.cache`).
  - Data should default to `$XDG_DATA_HOME` (falling back to `~/.local/share`).
  - Configuration should default to `$XDG_CONFIG_HOME` (falling back to `~/.config`).
- **macOS**: Follow Apple's Standard Directory Guidelines (typically `~/Library/Caches` and `~/Library/Application Support`).
- **Windows**: Use the standard Windows Known Folder API (e.g., `%LOCALAPPDATA%` / `AppData\Local`).

## Recommended Implementation

Instead of manually parsing environment variables and trying to handle complex multi-platform behavior, we can integrate the [`directories`](https://docs.rs/directories/latest/directories/) crate. 

The `directories` crate provides a clean, mid-level API (`ProjectDirs` and `BaseDirs`) that implements the correct standard behaviors seamlessly across Linux, macOS, and Windows.

```rust
use directories::ProjectDirs;

if let Some(proj_dirs) = ProjectDirs::from("com", "Google", "exocrate") {
    let cache_dir = proj_dirs.cache_dir();
    let data_dir = proj_dirs.data_local_dir();
    // Use these paths for storing exocrate artifacts/caches
}
```

## References

- [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir/)
- [Windows Known Folders documentation](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/legacy/bb776911(v=vs.85))
- [`directories` crate documentation](https://docs.rs/directories/latest/directories/)

Sagartaunk:
Use XDG cache dir instead of home dir on `UserGlobal` installs, to prevent polluting user home directories.

Fixes: #3420

### DIFF
--- a/exocrate/src/lib.rs
+++ b/exocrate/src/lib.rs
@@ -156,9 +156,9 @@ pub struct Config {
     /// sequence of path components for cross-platform compatibility.
     ///
     /// Dependencies are stored in `<rel_dir_path>`. In production use, this is
-    /// relative to `~`. In development, this is relative to Cargo's `target`
-    /// directory if it can be resolved, and otherwise relative to the current
-    /// working directory.
+    /// relative to the user cache directory. In development, this is relative
+    /// to Cargo's `target` directory if it can be resolved, and otherwise relative
+    /// relative to the current working directory.
     //
     // FIXME(#3408): Make this non-`pub` and validate at construction that each
     // item is valid for the OS.
@@ -169,7 +169,7 @@ pub struct Config {
 
 /// The location to install dependencies.
 pub enum Location {
-    /// A location in the user's home directory.
+    /// A location in the user's cache directory.
     UserGlobal,
     /// A location in the Cargo `target` directory if it can be resolved, and
     /// otherwise in the current working directory.
@@ -189,7 +189,7 @@ pub enum Source {
 impl Config {
     /// Resolves the dependency directory, failing if it doesn't exist.
     pub fn resolve_installation_dir(&self, location: Location) -> IoResult<PathBuf> {
-        let dir_path = self.dir_path(location);
+        let dir_path = self.dir_path(location)?;
         let _ = ManagedDirName::new(&dir_path).check_exists()?;
         Ok(dir_path)
     }
@@ -200,7 +200,7 @@ impl Config {
         location: Location,
         source: Source,
     ) -> IoResult<PathBuf> {
-        let dir_path = self.dir_path(location);
+        let dir_path = self.dir_path(location)?;
         if ManagedDirName::new(&dir_path).check_exists().is_ok() {
             return Ok(dir_path);
         }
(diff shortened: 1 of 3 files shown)


Sagartaunk:
Also, instead of adding `Directories` crate, I have just used `dirs::cache_dir()`. Do tell me if I should change it to use the `Directories` crate instead.

exocrate/src/lib.rs
```diff
@@ -242,13 +242,16 @@ impl Config {
...
     ///   `Local` (or `./target` if `CARGO_MANIFEST_DIR` is not set), and the
     ///   supplied path for `Custom`.
     /// - The remainder is `{self.rel_dir_path}/{self.version_slug}`.
-    fn dir_path(&self, location: Location) -> PathBuf {
+    fn dir_path(&self, location: Location) -> IoResult<PathBuf> {
         let mut parts = match location {
-            Location::UserGlobal => dirs::home_dir().expect("home dir not found"),
+            Location::UserGlobal => dirs::cache_dir().ok_or_else(|| {
+                std::io::Error::new(std::io::ErrorKind::NotFound, "cache dir not found")
+            })?,
```
joshlf, line 254: If we do this, we'll want to change the name of the next path segment to `anneal` instead of `.anneal`. IIRC that's currently only set in users of this crate, not in the crate itself, but you'll want to double check.

joshlf (changes requested):
Thanks for this!! One small comment.

cc @mdittmer who may also have feedback

exocrate/src/lib.rs
Sagartaunk: Yes, I did that in the initial commit but the anneal test failed because of it so I reverted it. I think there are some other areas where it needs to be changed as well though I was not sure about it. 
Should I resubmit after making the change ?

Sagartaunk:
@joshlf any updates on this?

mdittmer:
FWIW, the latest patch looks good to me. I think @joshlf needs to resolve the "Changes requested" though.

mdittmer (approved):

joshlf (approved):

PR #3241, opened 2026-04-12
[ui] Use --diagnostic-width=100 in UI tests

ksapru:
Pass --diagnostic-width=100 to rustc in UI tests to ensure consistent error message line-wrapping regardless of the terminal width of the environment in which the tests are run.

The flag is skipped for the 'msrv' toolchain (Rust 1.56.0) since it was only stabilized in Rust 1.70.0.

The TERM/COLUMNS environment variables are retained since they can still influence other tools, but a comment is updated to clarify that --diagnostic-width is now the primary and more reliable mechanism.

[Fixes #3113](https://github.com/google/zerocopy/issues/3113)

PR #3460, opened 2026-06-12
Use rustc's --diagnostic-width flag in UI tests

### LINKED ISSUE #3113: Use `rustc`'s `--diagnostic-width` flag in UI tests
This will prevent those tests from being sensitive to the terminal in which they're run, which often causes breakage.

joshlf:
Pass --diagnostic-width=100 to rustc in UI tests to ensure consistent error message line-wrapping regardless of the terminal width of the environment in which the tests are run.

The flag is skipped for the 'msrv' toolchain (Rust 1.56.0) since it was only stabilized in Rust 1.70.0.

The TERM/COLUMNS environment variables are retained since they can still influence other tools, but a comment is updated to clarify that --diagnostic-width is now the primary and more reliable mechanism.

Fixes #3113

### DIFF
--- a/zerocopy/tests/ui/diagnostic-not-implemented.stable.stderr
+++ b/zerocopy/tests/ui/diagnostic-not-implemented.stable.stderr
@@ -25,6 +25,8 @@ note: required by a bound in `takes_from_bytes`
    |
 77 | fn takes_from_bytes<T: FromBytes>() {}
    |                        ^^^^^^^^^ required by this bound in `takes_from_bytes`
+   = note: the full name for the type has been written to '/long-type-HASH.txt'
+   = note: consider using `--verbose` to print the full type name to the console
 
 error[E0277]: the trait bound `NotZerocopy: FromZeros` is not satisfied
   --> $DIR/diagnostic-not-implemented.rs:18:24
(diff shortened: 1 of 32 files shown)


joshlf:
This represents #3241 rebased on `main` since I can't force-push to #3241's upstream.

joshlf:
> The flag is skipped for the 'msrv' toolchain (Rust 1.56.0) since it was only stabilized in Rust 1.70.0.

@jswrenn – another reason to bump our MSRV; this will remain unstable because of the MSRV toolchain until we bump to/past 1.70.0.

jswrenn (approved):

zerocopy/tests/ui/diagnostic-not-implemented.stable.stderr
```diff
@@ -25,6 +25,8 @@ note: required by a bound in `takes_from_bytes`
    |
 77 | fn takes_from_bytes<T: FromBytes>() {}
    |                        ^^^^^^^^^ required by this bound in `takes_from_bytes`
+   = note: the full name for the type has been written to '/long-type-HASH.txt'
+   = note: consider using `--verbose` to print the full type name to the console
```
jswrenn, line 29: Can follow up and set the diagnostic width sufficiently high that truncation does not generally occur?

jswrenn (changes requested):

zerocopy/tests/ui/diagnostic-not-implemented.stable.stderr
joshlf: https://github.com/google/zerocopy/pull/3461

PR #3493, opened 2026-07-22
Upgrade zerocopy-derive to syn 3 and re-vendor dependencies

joshlf:
### Motivation
- Bring `zerocopy-derive` onto `syn` 3 so the derive crate can use the new `syn` API surface and avoid lifetime/AST mismatches between runtime and dev-time parsing.
- Ensure the dev/test toolchain (`prettyplease` used by output tests and `proc-macro2`) are aligned with `syn` 3 to keep generated-code tests working.
- Respect the repository's vendoring setup by updating `Cargo.lock` and re-vendoring the exact crates under `zerocopy/vendor/` rather than relying on crates.io at build time.

### Description
- Bumped `syn` in `zerocopy/zerocopy-derive/Cargo.toml` to `3.0.0` for both normal and dev dependencies and updated the dev `prettyplease` pin to `0.3.0` to match `syn` 3.
- Updated `zerocopy/Cargo.lock` so `zerocopy-derive` resolves to `syn 3.0.3` while retaining `syn 2.0.56` for other transitive users.
- Re-vendored and committed the affected crates into `zerocopy/vendor/`, notably `syn-3.0.3`, `prettyplease 0.3.0`, and `proc-macro2` updates, and adjusted vendor metadata files accordingly.
- Applied small compatibility updates in vendored `prettyplease`/`proc-macro2` sources (as part of the vendoring snapshot) so the workspace builds against the new `syn` major version.

### Testing
- Ran `CARGO_ZEROCOPY_AUTO_INSTALL_TOOLCHAIN=1 ./cargo.sh +stable check -p zerocopy-derive` and it completed successfully.
- Ran `CARGO_ZEROCOPY_AUTO_INSTALL_TOOLCHAIN=1 ./cargo.sh +stable check --tests -p zerocopy-derive` and it completed successfully.
- Ran `CARGO_ZEROCOPY_AUTO_INSTALL_TOOLCHAIN=1 ./cargo.sh +stable test -p zerocopy-derive --lib` and all derive unit tests passed (`23 passed`).
- Attempted to run `./githooks/pre-push`; the script ran but encountered missing host tools (`yq`) and an interactive toolchain install prompt which prevented a full automated run, so the full pre-push check was not completed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a603c4630308333b9dcef7278bf30fd)


Crovitche-1623:
Is something blocking this @joshlf ?

Downstream, `zerocopy-derive` is the last crate holding `syn` 2 in our tree.

I tried the bump on current `main` out of curiosity: it needs `prettyplease` 0.2 -> 0.3 alongside it, since `prettyplease::unparse(&syn::parse_file(..))` in `src/output_tests/mod.rs` stops type-checking otherwise, and no source change beyond that.

Happy to wait if it is just timing, it would help to know which.

PR #3543, opened 2026-08-04
[exocrate] Lift `parse_remote_archive` dependency on `toml_const`

### LINKED ISSUE #3409: `exocrate::parse_remote_archive`: Lift requirement that calling crate depend on `toml_const` directly

### LINKED ISSUE #3410: `exocrate::parse_remote_archive`: Don't require the user to specify os/arch pairs

Sagartaunk:
Changelog:
- `exocrate/lib.rs` : Write helper const fn's `bytes_eq_at`, `find_line` and `find_field` to replace `toml_const` from `parse_remote_archive` macro. This also resolves #3410 since the section header is now located by scanning rather than matching against a pre-declared list.

- `anneal/main.rs` , `anneal/setup.rs` : Fix the macro declaration.
- `anneal/cargo.toml` : Remove `toml_const` as a dependency.

Fixes: #3409 , #3410

### DIFF
--- a/anneal/Cargo.lock
+++ b/anneal/Cargo.lock
@@ -288,7 +288,6 @@ dependencies = [
  "sha2",
  "tempfile",
  "thiserror",
- "toml_const",
  "walkdir",
 ]
 
@@ -377,7 +376,7 @@ dependencies = [
  "strip-ansi-escapes",
  "take_mut",
  "tempfile",
- "toml 0.8.23",
+ "toml",
  "tracing",
  "tracing-subscriber",
  "tracing-tree",
@@ -1445,49 +1444,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "phf"
-version = "0.13.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c1562dc717473dbaa4c1f85a36410e03c047b2e7df7f45ee938fbef64ae7fadf"
-dependencies = [
- "phf_macros",
- "phf_shared",
- "serde",
-]
-
-[[package]]
-name = "phf_generator"
-version = "0.13.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "135ace3a761e564ec88c03a77317a7c6b80bb7f7135ef2544dbe054243b89737"
-dependencies = [
- "fastrand",
- "phf_shared",
-]
-
-[[package]]
-name = "phf_macros"
-version = "0.13.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "812f032b54b1e759ccd5f8b6677695d5268c588701effba24601f6932f8269ef"
-dependencies = [
- "phf_generator",
- "phf_shared",
- "proc-macro2",
- "quote",
- "syn 2.0.117",
-]
-
-[[package]]
-name = "phf_shared"
-version = "0.13.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e57fef6bc5981e38c2ce2d63bfa546861309f875b8a75f092d1d54ae2d64f266"
-dependencies = [
- "siphasher",
-]
-
-[[package]]
 name = "pin-project-lite"
 version = "0.2.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1846,15 +1802,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "serde_spanned"
-version = "1.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6662b5879511e06e8999a8a235d848113e942c9124f211511b16466ee2995f26"
-dependencies = [
- "serde_core",
-]
-
-[[package]]
 name = "serde_stacker"
 version = "0.1.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1921,12 +1868,6 @@ name = "simd-adler32"
 version = "0.3.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "703d5c7ef118737c72f1af64ad2f6f8c5e1921f818cdcb97b8fe6fc69bf66214"
-
-[[package]]
-name = "siphasher"
-version = "1.0.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8ee5873ec9cce0195efcb7a4e9507a04cd49aec9c83d0389df45b1ef7ba2e649"
 
 [[package]]
 name = "slab"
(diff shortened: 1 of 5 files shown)


joshlf:
Thanks for this!

I'm skeptical that reimplementing a parser by hand is the way to go – I had ChatGPT review, and it found some bugs that would probably be annoying to fix in a robust way.

I took a stab at fixing the underlying issue upstream, which turned out to be pretty straightforward: https://github.com/facesthe/toml_const/pull/12

IIUC that won't address https://github.com/google/zerocopy/issues/3410, so there's still interesting work to do here, but we probably shouldn't do that by reimplementing the functionality provided by toml_const. Maybe we could upstream support for this to toml_const itself?

I'd suggest that we wait for that to merge/get published on crates.io.

---

*Here's the ChatGPT review:*:

## Verdict

**Request changes.** Reviewed PR #3543 at head `942ad9c`. The dependency-removal goal is reasonable, but the replacement is a textual line scanner rather than a TOML parser. It can both reject ordinary valid TOML and, more seriously, silently construct a `RemoteArchive` from text that does not belong to the selected TOML table.

## Findings

### 1. High: table-looking text inside a multiline string is parsed as real configuration

**Anchor:** `exocrate/src/lib.rs:580–633`, particularly line 587.

`find_line` considers every physical line independently and does not track whether it is inside a TOML string. This valid manifest therefore selects the zero hash and `wrong.example`, even though TOML assigns the one hash and `right.example` to the requested table:

```toml
[package.metadata]
example = """
[package.metadata.exocrate.linux.x86_64]
sha256 = "0000000000000000000000000000000000000000000000000000000000000000"
url = "https://wrong.example/archive"
"""

[package.metadata.exocrate.linux.x86_64]
sha256 = "1111111111111111111111111111111111111111111111111111111111111111"
url = "https://right.example/archive"
```

The same problem exists in `find_field`: field-looking lines inside a multiline string within the real table can override later real fields. `package.metadata` has no Cargo-defined schema, and multiline strings are standard TOML, so this is not an invalid-manifest corner case. The macro can return a URL and hash that do not exist at that location in Cargo’s parsed metadata.  ([[TOML](https://toml.io/en/v1.0.0?utm_source=chatgpt.com)][1])

**Suggested review comment:**

> `find_line` cannot safely search physical lines without tracking TOML lexical state. A line matching this prefix may be content inside a multiline basic or literal string, in which case the macro silently reads configuration that does not exist in the TOML data model. Please keep semantic TOML parsing rather than recognizing table syntax textually.

---

### 2. High: fields can bleed in from another platform table

**Anchor:** `exocrate/src/lib.rs:657–659`.

`find_field` decides that the current table has ended only when the first byte of a line is `[`:

```rust
if bytes[i] == b'[' {
    return None;
}
```

TOML ignores indentation before a table header. Consequently, this valid manifest defines an empty Linux table and a populated macOS table, but a Linux build returns the macOS URL and hash:

```toml
[package.metadata.exocrate.linux.x86_64]
  [package.metadata.exocrate.macos.x86_64]
sha256 = "0000000000000000000000000000000000000000000000000000000000000000"
url = "https://mac.example/archive"
```

This is not merely an unnecessarily strict parser: it silently selects a different platform’s archive while pairing it with that archive’s valid hash. Trimming whitespace before checking for `[` would fix this example but would not fix table-looking lines inside multiline strings.  ([[TOML](https://toml.io/en/v1.0.0?utm_source=chatgpt.com)][1])

**Suggested review comment:**

> This table-boundary check misses valid indented headers, so a missing field in the selected table can be borrowed from a later table—potentially another platform’s table. Since indentation is TOML whitespace, this scan needs semantic table tracking rather than a byte-zero `[` check.

---

### 3. Medium: ordinary valid TOML spellings are rejected

**Anchor:** `exocrate/src/lib.rs:587–632` and `661–679`.

The implementation recognizes only:

* A header beginning in column zero, with no whitespace around dotted components.
* An unquoted bare key beginning in column zero.
* Exactly one space on each side of `=`.
* A double-quoted, single-line basic string.

For example, this is valid TOML but the new parser finds neither the header nor the fields:

```toml
[ package . metadata . exocrate . linux . x86_64 ]
sha256='0000000000000000000000000000000000000000000000000000000000000000'
url="https://example.com/archive"
```

Other valid representations that fail include tabs or multiple spaces, quoted keys, literal strings, and equivalent dotted-key or inline-table definitions. TOML explicitly ignores whitespace around keys and dotted components and supports bare, quoted, and dotted keys plus four string forms.  ([[TOML](https://toml.io/en/v1.0.0?utm_source=chatgpt.com)][1])

This is also brittle across packaging boundaries: Cargo rewrites and normalizes `Cargo.toml` when creating a distributable package, so source builds and builds from a packaged crate need not even inspect identical manifest text. ([[Rust Documentation](https://doc.rust-lang.org/nightly/cargo/commands/cargo-package.html)][2])

---

### 4. High: basic-string escapes are returned without TOML decoding

**Anchor:** `exocrate/src/lib.rs:670–679`.

`find_field` returns the raw bytes until the next `"` rather than the TOML string value. For example:

```toml
url = "https://example.com/\u0061.tar.zst"
```

The TOML value is:

```text
https://example.com/a.tar.zst
```

The macro instead stores the literal characters `\u0061`. An escaped quote such as `\"` terminates the scan early, and a multiline basic string beginning with `"""` is read as an empty value. A semantically valid escaped `sha256` similarly fails `decode_hex`. TOML basic strings require escape processing and quote recognition that accounts for escaping.  ([[TOML](https://toml.io/en/v1.0.0?utm_source=chatgpt.com)][1])

**Suggested review comment:**

> This slices the source representation of a TOML basic string rather than decoding its value. Unicode escapes and escaped quotes therefore produce a different `url` from Cargo’s parsed metadata. Supporting this correctly requires a TOML string parser, including multiline and literal string forms.

---

### 5. Low: `bytes_eq_at` can overflow despite its documented behavior

**Anchor:** `exocrate/src/lib.rs:553–555`.

```rust
if offset + to_search.len() > data.len()
```

The addition can overflow for an arbitrary caller-provided offset, instead of returning `false` as documented. A safe formulation is:

```rust
if offset > data.len() || to_search.len() > data.len() - offset {
    return false;
}
```

This does not affect current macro calls because their offsets are internally derived, but the helper is public, albeit hidden.

## Recommended implementation

I would not extend these `const fn`s into a partial TOML parser. A cleaner design is a private proc-macro crate that:

1. Reads and parses the invoking package’s manifest using the established `toml` parser.
2. Enumerates the discovered OS/architecture tables.
3. Emits target-side match arms and the final `RemoteArchive` literal.
4. Emits an `include_str!` dependency so manifest changes invalidate compilation.
5. Is re-exported through `exocrate`, so callers still need no direct parsing dependency.

That satisfies both intended issues: the caller no longer depends directly on `toml_const`, and OS/architecture pairs are discovered rather than repeated in the invocation. It also keeps parsing work on the host at macro-expansion time and avoids shipping a TOML parser in the target binary.

If the textual subset is intentionally part of the API, it needs a formally specified grammar and must fail closed whenever strings, arrays, inline tables, quoted keys, or ambiguous section boundaries appear. It still should not silently scan into another semantic TOML value or table.

## Tests needed before merge

The current in-tree manifests exercise the one canonical spelling expected by the scanner, but there are no adversarial parser cases in the PR.

At minimum, tests should cover:

* Decoy headers and fields inside both kinds of multiline string.
* Indented subsequent table and array-of-table headers.
* Missing fields that must not be borrowed from another table.
* Tabs, zero/multiple spaces, quoted keys, and whitespace around dotted components.
* Basic-string escapes, escaped quotes, literal strings, and multiline strings.
* Equivalent dotted-key and inline-table representations.
* LF and CRLF manifests.
* A published/normalized `Cargo.toml` fixture.

## Minor documentation cleanup

The macro documentation still says it parses from `$cargo_toml_path`, although that parameter was removed. The TOML code fence also remains open around the Rust invocation example, and `find_line` documents a different table path—`[metadata.exocrate.manifest.…]`—than the implementation searches.

Scope: source review against commit `942ad9c`; I did not run the repository’s Rust test suite in this environment.

[1]: https://toml.io/en/v1.0.0?utm_source=chatgpt.com "TOML: English v1.0.0"
[2]: https://doc.rust-lang.org/nightly/cargo/commands/cargo-package.html "https://doc.rust-lang.org/nightly/cargo/commands/cargo-package.html"

Sagartaunk:
Okay, So for now we should wait for the upstream to make changes right? 
Then should I close this PR or should I continue to improve this in case this might be required? 
I'll also look into toml_const once they upstream the changes if we'll need some changes there. 
Thanks.

joshlf:
You're welcome to work on this PR if you want, but I suspect we'll end up closing it, assuming upstream accepts the PR.

My guess is that the best use of your time would be to investigate, assuming that the PR is accepted upstream, how we can still support not requiring the user to specify OS/arch pairs. If need be, we could upstream more functionality to support that too.

Sagartaunk:
Okay, Thanks then I'll close this PR. 
Also, are there any other relatively low-priority issues in zerocopy I could look into?

joshlf:
> Also, are there any other relatively low-priority issues in zerocopy I could look into?

Unfortunately I have a good chunk of work queued up that is going to address many open issues, so I'd recommend checking back in a few weeks, as a lot may change in those few weeks.

PR #3544, opened 2026-08-04
#[zerocopy(crate = "...")] should parse paths, not idents

joshuaseaton:
Currently it parses the supplied literal as an identifier rather than a
path (contrary to its internal documentation).

This allows a crate to export an attribute macro of its own that deals
in re-exported zerocopy derive macros.

### DIFF
--- a/zerocopy/zerocopy-derive/src/util.rs
+++ b/zerocopy/zerocopy-derive/src/util.rs
@@ -13,9 +13,9 @@ use std::num::NonZeroU32;
 use proc_macro2::{Span, TokenStream};
 use quote::{quote, quote_spanned, ToTokens};
 use syn::{
-    parse_quote, spanned::Spanned as _, Data, DataEnum, DataStruct, DataUnion, DeriveInput, Error,
-    Expr, ExprLit, Field, GenericParam, Ident, Index, Lit, LitStr, Meta, Path, Type, Variant,
-    Visibility, WherePredicate,
+    parse::ParseBuffer, parse_quote, spanned::Spanned as _, token::PathSep, Data, DataEnum,
+    DataStruct, DataUnion, DeriveInput, Error, Expr, ExprLit, Field, GenericParam, Ident, Index,
+    Lit, LitStr, Meta, Path, Type, Variant, Visibility, WherePredicate,
 };
 
 use crate::repr::{CompoundRepr, EnumRepr, PrimitiveRepr, Repr, Spanned};
@@ -32,6 +32,56 @@ pub(crate) struct Ctx {
     pub(crate) on_error_span: Option<proc_macro2::Span>,
 }
 
+#[derive(Eq, PartialEq)]
+enum CratePath {
+    External,
+    CrateRelative,
+    ModuleRelative,
+}
+
+fn validate_crate_path(path: &Path) -> Result<CratePath, ()> {
+    if path.segments.is_empty() {
+        return Err(());
+    }
+
+    enum ModuleRelative {
+        Yes,
+        No,
+    }
+    let first = path.segments[0].ident.to_string();
+    let (mut prev_segment, path_type) = match first.as_str() {
+        "Self" => return Err(()),
+        "crate" => {
+            if path.leading_colon.is_some() {
+                return Err(());
+            }
+            (ModuleRelative::No, CratePath::CrateRelative)
+        }
+        "self" | "super" => {
+            if path.leading_colon.is_some() {
+                return Err(());
+            }
+            (ModuleRelative::Yes, CratePath::ModuleRelative)
+        }
+        _ => (ModuleRelative::No, CratePath::External),
+    };
+
+    for seg in path.segments.iter().skip(1) {
+        let ident = seg.ident.to_string();
+        match ident.as_str() {
+            "Self" | "crate" | "self" => return Err(()),
+            "super" => match prev_segment {
+                ModuleRelative::Yes => {}
+                ModuleRelative::No => return Err(()),
+            },
+            _ => {
+                prev_segment = ModuleRelative::No;
+            }
+        }
+    }
+    Ok(path_type)
+}
+
 impl Ctx {
     /// Attempt to extract a crate path from the provided attributes. Defaults to
     /// `::zerocopy` if not found.
@@ -45,16 +95,32 @@ impl Ctx {
                 if meta_list.path.is_ident("zerocopy") {
                     attr.parse_nested_meta(|meta| {
                         if meta.path.is_ident("crate") {
-                            let expr = meta.value().and_then(|value| value.parse());
+                            let expr = meta.value().and_then(ParseBuffer::parse);
                             if let Ok(Expr::Lit(ExprLit { lit: Lit::Str(lit), .. })) = expr {
-                                if let Ok(path_lit) = lit.parse::<Ident>() {
-                                    path = parse_quote!(::#path_lit);
-                                    return Ok(());
+                                if let Ok(mut path_lit) = lit.parse_with(Path::parse_mod_style) {
+                                    if let Ok(crate_path) = validate_crate_path(&path_lit) {
+                                        // If not expressly relative, absolutize.
+                                        if path_lit.leading_colon.is_none() && crate_path == CratePath::External {
+                                            path_lit.leading_colon = Some(PathSep::default());
+                                        }
+                                        path = path_lit;
+                                        return Ok(());
+                                    }
+
+                                    return Err(Error::new(
+                                        lit.span(),
+                                        "`crate` attribute requires a valid module path",
+                                    ));
                                 }
+
+                                return Err(Error::new(
+                                    lit.span(),
+                                    "`crate` attribute requires a path as the value",
+                                ));
                             }
 
                             return Err(Error::new(
-                                Span::call_site(),
+                                meta.path.span(),
                                 "`crate` attribute requires a path as the value",
                             ));
                         }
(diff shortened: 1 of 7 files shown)


joshuaseaton:
@joshlf or @djkoloski, if you could take a look when you get a chance, that'd be very appreciated! (The hypothetical use-case in the commit message is a real-life one over in fuchsia.git)

djkoloski:
Can you add a test case for path parsing? See `zerocopy/zerocopy-derive/tests` for examples of derive tests. For example, `hygiene.rs` uses `#[zerocopy(crate = "zerocopy_renamed")]`. So a similar test for a nested path should be very similar.

joshlf:
In lieu of the ChatGPT connector working, here's a manually-requested ChatGPT review. I'll follow up with my own separately.

EDIT: Looks like ChatGPT caught everything I would have mentioned.

---

## Review result

**Verdict: request changes.** The intended nested external-crate path works in principle, but the implementation mishandles other paths that `syn::Path` accepts. I found one blocking correctness issue and one smaller input-validation regression. The PR does not alter unsafe code or safety invariants. The current head remains `f41ebb1`, changing only `zerocopy-derive/src/util.rs`.

## Findings

### 1. **BLOCKING — valid relative paths are converted into invalid Rust**

**Location:** `zerocopy/zerocopy-derive/src/util.rs:52–54`

The code parses the literal as a `syn::Path`, then unconditionally executes:

```rust
path_lit.leading_colon = Some(PathSep::default());
```

However, Syn accepts paths beginning with `crate`, `self`, `super`, and `Self`. Consequently:

```rust
#[zerocopy(crate = "crate::reexports::zerocopy")]
```

is parsed successfully and then rewritten as:

```rust
::crate::reexports::zerocopy
```

Rust rejects that form: global paths cannot begin with `crate`. The same basic problem applies to `self::` and `super::`. Syn’s vendored parser explicitly accepts these qualifiers, while rustc has a regression test demonstrating that `crate::S` is valid but `::crate::S` is not.

This matters whenever the derive invocation needs to refer to a zerocopy re-export in the invoking crate, rather than through an external-crate name. It also means the implementation does not actually accept arbitrary Rust paths, despite switching to `syn::Path`.

The fix should either:

* preserve unrooted `crate::`, `self::`, and `super::` paths; or
* explicitly define this attribute as accepting only external-crate-style paths and reject relative qualifiers during attribute parsing.

Given the PR title and existing documentation’s use of “path,” preserving valid relative paths is the less surprising behavior. The normalization should resemble:

```rust
if let Ok(mut path_lit) = lit.parse_with(Path::parse_mod_style) {
    let first = path_lit.segments[0].ident.to_string();
    let is_relative =
        matches!(first.as_str(), "crate" | "self" | "super");

    if first == "Self"
        || (path_lit.leading_colon.is_some() && is_relative)
    {
        return Err(Error::new(
            lit.span(),
            "`crate` attribute requires a valid module path",
        ));
    }

    if path_lit.leading_colon.is_none() && !is_relative {
        path_lit.leading_colon = Some(PathSep::default());
    }

    path = path_lit;
    return Ok(());
}
```

The final implementation should also reject `crate`, `self`, `super`, or `Self` when they occur in an invalid later segment, such as `foo::crate::bar`.

**Ready-to-post inline comment:**

> `syn::Path` accepts `crate::`, `self::`, and `super::` prefixes, but unconditionally assigning `leading_colon` converts these into invalid paths such as `::crate::...`. For example, `#[zerocopy(crate = "crate::reexports::zerocopy")]` is a valid relative path before this transformation, but rustc rejects the generated path. Please preserve relative leading qualifiers, or explicitly reject them if this attribute is intended to accept only external-crate paths. A regression test for `crate::...` would catch this.

---

### 2. **LOW — `parse_str(lit.value())` drops existing literal validation and span information**

**Location:** `zerocopy/zerocopy-derive/src/util.rs:52`

This:

```rust
syn::parse_str::<Path>(&lit.value())
```

should be:

```rust
lit.parse_with(Path::parse_mod_style)
```

There are three reasons:

1. **It preserves the original literal’s span.** `syn::parse_str` gives every resulting token call-site hygiene, while `LitStr::parse_with` deliberately respans all parsed tokens to the source literal.
2. **It preserves suffix validation.** Rust permits an arbitrary literal suffix to reach a macro, so `"zerocopy"typo` is valid macro input that the macro must interpret or reject. The previous `lit.parse::<Ident>()` path rejected a suffix through `LitStr::parse_with`; calling `.value()` discards the suffix and silently accepts the path as `zerocopy`. ([[Rust Documentation](https://doc.rust-lang.org/reference/tokens.html?utm_source=chatgpt.com)][1])
3. **It parses the correct grammar.** Syn specifically documents `Path::parse_mod_style` for strings such as `"a::b::c"` where generic arguments are not permitted. A crate/module prefix should not accept values such as `"foo::<T>"`.

This can be addressed as part of the blocking fix rather than as a separate change.

## Tests

The PR currently adds no test; its only changed file is `util.rs`. The existing tests use only the single-identifier form:

```rust
#[zerocopy(crate = "zerocopy_renamed")]
```

so they execute the modified lines without proving that a path containing `::` works. This explains why line coverage remains green despite the new behavior being untested. Another reviewer has already requested nested-path coverage.

I would add both forms:

```rust
extern crate self as derive_path_test;
pub extern crate zerocopy_renamed as reexported_zerocopy;

#[derive(reexported_zerocopy::FromBytes)]
#[zerocopy(
    crate = "derive_path_test::reexported_zerocopy"
)]
#[repr(C)]
struct NestedExternalPath(u8);

#[derive(reexported_zerocopy::FromBytes)]
#[zerocopy(crate = "crate::reexported_zerocopy")]
#[repr(C)]
struct CrateRelativePath(u8);
```

The first covers the PR’s stated external-wrapper use case. The second catches the unconditional-`::` bug.

A UI test for an invalid value such as this would also preserve the old suffix-rejection behavior:

```rust
#[zerocopy(crate = "zerocopy_renamed"typo)]
```

## CI and formatting

At the reviewed head, the `Build & Tests` workflow is failing. The two failing jobs I inspected—`Check Rust formatting` and `Run Git hooks`—both fail because the new Syn imports are not rustfmt-formatted; the actual compilation and test matrix jobs passed. Running the repository formatter should merge the two standalone imports into the existing grouped `syn::{...}` import.

I did not rerun the suite locally because the available environment has no Rust toolchain. The review is based on the full changed file, surrounding derive tests, the vendored Syn implementation, rustc’s path tests, and the completed CI logs.

[1]: https://doc.rust-lang.org/reference/tokens.html?utm_source=chatgpt.com "Tokens - The Rust Reference"

joshuaseaton:
Thank you both for the quick response! Just updated things

joshlf:
Lmk once CI is passing and I'll re-review.

To address the current failures, run `./tools/update-expected-test-output.sh`.

joshuaseaton:
Alrighty - CI is passing now

joshlf:
Looks good! ChatGPT found one more issue – let's take its suggested approach of rejecting nested `crate`/`self`/`super`/`Self` path segments.

---

## Re-review result

**Verdict: request changes.** The updated head is `248cf2592c672f06661aeafa58854d9291544a04`. It addresses the two previous findings, adds substantial positive and negative test coverage, and all four GitHub Actions workflows now pass. One path-validation hole remains.

## Finding

### **BLOCKING — restricted path keywords are validated only in the first segment**

**Location:** `zerocopy/zerocopy-derive/src/util.rs`, immediately after `lit.parse_with(Path::parse_mod_style)`.

The updated code checks only:

```rust
let first = path_lit.segments[0].ident.to_string();
```

It then validates whether that first segment is `crate`, `self`, `super`, or `Self`. However, Syn’s `Path::parse_mod_style` loop accepts all four tokens at **every** segment position. Thus strings such as these successfully pass the new parser and validation:

```rust
"foo::crate::bar"
"foo::super::bar"
"foo::Self::bar"
"crate::foo::crate"
```

For example, `"foo::crate::bar"` is turned into:

```rust
::foo::crate::bar
```

and later used as a prefix for generated paths such as:

```rust
::foo::crate::bar::KnownLayout
```

Rust rejects this because `crate` may appear only as the first segment without a preceding `::`. Likewise, `Self` may appear only first, while `super` may appear only in the leading qualifier sequence.   ([[Rust Documentation](https://doc.rust-lang.org/reference/paths.html?utm_source=chatgpt.com)][1])

There is a related valid-input case:

```rust
"foo::self"
```

A trailing `self` can validly denote the preceding module. But because the derive treats the supplied path as a prefix and appends more segments, it produces something equivalent to:

```rust
::foo::self::KnownLayout
```

where `self` is no longer trailing and is therefore invalid. The implementation should either normalize away a valid trailing `self` or explicitly reject it as unsupported before storing the path.  ([[Rust Documentation](https://doc.rust-lang.org/reference/paths.html?utm_source=chatgpt.com)][1])

A complete validation should enforce roughly this grammar:

* `crate` may occur only as the first segment, without a leading `::`.
* `Self` should not occur anywhere in a crate/module prefix.
* `super` may occur only in an initial run of qualifiers, optionally following an initial `self`.
* `self` may be the first segment; a later trailing `self` must either be removed or rejected because generated code will append further segments.
* After the first ordinary identifier is encountered, every remaining segment must also be an ordinary identifier.

At minimum, add UI tests for:

```rust
#[zerocopy(crate = "foo::crate::bar")]
#[zerocopy(crate = "foo::super::bar")]
#[zerocopy(crate = "foo::Self::bar")]
#[zerocopy(crate = "foo::self")]
```

If `self::super::...` and repeated `super::super::...` are intentionally supported, positive tests for those forms would prevent an over-restrictive fix.

**Ready-to-post comment:**

> `Path::parse_mod_style` accepts `crate`, `self`, `super`, and `Self` at every segment position, but this validation examines only `segments[0]`. Consequently values such as `"foo::crate::bar"` and `"foo::super::bar"` pass here and are later emitted as invalid Rust paths. `"foo::self"` is also problematic: it can be a valid standalone module path, but appending `::KnownLayout` or `::util` makes `self` an invalid middle segment. Please validate the entire segment sequence—while retaining valid leading forms such as `super::super::...` and `self::super::...`—and add corresponding UI tests.

## Previously reported issues

Both prior issues are otherwise resolved:

* Plain external paths are absolutized, while leading `crate`, `self`, and `super` paths remain relative.
* Parsing now uses `LitStr::parse_with(Path::parse_mod_style)`, preserving literal suffix validation and source spans.
* Positive tests cover nested external, `crate::`, and `super::` paths.
* UI tests cover suffixed string literals, leading `Self`, and `::crate`.
* Formatting, MSRV, stable, nightly, UI, coverage, and the other CI workflows are passing.

The change does not touch unsafe code or alter zerocopy’s soundness invariants. I could not independently execute the Rust test suite in the local tool environment because no Rust toolchain is installed; the verification above uses the current source, vendored Syn parser, expected-output tests, and passing GitHub Actions results.

[1]: https://doc.rust-lang.org/reference/paths.html?utm_source=chatgpt.com "Paths - The Rust Reference"

joshuaseaton:
Done! PTAL

joshlf:
Pretty sure the test failures are due to string interpolation, which isn't supported on our MSRV (1.56):

```rust
assert!(
    super::validate_crate_path(&path).is_ok(),
    "expected valid path for `{path_str}`"
);
```

and

```rust
assert!(
    super::validate_crate_path(&path).is_err(),
    "expected invalid path for `{path_str}`"
);
```

You'll need to use the older `assert!(..., "expected valid path for {}", path_str)` syntax.

joshlf (approved):

joshuaseaton:
Done. Thanks!

joshlf:
Great, thanks for this! I'm going to let @jswrenn take a look before merging, but this LGTM!

joshlf:
Released in [0.8.56](https://crates.io/crates/zerocopy/0.8.56).

PR #3548, opened 2026-08-10
exocrate::Config::rel_dir_path: Make non-pub and validate at construc…

### LINKED ISSUE #3408: `exocrate::Config::rel_dir_path`: Make non-`pub` and validate at construction that each item is valid for the OS

Sagartaunk:
…tion that each item is valid

Make `Config`'s fields private and introduce `Config::new`, which validate both `version_slug` and `rel_dir_path` via `validate_path` before constructing.

Also, update `config` macro to make use of this new call.

Fixes: #3408

### DIFF
--- a/exocrate/src/lib.rs
+++ b/exocrate/src/lib.rs
@@ -41,15 +41,15 @@
 //! values directly, or derive them from the macros exported by `exocrate`:
 //!
 //! ```rust,no_run
-//! const CONFIG: exocrate::Config = exocrate::Config {
+//! const CONFIG: exocrate::Config = exocrate::Config::new (
 //!     // Path components that are joined to establish the base directory for all exocrate
 //!     // installations.
 //!     //
 //!     // For `my-tool` developers: `<my-tool-root>/target/.my-tool/exocrate/<version>`
 //!     // For `my-tool` users: `~/.my-tool/exocrate/<version>`
-//!     rel_dir_path: &[".my-tool", "exocrate"],
-//!     version_slug: "example-version",
-//! };
+//!     &[".my-tool", "exocrate"],
+//!     "example-version",
+//! );
 //!
 //! const REMOTE: exocrate::RemoteArchive = exocrate::RemoteArchive {
 //!     sha256: [0xaa; 32],
@@ -152,19 +152,8 @@ pub struct RemoteArchive {
 }
 
 pub struct Config {
-    /// The relative path of the directory containing dependencies, stored as a
-    /// sequence of path components for cross-platform compatibility.
-    ///
-    /// Dependencies are stored in `<rel_dir_path>`. In production use, this is
-    /// relative to the user cache directory. In development, this is relative
-    /// to Cargo's `target` directory if it can be resolved, and otherwise relative
-    /// relative to the current working directory.
-    //
-    // FIXME(#3408): Make this non-`pub` and validate at construction that each
-    // item is valid for the OS.
-    pub rel_dir_path: &'static [&'static str],
-    /// A unique identifier for this version of the dependencies.
-    pub version_slug: &'static str,
+    rel_dir_path: &'static [&'static str],
+    version_slug: &'static str,
 }
 
 /// The location to install dependencies.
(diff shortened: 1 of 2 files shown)


exocrate/src/lib.rs
```diff
@@ -159,12 +159,9 @@ pub struct Config {
     /// relative to the user cache directory. In development, this is relative
     /// to Cargo's `target` directory if it can be resolved, and otherwise relative
     /// relative to the current working directory.
-    //
-    // FIXME(#3408): Make this non-`pub` and validate at construction that each
-    // item is valid for the OS.
-    pub rel_dir_path: &'static [&'static str],
+    rel_dir_path: &'static [&'static str],
     /// A unique identifier for this version of the dependencies.
```
joshlf: These comments are meant to be public documentation. They aren't anymore now that these fields aren't `pub`. Can you move these comments to the new `Config::new` constructor?

joshlf (changes requested):

Sagartaunk:
@joshlf I have made the requested changes.

exocrate/src/lib.rs
```diff
@@ -276,6 +265,77 @@ impl Config {
...
+        // `rel_dir_path` component, so it needs the same validation.
+        validate_path(version_slug);
+
+        Self { rel_dir_path, version_slug }
+    }
+}
+
+/// Validate a single path component
+///
+/// This fn intentionally accepts only a small cross-platform subset
```
joshlf:
```suggestion
/// This function intentionally accepts only a small cross-platform subset
```

exocrate/src/lib.rs
```diff
@@ -276,6 +265,77 @@ impl Config {
...
+        }
+        // `version_slug` is joined onto the path in `dir_path` exactly like a
+        // `rel_dir_path` component, so it needs the same validation.
+        validate_path(version_slug);
+
+        Self { rel_dir_path, version_slug }
+    }
+}
+
+/// Validate a single path component
```
joshlf:
```suggestion
/// Validates a single path component.
```

exocrate/src/lib.rs
```diff
@@ -276,6 +265,77 @@ impl Config {
...
+///
+/// This fn intentionally accepts only a small cross-platform subset
+/// of path component names. Thus, ignoring platform-specific behaviour
+/// and ensuring that the same logical path is valid on Windows, macOS
+/// and Unix-like systems.
```
joshlf:
```suggestion
/// of path component names. Thus it ignores platform-specific behavior
/// and ensures that the same logical path is valid on Windows, macOS
/// and Unix-like systems.
```

Sorry to correct the spelling – since we've already standardized on American spellings, let's stick to that.

exocrate/src/lib.rs
```diff
@@ -276,6 +265,77 @@ impl Config {
...
+///
+/// Panics if the component is empty, is `.` or `..`, contains a path
+/// seperator, or a Windows-reserved character.
```
joshlf:
```suggestion
/// separator, or contains a Windows-reserved character.
```

joshlf (changes requested):
Can you also reflow all comments to 80 columns wide?

joshlf:
@codex review

Sagartaunk:
I also added 3 more test cases .

Also, sorry for not checking the spellings beforehand in the last commit.

joshlf:
@codex review

exocrate/src/lib.rs
```diff
@@ -276,6 +265,126 @@ impl Config {
...
+/// - is empty
+/// - is `.` or `..`
+/// - contains a path separator
+/// - contains a character reserved on Windows
+/// - is a Windows-reserved device name (matched by its stem)
+/// - contains a null byte
+const fn validate_path(part: &str) {
+    let bytes = part.as_bytes();
+
+    assert!(!bytes.is_empty(), "rel_dir_path component must not be empty");
```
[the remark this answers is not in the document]
Sagartaunk: @joshlf should I put a limit of 255 bytes on each path component for this ? 
Or is it fine without it?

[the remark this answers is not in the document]
joshlf: Yeah let's go ahead and add that limit. Seems a bit silly, but we're already trying to be anal here, so we might as well. Better to err on the side of caution and then consider lifting the restriction if this causes issues for anyone in practice.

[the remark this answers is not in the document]
Sagartaunk: > Yeah let's go ahead and add that limit. Seems a bit silly, but we're already trying to be anal here, so we might as well. Better to err on the side of caution and then consider lifting the restriction if this causes issues for anyone in practice.

Okay, made the edit.

joshlf (approved):
Awesome, thank you for this!

PR #3565, opened 2026-08-24
[derive] Stop emitting non-ASCII identifiers from TryFromBytes derive

### LINKED ISSUE #2880: Zerocopy 0.8.32 breaks combined with `deny(non_ascii_idents)`
For security reasons we cannot allow `non-ascii-identifiers` in our codebase, and as such we apply `#![deny(non_ascii_idents)]`. 

However, since upgrading to Zerocopy 0.8.32 (0.8.33 has same behavior) we are running into violations of the `non_ascii_idents` lint. 

`src/main.rs`:

```rust
#![deny(non_ascii_idents)]

use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};

#[derive(KnownLayout, FromBytes, IntoBytes, Immutable)]
#[repr(C)]
pub struct ifaddrmsg {
    pub ifa_family: u8,    /* Address type */
    pub ifa_prefixlen: u8, /* Prefixlength of address */
    pub ifa_flags: u8,     /* Address flags */
    pub ifa_scope: u8,     /* Address scope */
    pub ifa_index: u32,    /* Interface index */
}

fn main() {}
```

Cargo.toml:
```
[package]
name = "zerocopy-non-ascii-error"
version = "0.1.0"
edition = "2024"

[dependencies]
zerocopy = { version = "0.8.33", features = ["derive"] }
```

`bash`:

```bash
❯ cargo run
   Compiling zerocopy-non-ascii-error v0.1.0 (/home/kristof/projects/zerocopy-non-ascii-error)
error: identifier contains non-ASCII characters
 --> src/main.rs:8:9
  |
8 |     pub ifa_family: u8,    /* Address type */
  |         ^^^^^^^^^^
  |
note: the lint level is defined here
 --> src/main.rs:1:9
  |
1 | #![deny(non_ascii_idents)]
  |         ^^^^^^^^^^^^^^^^

error: identifier contains non-ASCII characters
 --> src/main.rs:9:9
  |
9 |     pub ifa_prefixlen: u8, /* Prefixlength of address */
  |         ^^^^^^^^^^^^^

error: identifier contains non-ASCII characters
  --> src/main.rs:10:9
   |
10 |     pub ifa_flags: u8,     /* Address flags */
   |         ^^^^^^^^^

error: identifier contains non-ASCII characters
  --> src/main.rs:11:9
   |
11 |     pub ifa_scope: u8,     /* Address scope */
   |         ^^^^^^^^^

error: identifier contains non-ASCII characters
  --> src/main.rs:12:9
   |
12 |     pub ifa_index: u32,    /* Interface index */
   |         ^^^^^^^^^

error: could not compile `zerocopy-non-ascii-error` (bin "zerocopy-non-ascii-error") due to 5 previous errors
```

Traced back to this PR: https://github.com/google/zerocopy/commit/002b97de9b73c1d012ae8241780931e717965219#diff-977c29e6cf079bd7307f811df24f99633df603962785c583145eb6250b0e12daR755

There was a comment about it: https://github.com/google/zerocopy/pull/2842#discussion_r2590153675 but no mention of `non_ascii_idents`.

Could an allow be applied to those lines, just like here: https://github.com/google/zerocopy/commit/002b97de9b73c1d012ae8241780931e717965219#diff-977c29e6cf079bd7307f811df24f99633df603962785c583145eb6250b0e12daR782 ?

rootkiller6788:
Fixes #2880.

## Motivation

When a user applies `#![deny(non_ascii_idents)]` at their crate root and derives `FromBytes`/`TryFromBytes`, compilation fails:

```
error: identifier contains non-ASCII characters
 --> src/main.rs:8:9
  |
8 |     pub ifa_family: u8,
  |         ^^^^^^^^^^
```

The cause is that the `TryFromBytes` derive generates per-field marker enum types whose names are prefixed with `ẕ` (U+1E95) - a non-ASCII character chosen to avoid colliding with user identifiers. These markers are emitted into the user's crate, and since their span points back at the user's field, the `non_ascii_idents` lint fires at the field even though the offending identifier lives in generated code.

Emitting `#[allow(non_ascii_idents)]` around the generated items doesn't work: rustc currently ignores `allow(non_ascii_idents)` at non-crate levels (see rust-lang/rust#151025), so the lint still fires.

## Change

Rename the generated marker types from `ẕ<field>` to `___ZerocopyHasField_<field>`, following the existing `___ZerocopyRawEnum` / `___ZcAlignment` naming convention used for other derive-generated items. All generated identifiers are now ASCII, so the lint can no longer fire regardless of lint configuration.

Additionally:
- Enable `FromBytes` derivation in the `issue_2880.rs` regression test (removing the FIXME) so the fix is covered by CI.
- Regenerate the `output_tests` expected files.
- Update the `privacy` UI-test annotations and `.stderr` files for the renamed marker types.

## Testing

- `cargo test -p zerocopy-derive` (output tests + derive integration tests) passes.
- The `issue_2880` test compiles under `#![deny(non_ascii_idents)]` with `FromBytes` derived on structs, unions, and enums.
- `cargo test -p zerocopy --lib` passes.


jswrenn:
See https://github.com/google/zerocopy/issues/2880#issuecomment-3740407237

rootkiller6788:
Thanks for the pointer to the discussion on #2880 - I'd read the thread and understood the preference to wait on rust-lang/rust#151025 before committing to the rename approach.

More than eight months later that issue (and the rustc-side fix in rust-lang/rust#157497) is still open, so users who apply `#![deny(non_ascii_idents)]` still can't derive these traits at all. This PR is the fallback that comment already anticipated - stopping the derive from emitting non-ASCII identifiers internally - and it follows the existing naming convention for derive-generated items (`___ZerocopyRawEnum`, `___ZcAlignment`), so I'd like to make the case for landing it rather than continuing to wait.

I've also added a follow-up commit on this branch in response to the review: the predictable ASCII marker names can now be shadowed by a type parameter, which would let code outside the module name the marker as a type argument and project a private field through `HasField`. The derive now detects a type parameter whose name matches a generated marker and emits a clear compile error instead of generating an impl that silently refers to the wrong item.

Could you please take another look? If you'd still prefer to hold off pending the rustc fix, I'm happy to leave this open until that direction changes.

PR #3568, opened 2026-08-24
[derive] Allow `missing_debug_implementations` on `KnownLayout` helpers

### LINKED ISSUE #3567: KnownLayout derive breaks with `deny(missing_debug_implementations)`
With `zerocopy 0.8.56`, this example code fails to compile:

```rust
#![deny(missing_debug_implementations)]

macro_rules! foo {
    ($ref_name:ident) => {
        #[derive(zerocopy::KnownLayout)]
        #[repr(C)]
        pub struct $ref_name([u8]);

        impl std::fmt::Debug for $ref_name {
            fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
                f.write_str("hello")
            }
        }
    };
}

foo! { Abc }
```
Output:
```
error: type does not implement `Debug`; consider adding `#[derive(Debug)]` or a manual implementation
  --> src/lib.rs:7:9
   |
 7 |         pub struct $ref_name([u8]);
   |         ^^^
...
17 | foo! { Abc }
   | ------------ in this macro invocation
```
IIUC, this is because the `KnownLayout` derive produces a public (albeit doc-hidden) type `__ZerocopyKnownLayoutMaybeUninit` that doesn't implement `Debug`. The same goes for the `__Zerocopy_Field_X` structs.

We should probably explicitly exempt those types using `#[allow(missing_debug_implementations)]`, since they are already doc-hidden.

apasel422:
Add `#[allow(missing_debug_implementations)]` to `__ZerocopyKnownLayoutMaybeUninit` and the generated field handle structs to prevent compile errors when the derive target is inside a macro and the crate has `#[deny(missing_debug_implementations)]`.

Fixes #3567

### DIFF
--- a/zerocopy/zerocopy-derive/tests/ui/issue_known_layout_missing_debug.rs
+++ b/zerocopy/zerocopy-derive/tests/ui/issue_known_layout_missing_debug.rs
@@ -1,0 +1,28 @@
+// Copyright 2026 The Fuchsia Authors
+//
+// Licensed under a BSD-style license <LICENSE-BSD>, Apache License, Version 2.0
+// <LICENSE-APACHE or https://www.apache.org/licenses/LICENSE-2.0>, or the MIT
+// license <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your option.
+// This file may not be copied, modified, or distributed except according to
+// those terms.
+
+//@[nightly,stable] check-pass
+
+#![no_implicit_prelude]
+#![deny(missing_debug_implementations)]
+#![cfg_attr(msrv, deny(private_in_public))]
+
+macro_rules! foo {
+    () => {
+        #[derive(::core::fmt::Debug, ::zerocopy_renamed::KnownLayout)]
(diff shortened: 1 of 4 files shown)


apasel422:
It looks like there's a separate issue blocking this from building, related to field visibility vs struct visibility.

zerocopy/zerocopy-derive/tests/ui/issue_known_layout_missing_debug.rs
jswrenn, line 1: This PR is fine, this test just needs to be reformulated as a UI test that produces no errors on nightly and stable but errors on MSRV. The visibility issue you're trying to fix in #3569 is a known bug in the type privacy checker on our MSRV toolchain.

apasel422: Done. Thanks for the guidance.

jswrenn (approved):

PR #3588, opened 2026-08-25
[ci] Require execution parity before planning

joshlf:
The typed command model helps only if every caller checks it before a
plan can escape. Run its full legacy-parity proof while constructing
CiInputs, after policy, inventory, workflow, and baselines are trusted.

This makes validation all-or-nothing. New planners and local commands
cannot omit the execution audit or consume a model which has drifted
from independently captured behavior.

Tests: cargo test -p zc --all-targets --offline

*Authored by an agent, posting via joshlf's account*


joshlf:
*Authored by an agent, posting via joshlf's account*

GHerrit's partial metadata sync left this PR targeting a later stack branch, so GitHub closed it with an empty diff and would not permit reopening it. The same managed head commit now continues as #3598, targeting its correct immediate parent.

PR #3596, opened 2026-08-25
[ci] Retire the generated-matrix toolchain check

joshlf:
The old shell check compared package.metadata.build-rs keys with the
handwritten workflow matrix. Typed planning now passes that matrix
through fromJSON, so yq sees an expression instead of the expected list.

Make checked inventory and policy the sole owner of this invariant.
Inventory requires manifest metadata keys and policy descriptors to be
exactly equal. Policy requires each descriptor to have nonempty scopes,
and ci audit plans every configured event. Both direct CI and pre-push
run that audit through check_tools.sh.

Remove the obsolete hosted job, yq installation, script, hook child,
workflow registry row, and aggregate dependency. Keep the frozen
execution specs unchanged because they describe their immutable source
commit, not the live workflow.

This removes one runner startup and tool installation from CI and one
duplicate local process from pre-push. It should reduce resource use and
a small amount of aggregate latency, although it is not on the matrix
critical path.

Tests: ci/check_tools.sh
Tests: ci/check_actions.sh
Tests: ci/check_job_dependencies.sh
Tests: githooks/test_pre_push.py
Tests: ci/check_fmt.sh
Tests: git diff --check

*Authored by an agent, posting via joshlf's account*

### DIFF
--- a/ci/check_tools.sh
+++ b/ci/check_tools.sh
@@ -68,5 +68,14 @@ env -u RUSTUP_TOOLCHAIN \
 # command must pin its own Cargo metadata invocation instead of inheriting this
 # deliberately invalid ambient override. Keep this command coordinated with
 # `tools/cargo-zerocopy/src/main.rs` and `tools/zc/src/cli.rs`.
+#
+# This audit establishes typed ownership of the build.rs toolchain invariant.
+# Inventory requires the manifest metadata keys and policy descriptors to be
+# exactly equal, policy requires every descriptor to have nonempty scopes, and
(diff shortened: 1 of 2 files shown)


joshlf:
*Authored by an agent, posting via joshlf's account*

GHerrit's partial metadata sync left this PR targeting a later stack branch, so GitHub marked it merged and deleted its managed head branch. I restored that branch at the unchanged commit; the change now continues as #3599, targeting its correct immediate parent.

PR #3625, opened 2026-09-04
Qualify KnownLayout metadata extraction

### LINKED ISSUE #3621: Security Advisory: Inherent method capture permits out-of-bounds splits of derived types

- **Vulnerability Type:** Unsafe derive expansion / incorrect pointer metadata / out-of-bounds read
- **Affected Software:** zerocopy-derive
- **Date:** September 2026

## Summary

`#[derive(KnownLayout)]` can call an application's safe inherent `pointer_to_metadata` method instead of the `KnownLayout` trait method when deriving an implementation for a structure with a concrete trailing field. The generated call does not explicitly select the trait. If the inherent method returns a different length, the enclosing structure's derived implementation reports incorrect pointer metadata to code that trusts the unsafe trait's guarantees.

The reproduction below contains nested structures ending in a byte slice. An inherent method on the inner structure deliberately returns eight instead of the actual one-byte length. A safe `split_at(8)` call on the outer structure succeeds, although that split is outside the object. Miri reports pointer arithmetic beyond a one-byte allocation. A native demonstration using a reused receive buffer returns seven synthetic secret bytes beyond the one-byte input passed to the parser.

**Threat model and qualification:** This is a library API soundness defect. Its security consequence requires a deployed application to already contain the colliding inherent method with incompatible semantics, use that concrete type as a trailing field in a derived structure, and call the affected splitting operation. The demonstration deliberately supplies an incorrect safe method to expose the derive's misplaced trust; we have not established that production applications contain one. An attacker controlling only packet bytes cannot add this method to an otherwise unaffected application. If the prerequisites are present, a short input can cause the application to expose bytes outside the received region. No remote code execution was demonstrated.

**Affected version and revision:** The original audit examined `zerocopy 0.8.56` and `zerocopy-derive 0.8.56` at public repository revision `2dad389b030e9268d6645ac0bf0626b867e96068`. The standalone reproduction below uses published version `0.8.56` of both crates, whose package metadata identifies revision `6dc429c451bdf1d7202ec1ec2cf426514e00d8eb`. The affected code is unchanged between these sources. We have not exhaustively tested other versions.

**Environment:** Native reproduction on macOS arm64 (`aarch64-apple-darwin`) with Rust 1.98.1 (`48a229cea`, LLVM 22.1.8). Miri reproduction with `nightly-2026-09-04`, Rust 1.100.0-nightly (`a69a63265`, 2026-09-03), and Miri 0.1.0 (`a69a63265c`).

## Sketch of the attack

The example models a service passing only the current request's received prefix from a reusable buffer to the parser:

```rust
let receive_buffer = *b"*SECRET!";
let wire = &receive_buffer[..1];
let packet = Outer::ref_from_bytes(wire).unwrap();

// The real trailing slice has one byte. This should return None.
let (expanded, _) = packet.split_at(8).unwrap().via_immutable();

// A response containing this slice also contains the unreceived suffix.
println!("{:?}", &expanded.inner.bytes);
```

Under the hood:

- `Outer` contains a concrete trailing `Inner` field. Both types derive `KnownLayout` normally.
- The implementation generated for `Outer` calls `Inner::pointer_to_metadata`. Rust selects `Inner`'s inherent method, which returns eight, instead of its derived trait implementation, which would report one.
- The safe split's bounds check trusts the outer type's metadata. It therefore accepts the requested length eight and subsequently performs unchecked pointer operations using that false bound.
- For a one-byte allocation, these operations are out of bounds. In the reusable-buffer demonstration, they also let the returned slice expose bytes excluded from the input slice.

## Full repro

```toml
[package]
name = "zerocopy-metadata-capture-repro"
version = "0.1.0"
edition = "2024"

[workspace]

[dependencies]
```

Save this complete program as `src/bin/metadata_override.rs`:

```rust
#![forbid(unsafe_code)]

use zerocopy::{FromBytes, Immutable, KnownLayout, SplitAt};

#[derive(FromBytes, Immutable, KnownLayout, SplitAt)]
#[repr(C)]
struct Inner {
    bytes: [u8],
}

impl Inner {
    fn pointer_to_metadata(_: *mut Self) -> usize {
        8
    }
}

#[derive(FromBytes, Immutable, KnownLayout, SplitAt)]
#[repr(C)]
struct Outer {
    inner: Inner,
}

fn respond(wire: &[u8]) {
    let packet = Outer::ref_from_bytes(wire).unwrap();
    println!("original length={}", packet.inner.bytes.len());
    let (packet, _) = packet.split_at(8).unwrap().via_immutable();
    println!("expanded bytes={:?}", &packet.inner.bytes);
}

fn main() {
    if std::env::args().nth(1).as_deref() == Some("leak") {
        let receive_buffer = std::hint::black_box(*b"*SECRET!");
        respond(&receive_buffer[..1]);
    } else {
        let wire = Box::new([42u8]);
        respond(wire.as_slice());
    }
}
```

After saving the program, generate the lockfile and run the one-byte-allocation case with Miri:

```sh
MIRIFLAGS='-Zmiri-strict-provenance' \
    cargo +nightly-2026-09-04 miri run --locked --bin metadata_override
```

Miri first prints the actual one-byte length and then reports an invalid eight-byte pointer offset. In the audited run, the diagnostic was:

```text
original length=1
error: Undefined Behavior: in-bounds pointer arithmetic failed: attempting to offset pointer by 8 bytes, but got alloc758 which is only 1 byte from the end of the allocation
```

The allocation identifier may vary between runs. The error occurs at `base.add(range.start)` in `zerocopy/src/pointer/inner.rs`, reached through the safe split. Miri's nonzero exit is the expected reproduction result.

Run the reusable-buffer demonstration in an optimized native build:

```sh
cargo +stable run --locked --release --bin metadata_override -- leak
```

Observed output:

```text
original length=1
expanded bytes=[42, 83, 69, 67, 82, 69, 84, 33]
```

Only the first byte, `42`, was supplied as the current input. The other seven bytes spell `SECRET!`. The fixture deliberately places this synthetic value in the buffer suffix; the result demonstrates exposure outside the received region, not that arbitrary production buffers always contain secrets.

## Root cause

The links below refer to the original audited revision `2dad389b030e9268d6645ac0bf0626b867e96068`. The same affected code is present in published version `0.8.56`.

1. The [`KnownLayout` derive](https://github.com/google/zerocopy/blob/2dad389b030e9268d6645ac0bf0626b867e96068/zerocopy/zerocopy-derive/src/derive/known_layout.rs#L71-L85) explicitly qualifies `raw_from_ptr_len` through `KnownLayout`, but emits `<TrailingType>::pointer_to_metadata(...)` without trait qualification. For a concrete trailing type, an inherent method takes precedence.
2. The selected safe inherent method has not promised to satisfy the unsafe trait's [`pointer_to_metadata` contract](https://github.com/google/zerocopy/blob/2dad389b030e9268d6645ac0bf0626b867e96068/zerocopy/src/lib.rs#L805-L812), which requires the actual metadata stored in the pointer. Nevertheless, the generated outer implementation reports its return value as trusted metadata.
3. [`PtrInner::meta`](https://github.com/google/zerocopy/blob/2dad389b030e9268d6645ac0bf0626b867e96068/zerocopy/src/pointer/inner.rs#L222-L229) obtains this metadata through `KnownLayout`. The safe [`SplitAt::split_at`](https://github.com/google/zerocopy/blob/2dad389b030e9268d6645ac0bf0626b867e96068/zerocopy/src/split_at.rs#L159-L169) relies on the metadata bounds check before invoking the unchecked split machinery.
4. The split uses the false length in [`split_at_unchecked`](https://github.com/google/zerocopy/blob/2dad389b030e9268d6645ac0bf0626b867e96068/zerocopy/src/pointer/inner.rs#L298-L314), eventually reaching [`base.add(range.start)`](https://github.com/google/zerocopy/blob/2dad389b030e9268d6645ac0bf0626b867e96068/zerocopy/src/pointer/inner.rs#L396-L406) outside the original allocation. This violates the standard library's [in-bounds pointer-arithmetic requirement](https://doc.rust-lang.org/std/primitive.pointer.html#method.add).

The generated call should explicitly select `<TrailingType as KnownLayout>::pointer_to_metadata(...)`, as the neighboring `raw_from_ptr_len` call already does.

## Disclaimer

joshlf:
Generated KnownLayout implementations called pointer_to_metadata through a
concrete trailing type. A same-named inherent method could therefore capture
the call and supply metadata that violates the unsafe trait's layout contract.

Emit a fully trait-qualified call for both the derived type and its generated
MaybeUninit form. Extend the output hygiene check to reject unqualified
associated-item syntax, and add compile-time and semantic regressions for an
inherent-method collision.

Closes #3621

---

- 　  #3629
- 👉 #3625

**Latest Update:** v2 — [Compare vs v1](/google/zerocopy/compare/gherrit/Gkpenpjjuex55evmm52puxsa34xtjgc7t/v1..gherrit/Gkpenpjjuex55evmm52puxsa34xtjgc7t/v2)

<details>

|Version| v1 |Base|
|:---|:---|:---|
|v2|[vs v1](/google/zerocopy/compare/gherrit/Gkpenpjjuex55evmm52puxsa34xtjgc7t/v1..gherrit/Gkpenpjjuex55evmm52puxsa34xtjgc7t/v2)|[vs Base](/google/zerocopy/compare/main..gherrit/Gkpenpjjuex55evmm52puxsa34xtjgc7t/v2)|
|v1||[vs Base](/google/zerocopy/compare/main..gherrit/Gkpenpjjuex55evmm52puxsa34xtjgc7t/v1)|

</details>
<details>

######

**Branch**
```bash
git fetch origin refs/heads/Gkpenpjjuex55evmm52puxsa34xtjgc7t && git checkout -b pr-Gkpenpjjuex55evmm52puxsa34xtjgc7t FETCH_HEAD
```

**Checkout**
```bash
git fetch origin refs/heads/Gkpenpjjuex55evmm52puxsa34xtjgc7t && git checkout FETCH_HEAD
```

**Cherry Pick**
```bash
git fetch origin refs/heads/Gkpenpjjuex55evmm52puxsa34xtjgc7t && git cherry-pick FETCH_HEAD
```

**Pull**
```bash
git pull origin refs/heads/Gkpenpjjuex55evmm52puxsa34xtjgc7t
```

</details>

### DIFF
--- a/zerocopy/zerocopy-derive/src/derive/known_layout.rs
+++ b/zerocopy/zerocopy-derive/src/derive/known_layout.rs
@@ -81,7 +81,7 @@ fn derive_known_layout_for_repr_c_struct<'a>(
 
             #[inline(always)]
             fn pointer_to_metadata(ptr: *mut Self) -> <Self as #zerocopy_crate::KnownLayout>::PointerMetadata {
-                <#trailing_field_ty>::pointer_to_metadata(ptr as *mut _)
+                <#trailing_field_ty as #zerocopy_crate::KnownLayout>::pointer_to_metadata(ptr as *mut _)
             }
         }
     };
--- a/zerocopy/zerocopy-derive/src/output_tests/expected/known_layout_repr_c_struct.expected.rs
+++ b/zerocopy/zerocopy-derive/src/output_tests/expected/known_layout_repr_c_struct.expected.rs
@@ -45,7 +45,7 @@ const _: () = {
         fn pointer_to_metadata(
             ptr: *mut Self,
         ) -> <Self as ::zerocopy::KnownLayout>::PointerMetadata {
-            <U>::pointer_to_metadata(ptr as *mut _)
+            <U as ::zerocopy::KnownLayout>::pointer_to_metadata(ptr as *mut _)
         }
     }
     #[allow(missing_debug_implementations)]
@@ -130,7 +130,7 @@ const _: () = {
                 U,
             > as ::zerocopy::util::macro_util::Field<
                 __Zerocopy_Field_1,
-            >>::Type as ::zerocopy::KnownLayout>::MaybeUninit>::pointer_to_metadata(
+            >>::Type as ::zerocopy::KnownLayout>::MaybeUninit as ::zerocopy::KnownLayout>::pointer_to_metadata(
                 ptr as *mut _,
             )
         }
--- a/zerocopy/zerocopy-derive/src/output_tests/mod.rs
+++ b/zerocopy/zerocopy-derive/src/output_tests/mod.rs
@@ -58,10 +58,11 @@ macro_rules! test {
             let ts: proc_macro2::TokenStream = quote::quote!( $($i)* );
             let ast = syn::parse2::<syn::DeriveInput>(ts).unwrap();
             let ctx = crate::Ctx::try_from_derive_input(ast).unwrap();
-            let res = $name(&ctx, crate::util::Trait::$name);
+            let res = $name(&ctx, crate::util::Trait::$name).into_ts();
+            crate::util::testutil::check_hygiene(res.clone());
             let expected_toks = quote::quote!( $($o)* );
             let expected = pretty_print(expected_toks);
-            let actual = pretty_print(res.into_ts().into());
+            let actual = pretty_print(res);
             assert_eq_or_diff(&expected, &actual);
         }
     };
@@ -75,8 +76,9 @@ macro_rules! test {
             let ts: proc_macro2::TokenStream = quote::quote!( $($i)* );
             let ast = syn::parse2::<syn::DeriveInput>(ts).unwrap();
             let ctx = crate::Ctx::try_from_derive_input(ast).unwrap();
-            let res = $name(&ctx, crate::util::Trait::$name);
-            let actual = pretty_print(res.into_ts().into());
+            let res = $name(&ctx, crate::util::Trait::$name).into_ts();
+            crate::util::testutil::check_hygiene(res.clone());
+            let actual = pretty_print(res);
 
             if std::env::var("ZEROCOPY_BLESS").is_ok() {
                 let path = Path::new(env!("CARGO_MANIFEST_DIR"))
--- a/zerocopy/zerocopy-derive/src/util.rs
+++ b/zerocopy/zerocopy-derive/src/util.rs
@@ -890,17 +890,39 @@ pub(crate) mod testutil {
     pub(crate) fn check_hygiene(ts: TokenStream) {
         struct AmbiguousItemVisitor;
 
+        fn panic_ambiguous_path(path: &impl quote::ToTokens) -> ! {
+            panic!(
+                "Found ambiguous path `{}` in generated output. \
+                 All associated item access must be fully qualified (e.g., `<Self as Trait>::Item`) \
+                 to prevent hygiene issues.",
+                quote::quote!(#path)
+            );
+        }
+
         impl<'ast> Visit<'ast> for AmbiguousItemVisitor {
+            fn visit_expr_path(&mut self, i: &'ast syn::ExprPath) {
+                if let Some(qself) = &i.qself {
+                    if qself.position == 0 {
+                        panic_ambiguous_path(i);
+                    }
+                }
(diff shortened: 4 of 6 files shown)


joshlf:
@codex review

jswrenn (approved):

PR #3629, opened 2026-09-05
Extract slice metadata without references

### LINKED ISSUE #3615: Soundness Advisory: Extracting metadata from a null raw slice creates an invalid reference

- **Vulnerability Type:** Unsound safe API / invalid reference construction
- **Affected Software:** zerocopy
- **Date:** September 2026

## Summary

The safe `KnownLayout::pointer_to_metadata` implementation for slices can cause undefined behavior when passed a null raw slice pointer. Its implementation creates a Rust reference to a slice of zero-sized elements before reading the slice length. Rust references must be non-null even when the slice is empty and no element bytes are accessed.

The demonstrated threat model is a Rust application directly calling this public, `#[doc(hidden)]` helper with a raw slice pointer whose data address is null. Constructing that raw pointer is legal in safe Rust, and the proof of concept contains no `unsafe` code. The method's safe signature does not require callers to establish a non-null pointer.

This is a confirmed library soundness bug. We have not demonstrated that an attacker controlling only packet contents can reach this condition through zerocopy's ordinary byte-slice parsing APIs, nor demonstrated information disclosure, memory corruption, or remote code execution from this particular issue. Application-specific reachability would be needed to establish a remotely exploitable security vulnerability.

**Affected version:** Reproduced using the published `zerocopy 0.8.56` crate. Its [release source](https://github.com/google/zerocopy/tree/6dc429c451bdf1d7202ec1ec2cf426514e00d8eb) contains the same affected code as audited revision `2dad389b030e9268d6645ac0bf0626b867e96068`, which the source references below use. Earlier releases have not been exhaustively tested.

**Environment:** The proof of concept was checked with Rust 1.98.1 and validated with default Miri from `nightly-2026-09-04` on macOS arm64 (`aarch64-apple-darwin`). No Miri checks were disabled.

## Sketch of the attack

The example exercises the unsafe boundary directly through safe Rust:

```rust
let raw = std::ptr::slice_from_raw_parts_mut(std::ptr::null_mut::<u8>(), 0);
let length = <[u8] as zerocopy::KnownLayout>::pointer_to_metadata(raw);
```

Under the hood:

- `slice_from_raw_parts_mut` constructs a raw slice pointer with a null data address and length zero. It does not create a reference or dereference the pointer.
- `pointer_to_metadata` casts that pointer to `*const [()]`, preserving the null data address and the length.
- The implementation creates `&[()]` with `&*slc`. This creates an invalid null reference before the subsequent `.len()` call, regardless of the slice's zero length.

The fact that `KnownLayout` is an unsafe trait restricts its implementations; it does not make its safe methods unsafe to call. Hiding the method from generated documentation also does not prevent safe callers from invoking it.

## Full repro

```toml
[package]
name = "zerocopy-null-metadata-repro"
version = "0.1.0"
edition = "2024"

[workspace]

[dependencies]
# Pin the derive dependencies to the tested versions.
```

Save this complete proof of concept as `src/bin/null_metadata.rs`:

```rust
#![forbid(unsafe_code)]

use zerocopy::KnownLayout;

fn main() {
    let null_slice = std::ptr::slice_from_raw_parts_mut(std::ptr::null_mut::<u8>(), 0);
    let metadata = <[u8] as KnownLayout>::pointer_to_metadata(null_slice);
    println!("metadata = {metadata}");
}
```

Generate a lockfile, check that the safe program compiles, and run Miri:

```sh
cargo +stable check --locked --bin null_metadata
cargo +nightly-2026-09-04 miri run --locked --bin null_metadata
```

The compilation check succeeds. The default Miri run fails with this exact diagnostic:

```text
error: Undefined Behavior: constructing invalid value of type &[()]: encountered a null reference
```

It identifies the following operation at `zerocopy/src/lib.rs:1085:28`:

```text
1085 |         let slc = unsafe { &*slc };
     |                            ^^^^^ Undefined Behavior occurred here
```

The backtrace leads from `<[u8] as zerocopy::KnownLayout>::pointer_to_metadata` to line 7 of the proof of concept. Miri exits with an error before the `println!` call. No memory read through the null pointer is necessary to trigger this diagnostic: construction of the reference is itself invalid.

## Root cause

The links below refer to repository revision `2dad389b030e9268d6645ac0bf0626b867e96068`.

1. [`KnownLayout::pointer_to_metadata`](https://github.com/google/zerocopy/blob/2dad389b030e9268d6645ac0bf0626b867e96068/zerocopy/src/lib.rs#L805-L812) is a safe method accepting an unrestricted raw pointer. Its signature neither requires `NonNull` nor establishes a caller-side unsafe precondition.
2. The [slice implementation](https://github.com/google/zerocopy/blob/2dad389b030e9268d6645ac0bf0626b867e96068/zerocopy/src/lib.rs#L1074-L1085) casts the pointer to `*const [()]` and creates a reference. Its safety comment assumes that the pointer is non-null, but no check establishes that fact for a direct safe call. Changing the element type to `()` removes the need to access element storage; it does not remove the reference's non-null requirement.
3. The [subsequent length extraction](https://github.com/google/zerocopy/blob/2dad389b030e9268d6645ac0bf0626b867e96068/zerocopy/src/lib.rs#L1087-L1098) occurs after the invalid reference has already been created. Metadata extraction needs to avoid creating such a reference or otherwise handle null input soundly.

The standard library explicitly permits safe construction of a raw slice pointer while requiring a non-null pointer before dereferencing it; its [`slice_from_raw_parts_mut` documentation](https://doc.rust-lang.org/std/ptr/fn.slice_from_raw_parts_mut.html) includes a null, empty raw slice example. The separate [reference safety requirements](https://doc.rust-lang.org/std/primitive.reference.html#safety) still apply to the `&[()]` created inside zerocopy.

## Suggested remediation

A small backportable fix is to check the data pointer before constructing the reference. This deliberately makes null input panic, including an empty null slice. The check must remain enabled in release builds. Applied to the audited revision, the complete source change is:

```diff
--- a/zerocopy/src/lib.rs
+++ b/zerocopy/src/lib.rs
@@ -1073,15 +1073,16 @@

     #[inline(always)]
     fn pointer_to_metadata(ptr: *mut [T]) -> usize {
+        assert!(!ptr.is_null(), "slice pointer must be non-null");
+
         #[allow(clippy::as_conversions)]
         let slc = ptr as *const [()];

         // SAFETY:
         // - `()` has alignment 1, so `slc` is trivially aligned.
-        // - `slc` was derived from a non-null pointer.
-        // - The size is 0 regardless of the length, so it is sound to
-        //   materialize a reference regardless of location.
-        // - By invariant, `self.ptr` has valid provenance.
+        // - `slc` is non-null by the assertion above.
+        // - `[()]` has size 0 for any length, and `()` has no invalid bit
+        //   patterns. No element storage is accessed or required.
         let slc = unsafe { &*slc };

         // This is correct because the preceding `as` cast preserves the number
```

The null check examines the data address, regardless of the slice length. After that check, the `[()]` reference is aligned and has size zero for every length. This justification does not require valid storage for any `T` elements, and it removes the previous reference to a nonexistent `self.ptr` invariant.

The exact patch was tested in an isolated copy of revision `2dad389b030e9268d6645ac0bf0626b867e96068`, using the toolchains and dependency versions above:

- The unmodified source reproduced the original null-reference UB under default Miri. With the patch, the unchanged proof of concept instead exited with the expected panic, `slice pointer must be non-null` (status 101), without a Miri UB diagnostic.
- Five regression tests passed under both default Miri and native Rust 1.98.1: catching the panic for null pointers with lengths 0, 1, and `usize::MAX`; preserving the length of a non-null empty slice; preserving the length and contents of a nonempty slice; extracting `usize::MAX` from a dangling `[()]` pointer; and extracting metadata from a non-null pointer without allocated or aligned `u64` element storage.
- `cargo +stable check --locked --all-targets`, formatting with the repository's configuration, and Clippy on the standalone regression harness with warnings denied passed. This is not a claim that the upstream Clippy or full CI matrix passed; upstream Clippy encounters an existing `clippy::needless_nonzero_get` diagnostic in `zerocopy-derive/src/repr.rs`.

The proposed guard uses APIs available at the declared Rust 1.56 MSRV, but that compiler and the full cross-platform CI matrix were not run. Null input now has a defined panic outcome, which callers must account for. A reference-free implementation using [`ptr.len()`](https://doc.rust-lang.org/std/primitive.pointer.html#method.len-1) could also accept null input, but that method was stabilized in Rust 1.79; adopting it would require an MSRV change or a compatible fallback.

## Disclaimer

joshlf:
KnownLayout's slice implementation materialized a shared reference solely to
read raw pointer metadata. That is invalid for null, dangling, unaligned, or
otherwise non-dereferenceable slice pointers even though metadata extraction
itself must accept them.

Use the safe raw-slice pointer length API on Rust 1.79 and newer. On older
supported compilers, capture a metadata candidate through raw-pointer hashing
and authenticate it with raw slice pointer equality before returning it.
Strengthen the implementer contract and cover null, dangling, deallocated,
unaligned, derived-DST, and boundary-toolchain cases.

Closes #3615

---

- 👉 #3629

**Latest Update:** v3 — [Compare vs v2](/google/zerocopy/compare/gherrit/Gro7m5vgpfdcy3pxtpl52hnbgworclnzw/v2..gherrit/Gro7m5vgpfdcy3pxtpl52hnbgworclnzw/v3)

<details>

|Version| v2 | v1 |Base|
|:---|:---|:---|:---|
|v3|[vs v2](/google/zerocopy/compare/gherrit/Gro7m5vgpfdcy3pxtpl52hnbgworclnzw/v2..gherrit/Gro7m5vgpfdcy3pxtpl52hnbgworclnzw/v3)|[vs v1](/google/zerocopy/compare/gherrit/Gro7m5vgpfdcy3pxtpl52hnbgworclnzw/v1..gherrit/Gro7m5vgpfdcy3pxtpl52hnbgworclnzw/v3)|[vs Base](/google/zerocopy/compare/main..gherrit/Gro7m5vgpfdcy3pxtpl52hnbgworclnzw/v3)|
|v2||[vs v1](/google/zerocopy/compare/gherrit/Gro7m5vgpfdcy3pxtpl52hnbgworclnzw/v1..gherrit/Gro7m5vgpfdcy3pxtpl52hnbgworclnzw/v2)|[vs Base](/google/zerocopy/compare/main..gherrit/Gro7m5vgpfdcy3pxtpl52hnbgworclnzw/v2)|
|v1|||[vs Base](/google/zerocopy/compare/main..gherrit/Gro7m5vgpfdcy3pxtpl52hnbgworclnzw/v1)|

</details>
<details>

######

**Branch**
```bash
git fetch origin refs/heads/Gro7m5vgpfdcy3pxtpl52hnbgworclnzw && git checkout -b pr-Gro7m5vgpfdcy3pxtpl52hnbgworclnzw FETCH_HEAD
```

**Checkout**
```bash
git fetch origin refs/heads/Gro7m5vgpfdcy3pxtpl52hnbgworclnzw && git checkout FETCH_HEAD
```

**Cherry Pick**
```bash
git fetch origin refs/heads/Gro7m5vgpfdcy3pxtpl52hnbgworclnzw && git cherry-pick FETCH_HEAD
```

**Pull**
```bash
git pull origin refs/heads/Gro7m5vgpfdcy3pxtpl52hnbgworclnzw
```

</details>

### DIFF
--- a/zerocopy/src/lib.rs
+++ b/zerocopy/src/lib.rs
@@ -1073,29 +1073,85 @@ unsafe impl<T> KnownLayout for [T] {
 
     #[inline(always)]
     fn pointer_to_metadata(ptr: *mut [T]) -> usize {
-        #[allow(clippy::as_conversions)]
-        let slc = ptr as *const [()];
+        #[cfg(not(no_zerocopy_slice_ptr_len_1_79_0))]
+        {
+            ptr.len()
+        }
 
-        // SAFETY:
-        // - `()` has alignment 1, so `slc` is trivially aligned.
-        // - `slc` was derived from a non-null pointer.
-        // - The size is 0 regardless of the length, so it is sound to
-        //   materialize a reference regardless of location.
-        // - By invariant, `self.ptr` has valid provenance.
-        let slc = unsafe { &*slc };
+        #[cfg(no_zerocopy_slice_ptr_len_1_79_0)]
+        {
+            // `*mut [T]::len` was not stable before Rust 1.79. In every Rust
+            // version from our 1.56 MSRV through 1.78, `Hash for *mut T`
+            // decomposes a raw pointer and passes its address and metadata to
+            // the hasher in that order [1]. `Hash for usize` passes its value
+            // to `Hasher::write_usize` [2]. Capture those two values to obtain
+            // a candidate for the slice length without dereferencing `ptr` or
+            // constructing a reference.
+            //
+            // This historical `Hash` implementation is only used to produce a
+            // candidate. Before returning it, we reconstruct a raw slice and
+            // authenticate the candidate with `ptr::eq`, which compares slice
+            // lengths as well as addresses [3]. Thus, an unexpected `Hash`
+            // implementation cannot cause us to return incorrect metadata; it
+            // can only fail to produce an authenticated candidate.
+            //
+            // [1] Per https://doc.rust-lang.org/1.56.0/src/core/hash/mod.rs.html#776-782:
+            //
+            //   let (address, metadata) = self.to_raw_parts();
+            //   state.write_usize(address as usize);
+            //   metadata.hash(state);
+            //
+            // [2] Per https://doc.rust-lang.org/1.56.0/src/core/hash/mod.rs.html#628-656:
+            //
+            //   fn hash<H: Hasher>(&self, state: &mut H) {
+            //       state.$meth(*self)
+            //   }
+            //   ...
+            //   (usize, write_usize),
+            //
+            // [3] Per https://doc.rust-lang.org/1.56.0/std/ptr/fn.eq.html:
+            //
+            //   Slices are also compared by their length (fat pointers).
+            struct MetadataHasher {
+                values: [usize; 2],
+                writes: usize,
+                valid: bool,
+            }
 
-        // This is correct because the preceding `as` cast preserves the number
-        // of slice elements. [1]
-        //
-        // [1] Per https://doc.rust-lang.org/reference/expressions/operator-expr.html#pointer-to-pointer-cast:
-        //
-        //   For slice types like `[T]` and `[U]`, the raw pointer types `*const
-        //   [T]`, `*mut [T]`, `*const [U]`, and `*mut [U]` encode the number of
-        //   elements in this slice. Casts between these raw pointer types
-        //   preserve the number of elements. ... The same holds for `str` and
-        //   any compound type whose unsized tail is a slice type, such as
-        //   struct `Foo(i32, [u8])` or `(u64, Foo)`.
-        slc.len()
+            impl Hasher for MetadataHasher {
+                #[inline(always)]
+                fn finish(&self) -> u64 {
+                    0
+                }
+
+                #[inline(always)]
+                fn write(&mut self, _bytes: &[u8]) {
+                    self.valid = false;
+                }
+
+                #[inline(always)]
+                fn write_usize(&mut self, value: usize) {
+                    match self.values.get_mut(self.writes) {
+                        Some(slot) => *slot = value,
+                        None => self.valid = false,
+                    }
+                    self.writes = self.writes.saturating_add(1);
+                }
+            }
+
+            let mut hasher = MetadataHasher { values: [0; 2], writes: 0, valid: true };
+            core::hash::Hash::hash(&ptr, &mut hasher);
+            assert!(
+                hasher.valid && hasher.writes == 2,
+                "unexpected raw-pointer Hash implementation"
+            );
+
+            let elems = hasher.values[1];
+            #[allow(clippy::as_conversions)]
+            let reconstructed = ptr::slice_from_raw_parts_mut(ptr as *mut T, elems);
+            assert!(ptr::eq(ptr, reconstructed), "captured value is not raw-slice metadata");
+            elems
+        }
     }
 }
 
@@ -6535,6 +6591,58 @@ mod tests {
         test!([()], layout(0, 1, Some(0), true));
         test!([u8], layout(0, 1, Some(1), true));
         test!(str, layout(0, 1, Some(1), true));
+    }
+
+    #[test]
+    fn test_known_layout_pointer_to_metadata() {
+        fn test<T>(data: *mut T, elems: usize) {
+            let ptr = ptr::slice_from_raw_parts_mut(data, elems);
+            assert_eq!(<[T] as KnownLayout>::pointer_to_metadata(ptr), elems);
+        }
+
+        for elems in [0, 1, usize::MAX] {
+            test(ptr::null_mut::<u8>(), elems);
+        }
+
+        test(NonNull::<u8>::dangling().as_ptr(), 0);
+
+        let mut bytes = [1, 2, 3];
+        test(bytes.as_mut_ptr(), bytes.len());
+        assert_eq!(bytes, [1, 2, 3]);
+
+        test(NonNull::<()>::dangling().as_ptr(), usize::MAX);
+
+        // Retaining and inspecting a raw pointer after freeing its allocation
+        // is safe so long as the pointer is not dereferenced.
+        let deallocated = {
+            let mut byte = Box::new(0u8);
+            let ptr: *mut u8 = &mut *byte;
+            drop(byte);
+            ptr
+        };
+        test(deallocated, 3);
+
+        // Metadata extraction requires neither element storage nor alignment
+        // for `T`; it does not access the pointer's referent.
+        test(NonNull::<u8>::dangling().as_ptr().cast::<u64>(), usize::MAX);
+    }
+
+    #[cfg(feature = "derive")]
+    #[test]
+    fn test_known_layout_pointer_to_metadata_derive() {
+        #[derive(KnownLayout)]
+        #[repr(C)]
+        struct Dst {
+            prefix: u8,
+            trailing: [u8],
+        }
+
+        for elems in [0, 1, usize::MAX] {
+            let ptr = ptr::slice_from_raw_parts_mut(ptr::null_mut::<u8>(), elems);
+            #[allow(clippy::as_conversions)]
+            let ptr = ptr as *mut Dst;
+            assert_eq!(Dst::pointer_to_metadata(ptr), elems);
+        }
     }
 
     #[cfg(feature = "derive")]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,6 +86,7 @@ jobs:
           # which a particular feature is supported.
           "no-zerocopy-simd-x86-avx12-1-89-0",
           "no-zerocopy-core-error-1-81-0",
+          "no-zerocopy-slice-ptr-len-1-79-0",
           "no-zerocopy-diagnostic-on-unimplemented-1-78-0",
           "no-zerocopy-generic-bounds-in-const-fn-1-61-0",
           "no-zerocopy-target-has-atomics-1-60-0",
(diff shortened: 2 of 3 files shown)


joshlf:
@codex review

zerocopy/src/lib.rs
```diff
@@ -1073,29 +1074,93 @@ unsafe impl<T> KnownLayout for [T] {
...
-        //   For slice types like `[T]` and `[U]`, the raw pointer types `*const
-        //   [T]`, `*mut [T]`, `*const [U]`, and `*mut [U]` encode the number of
-        //   elements in this slice. Casts between these raw pointer types
-        //   preserve the number of elements. ... The same holds for `str` and
-        //   any compound type whose unsized tail is a slice type, such as
-        //   struct `Foo(i32, [u8])` or `(u64, Foo)`.
-        slc.len()
+        #[cfg(no_zerocopy_slice_ptr_len_1_79_0)]
+        {
+            // `*mut [T]::len` was not stable before Rust 1.79. In every Rust
```
joshlf, line 1083: So long as we're relying on logic that is specific to semver-unstable details of a particular range of toolchains, could we instead just rely on the layout of a slice fat pointer and extract the length directly?

joshlf: I considered this, but I think the current implementation is safer despite its verbosity. The two dependencies on implementation detail are materially different:

- A direct union or transmute would make the undocumented wide-pointer component layout part of zerocopy’s unsafe proof before any check can run. Rust does not guarantee that layout to downstream code; in fact, the private `PtrRepr` implementation used by `core` in this toolchain range explicitly notes that only the standard library can rely on its pointer/components layout equivalence.
- Here, safe `core` performs the representation-sensitive decomposition through `Hash`. We treat its output only as a candidate, reconstruct a raw slice from the original data pointer, and accept it only when `ptr::eq` authenticates both address and metadata. If the historical `Hash` call shape changes, we panic rather than return incorrect metadata; every normal return is therefore correct.

So the current toolchain-specific fact is fail-closed rather than memory-safety-critical. I prefer to retain this fallback until the MSRV permits raw-slice `ptr.len()`; #3641 tracks deleting it then.

zerocopy/src/lib.rs
```diff
@@ -1073,29 +1074,93 @@ unsafe impl<T> KnownLayout for [T] {
...
+        #[cfg(not(no_zerocopy_slice_ptr_len_1_79_0))]
+        {
+            // `*mut [T]::len` is safe even if `ptr` cannot be converted to a
+            // reference because it is null, unaligned, or otherwise dangling
+            // [1]. It reads only the raw pointer's metadata.
+            //
+            // [1] Per https://doc.rust-lang.org/1.79.0/std/primitive.pointer.html#method.len-1:
+            //
+            //   This function is safe, even when the raw slice cannot be cast
+            //   to a slice reference because the pointer is null or unaligned.
```
joshlf: Why do we need what is in effect a safety comment on a not-`unsafe` block of code?

joshlf: Agreed; we do not need it. I added the block to spell out why the implementation satisfies `KnownLayout::pointer_to_metadata` for arbitrary raw slice pointers, but `ptr.len()` directly provides that behavior and the trait already states the requirement, so the comment only repeats the standard-library contract. I removed it. The longer pre-1.79 comment remains because that compatibility fallback and its fail-closed `ptr::eq` authentication are not obvious from the code.

joshlf:
@codex review

zerocopy/src/lib.rs
```diff
@@ -1073,29 +1074,85 @@ unsafe impl<T> KnownLayout for [T] {
...
+                hasher.valid && hasher.writes == 2,
+                "unexpected raw-pointer Hash implementation"
+            );
+
+            let elems = hasher.values[1];
+            #[allow(clippy::as_conversions)]
+            let reconstructed = ptr::slice_from_raw_parts_mut(ptr as *mut T, elems);
+            assert!(ptr::eq(ptr, reconstructed), "captured value is not raw-slice metadata");
+            elems
+        }
```
jswrenn, line 1154: This is, by far, my new favorite hack in the zerocopy codebase.

joshlf: Mine too

zerocopy/src/lib.rs
```diff
@@ -1073,29 +1074,85 @@ unsafe impl<T> KnownLayout for [T] {
...
-        #[allow(clippy::as_conversions)]
-        let slc = ptr as *const [()];
+        #[cfg(not(no_zerocopy_slice_ptr_len_1_79_0))]
+        {
+            ptr.len()
+        }
 
-        // SAFETY:
-        // - `()` has alignment 1, so `slc` is trivially aligned.
-        // - `slc` was derived from a non-null pointer.
```
jswrenn, line 1081: Just to check: We don't just want to have `pointer_to_metadata` consume a `NonNull`? 

And also that "The size is 0 regardless of the length, so it is sound to materialize a reference regardless of location." is also incorrect, since presumably that rule does not apply to location `0`?

jswrenn: Well the miri failure in the fixed issue handily answers that second question. Silly opsem!

IMO, the workaround is cool, and fixing this issue without reducing the expressive power of this method is *very* cool.

jswrenn (approved):
The alternative here is altering the contract of `pointer_to_metadata` to consume a `NonNull`, but I think that'd cause more churn than it's worth for what can instead be resolved by an extremely cute MSRV workaround.

zerocopy/src/lib.rs
```diff
@@ -807,7 +807,8 @@ pub unsafe trait KnownLayout {
     /// # Safety
     ///
     /// `pointer_to_metadata` always returns the correct metadata stored in
-    /// `ptr`.
+    /// `ptr`. Since this method is safe, implementations must not dereference
+    /// `ptr` or require it to be non-null, aligned, live, or dereferenceable.
```
joshlf: Redundant; remove. This is definitionally true of all safe implementations. The bug was internal, not a result of incomplete safety documentation.

joshlf: *Authored by an AI agent acting on Josh Liebow-Feeser's behalf.*

Addressed on exact head `70f05c5df8a2d9f036321845dc6b5b3198f52ae0`. Removed the redundant safe-implementation warning while retaining the actual `KnownLayout` safety contract that `pointer_to_metadata` return the correct metadata. The repository pre-push validation passes.

joshlf:

@codex review

Please review exact head `70f05c5df8a2d9f036321845dc6b5b3198f52ae0`, including the latest requested documentation removal.

PR #3631, opened 2026-09-05
Reject context-dependent enum discriminants

### LINKED ISSUE #3619: Security Advisory: Self-dependent enum discriminants admit invalid values and disclose adjacent memory

- **Vulnerability Type:** Incorrect validity checking / out-of-bounds read / information disclosure
- **Affected Software:** zerocopy-derive
- **Date:** September 2026

## Summary

`#[derive(TryFromBytes)]` can validate the wrong enum variant when a discriminant expression refers to `Self`. The derive copies that expression into a generated helper enum, where `Self` denotes the helper rather than the application's enum. If the expression resolves differently in the two contexts, the generated validation routine can accept bytes that are invalid for the actual variant.

The reproduction below defines an enum containing `Flag(bool)` and `Raw(u8)`. A safe blanket trait supplies a default `TAG` constant, while the application enum has an inherent constant of the same name. The actual enum and the generated helper consequently assign opposite tags to their variants. An attacker-supplied two-byte packet is validated as an unrestricted byte and returned as an invalid boolean. In an optimized native build, ordinary safe indexing with that boolean reads beyond a two-byte public response table and discloses adjacent synthetic secret bytes.

**Threat model:** The application must already contain an affected enum declaration, such as the constant-name collision demonstrated here, and parse attacker-controlled bytes into it. After compilation, the attacker only needs control of the packet bytes. Literal discriminants without this context-dependent resolution do not satisfy the demonstrated prerequisite. This is a library soundness defect with a validated, conditional information-disclosure consequence; we have not demonstrated remote code execution or established that a production service contains this schema.

**Affected version and revision:** The original audit examined `zerocopy 0.8.56` and `zerocopy-derive 0.8.56` at public repository revision `2dad389b030e9268d6645ac0bf0626b867e96068`. The standalone reproduction below uses published version `0.8.56` of both crates, whose package metadata identifies revision `6dc429c451bdf1d7202ec1ec2cf426514e00d8eb`. The affected code is unchanged between these sources. We have not exhaustively tested other versions.

**Environment:** Native reproduction on macOS arm64 (`aarch64-apple-darwin`) with Rust 1.98.1 (`48a229cea`, LLVM 22.1.8). Miri reproduction with `nightly-2026-09-04`, Rust 1.100.0-nightly (`a69a63265`, 2026-09-03), and Miri 0.1.0 (`a69a63265c`). The native out-of-bounds result depends on optimization; Miri independently confirms the invalid boolean.

## Sketch of the attack

The example models a server whose public responses are selected by a parsed boolean:

```rust
// In the actual enum, tag 1 means Flag(bool).
let wire = [1, 2];
let packet = Packet::try_ref_from_bytes(&wire).unwrap();

// The derive validated tag 1 as Raw(u8), so the invalid bool value 2 escaped.
match packet {
    Packet::Flag(flag) => replies.public[usize::from(*flag)],
    Packet::Raw(byte) => *byte,
};
```

Under the hood:

- `Self::TAG` resolves to the application's inherent `TAG = 1` in the original enum.
- The helper enum has no such inherent constant. The copied expression resolves to the blanket trait's `TAG = 0` instead.
- The helper therefore validates tag 1 as `Raw(u8)`, whose payload permits every byte value. The actual tag 1 variant is `Flag(bool)`, which permits only zero or one.
- The compiler can omit a bounds check when a valid boolean indexes a two-element array. Once the library supplies an invalid boolean, that safety premise is false. In the tested optimized build, payloads 2 through 7 read the six bytes immediately after the public array.

## Full repro

```toml
[package]
name = "zerocopy-enum-discriminant-repro"
version = "0.1.0"
edition = "2024"

[workspace]

[dependencies]
```

Save this complete program as `src/bin/enum_discriminant.rs`:

```rust
#![forbid(unsafe_code)]

use zerocopy::{Immutable, KnownLayout, TryFromBytes};

trait DefaultTag {
    const TAG: u8 = 0;
}
impl<T> DefaultTag for T {}

#[derive(Immutable, KnownLayout, TryFromBytes)]
#[repr(u8)]
enum Packet {
    Flag(bool) = Self::TAG,
    Raw(u8) = 1 - Self::TAG,
}

impl Packet {
    const TAG: u8 = 1;
}

#[repr(C)]
struct ServerReplies {
    public: [u8; 2],
    secret: [u8; 6],
}

#[inline(never)]
fn respond(packet: &Packet, replies: &ServerReplies) -> u8 {
    match packet {
        Packet::Flag(flag) => replies.public[usize::from(*flag)],
        Packet::Raw(byte) => *byte,
    }
}

fn main() {
    let index = std::env::args()
        .nth(1)
        .map_or(2, |s| s.parse::<u8>().unwrap());
    let wire = std::hint::black_box([1, index]);
    let replies = std::hint::black_box(ServerReplies {
        public: *b"AB",
        secret: *b"SECRET",
    });
    let packet = Packet::try_ref_from_bytes(&wire).unwrap();
    println!("wire={wire:?}; reply=0x{:02x}", respond(packet, &replies));
}
```

Generate the lockfile after saving the program, build, and run the native demonstration:

```sh
cargo +stable build --locked --release --bin enum_discriminant
for byte in 0 1 2 3 4 5 6 7; do
    ./target/release/enum_discriminant "$byte"
done
```

Observed output:

```text
wire=[1, 0]; reply=0x41
wire=[1, 1]; reply=0x42
wire=[1, 2]; reply=0x53
wire=[1, 3]; reply=0x45
wire=[1, 4]; reply=0x43
wire=[1, 5]; reply=0x52
wire=[1, 6]; reply=0x45
wire=[1, 7]; reply=0x54
```

The first two replies are the permitted public bytes `A` and `B`. The next six spell `SECRET`, which is outside the public response array. The fixture deliberately places synthetic secret data next to that array; this is not a claim that every application arranges sensitive data identically.

Run the same program with Miri:

```sh
MIRIFLAGS='-Zmiri-strict-provenance' \
    cargo +nightly-2026-09-04 miri run --locked --bin enum_discriminant -- 2
```

Miri exits with an error at the `*flag` read:

```text
error: Undefined Behavior: constructing invalid value of type bool: encountered 0x02, but expected a boolean
```

This error is the expected reproduction result. The Rust Reference requires a boolean to be zero or one and requires the fields of an enum's actual variant to be valid at their types. [Validity rules](https://doc.rust-lang.org/reference/behavior-considered-undefined.html#invalid-values).

## Root cause

The links below refer to the original audited revision `2dad389b030e9268d6645ac0bf0626b867e96068`. The same affected code is present in published version `0.8.56`.

1. [`generate_tag_enum`](https://github.com/google/zerocopy/blob/2dad389b030e9268d6645ac0bf0626b867e96068/zerocopy/zerocopy-derive/src/util.rs#L826-L850) copies each discriminant expression directly into `___ZerocopyTag`. It does not preserve the original enum as the meaning of `Self`.
2. [`generate_tag_consts`](https://github.com/google/zerocopy/blob/2dad389b030e9268d6645ac0bf0626b867e96068/zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs#L32-L55) computes validation constants from that helper enum. In the reproduction, these constants describe the opposite variant assignments from the original type.
3. The generated [`is_bit_valid` match arms](https://github.com/google/zerocopy/blob/2dad389b030e9268d6645ac0bf0626b867e96068/zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs#L304-L339) validate the payload selected by the helper tag. The [final match](https://github.com/google/zerocopy/blob/2dad389b030e9268d6645ac0bf0626b867e96068/zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs#L439-L442) can therefore report success without validating the actual variant's payload.
4. This breaks [`TryFromBytes::is_bit_valid`'s contract](https://github.com/google/zerocopy/blob/2dad389b030e9268d6645ac0bf0626b867e96068/zerocopy/src/lib.rs#L1752-L1777): a successful check must mean the bytes contain a valid instance of the target type.

Discriminant evaluation must preserve the original enum's context. Rejecting expressions whose meaning cannot be preserved is safer than evaluating them as members of a different type.

## Disclaimer

joshlf:
Derives that validate enum tags copy explicit discriminants into a helper
enum. Context-dependent expressions can evaluate differently in that helper,
allowing the generated validator to accept an invalid original-enum tag.

Restrict copied discriminants to a compiler-defined, repeatable grammar of
integer and byte literals with primitive arithmetic and bitwise operations.
Reject paths, macros, calls, casts, attributes, and other context-bearing
forms before emitting any unsafe impl. Preserve the skip-on-error dependency
chain for exhaustive enums, including fieldful variants whose zero tag cannot
be identified syntactically, and add cross-toolchain regressions.

Closes #3619

---

- 　  #3635
- 👉 #3631

**Latest Update:** v10 — [Compare vs v9](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v9..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v10)

<details>

|Version| v9 | v8 | v7 | v6 | v5 | v4 | v3 | v2 | v1 |Base|
|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|
|v10|[v9](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v9..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v10)|[v8](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v8..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v10)|[v7](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v7..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v10)|[v6](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v6..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v10)|[v5](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v5..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v10)|[v4](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v4..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v10)|[v3](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v3..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v10)|[v2](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v2..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v10)|[v1](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v1..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v10)|[Base](/google/zerocopy/compare/main..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v10)|
|v9||[v8](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v8..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v9)|[v7](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v7..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v9)|[v6](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v6..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v9)|[v5](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v5..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v9)|[v4](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v4..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v9)|[v3](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v3..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v9)|[v2](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v2..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v9)|[v1](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v1..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v9)|[Base](/google/zerocopy/compare/main..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v9)|
|v8|||[v7](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v7..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v8)|[v6](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v6..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v8)|[v5](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v5..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v8)|[v4](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v4..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v8)|[v3](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v3..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v8)|[v2](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v2..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v8)|[v1](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v1..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v8)|[Base](/google/zerocopy/compare/main..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v8)|
|v7||||[v6](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v6..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v7)|[v5](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v5..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v7)|[v4](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v4..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v7)|[v3](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v3..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v7)|[v2](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v2..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v7)|[v1](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v1..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v7)|[Base](/google/zerocopy/compare/main..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v7)|
|v6|||||[v5](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v5..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v6)|[v4](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v4..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v6)|[v3](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v3..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v6)|[v2](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v2..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v6)|[v1](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v1..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v6)|[Base](/google/zerocopy/compare/main..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v6)|
|v5||||||[v4](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v4..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v5)|[v3](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v3..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v5)|[v2](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v2..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v5)|[v1](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v1..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v5)|[Base](/google/zerocopy/compare/main..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v5)|
|v4|||||||[v3](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v3..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v4)|[v2](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v2..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v4)|[v1](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v1..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v4)|[Base](/google/zerocopy/compare/main..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v4)|
|v3||||||||[v2](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v2..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v3)|[v1](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v1..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v3)|[Base](/google/zerocopy/compare/main..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v3)|
|v2|||||||||[v1](/google/zerocopy/compare/gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v1..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v2)|[Base](/google/zerocopy/compare/main..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v2)|
|v1||||||||||[Base](/google/zerocopy/compare/main..gherrit/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck/v1)|

</details>
<details>

######

**Branch**
```bash
git fetch origin refs/heads/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck && git checkout -b pr-Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck FETCH_HEAD
```

**Checkout**
```bash
git fetch origin refs/heads/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck && git checkout FETCH_HEAD
```

**Cherry Pick**
```bash
git fetch origin refs/heads/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck && git cherry-pick FETCH_HEAD
```

**Pull**
```bash
git pull origin refs/heads/Gifjr6pbpu4fh6t7qbzuuqczj3odny3ck
```

</details>

### DIFF
--- a/zerocopy/src/lib.rs
+++ b/zerocopy/src/lib.rs
@@ -1411,6 +1411,17 @@ where
 ///     discriminant values are specified.
 ///   - The fields of that variant must be `FromZeros`.
 ///
+/// When this derive needs to reproduce an enum's explicit discriminants in
+/// generated code to implement its `TryFromBytes` supertrait, it accepts only
+/// integer or byte literals, arithmetic and bitwise operations whose operands
+/// recursively use this grammar, unary `-` and `!`, grouping, and `.to_le()` or
+/// `.to_be()` directly on an explicitly typed integer literal. Paths (including
+/// paths to constants), macros, general calls and method calls, casts, and
+/// control-flow expressions are rejected. A fieldless enum whose variants
+/// exhaust every discriminant of its representation uses a validator that
+/// accepts every bit pattern without inspecting or copying discriminants, so
+/// this restriction does not apply.
+///
 /// This analysis is subject to change. Unsafe code may *only* rely on the
 /// documented [safety conditions] of `FromZeros`, and must *not* rely on the
 /// implementation details of this derive.
@@ -1633,6 +1644,16 @@ pub unsafe trait Immutable {
 ///     V2 = 10u32.to_le(),
 /// }
 /// ```
+///
+/// When this derive needs to reproduce an enum's explicit discriminants in
+/// generated code, it accepts only integer or byte literals, arithmetic and
+/// bitwise operations whose operands recursively use this grammar, unary `-`
+/// and `!`, grouping, and `.to_le()` or `.to_be()` directly on an explicitly
+/// typed integer literal. Paths (including paths to constants), macros,
+/// general calls and method calls, casts, and control-flow expressions are
+/// rejected. The original enum and generated helper evaluate each copied
+/// discriminant independently, so the derive cannot assume that an arbitrary
+/// const expression produces the same value both times.
 ///
 /// [safety conditions]: trait@TryFromBytes#safety
 #[cfg(any(feature = "derive", test))]
@@ -3901,6 +3922,17 @@ pub unsafe trait FromZeros: TryFromBytes {
 ///     bit pattern is a valid one).
 ///   - Its fields must be `FromBytes`.
 ///
+/// When this derive needs to reproduce an enum's explicit discriminants in
+/// generated code to implement its `TryFromBytes` supertrait, it accepts only
+/// integer or byte literals, arithmetic and bitwise operations whose operands
+/// recursively use this grammar, unary `-` and `!`, grouping, and `.to_le()` or
+/// `.to_be()` directly on an explicitly typed integer literal. Paths (including
+/// paths to constants), macros, general calls and method calls, casts, and
+/// control-flow expressions are rejected. A fieldless enum whose variants
+/// exhaust every discriminant of its representation uses a validator that
+/// accepts every bit pattern without inspecting or copying discriminants, so
+/// this restriction does not apply.
+///
 /// This analysis is subject to change. Unsafe code may *only* rely on the
 /// documented [safety conditions] of `FromBytes`, and must *not* rely on the
 /// implementation details of this derive.
@@ -5529,6 +5561,16 @@ fn mut_from_prefix_suffix<T: FromBytes + IntoBytes + KnownLayout + ?Sized>(
 /// # */
 /// }
 /// ```
+///
+/// When this derive reproduces an enum's explicit discriminants in generated
+/// code, it accepts only integer or byte literals, arithmetic and bitwise
+/// operations whose operands recursively use this grammar, unary `-` and `!`,
+/// grouping, and `.to_le()` or `.to_be()` directly on an explicitly typed
+/// integer literal. Paths (including paths to constants), macros, general
+/// calls and method calls, casts, and control-flow expressions are rejected.
+/// The original enum and generated helper evaluate each copied discriminant
+/// independently, so the derive cannot assume that an arbitrary const
+/// expression produces the same value both times.
 ///
 /// [safety conditions]: trait@IntoBytes#safety
 ///
--- a/zerocopy/zerocopy-derive/src/util.rs
+++ b/zerocopy/zerocopy-derive/src/util.rs
@@ -835,16 +835,304 @@ pub(crate) fn const_block(items: impl IntoIterator<Item = Option<TokenStream>>) 
         };
     }
 }
-pub(crate) fn generate_tag_enum(ctx: &Ctx, repr: &EnumRepr, data: &DataEnum) -> TokenStream {
-    let zerocopy_crate = &ctx.zerocopy_crate;
-    let variants = data.variants.iter().map(|v| {
-        let ident = &v.ident;
-        if let Some((eq, discriminant)) = &v.discriminant {
-            quote! { #ident #eq #discriminant }
-        } else {
-            quote! { #ident }
+
+fn validate_tag_enum_discriminant(discriminant: &Expr) -> Result<(), Error> {
+    fn reject(syntax: impl ToTokens, message: &'static str) -> Result<(), Error> {
+        Err(Error::new_spanned(syntax, message))
+    }
+
+    fn validate_attrs(attrs: &[syn::Attribute]) -> Result<(), Error> {
+        match attrs.first() {
+            Some(attr) => reject(
+                attr,
+                "attributes are not supported in enum discriminants because their effect cannot \
+                 be preserved in Zerocopy's generated helper enum",
+            ),
+            None => Ok(()),
         }
-    });
+    }
+
+    fn has_supported_integer_suffix(integer: &syn::LitInt) -> bool {
+        matches!(
+            integer.suffix(),
+            "" | "u8"
+                | "u16"
+                | "u32"
+                | "u64"
+                | "u128"
+                | "usize"
+                | "i8"
+                | "i16"
+                | "i32"
+                | "i64"
+                | "i128"
+                | "isize"
+        )
+    }
+
+    fn validate_path(path: &Path) -> Result<(), Error> {
+        if path.segments.iter().any(|segment| segment.ident == "Self") {
+            // In the original discriminant, `Self` denotes the type being
+            // derived. In the copied discriminant, it would denote the
+            // generated tag enum instead.
+            return reject(
+                path,
+                "`Self` is not supported in enum discriminants because its meaning cannot be \
+                 preserved in Zerocopy's generated helper enum",
+            );
+        }
+
+        // The original enum and generated helper enum evaluate copied
+        // discriminants independently. Even a fully-qualified constant path
+        // is not guaranteed to evaluate repeatably: safe const evaluation can
+        // be nondeterministic, and a path can introduce caller-defined types
+        // whose operators have caller-defined semantics. Restrict every leaf
+        // to syntax whose value is defined directly by the compiler.
+        reject(
+            path,
+            "paths are not supported in enum discriminants because Zerocopy's generated helper \
+             enum evaluates copied discriminants independently",
+        )
+    }
+
+    fn validate_endian_method_call(call: &syn::ExprMethodCall) -> Result<(), Error> {
+        validate_attrs(&call.attrs)?;
+        let integer = match call.receiver.as_ref() {
+            Expr::Lit(ExprLit { attrs, lit: Lit::Int(integer), .. }) if attrs.is_empty() => integer,
+            _ => {
+                return reject(
+                    call,
+                    "only `.to_le()` and `.to_be()` on explicitly typed integer literals are \
+                     supported in enum discriminants",
+                )
+            }
+        };
+        if !integer.suffix().is_empty()
+            && has_supported_integer_suffix(integer)
+            && (call.method == "to_le" || call.method == "to_be")
+            && call.turbofish.is_none()
+            && call.args.is_empty()
+        {
+            // These inherent primitive methods are selected entirely by the
+            // explicit literal suffix and cannot be affected by the helper
+            // enum's surrounding item or trait scope.
+            Ok(())
+        } else {
+            reject(
+                call,
+                "only `.to_le()` and `.to_be()` on explicitly typed integer literals are \
+                 supported in enum discriminants",
+            )
+        }
+    }
+
+    fn validate_expr(expr: &Expr) -> Result<(), Error> {
+        match expr {
+            Expr::Binary(binary) => {
+                validate_attrs(&binary.attrs)?;
+                match &binary.op {
+                    syn::BinOp::Add(_)
+                    | syn::BinOp::Sub(_)
+                    | syn::BinOp::Mul(_)
+                    | syn::BinOp::Div(_)
+                    | syn::BinOp::Rem(_)
+                    | syn::BinOp::BitXor(_)
+                    | syn::BinOp::BitAnd(_)
+                    | syn::BinOp::BitOr(_)
+                    | syn::BinOp::Shl(_)
+                    | syn::BinOp::Shr(_) => {}
+                    _ => {
+                        return reject(
+                            binary.op,
+                            "only arithmetic and bitwise operators are supported in enum \
+                             discriminants",
+                        )
+                    }
+                }
+                validate_expr(&binary.left)?;
+                validate_expr(&binary.right)
+            }
+            Expr::Group(group) => {
+                validate_attrs(&group.attrs)?;
+                validate_expr(&group.expr)
+            }
+            Expr::Lit(lit) => {
+                validate_attrs(&lit.attrs)?;
+                match &lit.lit {
+                    Lit::Byte(_) => Ok(()),
+                    Lit::Int(integer) if has_supported_integer_suffix(integer) => Ok(()),
+                    Lit::Int(integer) => reject(
+                        integer,
+                        "only unsuffixed or primitive-integer-suffixed literals are supported in \
+                         enum discriminants",
+                    ),
+                    Lit::Verbatim(literal) => reject(
+                        literal,
+                        "unparsed syntax is not supported in enum discriminants because its \
+                         meaning cannot be preserved in Zerocopy's generated helper enum",
+                    ),
+                    _ => reject(
+                        &lit.lit,
+                        "only integer and byte literals are supported in enum discriminants",
+                    ),
+                }
+            }
+            Expr::MethodCall(call) => validate_endian_method_call(call),
+            Expr::Paren(paren) => {
+                validate_attrs(&paren.attrs)?;
+                validate_expr(&paren.expr)
+            }
+            Expr::Path(path) => {
+                validate_attrs(&path.attrs)?;
+                if let Some(qself) = &path.qself {
+                    if matches!(qself.ty.as_ref(), Type::Path(ty) if ty.qself.is_none() && ty.path.is_ident("Self"))
+                    {
+                        return reject(
+                            &qself.ty,
+                            "`Self` is not supported in enum discriminants because its meaning \
+                             cannot be preserved in Zerocopy's generated helper enum",
+                        );
+                    }
+                    return reject(
+                        path,
+                        "qualified type-relative paths are not supported in enum discriminants \
+                         because their meaning cannot be preserved in Zerocopy's generated \
+                         helper enum",
+                    );
+                }
+                validate_path(&path.path)
+            }
+            Expr::Unary(unary) => {
+                validate_attrs(&unary.attrs)?;
+                match &unary.op {
+                    syn::UnOp::Neg(_) | syn::UnOp::Not(_) => validate_expr(&unary.expr),
+                    _ => reject(
+                        unary.op,
+                        "only negation and bitwise-not unary operators are supported in enum \
+                         discriminants",
+                    ),
+                }
+            }
+            Expr::Macro(mac) => {
+                // A macro can emit `Self` even if its invocation does not
+                // contain a `Self` token. Expanding it once in the original
+                // enum and again in the helper enum can produce different
+                // values.
+                reject(
+                    mac,
+                    "macros are not supported in enum discriminants because their expansion \
+                     cannot be preserved in Zerocopy's generated helper enum",
+                )
+            }
+            Expr::Verbatim(tokens) => reject(
+                tokens,
+                "unparsed syntax is not supported in enum discriminants because its meaning \
+                 cannot be preserved in Zerocopy's generated helper enum",
+            ),
+            _ => {
+                // This is a positive grammar. In particular, it rejects every
+                // pattern, binding, control-flow construct, call, non-endian
+                // method call, block item, macro, and current or future
+                // unparsed AST variant. Those constructs can resolve names
+                // differently after the expression is copied into the
+                // generated helper enum.
+                reject(
+                    expr,
+                    "this expression is not supported in enum discriminants because its meaning \
+                     cannot be preserved in Zerocopy's generated helper enum",
+                )
+            }
+        }
+    }
+
+    validate_expr(discriminant)
+}
+
+pub(crate) fn validate_tag_enum_discriminants(data: &DataEnum) -> Result<(), Error> {
+    for variant in &data.variants {
+        if let Some((_, discriminant)) = &variant.discriminant {
+            validate_tag_enum_discriminant(discriminant)?;
+        }
+    }
+    Ok(())
+}
+
+fn tag_enum_discriminant_lint_attrs(attrs: &[syn::Attribute]) -> Vec<TokenStream> {
+    attrs
+        .iter()
+        .filter_map(|attr| {
+            // A lint level explicitly attached to the source enum or one of
+            // its variants applies to its discriminants, but not to the
+            // generated helper enum. Copy only source-provided levels which
+            // can permit the primitive expressions admitted by
+            // `validate_tag_enum_discriminant`.
+            //
+            // In particular, do not add an unconditional `allow`: lowering a
+            // lint which an enclosing scope forbids is itself an error, even
+            // when the copied discriminant would not trigger that lint.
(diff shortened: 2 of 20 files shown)


joshlf:
@codex review

joshlf:
@codex review

zerocopy/zerocopy-derive/src/util.rs
```diff
@@ -835,16 +835,303 @@ pub(crate) fn const_block(items: impl IntoIterator<Item = Option<TokenStream>>)
...
         }
-    });
+    }
+    Ok(())
+}
+
+fn tag_enum_discriminant_lint_attrs(ctx: &Ctx) -> Vec<TokenStream> {
+    ctx.ast
+        .attrs
+        .iter()
```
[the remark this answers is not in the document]
joshlf: Fixed. The existing narrow overflow-lint normalizer now runs independently on the source enum and on each source variant. Generated helper variants receive only normalized allowances for `overflowing_literals` and `arithmetic_overflow`; unrelated variant attributes are not replayed. The exact variant-scoped overflowing-literal case is covered on MSRV, stable, and nightly.

joshlf:
@codex review

joshlf:
We should be able to support this by scanning and normalizing any `Self` tokens that appear in the token stream, being careful to understand when they refer to the type in question and when they might be nested inside of other type definitions (e.g., inside a `[u8; { ... }]` where `...` is const code that itself contains a type definition and an `impl` block), and being careful to normalize raw vs non-raw identifiers as in the earlier PR. That will mean we don't have to reject the code that this PR currently rejects.

zerocopy/src/lib.rs
joshlf, line 1: Did you miss https://github.com/google/zerocopy/pull/3631#issuecomment-5593715738? Make sure to check top-level comments too.

joshlf:

I investigated the scope-aware rewrite. We can mechanically distinguish outer `Self` from nested item/impl scopes, normalize raw identifiers, and rewrite outer `Self` to a fresh alias of the source enum. Unfortunately, that fixes only name resolution; it does not make copying the discriminant sound. The helper still performs a second const evaluation (and potentially a second macro expansion) independently of the source enum.

A concrete accepted counterexample is:

```rust
#![feature(core_intrinsics)]
#![allow(internal_features)]
#[repr(u8)]
enum Packet {
    Flag(bool) = Self::tag(),
    Raw(u8),
}
impl Packet {
    const fn tag() -> u8 {
        core::intrinsics::is_val_statically_known(0) as u8
    }
}
```

The official intrinsic contract says this safe const fn nondeterministically returns `true` or `false`, that two identical calls need not agree, and that unsafe code may never rely on a particular answer. If the source enum gets 1 and the normalized helper gets 0, the real tags are `{Flag: 1, Raw: 2}` while the helper tags are `{Flag: 0, Raw: 1}`. Bytes with real tag 1 and an invalid `bool` payload are then checked as helper variant `Raw(u8)` and accepted. This recreates the soundness bug even though every `Self` resolves to `Packet`. The contract is documented in [`core::intrinsics::is_val_statically_known`](https://github.com/rust-lang/rust/blob/02c7f9bec0fd583160f8bcccb830216023b07bee/library/core/src/intrinsics/mod.rs#L2474-L2539); I also compile-checked this enum shape on that exact nightly.

The nested-item case has another independent problem: copying the block creates fresh nominal and anonymous types. Their `repr(Rust)` layouts need not agree across identities (I reproduced differing `size_of` results with randomized layout), while copied impls can also cause duplicate-definition/coherence errors. A macro is opaque to the derive before expansion and can emit either contextual `Self` or a different value on its second expansion.

I could not find a stable way around the second evaluation for data-bearing enums. The Reference specifies that [`mem::discriminant` is opaque and cannot yield the numeric value](https://doc.rust-lang.org/reference/items/enumerations.html#via-memdiscriminant); numeric casts are available only for [unit-only or otherwise qualifying fieldless enums](https://doc.rust-lang.org/reference/items/enumerations.html#casting). Reading a tag through the primitive-representation pointer technique requires an already-valid enum value, so it cannot bootstrap `TryFromBytes` validation.

For that reason, I think the current positive grammar is the necessary conservative boundary for this derive/helper architecture. We can add individual forms only when their repeatability is proved. Full support would require numeric discriminant reflection from the compiler or a source-owning macro/API that defines the source and validation tags from one canonical evaluation; a derive cannot retroactively rewrite the source enum. A separate fieldless-enum path could use casts of the original variants, but it would not solve the data-bearing case in #3619.

zerocopy/src/lib.rs
joshlf: *Authored by an AI agent acting on Josh Liebow-Feeser's behalf.*

I did miss the top-level comment on the earlier pass. I have now fetched the complete top-level conversation as well as inline threads, investigated the proposed normalization, and replied with the language-level counterexample here: https://github.com/google/zerocopy/pull/3631#issuecomment-5595317090

joshlf:
@codex review

Please review exact head `65f60cb70fe351add4707ada01320aa160a91a01`.

joshlf:
> _Authored by an AI agent acting on Josh Liebow-Feeser's behalf._
> 
> I investigated the scope-aware rewrite. We can mechanically distinguish outer `Self` from nested item/impl scopes, normalize raw identifiers, and rewrite outer `Self` to a fresh alias of the source enum. Unfortunately, that fixes only name resolution; it does not make copying the discriminant sound. The helper still performs a second const evaluation (and potentially a second macro expansion) independently of the source enum.
> 
> A concrete accepted counterexample is:
> 
> ```rust
> #![feature(core_intrinsics)]
> #![allow(internal_features)]
> #[repr(u8)]
> enum Packet {
>     Flag(bool) = Self::tag(),
>     Raw(u8),
> }
> impl Packet {
>     const fn tag() -> u8 {
>         core::intrinsics::is_val_statically_known(0) as u8
>     }
> }
> ```
> 
> The official intrinsic contract says this safe const fn nondeterministically returns `true` or `false`, that two identical calls need not agree, and that unsafe code may never rely on a particular answer. If the source enum gets 1 and the normalized helper gets 0, the real tags are `{Flag: 1, Raw: 2}` while the helper tags are `{Flag: 0, Raw: 1}`. Bytes with real tag 1 and an invalid `bool` payload are then checked as helper variant `Raw(u8)` and accepted. This recreates the soundness bug even though every `Self` resolves to `Packet`. The contract is documented in [`core::intrinsics::is_val_statically_known`](https://github.com/rust-lang/rust/blob/02c7f9bec0fd583160f8bcccb830216023b07bee/library/core/src/intrinsics/mod.rs#L2474-L2539); I also compile-checked this enum shape on that exact nightly.
> 
> The nested-item case has another independent problem: copying the block creates fresh nominal and anonymous types. Their `repr(Rust)` layouts need not agree across identities (I reproduced differing `size_of` results with randomized layout), while copied impls can also cause duplicate-definition/coherence errors. A macro is opaque to the derive before expansion and can emit either contextual `Self` or a different value on its second expansion.
> 
> I could not find a stable way around the second evaluation for data-bearing enums. The Reference specifies that [`mem::discriminant` is opaque and cannot yield the numeric value](https://doc.rust-lang.org/reference/items/enumerations.html#via-memdiscriminant); numeric casts are available only for [unit-only or otherwise qualifying fieldless enums](https://doc.rust-lang.org/reference/items/enumerations.html#casting). Reading a tag through the primitive-representation pointer technique requires an already-valid enum value, so it cannot bootstrap `TryFromBytes` validation.
> 
> For that reason, I think the current positive grammar is the necessary conservative boundary for this derive/helper architecture. We can add individual forms only when their repeatability is proved. Full support would require numeric discriminant reflection from the compiler or a source-owning macro/API that defines the source and validation tags from one canonical evaluation; a derive cannot retroactively rewrite the source enum. A separate fieldless-enum path could use casts of the original variants, but it would not solve the data-bearing case in #3619.

Is there some subset of `Self`-containing ASTs which we can accept? For example, IIUC, `Self::CONST` should be fine since it cannot invoke macro or function code, right? Help brainstorm the maximum such subset. We can reject the rest.

PR #3634, opened 2026-09-05
Preserve source bytes after failed try_transmute

### LINKED ISSUE #3632: Security Advisory: Failed transmutations can disclose data from a previous response

- **Vulnerability Type:** Uninitialized-memory read / information disclosure
- **Affected Software:** zerocopy
- **Date:** September 2026

## Summary

A failed `try_transmute!` call can return uninitialized bytes in place of the original input. The macro first moves the input through the destination type. If that type has padding, the move can discard bytes that the error path later treats as initialized input data.

A server that returns rejected input in an error response can consequently disclose data from a previous response. The reproduction below alternates a private response with a malformed 32-byte input. The malformed input contains only byte `2`, but its returned error contains fourteen bytes of the private response: `SERVER-SECRET!`. Miri independently reports undefined behavior while the library constructs that error.

The server must parse into a padded type and expose the rejected source bytes. The attacker supplies the malformed input; the private response belongs to an earlier authorized request. The example models this sequence in one process. The exact residual bytes depend on compiler optimization, target, and storage reuse.

**Affected version:** Confirmed with the published `zerocopy 0.8.56` crate. Earlier versions have not been exhaustively tested.

**Environment:** macOS arm64 (`aarch64-apple-darwin`), Rust 1.98.1 for the native build, and `nightly-2026-09-04` (`rustc 1.100.0-nightly`, `miri 0.1.0`) for Miri.

## Sketch of the attack

For the `WireRecord` defined below, a boolean value of `2` correctly fails validation. The error should preserve the supplied bytes:

```rust
let input = [2u8; 32];
let result: Result<WireRecord, _> = zerocopy::try_transmute!(input);
let response = result.unwrap_err().into_src();
```

Under the hood:

- `WireRecord` contains a `bool` followed by a `u128`. On the tested target, bytes 1 through 15 are padding.
- `try_transmute!` moves the byte array into storage typed as `MaybeUninit<ReadOnly<WireRecord>>`. A typed move need not preserve destination padding.
- Validation rejects the boolean. The error path moves that storage back into a byte array and calls `assume_init()`. Padding positions are now array elements, which must contain initialized integers.
- A native build can leave prior response data in those positions and return it to the caller.

## Full repro

```toml
[package]
name = "zerocopy-transmute-disclosure"
version = "0.1.0"
edition = "2024"

[workspace]

[dependencies]
zerocopy = { version = "=0.8.56", features = ["derive", "std"] }
```

Save the following as `src/bin/try_transmute_disclosure.rs`:

```rust
#![forbid(unsafe_code)]

use std::hint::black_box;
use zerocopy::TryFromBytes;

#[derive(Debug, TryFromBytes)]
#[repr(C)]
struct WireRecord {
    flag: bool,
    identifier: u128,
}

#[inline(never)]
fn reply(private: bool, input: [u8; 32]) -> [u8; 32] {
    if private {
        return *b"!SERVER-SECRET!0123456789abcdefg";
    }
    let result: Result<WireRecord, _> = zerocopy::try_transmute!(input);
    result.unwrap_err().into_src()
}

fn main() {
    let input = black_box([2u8; 32]);
    let mut replies = [[0u8; 32]; 4];
    for (index, saved) in replies.iter_mut().enumerate() {
        let output = black_box(reply(black_box(index % 2 == 0), input));
        *saved = output;
    }
    for (index, output) in replies.iter().enumerate() {
        println!("private={} output={output:02x?}", index % 2 == 0);
    }
    let leaked = std::str::from_utf8(&replies[1][1..15]).expect("ASCII fixture");
    println!("first unauthenticated error contains: {leaked}");
    assert_eq!(leaked, "SERVER-SECRET!");
}
```

The `private` flag represents the server's authorization decision, not attacker input. The loop calls the same handler first for an authorized response and then for a malformed request, allowing its return storage to be reused.

Run the native reproduction:

```sh
cargo +stable run --locked --release --bin try_transmute_disclosure
```

The final output line is:

```text
first unauthenticated error contains: SERVER-SECRET!
```

The assertion verifies those fourteen bytes. They were absent from the malformed input and came from the preceding private response.

Run the same program under Miri:

```sh
cargo +nightly-2026-09-04 miri run --locked --bin try_transmute_disclosure
```

Miri reports the error at `zerocopy/src/util/macro_util.rs:555`, before the caller receives the error value:

```text
error: Undefined Behavior: constructing invalid value of type [u8; 32]: at [1], encountered uninitialized memory, but expected an integer
555 |         Err(ValidityError::new(unsafe { mu_src.assume_init() }))
    |                                         ^^^^^^^^^^^^^^^^^^^^ Undefined Behavior occurred here
```

## Root cause

The source links refer to release revision `6dc429c451bdf1d7202ec1ec2cf426514e00d8eb`.

1. [`try_transmute`](https://github.com/google/zerocopy/blob/6dc429c451bdf1d7202ec1ec2cf426514e00d8eb/zerocopy/src/util/macro_util.rs#L525-L535) requires `Src: IntoBytes` and `Dst: TryFromBytes`. These bounds permit a padding-free source and a padded destination. It converts `MaybeUninit<Src>` to `MaybeUninit<ReadOnly<Dst>>` by value.
2. The [initialization assumption](https://github.com/google/zerocopy/blob/6dc429c451bdf1d7202ec1ec2cf426514e00d8eb/zerocopy/src/util/macro_util.rs#L539-L545) relies on the original source bytes still being initialized. Destination padding is not guaranteed to survive the preceding move.
3. The [failure branch](https://github.com/google/zerocopy/blob/6dc429c451bdf1d7202ec1ec2cf426514e00d8eb/zerocopy/src/util/macro_util.rs#L550-L555) reconstructs `Src` from the destination storage. Its justification that the storage was never modified does not account for padding lost during typed moves. `assume_init()` then requires initialization that the round trip did not preserve.
4. The standard library's [`MaybeUninit` validity documentation](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#validity) explicitly limits round-trip transmutation when initialized source bytes overlap padding in the intermediate type.

`[u8; 32]` correctly implements `IntoBytes`, and the destination's validator correctly rejects the input. This defect does not require a false padding witness or an incorrectly accepted `IntoBytes` derive. Repairs to those checks would leave this error-recovery path unchanged.

## Disclaimer

joshlf:
Validate the destination in place so a typed move cannot discard destination
padding before the original source is returned.

Closes #3632

---

- 👉 #3634

**Latest Update:** v5 — [Compare vs v4](/google/zerocopy/compare/gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v4..gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v5)

<details>

|Version| v4 | v3 | v2 | v1 |Base|
|:---|:---|:---|:---|:---|:---|
|v5|[vs v4](/google/zerocopy/compare/gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v4..gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v5)|[vs v3](/google/zerocopy/compare/gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v3..gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v5)|[vs v2](/google/zerocopy/compare/gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v2..gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v5)|[vs v1](/google/zerocopy/compare/gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v1..gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v5)|[vs Base](/google/zerocopy/compare/main..gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v5)|
|v4||[vs v3](/google/zerocopy/compare/gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v3..gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v4)|[vs v2](/google/zerocopy/compare/gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v2..gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v4)|[vs v1](/google/zerocopy/compare/gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v1..gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v4)|[vs Base](/google/zerocopy/compare/main..gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v4)|
|v3|||[vs v2](/google/zerocopy/compare/gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v2..gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v3)|[vs v1](/google/zerocopy/compare/gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v1..gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v3)|[vs Base](/google/zerocopy/compare/main..gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v3)|
|v2||||[vs v1](/google/zerocopy/compare/gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v1..gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v2)|[vs Base](/google/zerocopy/compare/main..gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v2)|
|v1|||||[vs Base](/google/zerocopy/compare/main..gherrit/Gosc7u5x37moad4qkehnrpfuziq62ctz3/v1)|

</details>
<details>

######

**Branch**
```bash
git fetch origin refs/heads/Gosc7u5x37moad4qkehnrpfuziq62ctz3 && git checkout -b pr-Gosc7u5x37moad4qkehnrpfuziq62ctz3 FETCH_HEAD
```

**Checkout**
```bash
git fetch origin refs/heads/Gosc7u5x37moad4qkehnrpfuziq62ctz3 && git checkout FETCH_HEAD
```

**Cherry Pick**
```bash
git fetch origin refs/heads/Gosc7u5x37moad4qkehnrpfuziq62ctz3 && git cherry-pick FETCH_HEAD
```

**Pull**
```bash
git pull origin refs/heads/Gosc7u5x37moad4qkehnrpfuziq62ctz3
```

</details>

### DIFF
--- a/zerocopy/src/macros.rs
+++ b/zerocopy/src/macros.rs
@@ -1471,6 +1471,19 @@ mod tests {
         // its `PanicOnDrop` temporary, which would cause a panic.
         assert_eq!(y.as_ref().map_err(|p| &p.src.0), Err::<&bool, _>(&2u8));
         mem::forget(y);
+
+        #[derive(Debug, TryFromBytes)]
+        #[repr(C, align(8))]
+        struct PaddedBool(bool);
+
+        // A failed transmutation must return every source byte, including
+        // bytes which would be padding in the destination type.
+        let src = [2u8, 1, 2, 3, 4, 5, 6, 7];
+        let x: Result<PaddedBool, _> = try_transmute!(src);
+        assert_eq!(x.unwrap_err().into_src(), src);
+
+        let x: Result<PaddedBool, _> = try_transmute!([0u8; 8]);
+        assert!(!x.unwrap().0);
     }
 
     #[test]
--- a/zerocopy/src/util/macro_util.rs
+++ b/zerocopy/src/util/macro_util.rs
@@ -28,7 +28,7 @@ use core::{marker::PhantomData, mem, num::Wrapping};
 
 use crate::{
     pointer::{
-        cast::CastSized,
+        cast::CastSizedExact,
         invariant::{Aligned, Initialized, Valid},
         BecauseImmutable,
     },
@@ -529,30 +529,64 @@ where
 {
     static_assert!(Src, Dst => mem::size_of::<Dst>() == mem::size_of::<Src>());
 
-    let mu_src = mem::MaybeUninit::new(src);
-    // SAFETY: `MaybeUninit` has no validity requirements.
-    let mu_dst: mem::MaybeUninit<ReadOnly<Dst>> =
-        unsafe { crate::util::transmute_unchecked(mu_src) };
-
-    let ptr = Ptr::from_ref(&mu_dst);
-
-    // SAFETY: Since `Src: IntoBytes`, and since `size_of::<Src>() ==
-    // size_of::<Dst>()` by the preceding assertion, all of `mu_dst`'s bytes are
-    // initialized. `MaybeUninit` has no validity requirements, so even if
-    // `ptr` is used to mutate its referent (which it actually can't be - it's
-    // a shared `ReadOnly` pointer), that won't violate its referent's validity.
-    let ptr = unsafe { ptr.assume_validity::<Initialized>() };
-    if Dst::is_bit_valid(ptr.cast::<_, CastSized, _>()) {
-        // SAFETY: Since `Dst::is_bit_valid`, we know that `ptr`'s referent is
-        // bit-valid for `Dst`. `ptr` points to `mu_dst`, and no intervening
-        // operations have mutated it, so it is a bit-valid `Dst`.
-        Ok(ReadOnly::into_inner(unsafe { mu_dst.assume_init() }))
+    let src = mem::ManuallyDrop::new(ReadOnly::new(src));
+    let mut ptr =
+        Ptr::from_ref(&*src).transmute_with::<ReadOnly<Dst>, Initialized, CastSizedExact, _>();
+    if Dst::is_bit_valid(ptr.reborrow_shared()) {
+        // SAFETY: `read_unaligned` requires that `ptr` be valid for reads and
+        // point to a properly initialized `ReadOnly<Dst>` [1]. `ptr` was
+        // derived from a live reference to `src`; `CastSizedExact` preserves
+        // that reference's address, provenance, and exact byte extent. Thus,
+        // the entire range remains live and valid for reads. The reference is
+        // non-null even when `ReadOnly<Dst>` is zero-sized, and the cast
+        // preserves that address, satisfying [1]'s explicit ZST requirement.
+        //
+        // `Src: IntoBytes` and the `ReadOnly<Src>` bridge let `transmute_with`
+        // establish that every byte in that exact range is initialized.
+        // `Dst::is_bit_valid` then established that the same bytes are
+        // bit-valid for `Dst`, and thus for `ReadOnly<Dst>`. The candidate was
+        // a shared `ReadOnly` pointer, so it could not mutate the bytes; no
+        // operation occurs between validation and this read which could do so.
+        // Alignment is not an additional precondition because [1] expressly
+        // permits unaligned pointers.
+        //
+        // `read_unaligned` leaves the source memory unchanged and makes a
+        // bitwise copy [1]. The source is wrapped in `ManuallyDrop`, which
+        // inhibits its destructor [2]. After this read, neither `src` nor
+        // `ptr` is used to access the inner value, and the wrapper's eventual
+        // drop does not drop it. Thus, only the returned `Dst` is subsequently
+        // used or dropped; there is no duplicate ownership use.
+        //
+        // [1] Per https://doc.rust-lang.org/1.56.0/std/ptr/fn.read_unaligned.html:
+        //
+        //     Reads the value from `src` without moving it. This leaves the
+        //     memory in `src` unchanged.
+        //
+        //     Unlike `read`, `read_unaligned` works with unaligned pointers.
+        //
+        //     Behavior is undefined if any of the following conditions are
+        //     violated:
+        //
+        //     * `src` must be valid for reads.
+        //
+        //     * `src` must point to a properly initialized value of type `T`.
+        //
+        //     Like `read`, `read_unaligned` creates a bitwise copy of `T`,
+        //     regardless of whether `T` is `Copy`. If `T` is not `Copy`, using
+        //     both the returned value and the value at `*src` can violate
+        //     memory safety.
+        //
+        //     Note that even if `T` has size `0`, the pointer must be non-null.
+        //
+        // [2] Per https://doc.rust-lang.org/1.56.0/std/mem/struct.ManuallyDrop.html:
+        //
+        //     A wrapper to inhibit compiler from automatically calling `T`'s
+        //     destructor.
+        let dst = unsafe { core::ptr::read_unaligned(ptr.as_inner().as_ptr()) };
+        Ok(ReadOnly::into_inner(dst))
     } else {
-        // SAFETY: `MaybeUninit` has no validity requirements.
-        let mu_src: mem::MaybeUninit<Src> = unsafe { crate::util::transmute_unchecked(mu_dst) };
-        // SAFETY: `mu_dst`/`mu_src` was constructed from `src` and never
-        // modified, so it is still bit-valid.
-        Err(ValidityError::new(unsafe { mu_src.assume_init() }))
+        let src = ReadOnly::into_inner(mem::ManuallyDrop::into_inner(src));
+        Err(ValidityError::new(src))
     }
 }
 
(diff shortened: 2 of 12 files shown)


joshlf:
@codex review

joshlf:
@codex review

joshlf (commented):
This looks completely sound. However, I wonder whether we could slightly simplify by taking a `Ptr::from_ref(&Src)` and then using `Ptr` transmute operations to produce the argument to `Dst::is_bit_valid` (as opposed to using a union). It's also possible that, since we need to do a by-value transmute in the success path anyway, that approach wouldn't actually save any logic or code.

joshlf:
@codex review

joshlf:

I adopted the suggested `Ptr`-based design in 30c0968168fb47dd40cfefa88cb5740f77dbc855. A literal `Ptr::from_ref(&Src)` is insufficient because `Src: IntoBytes` does not imply `Src: Immutable`, so the implementation places `ReadOnly<Src>` in `ManuallyDrop`, constructs a `Ptr` from that shared reference, and uses an exact-size `transmute_with` to obtain the candidate `ReadOnly<Dst>`. The valid branch performs one unaligned by-value read; the invalid branch extracts and returns the unchanged original `Src`.

This removes the bespoke union and most of its custom safety proof while preserving the required success and failure semantics. Nightly, stable, and MSRV UI tests, targeted regression tests, full nightly library tests, clippy, formatting, and an optimized-code comparison all pass.

@codex review

zerocopy/src/util/macro_util.rs
```diff
@@ -529,30 +529,64 @@ where
...
+        //     both the returned value and the value at `*src` can violate
+        //     memory safety.
+        //
+        //     Note that even if `T` has size `0`, the pointer must be non-null.
+        //
+        // [2] Per https://doc.rust-lang.org/1.56.0/std/mem/struct.ManuallyDrop.html:
+        //
+        //     A wrapper to inhibit compiler from automatically calling `T`'s
+        //     destructor.
+        let dst = unsafe { core::ptr::read_unaligned(ptr.as_inner().as_ptr()) };
```
joshlf, line 585: Could some combination of `Ptr` methods which convert to `Unalign` and transmute to `MaybeUninit` allow us to safely transmute to a `Ptr` to `ReadOnly<Unalign<MaybeUninit<Dst>>>` and thus allow a safe `Ptr::read`? That way the only remaining `unsafe` would be `MaybeUninit::assume_init`.

joshlf: Good thought. The current implementation already uses the `Ptr` part of this approach for the exact-size cast and validation, but the final safe `Ptr::read` step does not work for this API. `Ptr::read` requires its referent type to be `Copy`; `Unalign<T>` is `Copy` only when `T: Copy`, and `core::mem::MaybeUninit<Dst>` is `Copy` only when `Dst: Copy` (the crate's unsized-capable `MaybeUninit<Dst>` is not `Copy` at all). `try_transmute!` intentionally supports ownership-bearing, non-`Copy` destinations, so adding that bound would be an API regression.

There is also no ownership-transfer invariant in `Ptr` that would make a generic non-`Copy` move-out safe: the source storage would still appear to own the value. The current `ManuallyDrop` protocol supplies exactly that missing fact. After the `Ptr` cast and validation prove the bytes form a valid `ReadOnly<Dst>`, the one `read_unaligned` moves those bytes out without imposing `Copy`; `ManuallyDrop` prevents a second drop/use on success, while the error branch returns the original `Src` untouched. `Unalign`/`MaybeUninit` would add typestate transitions without eliminating that ownership-moving unsafe operation, so I think the existing implementation is the smaller and easier proof.

zerocopy/src/macros.rs
```diff
@@ -1471,6 +1471,43 @@ mod tests {
...
+        let src = [2u8, 1, 2, 3, 4, 5, 6, 7];
+        let x: Result<PaddedBool, _> = try_transmute!(src);
+        assert_eq!(x.unwrap_err().into_src(), src);
+
+        let x: Result<PaddedZero, _> = try_transmute!([0u8; 8]);
+        assert!(!x.unwrap().0);
+
+        let src = [0u8, 1, 2, 3, 4, 5, 6, 7];
+        let x: Result<PaddedZero, _> = try_transmute!(src);
+        assert_eq!(x.unwrap_err().into_src(), src);
```
jswrenn: Can we pose this test without manually implementing `TryFromBytes`? Either use the `derive`, or use `(u16, u8)` as the test case.

jswrenn (changes requested):

zerocopy/src/macros.rs
joshlf: Done—this now uses the derived `PaddedBool` for both success and failure. The failing case still verifies that every source byte corresponding to destination padding is returned unchanged.

joshlf:
@codex review

jswrenn (approved):

PR #3672, opened 2026-09-12
Fix `Ref::into_mut` internal error message

joshlf:
The unsized `Ref::into_mut` path reports `into_ref` if its internal cast
unexpectedly fails. Name the operation that actually failed so the diagnostic
points at the correct code path.

This is a one-line diagnostic-only change.


joshlf:
@codex review

PR #3673, opened 2026-09-12
Strengthen byteorder and split tests

joshlf:
Classify `Isize` as signed in the byteorder test helpers so its test-only
traits match the production type. Add deterministic coverage for numeric
ordering, valid shift counts, and floating-point bit-pattern preservation.

Also make the mutable `RefMut<[u8]>` split test verify writes through both
returned halves reach the backing slice.


joshlf:
@codex review

PR #3678, opened 2026-09-12
Add public constructors for SizeError and ValidityError

### LINKED ISSUE #3666: Add a way to construct Size Errors, etc
In a private project, I have functions that return things like `Result<T, TryReadError<(), T>>` from using `map_src`. I think the `zerocopy` errors are excellent enough that I want to reuse and forward them. These functions can fail partway through initializing the buffer for which `<T as TryFromBytes>::try_read_from_bytes` is called on. I have to go through this round about way of generating a size error if it fails and doesn't get enough size:
```
            // generate a size error using this trick because alignment errors
            // come first
            let mut buf: [O; 0] = [];
            let Err(e) = O::mut_from_bytes(buf.as_mut_bytes()).map_err(|e| e.map_src(drop)) else {
                unreachable!();
            };
            return Err(e);
```
Which is unnecessarily risky and may have weird edge cases in the future. I would rather be able to generate the size error directly.

barunaniket:
## Summary

Callers who know, out-of-band, that a conversion would fail due to incorrect source size or invalid source data currently have no way to construct a `SizeError` or `ValidityError` directly, and must instead perform a contrived failing conversion just to obtain one (see #3666).

This makes `SizeError::new` and `ValidityError::new` public so these errors can be constructed directly and forwarded via the existing `From` impls into `ConvertError`, `TryReadError`, etc.

`AlignmentError` is intentionally left as-is: its constructor is `unsafe` because of a soundness invariant (`Dst`'s alignment requirement must be greater than one, relied on elsewhere via `unreachable_unchecked`), so a safe public constructor would need a checked design (e.g. panicking or `Option`-returning) rather than a simple visibility change — that seems like a separate decision for maintainers to weigh in on.

Fixes #3666

## Test plan
- [x] `cargo test --doc` — new doc examples for both constructors compile and pass
- [x] `cargo test --lib` — full library test suite passes (112 tests)
- [x] `cargo test --doc` (full suite) — 128 doc tests pass
- [x] `cargo clippy` — no new warnings in `error.rs`
- [x] `cargo build --no-default-features` — no_std build succeeds
- [x] `rustfmt +nightly --check` on the changed file — no diff introduced by this change


barunaniket:
@joshlf This is ready for review whenever you get a chance — CI is green. Let me know if you'd like any changes.

Issue #1330, opened 2024-05-20
Support custom validators for `TryFromBytes`


kupiakos:
These are kinds of validity that users may need to have checked before transmutation from `&[u8]` to `&T`:

- The language-level validity of the bits for the type of each field in `T`, e.g. a `bool` must be either `0` or `1`. This is implemented by the `derive`.
- The library-level validity of the bits in `T`, e.g. an invariant that the first field is less than the second. This can be referenced by the `derive` but inherently must be user-controlled.
- The library-level validity of the individual fields in `T`, based on the above library-level check applied to each field. This is also implemented by the `derive`.
- The library-level validity of the length of the struct given the header contents, e.g. the `length` field is equal to the size of the tail slice. This only applies to dynamically sized structs ending in a slice (slice DSTs).

The plan discussed in #5 and #372 is to support the concept of a _custom validator_, a function or closure provided to `derive(TryFromBytes)` that will always be called before allowing a `TryFromBytes` transmute to succeed.

### Open Questions

- Should users be able to provide their own error type, to communicate the way in which the validation failed? How would this be exposed to the higher-level `TryFromBytes` APIs?
- How will data validation interact with custom DSTs? Is the validator also responsible for validating the correct length of the tail slice, or even deriving the correct length (as proposed in https://github.com/google/zerocopy/issues/1289#issuecomment-2121244709)?
  - What happens if the user provides both a [`#[length]`](https://github.com/google/zerocopy/issues/1328) and custom validator function and they disagree?
- What is the required signature(s) for the custom validators?

joshlf:
See also: #590

joshlf:
@jswrenn and I met and discussed potential designs. He has his own design proposal that he'll share at some point. Here's mine.

This design is based on the observation that some users may want to not only validate safety conditions, but actively transform their type - for example to construct a new witness wrapper type. It permits a distinction between the type being constructed and its "raw" equivalent.

This design is also explicitly meant to support custom length fields (#1289). In order to do that, validation can return a rich error message which might be required when a length field cannot be parsed.

We didn't have time during our discussion to think deeply about how this composes with a `#[length]` attribute. For example, is the length extracted before or after calling the user's custom validator? Can the `#[length]` attribute itself provide a "custom extractor" that also has error cases? These will need to be thought through.

```rust
// Used in the following examples
type MaybeValid<T, A> = Ptr<T, (A, Any, AsInitialized)>;
[...]

PR #3665, opened 2026-09-09
Rename `invariant::Valid` to `invariant::Safe`

jswrenn:
---

- 👉 #3665

**Latest Update:** v8 — [Compare vs v7](/google/zerocopy/compare/gherrit/G5fa632801df03c070c991b72a15721a1f2d66490/v7..gherrit/G5fa632801df03c070c991b72a15721a1f2d66490/v8)

<details>
<summary><strong>📚 Full Patch History</strong></summary>

*Links show the diff between the row version and the column version.*

|Version| v7 | v6 | v5 | v4 | v3 | v2 | v1 |Base|
|:---|:---|:---|:---|:---|:---|:---|:---|:---|
|v8|[vs v7](/google/zerocopy/compare/gherrit/G5fa632801df03c070c991b72a15721a1f2d66490/v7..gherrit/G5fa632801df03c070c991b72a15721a1f2d66490/v8)|[vs v6](/google/zerocopy/compare/gherrit/G5fa632801df03c070c991b72a15721a1f2d66490/v6..gherrit/G5fa632801df03c070c991b72a15721a1f2d66490/v8)|[vs v5](/google/zerocopy/compare/gherrit/G5fa632801df03c070c991b72a15721a1f2d66490/v5..gherrit/G5fa632801df03c070c991b72a15721a1f2d66490/v8)|[vs v4](/google/zerocopy/compare/gherrit/G5fa632801df03c070c991b72a15721a1f2d66490/v4..gherrit/G5fa632801df03c070c991b72a15721a1f2d66490/v8)|[vs v3](/google/zerocopy/compare/gherrit/G5fa632801df03c070c991b72a15721a1f2d66490/v3..gherrit/G5fa632801df03c070c991b72a15721a1f2d66490/v8)|[vs v2](/google/zerocopy/compare/gherrit/G5fa632801df03c070c991b72a15721a1f2d66490/v2..gherrit/G5fa632801df03c070c991b72a15721a1f2d66490/v8)|[vs v1](/google/zerocopy/compare/gherrit/G5fa632801df03c070c991b72a15721a1f2d66490/v1..gherrit/G5fa632801df03c070c991b72a15721a1f2d66490/v8)|[vs Base](/google/zerocopy/compare/main..gherrit/G5fa632801df03c070c991b72a15721a1f2d66490/v8)|
[...]

Issue #2762, opened 2025-11-10
Derive macros are unsound due to types being mentioned twice


theemathas:
This is a different unsoundness than https://github.com/google/zerocopy/issues/388.

The derive macros expand to code that mentions the field types, in order to validate that they implement the expected traits. This doesn't work if mentioning the same type name twice produces a different type. This can happen if the type is defined as a macro invocation that's nondeterministic.

The code below causes a segmentation fault. (Also [in a zip file](https://github.com/user-attachments/files/23456453/repro.zip) with all files necessary to reproduce, for your convenience.)

dep/src/lib.rs:
```rust
use proc_macro::TokenStream;
use std::sync::atomic::{AtomicI32, Ordering::SeqCst};

static COUNTER: AtomicI32 = AtomicI32::new(0);

#[proc_macro]
pub fn make_type(_: TokenStream) -> TokenStream {
    let index = COUNTER.fetch_add(1, SeqCst);
    match index {
        0..=2 => "i32",
        3 => "&'static i32",
        _ => panic!(),
    }
    .parse()
    .unwrap()
}
```

src/main.rs:
```rust
use dep::make_type;
use zerocopy::FromBytes;

#[derive(FromBytes)]
struct Thing {
    field: make_type!(),
}

fn main() {
    let thing: Thing = Thing::read_from_bytes(&1usize.to_ne_bytes()).unwrap();
    let field: &'static i32 = thing.field;
    println!("{field}");
}
```

<details>
<summary>Macro expansion</summary>

```rust
#![feature(prelude_import)]
#[macro_use]
extern crate std;
#[prelude_import]
use std::prelude::rust_2024::*;
use dep::make_type;
use zerocopy::FromBytes;

struct Thing {
    field: &'static i32,
}
#[allow(deprecated)]
#[automatically_derived]
unsafe impl ::zerocopy::TryFromBytes for Thing<> where
    i32: ::zerocopy::TryFromBytes {
    fn only_derive_is_allowed_to_implement_this_trait() {}
    fn is_bit_valid<___ZerocopyAliasing>(_candidate:
            ::zerocopy::Maybe<Self, ___ZerocopyAliasing>)
        -> ::zerocopy::util::macro_util::core_reexport::primitive::bool where     
        ___ZerocopyAliasing: ::zerocopy::pointer::invariant::Reference {
        if false {
            fn assert_is_from_bytes<T>() where T: ::zerocopy::FromBytes,
[...]

Issue #388, opened 2023-09-17
Could attributes cause unsoundness in our derives?


joshlf:
I haven't investigated enough to confirm that this is actually a problem, but it occurred to me that a few things might cause us problems:
- Does `#[cfg_attr(foo, repr(...)]` get evaluated before our derive? Would we observe a normal `#[repr(...)]` attribute as a result?
- Is it possible that attributes could be used - either new attributes added later to the language, or attributes defined as external proc macros - that change the semantics of the type after our code evaluates? E.g., imagine a hypothetical `#[add_trailing_field]` attribute that modifies the type's AST after we've already executed, and so the impls we emit don't "know about" the modified AST.

In the worst case, we might need to fail compilation if we detect attributes we don't recognize. Currently, proc macro derives do not have the ability to modify the type itself, so they should be fine (ie, `#[derive(UnknownTrait)]` isn't a problem), but we should make sure this is guaranteed by the language.

Another potential avenue is to consider the order of evaluation. However, it may be the case that we can just fail compilation if we detect an attribute, and suggest that the user reorder attributes - though not guaranteed, order of [evaluation is currently lexical](https://github.com/rust-lang/rust/issues/63336). Since it's not guaranteed, it would pose a stability hazard, but not a soundness hazard. We'd of course want to get the order of evaluation guaranteed as well to avoid the stability hazard.

djkoloski:
I verified that attribute macros can change the definition of an item _after_ derives on it have already run:

```rust
use transform::insert_field;

trait A {}
trait B {}

#[derive(Debug)]
#[insert_field(baz = "u32")]
struct Foo {
    bar: i32,
}

fn main() {
    println!("{:?}", Foo {
        bar: 10,
        baz: 100,
    });
}
```

prints:

```
Foo { bar: 10 }
```

Which demonstrates that even though `Foo` had an extra `baz` field inserted via attribute proc-macro, the item gets passed to the `Debug` derive before that modification takes place. Expanding the source with `cargo expand` also demonstrates the issue. For completeness, the source of the `insert_field` proc-macro is:

```rust
use proc_macro2::{Ident, Span};
use quote::quote;
use syn::{parse_macro_input, DeriveInput, Meta, Lit, Expr, Visibility, token::Colon, Data, Fields, Field, parse_quote};

#[proc_macro_attribute]
pub fn insert_field(
    attr: proc_macro::TokenStream,
    item: proc_macro::TokenStream,
) -> proc_macro::TokenStream {
[...]

PR #3681, opened 2026-09-13
Introduce unstable `#[zerocopy(invariant(...))]` library invariant checking

jswrenn:
Makes progress on #1330.

---

- 　  #3682
- 👉 #3681

**Latest Update:** v7 — [Compare vs v6](/google/zerocopy/compare/gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v6..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v7)

<details>

|Version| v6 | v5 | v4 | v3 | v2 | v1 |Base|
|:---|:---|:---|:---|:---|:---|:---|:---|
|v7|[vs v6](/google/zerocopy/compare/gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v6..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v7)|[vs v5](/google/zerocopy/compare/gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v5..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v7)|[vs v4](/google/zerocopy/compare/gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v4..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v7)|[vs v3](/google/zerocopy/compare/gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v3..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v7)|[vs v2](/google/zerocopy/compare/gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v2..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v7)|[vs v1](/google/zerocopy/compare/gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v1..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v7)|[vs Base](/google/zerocopy/compare/main..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v7)|
|v6||[vs v5](/google/zerocopy/compare/gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v5..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v6)|[vs v4](/google/zerocopy/compare/gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v4..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v6)|[vs v3](/google/zerocopy/compare/gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v3..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v6)|[vs v2](/google/zerocopy/compare/gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v2..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v6)|[vs v1](/google/zerocopy/compare/gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v1..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v6)|[vs Base](/google/zerocopy/compare/main..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v6)|
|v5|||[vs v4](/google/zerocopy/compare/gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v4..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v5)|[vs v3](/google/zerocopy/compare/gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v3..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v5)|[vs v2](/google/zerocopy/compare/gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v2..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v5)|[vs v1](/google/zerocopy/compare/gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v1..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v5)|[vs Base](/google/zerocopy/compare/main..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v5)|
|v4||||[vs v3](/google/zerocopy/compare/gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v3..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v4)|[vs v2](/google/zerocopy/compare/gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v2..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v4)|[vs v1](/google/zerocopy/compare/gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v1..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v4)|[vs Base](/google/zerocopy/compare/main..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v4)|
|v3|||||[vs v2](/google/zerocopy/compare/gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v2..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v3)|[vs v1](/google/zerocopy/compare/gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v1..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v3)|[vs Base](/google/zerocopy/compare/main..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v3)|
|v2||||||[vs v1](/google/zerocopy/compare/gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v1..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v2)|[vs Base](/google/zerocopy/compare/main..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v2)|
|v1|||||||[vs Base](/google/zerocopy/compare/main..gherrit/G2cef62dbb236b18953ded9a3d514bf0913c0e49c/v1)|

</details>
<details>

######

**Branch**
```bash
git fetch origin refs/heads/G2cef62dbb236b18953ded9a3d514bf0913c0e49c && git checkout -b pr-G2cef62dbb236b18953ded9a3d514bf0913c0e49c FETCH_HEAD
```

**Checkout**
```bash
git fetch origin refs/heads/G2cef62dbb236b18953ded9a3d514bf0913c0e49c && git checkout FETCH_HEAD
```

**Cherry Pick**
```bash
git fetch origin refs/heads/G2cef62dbb236b18953ded9a3d514bf0913c0e49c && git cherry-pick FETCH_HEAD
```

**Pull**
```bash
git pull origin refs/heads/G2cef62dbb236b18953ded9a3d514bf0913c0e49c
```

</details>


zerocopy/src/lib.rs
```diff
@@ -1708,6 +1708,53 @@ pub unsafe trait Immutable {
...
+
+Fields are checked in declaration order. Each field's bit validity is
+checked before its invariants run. Multiple invariants on a field run in
+attribute order. For structs and enums, validation stops at the first invalid
+field or invariant that returns `false`. For unions, a failed bit-validity
+check or invariant causes validation to try the next field; validation
+succeeds as soon as one field and all its invariants pass. Each expression
+runs in its own closure; `return` returns from that expression, and panics
+propagate to the caller.
+Only the selected enum variant's fields and invariants are checked.
```
joshlf, line 1747:
```suggestion
propagate to the caller. Only the selected enum variant's fields and
invariants are checked.
```

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
```diff
@@ -16,29 +16,94 @@ use crate::{
...
+                        return false;
+                    }
+                )*
+            })
+        })
+        .collect::<Result<Vec<_>, Error>>()?;
+    Ok(quote! {
+        #[allow(unused_imports)]
+        use #core::mem::drop as #candidate;
+        let #candidate = candidate;
```
joshlf, line 76: Is this necessary?

joshlf (changes requested):

joshlf (commented):
Test-coverage pass: the current suite covers the main container kinds, field ordering and short-circuiting, `return` containment, basic DST bit-validity, generics, non-`Immutable` union fields, experimental-cfg gating, and several hygiene cases. I would add the following before treating the feature as exhaustively covered:

1. **Panic propagation.** Add a struct test whose invariant panics and assert that `catch_unwind` observes the panic. Also add a union case whose first field invariant panics while a later field would pass, to prove that union fallback does not turn a panic into `false` and continue.
2. **An invariant on the DST field itself.** `Unsized` currently puts the invariant on `a`; the `[bool]` tail only exercises bit validation. Put an invariant on `b` that inspects its metadata/elements (and preferably also `a`) so the user-visible `Ptr` for a dynamically sized current field is exercised.
3. **Nested invariant-bearing fields.** Have `Outer` contain an `Inner` with its own invariant, then verify that an `Inner` invariant failure prevents `Outer`'s field invariant and later-field validation from running. This locks in compositional validation through `ReadOnly<T>::TryFromBytes` rather than testing only primitive field validators.
4. **Mutable conversion semantics.** Exercise an invariant-bearing `IntoBytes` type end-to-end through `try_mut_from_bytes`: invalid input should be rejected, valid input should be accepted, and the returned value should then be mutable into a state that violates the invariant. That last case is part of the documented contract that invariants validate conversions rather than constrain later mutation.
5. **A multi-variant `repr(C)` enum.** `CEnum` has only one fieldful variant. Add two fieldful variants with different predicates so the C-layout variant-to-field projection path is tested independently of the existing multi-variant primitive-repr enum.
6. **ZST/overlapping projections.** Add a named zero-sized field with an invariant before another field, so multiple retained shared `Safe` field projections can have the same address. This would be a useful Miri regression case for the new behavior of keeping prior field pointers in scope.

I am not repeating the existing review threads for `most_traits`, skipped experimental-cfg emission, or generated-name hygiene; those already identify separate missing cases.

*Authored by an AI agent acting on Josh's behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
```diff
@@ -16,29 +16,94 @@ use crate::{
...
+        {
+            #(
+                // Prevent an outer constant or unit struct with this name from
+                // turning the binding below into a path pattern.
+                #[allow(unused_imports)]
+                use #core::mem::drop as #field_bindings;
```
joshlf, line 82: These poison imports are block-scoped items, so each import is in scope from the **start** of this block, not only from its textual position. As a result, names for later fields affect earlier invariants even though the documented DSL exposes only the current and preceding fields.

For example, with an outer `fn b(_: bool) { panic!() }`, an invariant on `a` such as `{ b(false); true }`, and a later field named `b`, the generated `use core::mem::drop as b` makes that earlier call resolve to `drop(false)`. The invariant silently returns `true` instead of propagating the intended panic. This is stronger than a diagnostic-only collision: generated syntax changes the executed predicate.

Please arrange the lowering so a field name enters the invariant's lexical scope only after that field has been validated. In particular, avoid block-wide item aliases for future fields; add a regression where an earlier invariant refers to an outer helper whose name matches a later field.

*Authored by an AI agent acting on Josh's behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
```diff
@@ -16,29 +16,94 @@ use crate::{
...
+        field
+            .ident
+            .as_ref()
+            .map(|name| name.to_token_stream())
+            .unwrap_or_else(|| syn::Index::from(idx).to_token_stream())
+    });
+    let field_bindings = fields.iter().enumerate().map(|(idx, field)| {
+        field
+            .ident
+            .clone()
```
joshlf, line 56: The field-name DSL should not depend on the source identifier's syntax context. Cloning `field.ident` also clones its hygiene context, but the same-spelling identifier inside `#[zerocopy(invariant(...))]` can have a different context when an enclosing macro constructs the item. In that case the invariant token is not guaranteed to resolve to this generated local; it can fail to bind here or resolve to another same-spelling call-site name.

This repository already has `__test_hygienically_mixed_into_bytes` specifically to construct identically printed `def_site`/`call_site` identifiers. Please add the analogous adversarial fixture for an invariant-bearing named field, with the field declaration and invariant reference deliberately given different contexts, and lower field references according to the invariant DSL's semantic field names rather than assuming their original spans make ordinary local-variable lookup work.

*Authored by an AI agent acting on Josh's behalf.*

zerocopy/zerocopy-derive/src/invariant.rs
```diff
@@ -0,0 +1,66 @@
...
+pub(crate) fn parse(attrs: &[Attribute]) -> Result<Vec<Expr>, Error> {
+    let mut invariants = Vec::new();
+    for attr in attrs.iter().filter(|attr| path_is_ident(attr.path(), "zerocopy")) {
+        let options = attr.parse_args_with(Punctuated::<Meta, Token![,]>::parse_terminated)?;
+        for option in options {
+            if !path_is_ident(option.path(), "invariant") {
+                return Err(Error::new_spanned(option, "expected `invariant(...)`"));
+            }
+            match option {
+                Meta::List(list) => invariants.push(list.parse_args::<Expr>()?),
```
joshlf, line 27: `Meta::List` does not imply parentheses; Syn preserves `()`, `[]`, and `{}` as distinct `MacroDelimiter`s. This therefore accepts `#[zerocopy(invariant[true])]` and `#[zerocopy(invariant{true})]`, even though the public syntax and this parser's own diagnostics specify `invariant(...)`.

If parentheses are the intended grammar, require `MacroDelimiter::Paren` here and add brace/bracket UI cases. Otherwise the broader grammar should be documented and tested explicitly rather than accepted accidentally by the AST representation.

*Authored by an AI agent acting on Josh's behalf.*

joshlf (commented):
Generated-syntax audit on exact head `6a2d740b5d810f435af98a5cd18a0a989a02ebd4`. I traced the new invariant expression lowering through the projection generator, generic/enum helper types, identifier hygiene, attribute grammar, and the accepted `syn` AST surface. I am not duplicating the known repository-wide derive unsoundness tracked by #2762 and #388; those risks predate this PR.

*Authored by an AI agent acting on Josh's behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
```diff
@@ -16,29 +16,94 @@ use crate::{
...
+                // Disambiguate the available `TryTransmuteFromPtr` impls.
+                #[allow(unused_variables)]
+                let #field_bindings = match #field_bindings.try_into_safe::<
+                    _,
+                    #zerocopy_crate::BecauseImmutable,
+                >() {
+                    #core::result::Result::Ok(#field_bindings) => #field_bindings,
+                    #core::result::Result::Err(_) => return false,
+                };
+                #field_validations
```
joshlf, line 102: Caller-authored invariant code is inserted under `ImplBlockBuilder`'s derive-wide `const_block`, which carries `#[allow(deprecated, non_snake_case, non_local_definitions, ...)]`. Unlike ordinary generated glue, the invariant expression is executable source supplied by the caller and is only semantically checked after this expansion. The derive can therefore silently weaken the caller's lint policy—for example, a crate with `#![deny(deprecated)]` can call a deprecated helper solely from `#[zerocopy(invariant(...))]`, while the enclosing `allow(deprecated)` suppresses that error.

Please keep caller-authored invariant expressions outside the blanket generated-code lint suppressions, or narrow those suppressions so they do not cover the expression. Add a UI regression with at least `deny(deprecated)`.

*Authored by an AI agent acting on Josh's behalf.*

joshlf (commented):
Independent bug-finding pass: one additional issue found below. I avoided repeating the existing `most_traits`, feature-gate, generated-name, future-field-scope, hygiene, and delimiter findings.

*Authored by an AI agent acting on Josh's behalf.*

joshlf (commented):
Round 2 found one additional behavioral-contract issue. I am not repeating any previously reported findings.

The new invariant expressions are arbitrary executable user code. They may mutate external state as well as panic; for example, an invariant can increment an `AtomicUsize` and return `true`. However, the existing `TryFromBytes` conversion entry points are still annotated `#[must_use = "has no side effects"]`. That diagnostic is now false for invariant-bearing types: merely calling `try_ref_from_bytes`, `try_mut_from_bytes`, `try_read_from_bytes`, and the prefix/suffix variants can execute user side effects even when their result is discarded.

Please update the affected `TryFromBytes` `must_use` reasons (or otherwise restrict/document invariant effects if purity is intended). A regression with a side-effecting invariant would also make the changed behavior explicit. The new field-invariant documentation should mention arbitrary side effects if they are intentionally supported, alongside the existing panic behavior.

*Authored by an AI agent acting on Josh's behalf.*

joshlf (commented):
A second new round-2 issue is in test integration rather than product semantics.

The new positive/runtime test target `zerocopy-derive/tests/invariant.rs` is guarded wholesale by `#![cfg(zerocopy_unstable_ptr)]`. CI runs tests through `./cargo.sh`, whose `get_rustflags` currently injects `zerocopy_unstable_linux`, `zerocopy_derive_union_into_bytes`, and internal cfgs, but **not** `zerocopy_unstable_ptr`. The workflow-level `RUSTFLAGS` is only `-Dwarnings`; `zerocopy_unstable_ptr` appears in `RUSTDOCFLAGS`, which does not enable this test target for `cargo test`. Thus the main positive invariant tests compile as an empty integration target in ordinary CI, while only the separately driven UI cases exercise the feature.

Please add `--cfg zerocopy_unstable_ptr` to the test wrapper's enabled experimental cfg set (or otherwise run this integration target explicitly with the cfg) so these runtime tests actually execute in CI. A small assertion/check that the target contains or runs at least one invariant test would make this harder to regress silently.

*Authored by an AI agent acting on Josh's behalf.*

zerocopy/src/lib.rs
```diff
@@ -1708,6 +1708,53 @@ pub unsafe trait Immutable {
...
+
+Invariants are not supported on tuple fields. Types with
+invariants cannot derive [`FromZeros`] or [`FromBytes`], whose conversions
+do not perform runtime validation. These checks apply to conversions
+through [`TryFromBytes`]; they do not restrict ordinary construction or
```
joshlf, line 1752: The existing `TryFromBytes` documentation defines a “valid instance” in terms of Rust bit validity, and the conversion methods say that if the source bytes are a valid instance they return `Ok`. Field invariants make that promise false unless “valid instance” is explicitly broadened: for example, `struct Foo { #[zerocopy(invariant(false))] a: u8 }` is bit-valid for every `u8`, but every `TryFromBytes` conversion rejects it.

Please update the controlling `TryFromBytes`/conversion documentation to distinguish Rust bit validity from these additional user-defined acceptance predicates—either redefine the relevant notion of validity to include configured invariants, or state that bit-valid inputs may still be rejected when an invariant returns `false`. The new section should link that distinction back to the existing “What is a valid instance?” text so the public contract is internally consistent.

*Authored by an AI agent acting on Josh's behalf.*

joshlf (commented):
Round 3 found one additional public-contract/documentation issue. I am not repeating any previously reported finding.

*Authored by an AI agent acting on Josh's behalf.*

zerocopy/zerocopy-derive/src/derive/from_bytes.rs
```diff
@@ -98,6 +104,12 @@ pub(crate) fn derive_from_zeros(ctx: &Ctx, top_level: Trait) -> Result<TokenStre
     Ok(IntoIterator::into_iter([try_from_bytes, from_zeros]).collect())
 }
 pub(crate) fn derive_from_bytes(ctx: &Ctx, top_level: Trait) -> Result<TokenStream, Error> {
+    if let Some(span) = ctx.invariant_span {
+        return ctx.error_or_skip(Error::new(
+            span,
+            "cannot derive `FromBytes` for a type with invariants",
+        ));
```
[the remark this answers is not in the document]
jswrenn: Confirmed. `most_traits` now explicitly derives `TryFromBytes` for invariant-bearing inputs while continuing to skip `FromBytes` and `FromZeros`. Added struct, enum, and union regressions that check trait availability and acceptance/rejection by the invariant.

Implemented in [32a19cb2f](https://github.com/google/zerocopy/commit/32a19cb2f9451a680e866c8bef6a2382a3346694) (originally prepared as local amend `c1dc32796152180aa0bd66465fba53648d12f3ec`).

Tests: With `RUSTFLAGS="--cfg zerocopy_unstable_ptr"`, `./cargo.sh +all test -p zerocopy-derive --test invariant --test hygiene --test on_error` passed on MSRV, stable, and nightly (16 invariant tests and the existing hygiene regression on each; the on-error target provides compile-time assertions). The complete nightly derive suite and the repository pre-push checks also passed.

Update: the fix is now present in the published PR head linked above. The tests reported here ran on the original local amend; the derive sources and tests are unchanged in the published commit. Leaving this thread unresolved as requested.

*Authored by Codex, an AI agent acting on Jack Wrenn’s behalf.*

zerocopy/zerocopy-derive/src/util.rs
```diff
@@ -197,7 +201,24 @@ impl Ctx {
...
         } else {
             quote!()
-        }
+        };
+        let invariant = if cfg!(zerocopy_unstable_ptr) {
+            quote!()
+        } else if let Some(span) = self.invariant_span {
+            let core = self.core_path();
+            let error_message =
+                "`invariant` is experimental; pass '--cfg zerocopy_unstable_ptr' to enable";
```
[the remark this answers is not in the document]
jswrenn: Confirmed. The unsupported-enum exit now uses `ctx.error_or_skip(e)`, preserving the experimental cfg diagnostics instead of returning an empty token stream. Added a no-repr enum to the cfg UI test and an enabled-cfg compile-time regression verifying that its unsupported derive is still skipped.

Implemented in [32a19cb2f](https://github.com/google/zerocopy/commit/32a19cb2f9451a680e866c8bef6a2382a3346694) (originally prepared as local amend `c1dc32796152180aa0bd66465fba53648d12f3ec`).

Tests: `./cargo.sh +all test -p zerocopy-derive --test ui` passed on MSRV, stable, and nightly, including the new missing-cfg diagnostics. Regenerated expectations with `../tools/update-expected-test-output.sh`. The complete nightly derive suite with `zerocopy_unstable_ptr` and the repository pre-push checks also passed.

Update: the fix is now present in the published PR head linked above. The tests reported here ran on the original local amend; the derive sources and tests are unchanged in the published commit. Leaving this thread unresolved as requested.

*Authored by Codex, an AI agent acting on Jack Wrenn’s behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
```diff
@@ -16,29 +16,94 @@ use crate::{
...
+            .unwrap_or_else(|| format_ident!("field_{}", idx, span = Span::mixed_site()))
+    });
+    let field_validations = fields
+        .iter()
+        .map(|field| {
+            let invariants = crate::invariant::parse(&field.attrs)?;
+            Ok(quote! {
+                #(
+                    // Keep `return` in an invariant from bypassing later fields.
+                    if !{ #[inline(always)] || -> #core::primitive::bool { #invariants } }() {
```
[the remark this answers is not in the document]
jswrenn: Confirmed. The generated candidate parameter and enum tag now use mixed-site identifiers, and the extra candidate import/rebinding has been removed. Candidate spelling still avoids field names so field imports cannot shadow the source pointer. Added struct, enum, and union regressions calling outer helpers named `candidate`, `candidate_`, and `tag`, while retaining the existing colliding-field-name tests.

Implemented in [32a19cb2f](https://github.com/google/zerocopy/commit/32a19cb2f9451a680e866c8bef6a2382a3346694) (originally prepared as local amend `c1dc32796152180aa0bd66465fba53648d12f3ec`).

Tests: The name-collision regression failed against the original implementation and now passes. With `RUSTFLAGS="--cfg zerocopy_unstable_ptr"`, the invariant and hygiene targets passed on MSRV, stable, and nightly (16 invariant tests plus the existing hygiene regression per toolchain). The complete nightly derive suite, regenerated output snapshots, nightly Clippy, and repository pre-push checks also passed.

Update: the fix is now present in the published PR head linked above. The tests reported here ran on the original local amend; the derive sources and tests are unchanged in the published commit. Leaving this thread unresolved as requested.

*Authored by Codex, an AI agent acting on Jack Wrenn’s behalf.*

zerocopy/src/lib.rs
jswrenn: I missed this formatting suggestion in the earlier pass. The current PR still splits these two sentences across separate source lines, so this was not applied. Your proposed wrapping preserves the wording and is appropriate; leaving this thread open. No code or test changes were made for this comment.

*Authored by Codex, an AI agent acting on Jack Wrenn’s behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
jswrenn: The rebinding is no longer needed with the revised lowering. It and the accompanying candidate import were removed in [32a19cb2f](https://github.com/google/zerocopy/commit/32a19cb2f9451a680e866c8bef6a2382a3346694). The method parameter now uses a mixed-site identifier whose spelling avoids field names, and field projections use that parameter directly.

The invariant and hygiene regressions passed on MSRV, stable, and nightly with `RUSTFLAGS="--cfg zerocopy_unstable_ptr"`; the full nightly derive suite also passed. Those tests ran on the original local amend `c1dc32796`; the derive sources and tests are unchanged in the published commit linked above. Leaving your thread unresolved for your review.

*Authored by Codex, an AI agent acting on Jack Wrenn’s behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
jswrenn: This remains unaddressed. The current generator still emits all field-name `use ... as ...` items in one block. Removing the separate candidate alias only addresses the generated candidate-name collision; it does not prevent an import for a later field from affecting an earlier invariant. The field imports need narrower scopes, with a regression for an earlier invariant calling a helper named after a later field. The existing name-collision tests do not cover that case, so their passing results do not establish that this concern is fixed. Leaving this thread open.

*Authored by Codex, an AI agent acting on Jack Wrenn’s behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
jswrenn: This remains unaddressed. The generator still clones `field.ident` for named field bindings. The earlier fix gives generated candidate/tag locals mixed-site identifiers, but it does not change the relationship between a caller-supplied field declaration’s syntax context and identifiers in its invariant. I have not added or run the requested adversarial mixed-context fixture, so I cannot claim that the current field-name binding behavior handles it. Leaving this separate hygiene concern open.

*Authored by Codex, an AI agent acting on Jack Wrenn’s behalf.*

zerocopy/zerocopy-derive/src/invariant.rs
jswrenn: The parser still accepts `Meta::List` without inspecting its delimiter, while its diagnostics and public examples specify `invariant(...)`. The previous fix did not change this. I would enforce the documented parentheses grammar with a `MacroDelimiter::Paren` check and brace/bracket UI regressions; accepting the broader grammar instead needs an explicit documentation decision. No delimiter regression was added or run in the earlier pass. Leaving this thread open.

*Authored by Codex, an AI agent acting on Jack Wrenn’s behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
jswrenn: This remains unaddressed. The invariant expressions are still emitted inside the impl wrapped by `const_block`, and that wrapper still carries `allow(deprecated)` and the other broad lint allowances. The candidate-name hygiene fix did not narrow those attributes or move caller expressions outside them. The requested `deny(deprecated)` UI regression was not part of the prior test run; narrowing generated-code lint suppression still needs a separate fix. Leaving this thread open.

*Authored by Codex, an AI agent acting on Jack Wrenn’s behalf.*

zerocopy/src/lib.rs
jswrenn: The documentation gap remains. The trait already warns that `TryFromBytes` need not accept every byte sequence produced by `IntoBytes`, and the new invariant section explains the additional checks. However, “What is a valid instance?” still discusses Rust bit validity, and the conversion-method wording does not explicitly connect that definition to invariant-based rejection. The controlling docs should state that bit-valid input may still be rejected by a configured invariant and link that distinction to the invariant section. The previous fixes did not update these docs or add a regression specifically for this contract clarification. Leaving this thread open.

*Authored by Codex, an AI agent acting on Jack Wrenn’s behalf.*

zerocopy/src/lib.rs
joshlf: Please apply this – we need to reflow everything consistently.

zerocopy/src/lib.rs
joshlf: Maybe we should update the `TryFromBytes` docs to refer to safety more generally? We've recently done something similar internally: #3665

joshlf (changes requested):

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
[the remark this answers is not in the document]
joshlf: Follow-up on current head `32a19cb2f`: the `candidate` parameter and local enum `tag` are now hygienic, but invariant expressions still sit inside scopes containing other implementation-only call-site names.

`is_safe` still declares the literal type parameter `___ZcAlignment`, so an invariant that intends an outer `___ZcAlignment` type resolves to the derive's generic parameter instead. Enums have the same problem with generated items: `generate_tag_consts` creates `___ZEROCOPY_TAG_<Variant>` using `variant.span()`. For example, an outer `fn ___ZEROCOPY_TAG_A() -> bool { true }` plus `#[zerocopy(invariant(___ZEROCOPY_TAG_A()))]` on variant `A` resolves to the generated tag constant rather than the caller's function.

Please freshen or otherwise isolate every implementation-only identifier visible from an invariant expression, not just these two local bindings, and add regressions for at least a method-generic collision and an enum tag-constant collision. `Span::mixed_site()` is sufficient for the local bindings, but it should not be assumed to hide item or type names.

*Authored by an AI agent acting on Josh's behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
joshlf: The implication here is stronger than a name-resolution nuisance: this can undermine the safety purpose of `invariant(...)`.

A caller can use an invariant to establish a library invariant that later safe methods rely on when they execute `unsafe` internally. If a later field's generated `use ... as ...` changes what an earlier invariant expression calls, the derive can evaluate a different predicate from the one the author wrote. That can let `TryFromBytes` accept bytes that violate the intended library invariant. A later safe method may then reach `unsafe` code with its preconditions false.

So I think this is a real soundness issue, not just a hygiene or diagnostics issue. The fix needs to ensure that later fields cannot affect name resolution in earlier invariants.

*Authored by an AI agent acting on Josh's behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
joshlf: The implication is similar, but the evidence is weaker until we have the adversarial macro test.

The feature promises that a field name inside `invariant(...)` refers to that field's validated pointer. If macro hygiene lets the invariant token and the generated binding have different syntax contexts, that promise may fail: the expression can fail to bind to the field or bind to some other same-spelling name.

If it merely fails to compile, this is a usability bug. If it resolves to another value and the predicate returns `true`, it becomes soundness-relevant for the same reason as the later-field collision: `TryFromBytes` can accept bytes without actually checking the library invariant that downstream `unsafe` code relies on.

I would therefore treat this as a soundness blocker until the mixed-context fixture shows which behavior rustc actually gives us.

*Authored by an AI agent acting on Josh's behalf.*

zerocopy/zerocopy-derive/src/invariant.rs
joshlf: The practical implication here is API/grammar consistency, not memory safety by itself.

Today the parser accepts `invariant(...)`, `invariant[...]`, and `invariant{...}` because all three arrive as `Meta::List`, even though the documentation and diagnostics describe only the parenthesized form. Once accepted, they all parse the enclosed tokens as the same Rust expression, so this does not appear to change the safety check that runs.

The risk is that we accidentally make undocumented syntax part of the de facto interface, then have to support it or break users later. I would still fix or explicitly document it, but I would not block this PR on soundness grounds for this item alone.

*Authored by an AI agent acting on Josh's behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
joshlf: The implication here is that caller-written executable code is being compiled under a different lint policy than the caller requested.

For example, `#![deny(deprecated)]` normally makes a deprecated call a hard error, but the derive's surrounding `#[allow(deprecated)]` can make the same call compile when it appears inside `invariant(...)`. The same principle applies to the other blanket allowances: they were intended for generated glue, but they also cover user code copied into that glue.

I do not see a direct memory-safety hole from this alone. These attributes suppress lints, not Rust's type system, validity rules, or unsafe-context checks. The concrete problem is that the macro can silently defeat project policy and admit code the caller explicitly asked rustc to reject. I would fix it, but classify it separately from the soundness blockers above.

*Authored by an AI agent acting on Josh's behalf.*

zerocopy/src/lib.rs
joshlf: The implication here is a broken public contract, even if the implementation remains memory-safe.

The current docs define a "valid instance" in terms of Rust bit validity, then say the conversion succeeds when the source contains such an instance. With field invariants, that is no longer true: bytes can be fully bit-valid and still be rejected because an invariant returns `false`.

That matters because users should be able to reason from the documented contract, including in unsafe code. Returning `Err` for a bit-valid value does not itself create UB, but a false documented guarantee can invalidate downstream reasoning that relied on the stated success condition. The docs should therefore describe the actual acceptance condition explicitly rather than overloading "valid instance" to mean two different things.

*Authored by an AI agent acting on Josh's behalf.*

zerocopy/src/lib.rs
joshlf: No semantic implication here. This is only source formatting: the proposed reflow preserves the exact wording and behavior.

It is worth fixing for consistency, but it does not affect the generated code, API contract, or soundness argument.

*Authored by an AI agent acting on Josh's behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
joshlf: I prototyped a minimal witness for the soundness consequence here. The key is that the later field can suppress the check that a safe method later relies on:

```rust
fn require_nonzero(ok: bool) {
    assert!(ok);
}

#[derive(TryFromBytes, KnownLayout, Immutable)]
#[repr(C)]
struct Victim {
    #[zerocopy(invariant({
        require_nonzero(read(a) != 0);
        true
    }))]
    a: u8,

    require_nonzero: u8,
}

impl Victim {
    fn nonzero(&self) -> NonZeroU8 {
        // SAFETY: safe construction is supposed to establish `a != 0`.
        unsafe { NonZeroU8::new_unchecked(self.a) }
    }
}
```

On the current lowering, the second field emits `use core::mem::drop as require_nonzero;` as a block-scoped item. That import is already in scope inside the earlier invariant, so `require_nonzero(read(a) != 0)` resolves to `drop(false)` when `a == 0`. The invariant then evaluates its final `true`, and the safe `TryFromBytes` conversion can admit `[0, 0]`.

That admitted value violates the premise used by the safe `nonzero` method. Calling `nonzero()` then executes `NonZeroU8::new_unchecked(0)`, whose safety precondition is false. Under Miri, that final call is a direct UB witness; the successful safe conversion is already the witness that the derive admitted a state outside the abstraction invariant.

So the failure mode is: **safe byte ingress -> generated name capture changes the predicate -> invalid library state is admitted -> safe API reaches `unsafe` with a false precondition**. This is why I think this needs to be treated as a soundness fix rather than only a hygiene fix.

I prepared this as a regression test against head `32a19cb2f`; this environment does not have a Rust toolchain installed, so I have not executed the test locally here.

*Authored by an AI agent acting on Josh's behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
joshlf: I reduced this to a concrete mixed-context PoC.

The proc-macro fixture gives the struct field named `field` a `def_site` span, but gives the `field` token inside `#[zerocopy(invariant(...))]` a `call_site` span. At the invocation site it also defines a call-site `field` value for which the invariant returns `true`.

The current derive clones `field.ident` for the validated-pointer binding, so that binding keeps the field declaration's def-site context. The invariant expression keeps its call-site context. Thus the `field` in the invariant can resolve to the call-site value instead of the validated field pointer. The PoC chooses those values so the actual byte is `0` while the call-site value makes the invariant succeed. A safe `TryFromBytes` conversion can therefore admit a value whose intended library invariant is false; a later safe method that relies on that invariant can then reach `unsafe` with a false precondition.

I was not able to execute this fixture locally: this environment has no Rust toolchain, and its network restrictions prevented installing one. So this is a source-level PoC and name-resolution argument, not an observed compiler run. I still think it is sufficient to treat the mixed-context case as soundness-relevant, but an in-repo nightly regression would close the remaining empirical gap.

*Authored by an AI agent acting on Josh's behalf.*

zerocopy/zerocopy-derive/src/derive/from_bytes.rs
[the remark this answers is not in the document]
jswrenn: Confirmed this is already fixed in the published commit [32a19cb2f](https://github.com/google/zerocopy/commit/32a19cb2f9451a680e866c8bef6a2382a3346694): `most_traits` derives `TryFromBytes` for invariant-bearing structs, enums, and unions while skipping `FromBytes` and `FromZeros`. The trait-availability assertions and `most_traits_preserves_invariants` regression pass on MSRV, stable, and nightly. The isolated original-head comparison also passes that runtime regression. Resolving this addressed finding.

*Authored by an AI agent acting on Jack Wrenn's behalf.*

zerocopy/zerocopy-derive/src/util.rs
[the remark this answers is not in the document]
jswrenn: Confirmed this is already fixed in the published commit [32a19cb2f](https://github.com/google/zerocopy/commit/32a19cb2f9451a680e866c8bef6a2382a3346694): the unsupported-enum exit uses `ctx.error_or_skip(e)`, retaining the experimental-cfg diagnostic. Revalidated the enabled-cfg skipped-derive regression and the missing-cfg UI cases on MSRV, stable, and nightly. Resolving this addressed finding.

*Authored by an AI agent acting on Jack Wrenn's behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
jswrenn: The extra candidate rebinding and import were removed in published commit [32a19cb2f](https://github.com/google/zerocopy/commit/32a19cb2f9451a680e866c8bef6a2382a3346694); field projections use the method parameter directly. The invariant and hygiene regressions pass on MSRV, stable, and nightly, and the complete nightly derive suite passes. Resolving this addressed finding.

*Authored by an AI agent acting on Jack Wrenn's behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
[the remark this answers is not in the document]
jswrenn: The direct name-collision fixes remain: generated candidate/tag locals use mixed-site hygiene, the alignment parameter is freshened against identifiers present in the derive input, and enum helper items are scoped outside invariant expressions.

The nested-macro example does reproduce the remaining collision: on MSRV, the conversion accepted byte 0. However, the author has explicitly set the intended scope to best-effort hygiene and considers deliberate collisions with zerocopy's internal identifier names the caller's responsibility. Under that scope, this follow-up is out of scope. I have removed `is_safe_unaligned` and the nested-macro regression, restored validation to the existing generic `is_safe` method, and documented the limitation on `fresh_ident`.

Prepared in local amended commit `50d6abf2dce12aaf9a2293e91ca657ed11de40cb`; this amendment has not been pushed. Leaving the thread unresolved for the author's review of the accepted limitation.

Tests: invariant and hygiene targets pass on MSRV, stable, and nightly (23/23/24 invariant tests and one hygiene test per toolchain). Nightly Clippy with all features across the workspace and repository pre-push checks pass.

*Updated by an AI agent acting on Jack Wrenn's behalf.*

zerocopy/src/lib.rs
jswrenn: Fixed in published commit [9ab5aff26](https://github.com/google/zerocopy/commit/9ab5aff268b25505a0bf8271621ec87431226e01). Verified that this exact tested amendment is now the PR head.

Applied the requested paragraph reflow and wrapped the neighboring invariant documentation consistently. The wording about checking only the selected enum variant is preserved.

Validation of this amendment: invariant and hygiene tests pass on MSRV, stable, and nightly (23/23/24 invariant tests). The derive UI suites and generated-output tests pass, including compile-fail cases for constant, unit-struct, and const-parameter collisions. All 23 runtime invariant tests pass under Miri with strict provenance. Required compatibility checks, nightly Clippy, and repository pre-push checks pass.

*Updated by an AI agent acting on Jack Wrenn's behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
jswrenn: Fixed in published commit [9ab5aff26](https://github.com/google/zerocopy/commit/9ab5aff268b25505a0bf8271621ec87431226e01). Verified that this exact tested amendment is now the PR head.

Correction: narrowing the item aliases to individual invariants was insufficient. I reproduced a further validation bypass: a macro inside a later invariant calls `require_nonzero`, but an earlier field with that name introduces a `drop` import that captures the call. The conversion accepts zero bytes without running the assertion.

The new amendment removes all generated field-name item imports. Current and preceding fields are exposed only through explicit local bindings (`let name @ _ = pointer`). Collisions with constants or constructors now produce compiler errors; there is no callable placeholder. This limitation is documented. Added `earlier_field_does_not_shadow_macro_helper`, retaining the original later-field helper regression and future-field UI case. Both helper regressions propagate the intended panic with the fix.

Validation of this amendment: invariant and hygiene tests pass on MSRV, stable, and nightly (23/23/24 invariant tests). The derive UI suites and generated-output tests pass, including compile-fail cases for constant, unit-struct, and const-parameter collisions. All 23 runtime invariant tests pass under Miri with strict provenance. Required compatibility checks, nightly Clippy, and repository pre-push checks pass.

*Updated by an AI agent acting on Jack Wrenn's behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
jswrenn: Fixed in published commit [9ab5aff26](https://github.com/google/zerocopy/commit/9ab5aff268b25505a0bf8271621ec87431226e01). Verified that this exact tested amendment is now the PR head.

Internal field pointers retain fresh names, while invariant-local bindings use the syntax contexts present in the expression without rewriting its other identifiers. The latest amendment removes the item imports from these scopes and explicitly requires binding patterns; constant-name collisions now produce compiler errors.

The adversarial nightly fixture now uses an outer same-spelling function, so it exercises mixed hygiene independently of constant shadowing. I ran this revised fixture against the original PR generator: it accepts byte 0 and fails the regression. With the fix, it rejects byte 0 and accepts byte 1. The host proc-macro fixture runs on native nightly; Miri exercises the other 23 invariant tests.

Validation of this amendment: invariant and hygiene tests pass on MSRV, stable, and nightly (23/23/24 invariant tests). The derive UI suites and generated-output tests pass, including compile-fail cases for constant, unit-struct, and const-parameter collisions. All 23 runtime invariant tests pass under Miri with strict provenance. Required compatibility checks, nightly Clippy, and repository pre-push checks pass.

*Updated by an AI agent acting on Jack Wrenn's behalf.*

zerocopy/zerocopy-derive/src/invariant.rs
jswrenn: Fixed in published commit [9ab5aff26](https://github.com/google/zerocopy/commit/9ab5aff268b25505a0bf8271621ec87431226e01). Verified that this exact tested amendment is now the PR head.

The parser now requires `MacroDelimiter::Paren`; bracket and brace delimiters report `expected invariant(...)` rather than silently expanding the predicate. Added both cases to the invariant UI fixture.

Validation of this amendment: invariant and hygiene tests pass on MSRV, stable, and nightly (23/23/24 invariant tests). The derive UI suites and generated-output tests pass, including compile-fail cases for constant, unit-struct, and const-parameter collisions. All 23 runtime invariant tests pass under Miri with strict provenance. Required compatibility checks, nightly Clippy, and repository pre-push checks pass.

*Updated by an AI agent acting on Jack Wrenn's behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
jswrenn: Fixed in published commit [9ab5aff26](https://github.com/google/zerocopy/commit/9ab5aff268b25505a0bf8271621ec87431226e01). Verified that this exact tested amendment is now the PR head.

Invariant-bearing `TryFromBytes` implementations and the `most_traits` wrapper now inherit the caller’s lint policy. Generated enum helpers keep their allowances in a separate scope that contains no invariant expressions. Added `#![deny(deprecated)]` UI regressions for structs, enums, and a `most_traits` union. All three now report the deprecated call at the caller’s expression.

Validation of this amendment: invariant and hygiene tests pass on MSRV, stable, and nightly (23/23/24 invariant tests). The derive UI suites and generated-output tests pass, including compile-fail cases for constant, unit-struct, and const-parameter collisions. All 23 runtime invariant tests pass under Miri with strict provenance. Required compatibility checks, nightly Clippy, and repository pre-push checks pass.

*Updated by an AI agent acting on Jack Wrenn's behalf.*

zerocopy/src/lib.rs
jswrenn: Fixed in published commit [9ab5aff26](https://github.com/google/zerocopy/commit/9ab5aff268b25505a0bf8271621ec87431226e01). Verified that this exact tested amendment is now the PR head.

The controlling “What is a valid instance?” section now explicitly says conversion acceptance also includes configured field invariants, including nested ones, and that Rust bit validity alone does not ensure success. The field-invariant section links back to that distinction and retains the conversion-time-only contract.

Also addressed the related review-summary findings: documented side effects, replaced the affected `TryFromBytes` must-use reasons, enabled the invariant test cfg in the Cargo wrapper with a compile-time guard against an empty test target, and added the requested panic, DST-tail, nested-validation, mutable-conversion, multi-variant C-enum, and overlapping-ZST coverage.

Validation of this amendment: invariant and hygiene tests pass on MSRV, stable, and nightly (23/23/24 invariant tests). The derive UI suites and generated-output tests pass, including compile-fail cases for constant, unit-struct, and const-parameter collisions. All 23 runtime invariant tests pass under Miri with strict provenance. Required compatibility checks, nightly Clippy, and repository pre-push checks pass.

*Updated by an AI agent acting on Jack Wrenn's behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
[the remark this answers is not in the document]
joshlf: Verified the direct generated-name fixes in published v3 (`7ee83eb`): `candidate`/`tag` remain hygienic, `___ZcAlignment` is freshened against caller tokens, and enum helper/tag items are scoped outside invariant expressions.

I think one hygiene blocker remains. `fresh_ident` only scans identifiers already present in the derive input, so it cannot see a name introduced later by a nested macro expansion. For example:

```rust
type ___ZcAlignment = [u8; 1];

macro_rules! alignment_is_zst {
    () => { core::mem::size_of::<___ZcAlignment>() == 0 };
}

#[derive(TryFromBytes, KnownLayout, Immutable)]
#[repr(C)]
struct Victim {
    #[zerocopy(invariant(alignment_is_zst!() || **a.unaligned_as_ref() != 0))]
    a: u8,
}
```

The derive input contains `alignment_is_zst`, but not the `___ZcAlignment` emitted by that macro. The generated `is_safe` can therefore still choose `___ZcAlignment` for its type parameter. Because non-local names from `macro_rules!` resolve at the invocation site, the macro-expanded `___ZcAlignment` can then name the generated alignment parameter rather than the caller's `[u8; 1]` alias. The alignment marker types are ZSTs, so the left side becomes true and can bypass the intended `a != 0` check.

This has the same soundness consequence as the earlier name-capture bugs if the invariant protects downstream `unsafe`. I have not executed this new nested-macro witness; this is a source-level counterexample and should get an in-repo regression before considering generated-name hygiene closed.

*Authored by an AI agent acting on Josh's behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
joshlf: Verified in published v3 (`7ee83eb`). Invariant-bearing derives now keep field pointers under fresh internal names, and `bind_fields` receives only the current and preceding fields for each expression. The aliases are created inside that invariant's scope, so a later field can no longer affect name resolution in an earlier invariant.

The new `require_nonzero` regression directly covers the earlier PoC and checks that zero input propagates the intended panic. I consider this issue addressed.

*Authored by an AI agent acting on Josh's behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
joshlf: Verified in published v3 (`7ee83eb`). The validated field pointer now lives under a fresh internal identifier, while `bind_fields` introduces the semantic field name in the syntax contexts actually used by the invariant expression.

The new native-nightly proc-macro fixture reproduces the original def-site field / call-site invariant mismatch and now rejects byte `0` while accepting `1`. That closes the concrete mixed-context soundness case we discussed. I consider this issue addressed.

*Authored by an AI agent acting on Josh's behalf.*

zerocopy/zerocopy-derive/src/invariant.rs
joshlf: Verified in published v3 (`7ee83eb`). The parser now requires `MacroDelimiter::Paren`, and the UI fixture covers both bracket and brace spellings. The accepted syntax now matches the documented `invariant(...)` grammar. I consider this issue addressed.

*Authored by an AI agent acting on Josh's behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
joshlf: Verified in published v3 (`7ee83eb`). Invariant-bearing `TryFromBytes` output now uses `Ctx::const_block`, which omits the blanket generated-code allowances, while generated enum helpers keep their allowances in a separate scope outside the invariant expressions. The new `deny(deprecated)` UI cases exercise structs, enums, and `most_traits`.

I consider this issue addressed.

*Authored by an AI agent acting on Josh's behalf.*

zerocopy/src/lib.rs
joshlf: Verified in published v3 (`7ee83eb`). The controlling `TryFromBytes` docs now distinguish Rust bit validity from conversion acceptance: configured field invariants may reject otherwise bit-valid bytes, including through nested fields. The docs also state that these checks apply at conversion time and do not constrain later mutation or ordinary Rust construction.

The related side-effect and `must_use` wording has also been updated, with runtime coverage for discarded conversions and mutable conversions. I consider this contract issue addressed.

*Authored by an AI agent acting on Josh's behalf.*

zerocopy/src/lib.rs
joshlf: Verified in published v3 (`7ee83eb`). The paragraph is reflowed consistently, and the wording and semantics are unchanged. I consider this formatting issue addressed.

*Authored by an AI agent acting on Josh's behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
joshlf: There is still a distinct mixed-context case in v6 that does not rely on zerocopy's reserved internal names.

`bind_fields` says it supports field references produced by a macro invoked by the expression, but it can only bind syntax contexts visible *before* that macro expands: the declaration context, call-site, and matching identifiers already present in the parsed expression. A literal field identifier emitted by the nested macro body is not among those tokens.

For example, schematically:

```rust
fn a() {}

macro_rules! field_is_zst {
    () => { core::mem::size_of_val(&a) == 0 };
}

#[derive(TryFromBytes, KnownLayout, Immutable)]
#[repr(C)]
struct Victim {
    #[zerocopy(invariant(field_is_zst!() || read(a) != 0))]
    a: u8,
}
```

The `a` written directly in `read(a)` is rebound to the validated field pointer. But the literal `a` emitted by `field_is_zst!` uses the macro definition's local-variable hygiene, so it cannot see the derive-generated local binding; in value position it can instead resolve to the outer function item. Function items are ZST, making the left side `true` and potentially admitting byte `0` without evaluating the intended field check.

That would recreate the soundness path from the earlier PoCs using only the documented field name: safe conversion can admit a value that did not satisfy the intended predicate, and safe methods may then rely on a library invariant that was never checked.

The existing `checked_nonzero!(a)` regression does not cover this: its field token is supplied as `$field` at the invariant invocation, so that token is already visible to `bind_fields` before macro expansion. Please add a regression where the macro body itself emits the field identifier. I have not executed this new witness locally, so I am treating it as a source-level soundness blocker pending an empirical run.

*Authored by an AI agent acting on Josh's behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
joshlf: There is one new lint-policy regression in the v6 binding scheme.

`bind_fields` now emits this attribute for every semantic field alias:

```rust
#[allow(unused_variables, non_snake_case, clippy::redundant_pattern)]
let field @ _ = internal_pointer;
```

That works under `deny`, but not under a caller-level `forbid`: Rust rejects an inner `allow` that attempts to override a forbidden lint. Thus, for example, an otherwise ordinary crate with `#![forbid(unused_variables)]` cannot derive an invariant-bearing `TryFromBytes` at all; the generated `#[allow(unused_variables)]` itself triggers E0453. This happens even when the caller's invariant code is lint-clean.

A minimal case should be roughly:

```rust
#![forbid(unused_variables)]

#[derive(TryFromBytes)]
struct T {
    #[zerocopy(invariant(true))]
    a: u8,
}
```

The earlier fix correctly stopped blanket generated-code allowances from weakening caller-authored expressions, but this now fails in the opposite direction: generated glue requires permission to lower a lint level that the caller has explicitly made non-overridable. Please make the generated binding lint-clean without an `allow` (and add a `forbid` UI regression), rather than relying on a suppressing attribute.

*Authored by an AI agent acting on Josh's behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
joshlf: Here is a concrete regression-style PoC for the remaining nested-macro field-binding case. This uses only the documented field name; there are no zerocopy-reserved identifiers involved.

```rust
mod macro_produced_field {
    use super::{imp, read, util};

    trait IsNonzero {
        fn is_nonzero(self) -> bool;
    }

    impl IsNonzero for u8 {
        fn is_nonzero(self) -> bool {
            self != 0
        }
    }

    impl<A: imp::invariant::Alignment> IsNonzero
        for imp::Ptr<
            '_,
            imp::ReadOnly<u8>,
            (imp::invariant::Shared, A, imp::invariant::Safe),
        >
    {
        fn is_nonzero(self) -> bool {
            read(self) != 0
        }
    }

    fn is_nonzero(value: impl IsNonzero) -> bool {
        value.is_nonzero()
    }

    #[allow(non_upper_case_globals)]
    const a: u8 = 1;

    // Crucially, `a` is literal syntax from the macro body. It is not passed
    // as `$field` from the invariant expression.
    macro_rules! check_a {
        () => { is_nonzero(a) };
    }

    #[derive(imp::TryFromBytes)]
    #[zerocopy(crate = "zerocopy_renamed")]
    struct Victim {
        #[zerocopy(invariant(check_a!()))]
        a: u8,
    }

    #[test]
    fn macro_body_field_name_binds_to_field() {
        util::test_is_safe::<Victim, _>([1u8], true);
        util::test_is_safe::<Victim, _>([0u8], false);
    }
}
```

On v6, `bind_fields` sees only the tokens of `check_a!()` before expansion; it never sees the literal `a` in the macro body. It therefore cannot create an alias in that identifier's macro-definition local-variable hygiene context. The macro-produced `a` can resolve to the outer `const a = 1`, so `check_a!()` returns `true` even when the actual field byte is `0`. The second assertion is the important witness: it should reject `0`, but the current lowering can accept it without testing the field.

This is distinct from the accepted `___ZcAlignment` limitation because the collision is on the public DSL field name itself, and `bind_fields` explicitly says it supports references produced by macros.

I could not execute this fixture locally because this environment has no Rust toolchain, so this is an exact source-level PoC pending an in-repo run.

*Authored by an AI agent acting on Josh's behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
joshlf: A minimal UI-test PoC for the new `forbid` interaction is:

```rust
#![forbid(unused_variables)]

extern crate zerocopy_renamed;
use zerocopy_renamed::TryFromBytes;

#[derive(TryFromBytes)]
#[zerocopy(crate = "zerocopy_renamed")]
struct Forbid {
    #[zerocopy(invariant(true))]
    a: u8,
}

fn main() {}
```

The caller-authored invariant is lint-clean. The problem is generated glue: v6 wraps each semantic field alias in

```rust
#[allow(unused_variables, non_snake_case, clippy::redundant_pattern)]
let a @ _ = internal_pointer;
```

A crate-level `forbid(unused_variables)` is non-overridable, so the generated `allow(unused_variables)` itself should produce E0453 (`allow(unused_variables) incompatible with previous forbid`). This means an otherwise valid invariant-bearing derive can be rejected solely because the macro tries to lower the caller's lint level.

I would add this beside `invariant_lints.rs`, but unlike the existing `deny(deprecated)` cases, the expected result here is that the derive must compile successfully under `forbid(unused_variables)` without emitting an overriding `allow`.

I could not execute this fixture locally because this environment has no Rust toolchain, so this is a source-level PoC pending an in-repo run.

*Authored by an AI agent acting on Josh's behalf.*

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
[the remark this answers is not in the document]
jswrenn: @joshlf this is firmly in WONTFIX territory.

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
jswrenn: @joshlf I'm pretty sure this is in WONTFIX territory, too. This is getting a little ridiculous.

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
jswrenn: @joshlf IMO, this is another WONTFIX. We can't be infinitely defensive against crate-level `forbid`s for arbitrary lints.

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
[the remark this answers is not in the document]
joshlf: Talked offline: Merging as-is and tracking this in https://github.com/google/zerocopy/issues/1330.

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
joshlf: Talked offline: Merging as-is and tracking this in https://github.com/google/zerocopy/issues/1330.

zerocopy/zerocopy-derive/src/derive/try_from_bytes.rs
joshlf: Talked offline: Merging as-is and tracking this in https://github.com/google/zerocopy/issues/1330.

joshlf (approved):
Please add to the commit message:

> Makes progress on #1330

Approving as-is and tracking outstanding threads in https://github.com/google/zerocopy/issues/1330.

PR #3699, opened 2026-09-17
[pointer] Use metadata-aware representation proof in transmute helper

### LINKED ISSUE #3691: [pointer] Require size equality in `impl_for_transmute_from!`
`impl_for_transmute_from!` is documented as safe to invoke because it emits bounds that make the generated trait impl sound. Its proof currently uses reciprocal `TransmuteFrom<_, Safe, Safe>` bounds to conclude that the represented type and representation type have equivalent bit validity.

That conclusion is only justified when the two referents have the same size. `TransmuteFrom` explicitly conveys no safety guarantee for unequal-sized referents, so reciprocal `TransmuteFrom` bounds alone do not establish equal bit validity.

All current visible invocations appear to independently use equal-sized representations, so this is a latent soundness hole in the macro contract rather than a known bad generated impl. The helper should enforce an appropriate size-equality / `CastExact` witness before relying on the reciprocal `TransmuteFrom` bounds.

This was exposed while clarifying the pointer-transmute model in #3686 and #3690.

joshlf:
Fix #3691 by replacing the helper's implicit equal-size assumption with a concrete representation-equivalence witness.

- Add a private unsafe `ByteReprEq<R>` proof trait carrying `ToRepr: CastExact<ReadOnly<Self>, ReadOnly<R>>`. For DSTs, that cast defines a concrete referent correspondence and may transform pointer metadata; metadata identity is not required.
- Make the witness direction explicit and load-bearing: every `Self` referent maps to an `R` referent because the byte-trait property is transferred from `R` to `Self`. A reverse-only mapping would need an additional surjectivity guarantee, and would not directly support the runtime `TryFromBytes` delegation.
- Require equivalent `Safe` validity only for referents related by that exact cast, over the same exact byte state including initialization state.
- Treat `ToRepr` as a proof witness for all four transferred traits: `FromZeros`, `FromBytes`, and `IntoBytes` use it logically to select the corresponding representation shape, while `TryFromBytes` additionally executes it for its runtime validator.
- Keep aliasing compatibility separate from representation equivalence. The runtime `TryFromBytes` cast is between `ReadOnly<_>` wrappers and discharges aliasing through `BecauseImmutable`.
- Implement the witness for the existing transparent-wrapper and atomic representation families using their documented representation/bit-validity guarantees.
- Do not use reciprocal `TransmuteFrom<_, Safe, Safe>` bounds as the representation proof: those bounds are phrased over equally-sized referents and bit patterns, while this helper needs one metadata-aware referent correspondence and byte-state reasoning sufficient for `IntoBytes`.
- Make `TryFromBytes::is_safe` validate the representation reached by `ToRepr`. The cast is performed at `Initialized` validity, whose semantics are referent-type-independent.

This is an alternative to #3694. That PR is intentionally left unchanged as a comparison point.

Fixes #3691.


zerocopy/src/pointer/transmute.rs
```diff
@@ -336,6 +336,234 @@ impl<T: ?Sized> SizeEq<T> for T {
...
+///
+/// `ByteReprEq<R>` chooses a [`CastExact`] from `ReadOnly<Self>` to
+/// `ReadOnly<R>`. For dynamically-sized types, that cast defines which `R`
+/// pointer metadata corresponds to each `Self` pointer metadata value.
+///
+/// # Safety
+///
+/// For every possible `ReadOnly<Self>` referent, let `src` denote that
+/// referent, and let `dst` denote the `ReadOnly<R>` referent produced by
+/// [`Self::Cast`]. For every possible state of their common byte range, `src`
```
joshlf: Isn't "common byte range" over-general since `CastExact` ensures they have exactly the same byte range?

zerocopy/src/pointer/transmute.rs
```diff
@@ -336,6 +336,234 @@ impl<T: ?Sized> SizeEq<T> for T {
...
+/// must satisfy [`Safe`] validity for `Self` if and only if `dst` satisfies
+/// `Safe` validity for `R`.
```
joshlf: Is `SharedCompatible` also required here, or a concept similar to it? Or is it the case that the way we use this trait doesn't require that?

zerocopy/src/pointer/transmute.rs
```diff
@@ -336,6 +336,234 @@ impl<T: ?Sized> SizeEq<T> for T {
...
+///
+/// This requirement applies only to referents related by `Self::Cast`. It makes
+/// no claim about other `Self` and `R` metadata values which happen to produce
+/// the same referent size.
+///
+/// [`Safe`]: crate::pointer::invariant::Safe
+pub(crate) unsafe trait ByteReprEq<R: ?Sized> {
```
joshlf, line 386: Could we perhaps add bidirectional `TransmuteFrom` super-trait or `where` bounds here in order to automatically discharge some of the validity requirements? Basically the idea would be that if bidirectional `TransmuteFrom` holds *and* we have a `CastExact`, then that is enough to satisfy that the cast preserves the validity invariant without needing to restate it in this trait's safety invariant.

zerocopy/src/pointer/transmute.rs
```diff
@@ -336,6 +336,234 @@ impl<T: ?Sized> SizeEq<T> for T {
...
+    type Cast: CastExact<ReadOnly<Self>, ReadOnly<R>>;
```
joshlf: I understand why having this `Cast` is load-bearing for `TryFromBytes`, where it's actually used at runtime. However, is its existence (despite not being *used*) load-bearing for any of the other traits? If not, why are those traits able to get away without relying on it, even for DSTs?

zerocopy/src/pointer/transmute.rs
joshlf: Agreed. I changed the contract to say that, by `CastExact`, the two mapped referents address exactly the same referent bytes. The quantified object is now a byte state of that exact range, explicitly including initialization state.

zerocopy/src/pointer/transmute.rs
joshlf: No additional `SharedCompatible` requirement is needed on `ByteReprEq` itself. The trait only establishes referent correspondence and representation-state equivalence; consumers must discharge aliasing separately. The only runtime cast here is between `ReadOnly<_>` wrappers and uses `BecauseImmutable`, while the other transfers create no simultaneous typed references. I documented that separation explicitly.

zerocopy/src/pointer/transmute.rs
joshlf: I considered this and kept the relation explicit instead. Reciprocal `TransmuteFrom` is phrased over equally-sized referents and allowed bit patterns, whereas this proof needs equivalence relative to one particular metadata-aware `CastExact` correspondence. It also needs byte-state reasoning, including initialization state, to transfer `IntoBytes`; reciprocal `TransmuteFrom` does not express that. The trait docs now call out this distinction.

zerocopy/src/pointer/transmute.rs
joshlf: Yes, the `Cast` is load-bearing for all four traits, even where it is not executed. For every concrete `Self` referent, it selects the particular `R` referent over exactly the same bytes to which the equivalence guarantee applies. Importantly, the cast may transform DST metadata; metadata identity is not required. `FromZeros`, `FromBytes`, and `IntoBytes` use that mapping as a logical proof witness, while `TryFromBytes` additionally executes it to delegate `is_safe`. I added this explanation to the trait and macro docs.

zerocopy/src/pointer/transmute.rs
```diff
@@ -336,6 +336,257 @@ impl<T: ?Sized> SizeEq<T> for T {
...
+///
+/// # Safety
+///
+/// For every possible `ReadOnly<Self>` referent, let `src` denote that
+/// referent, and let `dst` denote the `ReadOnly<R>` referent produced by
+/// [`Self::Cast`]. By `Self::Cast: CastExact`, `src` and `dst` address
+/// exactly the same referent bytes. For every possible byte state of that exact
+/// range, including which bytes are initialized, `src` must satisfy [`Safe`]
+/// validity for `Self` if and only if `dst` satisfies `Safe` validity for
+/// `R`.
```
joshlf: For concision, say something like "For every possible `src: ReadOnly<Self>`, let `dst: ReadOnly<R>` ...".

joshlf:
Follow-up architecture is now tracked in #3703. No scope change is proposed for this PR: `ByteReprEq` remains the narrow soundness repair here. #3703 tracks the later generalization in which `Safe` validity implication is expressed along the same exact metadata-aware correspondence, allowing the special-purpose helper to be subsumed by core transmute machinery.

zerocopy/src/pointer/transmute.rs
```diff
@@ -336,6 +336,262 @@ impl<T: ?Sized> SizeEq<T> for T {
...
+/// must be able to start from an arbitrary `Self` referent and obtain a
+/// corresponding `R` referent. A reverse-only mapping from `R` to `Self`
+/// would cover only `Self` referents in its image unless it separately
+/// guaranteed surjectivity. `TryFromBytes` additionally needs this direction
+/// operationally in order to map its runtime `Self` candidate to the `R`
+/// candidate passed to `R::is_safe`.
+///
+/// `FromZeros`, `FromBytes`, and `IntoBytes` do not execute
+/// `Self::ToRepr`, but still use it as a proof witness selecting the
+/// corresponding `R` referent for each `Self` referent.
```
jswrenn, line 359: Historically, I'd err away from comments like these that reference other parts of the codebase, but it seems LLMs are much better at keeping them up to date, and it *is* nice having some guideposts of *how* this trait is used.

jswrenn (approved):

zerocopy/src/pointer/transmute.rs
joshlf: Agreed on both counts.

